PR #27, opened 2021-12-20
WIP: p2w_autoattest to use config file instead of env variables

erancx:
This is mainly due to security concerns.

Fixes https://github.com/pyth-network/pyth2wormhole/issues/25 (2)


ali-bahjati:
This is outdated.

PR #62, opened 2022-03-07
Adding and splitting p2w relay and attest Dockerfile

erancx:
This PR adds an improved Dockerfiles for p2w-attest and relay.
Next step is adding CI/CD for these images.

Ticket: CU-22k4cj4


ali-bahjati:
This is outdated now.

PR #219, opened 2022-05-20
pyth2wormhole-client: refactor into fully-async futures-based approach

drozdziak1:
The main scope of this changeset lives in `main.rs`. We start using Rust's async/await paradigm to improve on the crude round-robin concurrency we got from the previous polling approach.  At the cost of increased complexity, this approach wins us approximately 2x faster batch attestation life cycle (send + confirm transaction). From now on, we become bottlenecked by Solana's devnet rate-limiting, which we can address in subsequent changesets and stricter attestation conditions in prod configuration.

### DIFF
--- a/solana/pyth2wormhole/client/src/main.rs
+++ b/solana/pyth2wormhole/client/src/main.rs
@@ -145,10 +145,10 @@ async fn main() -> Result<(), ErrBox> {
                 read_keypair_file(&*shellexpand::tilde(&owner))?,
                 latest_blockhash,
             )?;
-            rpc_client.send_and_confirm_transaction_with_spinner(&tx)?;
+            rpc_client.send_and_confirm_transaction_with_spinner(&tx).await?;
             println!(
                 "Applied conifg:\n{:?}",
-                get_config_account(&rpc_client, &p2w_addr)?
+                get_config_account(&rpc_client, &p2w_addr).await?
             );
         }
         Action::Attest {


solana/pyth2wormhole/client/src/main.rs
```diff
@@ -171,301 +206,214 @@ fn handle_attest(
...
+    // batch. Without daemon mode, we return after the first job is
+    // complete on the batch. In daemon mode, we never return and
+    // resend each batch continuously according to attestation conditions.
+    let attestation_sched_futs = batches.into_iter().map(|(batch_no, b)| {
+        let config4fut = config.clone();
+        let p2w_addr4fut = p2w_addr.clone();
+        let payer4fut = Keypair::from_bytes(&payer.to_bytes()).unwrap(); // Keypair has no clone
+        let rpc_url4fut = rpc_url.clone();
+        let commitment4fut = commitment.clone();
+        let batch_count4fut = batch_count;
```
ali-bahjati, line 219: @Reisen can these be avoided in some way?

ali-bahjati:
I think it looks very good. I am not sure how errors are being handled in the async state and also using lock. are the lock released? will errors fails or retry? :think:

ali-bahjati:
In this structure we allow `x` concurrent batches from same attestation, I think it's not that necessary to have x concurrent and wait if it's not meeting it. If we have 1 batch and max_per_batch of 2. then we have transaction `a` pending, then transaction `b` comes and it's again pending it's better to kill transaction `a` thread and start transaction `c` instead. 

I don't know how complex it is and how much we actually pend on sending the transaction.

solana/pyth2wormhole/client/src/batch_state.rs
```diff
@@ -40,63 +42,59 @@ impl<'a> BatchState<'a> {
...
                 .get_account_data(&sym.price_addr)
                 .map_err(|e| e.to_string())
-                .and_then(|bytes| {
+                .and_then(|bytes| async move {
                     pyth_sdk_solana::state::load_price_account(&bytes)
                         .map(|state| state.clone())
                         .map_err(|e| e.to_string())
-                }) {
+                })
+                .await
```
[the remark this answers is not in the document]
drozdziak1: You are correct, I intentionally only refactored the core event loop so that only essential complexity is introduced. The account state lookups are reasonably quick and I don't consider them important to parallelize. They're already concurrent across batches thanks to the main.rs changes

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -171,301 +206,214 @@ fn handle_attest(
...
+                let payer4job = Keypair::from_bytes(&payer4fut.to_bytes()).unwrap(); // Keypair has no clone
+                let group_name4job = b4fut.group_name.clone();
+                let symbols4job = b4fut.symbols.to_vec();
+                let rpc4job =
+                    RpcClient::new_with_timeout_and_commitment(rpc_url4fut.clone(),confirmation_timeout.clone(), commitment4fut.clone());
+                let sema4job = sema.clone();
+
+                // A future for a single attempt to attest a batch on
+                // Solana.
+                let attestation_job = async move {
```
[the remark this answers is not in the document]
drozdziak1: Yeah, I usually haven't had good results with making async routines into their own functions, but I'll give it a shot. I may have slight prejudice from the dark age before async/await :sweat_smile:

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -171,301 +206,214 @@ fn handle_attest(
...
+                };
+
+                debug!(
+                    "Batch {}/{}, group {:?}: Scheduling attestation job",
+                    batch_no, batch_count4fut, b4fut.group_name
+                );
+                if daemon {
+                    // park this routine until a resend condition is met
+                    loop {
+                        if let Some(reason) = b4fut.should_resend(&rpc4fut).await {
```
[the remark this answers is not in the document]
drozdziak1: In my opinion, async/await in Rust makes async polling very cheap, because every `await` is essentially a preemption callback to the runtime which can park this task and do something else until a poll determines the awaited task is done. If anything, I think this code needs rate-limiting in this and some other places. Unfortunately, I wasn't able to come up with a good way to tie all RPC requests to a mandatory rate-limiting delay. At this point I think a global lock on a single RPC instance might be the way to go. Ideally, with a way to tie a rate-limiting delay to every lock release.

drozdziak1:
> 

I thought about this and I don't believe we should cancel attestation jobs if max jobs are reached.  This would require much more state for the attestation_sched futures. Like I mentioned in the `attestation_cfg` comment for the max job count, this is just a failsafe to make sure the job count does not cause the an OOM condition.

solana/pyth2wormhole/client/src/util.rs
```diff
@@ -0,0 +1,84 @@
+use log::trace;
+
+use tokio::sync::{Mutex, MutexGuard};
+use std::{ops::{Deref, DerefMut}, time::{Duration, Instant}};
+
+/// Rate-limited mutex. Ensures there's a period of minimum rl_interval between lock acquisitions
+pub struct RLMutex<T> {
```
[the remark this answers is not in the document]
drozdziak1: I'm really torn about this, because your snippet is significantly smaller, at the cost of some precision and user convenience I was really looking for. I think I may be falling for a sunken cost thing too, but here goes nothing :sweat_smile: 

This implementation may not be useful _right now_, but it may change when we decide to attest bigger symbol counts, which I think will happen long before the attester contract becomes obsolete. I'd argue the RLMutex take is more precise and harder to misuse, which may age better in that case.

Starting with the volume of the RLMutex, most is contributed by `RLMutexGuard` and `RLMutexState`, which primarily add one- or two-liner trait implementations.

Logic-wise, I don't think `rate_limit_lock` will achieve the desired behavior. Functionally, `RLMutex`  sleeps only if the desired duration hadn't passed _already_ since last lock release. If we do sleep, the wait time is precisely the remaining time until target rate limit is met, while `rate_limit_lock` will always slap a forced sleep onto each lock acquisition, no matter when it happens. Any attempt to match the desired behavior would result in keeping a mutable state equivalent to that present in `RLMutex`.

Finally, I would still prefer not to keep the rate value as a separate variable. I do not like that, because the job of applying the correct rate would move to each `rate_limit_lock` call site, and that makes each use prone to cause concurrency bugs, and IMO those tend to be a huge waste of time. If we were to keep the rate elsewhere, we would still end up with a wrapper for a `Mutex`, one way or another.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I think the above only works because the internal reference is `'static`. Any non-copy by-value member would cause a `doesn't live long enough`. The read-only data I'm using isn't `'static`, so I'd probably need a bunch of `Arc`s. Not ideal, but may still be cheaper and cleaner than the clones. A regular attempt to borrow read-only  for `config` gives me this:
```rust
error[E0597]: `config` does not live long enough
   --> client/src/main.rs:216:27
    |
215 |       let attestation_sched_futs = batches.into_iter().map(|(batch_no, b)| {
    |                                                            --------------- value captured here
216 |           let config4fut = &config;
    |                             ^^^^^^ borrowed value does not live long enough
...
236 |                   let job = attestation_job(
    |  ___________________________-
237 | |                     rpc4job,
238 | |                     batch_no,
239 | |                     batch_count4fut,
...   |
245 | |                     sema.clone(),
246 | |                 );
    | |_________________- argument requires that `config` is borrowed for `'static`
...
363 |   }
    |   - `config` dropped here while still borrowed

```

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -45,16 +61,24 @@ use pyth2wormhole_client::*;
...
 
-fn main() -> Result<(), ErrBox> {
+#[tokio::main]
+async fn main() -> Result<(), ErrBox> {
     let cli = Cli::parse();
     init_logging(cli.log_level);
 
     let payer = read_keypair_file(&*shellexpand::tilde(&cli.payer))?;
-    let rpc_client = RpcClient::new_with_commitment(cli.rpc_url, CommitmentConfig::confirmed());
+    let commitment = match cli.commitment.as_str() {
```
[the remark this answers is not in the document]
drozdziak1: Yep, used their trait impl directly in `cli.rs`

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -171,301 +210,240 @@ fn handle_attest(
...
-                                .map_err(|e| -> ErrBox { e.into() })
-                        });
-
-                    // Individual batch errors mustn't prevent other batches from being sent.
-                    match res {
-                        Ok(signature) => {
+    let rpc = Arc::new(RLMutex::new(RpcCfg {url: rpc_url, timeout: confirmation_timeout, commitment: commitment.clone() }, rpc_interval));
+
+    // Each future schedules individual attestation attempts per
+    // batch. Without daemon mode, we return after the first job is
```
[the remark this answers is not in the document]
drozdziak1: Reworded the comment and moved it to a helper function that replaces this closure

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -171,301 +210,240 @@ fn handle_attest(
...
+    // complete on the batch. In daemon mode, we never return and
+    // resend each batch continuously according to attestation conditions.
+    let attestation_sched_futs = batches.into_iter().map(|(batch_no, b)| {
```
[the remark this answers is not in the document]
drozdziak1: becomes `mut batch` on `attestation_sched_job`

solana/pyth2wormhole/client/src/batch_state.rs
```diff
@@ -40,63 +43,46 @@ impl<'a> BatchState<'a> {
...
-                }
-            };
-
-            new_symbol_states.push(new_state);
+        let mut new_symbol_states: Vec<Option<PriceAccount>> = vec![None; sym_count];
+        let pubkeys: Vec<_> = self.symbols.iter().map(|s| s.price_addr).collect();
+
+        // Always learn the current on-chain state for each symbol
+        match c.get_multiple_accounts(&pubkeys).await {
+            Ok(acc_opts) => {
```
tompntn, line 63: Changing `new_symbol_states` to be a mapping of `Pubkey` to `PriceAccount`s might let you simplify this logic.

solana/pyth2wormhole/client/src/lib.rs
```diff
@@ -50,19 +51,24 @@ use pyth2wormhole::{
...
-    BatchState,
-    BatchTxStatus,
+pub use batch_state::BatchState;
+pub use util::{
+    RLMutex,
+    RLMutexGuard,
 };
 
+/// Future-friendly version of solitaire::ErrBox
+pub type ErrBoxSend = Box<dyn std::error::Error + Send + Sync>;
```
tompntn, line 70: Have you seen anyhow (https://docs.rs/anyhow/latest/anyhow/)? It allows you to easily propagate any error which implements `std::error::Error` by converting it to an `anyhow::Error`. You can then return `Result<T, anyhow::Error>` from your functions, and across threads, without having to do manual error conversion using `map_err`. It makes dealing with errors in Rust much nicer.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -190,301 +225,276 @@ fn handle_attest(
...
-    // NOTE(2022-04-26): only increment this if `daemon` is false
-    let mut finished_count = 0;
+    let rpc_cfg = Arc::new(RLMutex::new(
+        RpcCfg {
+            url: rpc_url,
+            timeout: confirmation_timeout,
+            commitment: commitment.clone(),
+        },
+        rpc_interval,
+    ));
```
tompntn, line 235: From a readibility perspective, the fact that an `RLMutex` is used here may be missed at first glance by someone skimming over the code. It may be more obvious to make the `RLMutex` a field and highlight this with a comment.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -190,301 +225,276 @@ fn handle_attest(
...
+    // With daemon mode off, the sched jobs return from the
+    // join_all. We filter out errors and report them
+    let errors: Vec<_> = results
+        .iter()
+        .filter_map(|r| {
+            if let Err(e) = r {
+                Some(e.to_string())
+            } else {
+                None
             }
```
tompntn, line 268: You could use `Result::err` here: https://doc.rust-lang.org/std/result/enum.Result.html#method.err

solana/pyth2wormhole/client/src/batch_state.rs
```diff
@@ -40,63 +43,46 @@ impl<'a> BatchState<'a> {
...
+                                .map_err(|e| {
+                                    warn!("Could not parse symbol {}/{}: {}", idx, sym_count, e);
+                                    e
+                                })
+                                .ok() // Err becomes None
+                        })
+                    })
+                    .collect();
+            }
+            Err(e) => warn!("Could not look up any symbols on-chain: {}", e),
```
tompntn, line 81: You could move this below the "min interval" check, which will potentially reduce the number of RPC calls.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -190,301 +225,276 @@ fn handle_attest(
...
-    for (batch_no, state) in batches {
-        match state.get_status() {
-            Success { .. } => {}
-            FailedSend { last_err, .. } | FailedConfirm { last_err, .. } => {
-                errors.push(last_err.to_string())
+        if daemon {
+            // park this routine until a resend condition is met
+            loop {
+                if let Some(reason) = batch
+                    .should_resend(&lock_and_make_rpc(&rpc_cfg).await)
```
tompntn, line 351: This will create a new RPC client, and so a new connection, every time, which could potentially be problematic. Instead could you create one RPC client at the top of `attestation_sched_job` and then sleep inside this loop to rate-limit the calls?

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -190,301 +225,276 @@ fn handle_attest(
...
+    retry_interval: Duration,
+    daemon: bool,
+    rpc_cfg: Arc<RLMutex<RpcCfg>>,
+    p2w_addr: Pubkey,
+    config: Pyth2WormholeConfig,
+    payer: Keypair,
+) -> Result<(), ErrBoxSend> {
+    let mut retries_left = n_retries;
+    // Enforces the max batch job count
+    let sema = Arc::new(Semaphore::new(batch.conditions.max_batch_jobs));
```
tompntn, line 328: Having multiple in-flight instances of the same request (well, technically batch but you know what I mean) is possibly overkill - what's the rationale here? Did it solve a particular problem you ran into?

solana/pyth2wormhole/client/src/lib.rs
drozdziak1: I've heard of it, but trading this one-liner for a whole dependency did not seem justified. I checked and we have 3-4 trivial map_err's in the whole pyth2wormhole cargo workspace. I'm sure this could be a good step to take in the future

solana/pyth2wormhole/client/src/batch_state.rs
drozdziak1: This does seem nice, but a map will obfuscate the order in which the symbols were configured (which could be useful for cases where we want to rule out typos in failing batch configs)

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I think that's a good idea, thanks

solana/pyth2wormhole/client/src/main.rs
drozdziak1: The period from sending to confirmation will often take longer than the duration between subsequent crank turns of a batch. This changeset decouples scheduling a crank turn from waiting for its confirmations on Solana. This enables better attestation rates per-symbol.

solana/pyth2wormhole/client/src/batch_state.rs
tompntn: Sorry if I'm misunderstanding, but I think all this loop is doing is updating `new_symbol_states`. We don't reference `new_symbol_states` unless the "min interval" check below passes, so I'm not sure what we gain by updating it every time? What "other conditions" are you referring to? Sorry if I'm missing something :)

drozdziak1: The general idea is to populate `ret` instead of returning and let it be returned at the very end. `new_symbol_states`  is referenced when we check that a condition was found. We use it update the state only if a condition is met. This way we can evaluate conditions since last attestation as opposed to since last lookup. The other conditions are publish_time change and price value change

drozdziak1: i.e. we have three possible conditions that can trigger an attestation:
1. min_interval elapsed since last attestation
2. publish_time changed by at least the specified value
3. price value changed by at least the specified percentage

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I will try rewording the comment for this function. But it's very hard to explain concisely

solana/pyth2wormhole/client/src/batch_state.rs
tompntn: I see, thank-you!

tompntn (approved):

PR #221, opened 2022-05-24
Add Websocket to price service

ali-bahjati:
This adds web socket support for the price service to subscribe to a list of price feeds. It logs their ip and assigns an id for better tracking/debugging. 

I think the client side (on pyth-common-js) is relatively easy, just we should make sure to use a library that supports browser and node at the same time (like `isomorphic-ws`)


third_party/pyth/price-service/src/promClient.ts
```diff
@@ -31,6 +31,11 @@ export class PromClient {
     buckets: [1, 5, 10, 15, 30, 60, 120, 180],
     labelNames: ["path", "price_id"]
   });
+  private webSocketInteractionCounter = new client.Counter({
+    name: `${SERVICE_PREFIX}websocker_interaction`,
```
jayantk, line 35: typo: websocker

third_party/pyth/price-service/src/listen.ts
```diff
@@ -27,7 +27,9 @@ export type PriceInfo = {
 };
 
 export interface PriceFeedPriceInfo {
-  getLatestPriceInfo(priceFeedId: string): PriceInfo | undefined;
+  getPriceIds(): Set<HexString>;
```
jayantk, line 30: nitpick: the name `PriceFeedPriceInfo` is not great. this interface is basically the API for the spy right? maybe something like SpyApi would be better?

jayantk (approved):
this is really nicely done!

third_party/pyth/price-service/src/ws.ts
```diff
@@ -0,0 +1,211 @@
...
+        logger.info(`client ${this.wsId.get(ws)} closed the connection.`);
+        this.promClient?.addWebSocketInteraction("close", "ok");
+
+        this.clientClose(ws);
+      });
+
+      this.promClient?.addWebSocketInteraction("connection", "ok");
+    });
+
+    const pingInterval = setInterval( () => {
```
Reisen, line 186: Overall looks great, a quick format with `prettier` or similar run over the files to fix up little spacing issues is the only nitpick.

Reisen (dismissed):

third_party/pyth/price-service/src/promClient.ts
ali-bahjati: Thanks!

Reisen (approved):

PR #222, opened 2022-06-13
ethereum: Add setContractConfig for on-chain state

drozdziak1:
This method lets us alter the emitter/wormhole/chain IDs on-chain using the owner

### DIFF
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -28,7 +28,9 @@ contract Pyth is PythGetters, PythSetters, AbstractPyth {
         (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(encodedVm);
 
         require(valid, reason);
-        require(verifyPythVM(vm), "invalid emitter");
+
+	(bool pythValid, string memory pythReason) = verifyPythVM(vm);
+	require(pythValid, pythReason);
 
         PythInternalStructs.BatchPriceAttestation memory batch = parseBatchPriceAttestation(vm.payload);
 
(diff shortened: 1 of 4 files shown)


ali-bahjati (approved):
lgtm!

ethereum/contracts/pyth/Pyth.sol
```diff
@@ -73,14 +75,14 @@ contract Pyth is PythGetters, PythSetters, AbstractPyth {
         return info;
     }
 
-    function verifyPythVM(IWormhole.VM memory vm) private view returns (bool valid) {
+    function verifyPythVM(IWormhole.VM memory vm) private view returns (bool valid, string memory reason) {
         if (vm.emitterChainId != pyth2WormholeChainId()) {
-            return false;
+            return (false, "Invalid Chain ID");
```
drozdziak1, line 80: This was a bit annoying to deal with when writing the test, and the existing error message was outdated anyway

ethereum/test/pyth.js
```diff
@@ -362,6 +362,60 @@ contract("Pyth", function () {
...
+        let wrongChainIdFails = false;
+
+        try {
+            // Should result in expected failure
+            await updatePriceFeeds(this.pythProxy, [rawBatch]);
+        } catch (e) {
+            wrongChainIdFails = true;
+        }
+
+        assert.equal(wrongChainIdFails, true);
```
ali-bahjati, line 383: I'm not sure what is the assert library here, but don't we have `assert.throws` or sth similar to make it more clear?

ali-bahjati (approved):

drozdziak1:
This change is similar to #228 except for the wormhole on-chain field. We can revisit this if we think there's going to be a change to Wormhole's address

PR #225, opened 2022-06-22
Add image push ci

ali-bahjati:


.github/workflows/image-push.yml
```diff
@@ -0,0 +1,29 @@
...
+          aws-region: eu-west-2
+      - uses: aws-actions/amazon-ecr-login@v1
+        id: ecr_login
+      - run: |
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile.wasm -o type=local,dest=. .
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f third_party/pyth/price-service/Dockerfile.price_service .
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
+        env:
+          ECR_REGISTRY: ${{ steps.ecr_login.outputs.registry }}
+          ECR_REPOSITORY: price-service
```
thmzlt, line 28: Can you set this to `pyth-price` so we can reuse the repository that has already been created? We can then later revisit the names for the three p2w repositories.

thmzlt (changes requested):

.github/workflows/image-push.yml
ali-bahjati: Ah you are right. i thought it was price-service in first place :D fixed.

thmzlt (approved):

PR #226, opened 2022-06-22
Price Service Multisource: Use attestationTime instead of seqnum

ali-bahjati:
SeqNum is not suitable for multiple sources, attestationTime is good
but we should be careful of solana clock falling behind

Also use this time for freshness metric because it's more useful.


drozdziak1 (approved):
lgtm! I think attestation time is a good choice here

PR #227, opened 2022-06-23
Fix wrong branch names

ali-bahjati:


thmzlt:
Looks like there's an issue with a command from the Dockerfile?

thmzlt (approved):
Docker build is failing, but looks good otherwise.

PR #229, opened 2022-06-24
Abehjati/pyth-evm-contract-fire-event

ali-bahjati:
It also fixes a gas issue in evm relayer and deletes an unused open zeppelin file (for a test network)


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -39,6 +39,7 @@ contract Pyth is PythGetters, PythSetters, AbstractPyth {
 
             if(attestation.attestationTime > latestPrice.attestationTime) {
                 setLatestPriceInfo(attestation.priceId, newPriceInfo(attestation));
+                emit PriceUpdate(attestation.priceId, attestation.publishTime);
```
tompntn, line 42: Do we need to declare the `PriceUpdate` event anywhere?

ethereum/contracts/pyth/Pyth.sol
```diff
@@ -39,6 +39,7 @@ contract Pyth is PythGetters, PythSetters, AbstractPyth {
...
             }
```
tompntn, line 43: Could we add a test for this?

tompntn (commented):
Nice!

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: It's in the solidity sdk.

drozdziak1, line 42: Small nit: I noticed that you had to raise the gas limit, which seems like the event might cost a fair bit. Do we need the publish time inside the event? If we're already giving them the price ID, consumers can look the rest of the metadata up, which could make this method cheaper for relayers. My thinking is that smaller events may be cheaper (just a hunch).

ali-bahjati: I am not sure it was working before or not. It is ok on 4 price feeds and not ok on 5 price feeds. I will try it out and will let you know.

The rationale behind having publish time is to make monitoring easier for us (specially I like to compare publish time vs block time)

drozdziak1: That makes sense, we can always come back to optimize this.

ali-bahjati: I just looked back and it doesn't work on 5 symbols batch in main branch. I also checked with an older change of switching from `updatePriceBatchFromVm` to `updatePriceFeeds` and it wasn't still working there. So I think this change to 200k is legit here.

tompntn (approved):

PR #230, opened 2022-06-29
Use single server for both ws and http

ali-bahjati:


third_party/pyth/price-service/src/rest.ts
```diff
@@ -126,7 +127,7 @@ export class RestAPI {
       }
     );
     endpoints.push(
-      "latest_vaas?ids[]=<price_feed_id>&ids[]=<price_feed_id_2>&.."
+      "api/latest_vaas?ids[]=<price_feed_id>&ids[]=<price_feed_id_2>&.."
```
thmzlt, line 130: These new endpoints with the `api/` prefix, it looks like the consumers on the other side will need to be aware of the change. Just wanted to make sure that this is being taken in account.

thmzlt:
I'm not familiar with the p2w codebase, but this looks good. I'll leave it to another reviewer to give explicit approval.

jayantk (approved):
nice

third_party/pyth/price-service/src/rest.ts
ali-bahjati: Yes, the next PR fixes them.

PR #235, opened 2022-07-12
Remove unnecessary payable marker from wormhole address

tompntn:
The Wormhole address is not designed to receive payment and we do not need to pay it.


ali-bahjati (approved):
The only reason that I can think of its existence is wormhole possibly charging us in the future (for their verification). But that is very unlikely as anyone can verify them.

Probably it was there because protocols that interact with wormhole to post a message should pay a small amount of fee for posting messages which is not our case.

PR #236, opened 2022-07-13
guardiand: Bump wormhole sources

drozdziak1:
Before this change, a fresh build from scratch would cause a guardian
software compilation error ("Failure: empty digest") that would
normally be hidden by caching in CI.


ali-bahjati (approved):
What do you mean by fresh build?

I think it should work fine with wormhole releases. But problem is that we can't access newer codes that are not released. So it is fine for now. 

In future I think it's better to stick to versions rather than commit hashes.

PR #241, opened 2022-07-25
Implement the new IPyth events and methods

ali-bahjati:
Changes:
- Emits multiple events that capture price updates.
- Adds update fee, with setter and getter and the check. Also makes `updatePriceBatchFromVm` private.

p.s: I tried to separate the changes but it was difficult as the new sdk version contained both changes.

### DIFF
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -24,7 +24,7 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
         setPyth2WormholeEmitter(pyth2WormholeEmitter);
     }
 
-    function updatePriceBatchFromVm(bytes memory encodedVm) public returns (PythInternalStructs.BatchPriceAttestation memory bpa) {
+    function updatePriceBatchFromVm(bytes memory encodedVm) private returns (PythInternalStructs.BatchPriceAttestation memory bpa) {
         (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(encodedVm);
 
         require(valid, reason);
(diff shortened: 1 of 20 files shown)


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -24,32 +24,51 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
...
+        emit BatchPriceFeedUpdate(vm.emitterChainId, vm.sequence, batch.attestations.length, freshPrices);
+
         return batch;
     }
 
-    function updatePriceFeeds(bytes[] calldata updateData) public override {
+    function updatePriceFeeds(bytes[] calldata updateData) public override payable {
+        uint requiredFee = getUpdateFee(updateData.length);
+        require(msg.value >= requiredFee, "Insufficient paid fee amount");
+        payable(msg.sender).transfer(msg.value - requiredFee);
```
jayantk, line 61: I don't think you need this. I think `msg.value` gets transferred to this contract automatically when the payable function is called.

https://solidity-by-example.org/payable/

ethereum/test/pyth.js
```diff
@@ -132,6 +135,38 @@ contract("Pyth", function () {
...
+
+        // Check initial fee is zero
+        assert.equal(await this.pythProxy.singleUpdateFeeInWei(), 0);
+
+        // Checks setting fee using another account reverts.
+        await expectRevert(
+            this.pythProxy.updateSingleUpdateFeeInWei(10, {from: accounts[1]}),
+            notOwnerError,
+        );
+    });
```
jayantk, line 168: 👍 good test

ethereum/test/pyth.js
```diff
@@ -260,23 +295,85 @@ contract("Pyth", function () {
...
+        // When a smaller fee is payed it reverts
+        await expectRevert(
+            updatePriceFeeds(this.pythProxy, [rawBatch1, rawBatch2], feeInWei - 1), 
+            insufficientFeeError
+        );
+
+        const receipt = await updatePriceFeeds(this.pythProxy, [rawBatch1, rawBatch2], feeInWei);
+        expectEvent(receipt, 'UpdatePriceFeeds', {
+            fee: feeInWei
+        });
```
jayantk, line 376: can you add a test of doing the update that validates that the contract balance is updated properly?

jayantk (commented):
nice, just one question here about the right way to accept payment in eth. see inline

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: This is returning the exceeded value back to the sender.

ethereum/test/pyth.js
ali-bahjati: Good idea. Will do.

ethereum/test/pyth.js
```diff
@@ -260,23 +295,141 @@ contract("Pyth", function () {
...
+
+        let ts = 1647273460;
+        let rawBatch1 = generateRawBatchAttestation(ts - 5, ts, 1337);
+        let rawBatch2 = generateRawBatchAttestation(ts + 5, ts + 10, 1338);
+
+        // Getting the fee from the contract
+        let feeInWei = await this.pythProxy.getUpdateFee(2);
+        assert.equal(feeInWei, 20);
+
+        const receipt = await updatePriceFeeds(this.pythProxy, [rawBatch1, rawBatch2], feeInWei + 10);
```
jayantk, line 427: good test

jayantk (approved):
very nicely done

ethereum/contracts/pyth/Pyth.sol
```diff
@@ -24,32 +24,51 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
...
+                fresh = true;
                 setLatestPriceInfo(attestation.priceId, newPriceInfo(attestation));
-                emit PriceUpdate(attestation.priceId, attestation.publishTime);
             }
+
+            emit PriceFeedUpdate(attestation.priceId, fresh, vm.emitterChainId, vm.sequence, latestPrice.priceFeed.publishTime,
+                attestation.publishTime, attestation.price, attestation.conf);
         }
 
+        emit BatchPriceFeedUpdate(vm.emitterChainId, vm.sequence, batch.attestations.length, freshPrices);
```
drozdziak1, line 53: I think there's a lot of overlap between these two event types. Is there an important difference between them that prevents us from emitting a single bigger event?

drozdziak1 (commented):
Nice! Just a small question around feed update event structure

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: The overlap helps me to match the single price update to its corresponding batch. 

One idea could be have only single event for all of the batch info but I think there are some people that are interested in single price feed.

ethereum/contracts/pyth/PythState.sol
```diff
@@ -21,6 +21,8 @@ contract PythStorage {
         // (chainId, emitterAddress) => isValid; takes advantage of
         // constant-time mapping lookup for VM verification
         mapping(bytes32 => bool) isValidDataSource;
+
+        uint singleUpdateFeeInWei;
```
tompntn, line 25: Do we want to add `singleUpdateFeeInWei` to `Pyth.sol`'s `initialize` function, so that when we deploy on new chains we can initialize it atomically with the other state variables? I don't think it's strictly necessary but it's nice to treat all the state variables equally.

ali-bahjati: This way is actually contradicts with truffle migration philosophy and can cause inconsistent updates. We are not setting any value here, but if any value should be set, it should be set during the same migration step that the deployed one will go for. 

Once we go for governance, the sequence of governance messages will do that.

PR #246, opened 2022-07-29
Add pythnet migration files

ali-bahjati:
Also:
- Add related env vars for all networks.
- Rename Pyth2Wormhole env vars to solana


jayantk (approved):

ethereum/.env.template
```diff
@@ -15,8 +15,10 @@ INIT_GOV_CONTRACT=                # 0x000000000000000000000000000000000000000000
 
 # Pyth Migrations           # Example Format
 WORMHOLE_BRIDGE_ADDRESS     # 0x68605AD7b15c732a30b1BbC62BE8F2A509D74b4D (only if wormhole exists)
-PYTH_TO_WORMHOLE_CHAIN_ID=  # 0x1
-PYTH_TO_WORMHOLE_EMITTER=   # 8fuAZUxHecYLMC76ZNjYzwRybUiDv9LhkRQsAccEykLr
+SOLANA_CHAIN_ID=  # 0x1
+SOLANA_EMITTER=   # 0x8fuAZUxHecYLMC76ZNjYzwRybUiDv9LhkRQsAccEykLr
```
jayantk, line 19: this is the wrong kind of address

jayantk (approved):
LGTM, though I didn't check the addresses for the emitters.

Also, i really dislike the amount of duplication across these .env files. it's probably worthwhile to come up with a less repetitive way to do this configuration in the future. (.env is actually a really terrible config format because it's super inflexible. If we used something more powerful like HOCON, that has simple solutions to exactly this kind of problem)

ali-bahjati:
I agree, the env vars are not a good option here.. we should definitely revisit it.

PR #249, opened 2022-08-11
Make validTimePeriod configurable

ali-bahjati:
I will not deploy this change to testnet until we know the proper value, and that mean we can measure time from a pyth published to a client subscribed to the price service getting that price. 

The ideal value is derived like this:

```
2*price_update_frequency + e2e(price published, a price service client receives it) + 2*target_network_latency + target_network_block_time
```

In solana we have a ~12 seconds validity time threshold which was set for the congestion periods. We might also consider setting in acceptable minimum here, I can change the contract to instead of being initialized with 0, be initialized with that value.


ethereum/.env.prod.aurora
```diff
@@ -5,3 +5,5 @@ MIGRATIONS_NETWORK=aurora
 WORMHOLE_BRIDGE_ADDRESS=0xa321448d90d4e5b0A732867c18eA198e75CAC48E
 PYTH_TO_WORMHOLE_CHAIN_ID=0x1
 PYTH_TO_WORMHOLE_EMITTER=0x6bb14509a612f01fbbc4cffeebd4bbfb492a86df717ebe92eb6df432a3f00a25
+
+VALID_TIME_PERIOD_SECONDS=15
```
jayantk, line 9: Can we set the initial time periods to err on the high side? I think some of these thresholds are likely too aggressive -- would suggest doubling all of them as a baseline, and using 60 seconds as a minimum. (Note that your calculation also doesn't account for clock drift between the chains, which could be substantial)

I suggest this because we are currently not in a position to measure how many failures occur because the threshold is too aggressive. Without that measurement, we could end up in a situation where Pyth is extremely flaky on a target chain, and we wouldn't know it. We can optimize these thresholds later once our measurement infrastructure is in place.

jayantk (commented):
change is generally fine. I think the configured validity times are likely too low though (see inline)

jayantk:
Also, no need to overcomplicate this by putting in a minimum validity threshold. We can guard against that in code that's easier for us to change, especially since we don't know the right minimum.

ethereum/contracts/pyth/PythState.sol
```diff
@@ -23,6 +23,10 @@ contract PythStorage {
         mapping(bytes32 => bool) isValidDataSource;
 
         uint singleUpdateFeeInWei;
+        
+        /// Maximum acceptable time period before price is considered to be stale.
+        /// This includes attestation delay and block time.
```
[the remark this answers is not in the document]
ali-bahjati: I don't think we need to consider it given 1) solana made a patch to solve it + 2) we are gonna use pythnet. 
It seems that evm chains have reliable utc time.

ethereum/test/pyth.js
```diff
@@ -44,6 +45,9 @@ contract("Pyth", function () {
             testPyth2WormholeChainId,
             testPyth2WormholeEmitter
         );
+
+        // Setting the validity time to 10 seconds
```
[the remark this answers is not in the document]
ali-bahjati: Thanks.

ethereum/test/pyth.js
```diff
@@ -157,16 +161,48 @@ contract("Pyth", function () {
...
+    });
+
+    it("should not allow updating validTimePeriodSeconds by another account", async function () {
+        // Check that the owner is the default account Truffle
+        // has configured for the network.
+        const accounts = await web3.eth.getAccounts();
+        const defaultAccount = accounts[0];
+        assert.equal(await this.pythProxy.owner(), defaultAccount);
+
+        // Check initial fee is zero
```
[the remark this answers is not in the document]
ali-bahjati: +1

ethereum/.env.prod.aurora
ali-bahjati: These are my initial estimates based on the block time of the given chain. But we can measure it once we deploy attester in Pythnet. 

I shared my opinion about clock drift in the above thread too. I don't think it be an issue.

ethereum/contracts/pyth/PythState.sol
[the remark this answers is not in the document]
ali-bahjati: Added though.

ethereum/.env.prod.aurora
ali-bahjati: I applied your suggestion.

ethereum/contracts/pyth/Pyth.sol
```diff
@@ -218,7 +218,8 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
 
         // Check that there is not a significant difference between this chain's time
         // and the price publish time.
-        if (diff(block.timestamp, info.priceFeed.publishTime) > validTimePeriodSeconds()) {
+        if (info.priceFeed.status == PythStructs.PriceStatus.TRADING &&
```
jayantk, line 221: is this trading check right? what if we add a new status flag in the future -- do we not want to change that status to unknown?

I think dropping this check is probably better.

jayantk (commented):
not sure about this trading status change, but LGTM aside from that

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: I think when we add a new status we will probably handle all cases at that time 😅

jayantk (approved):

PR #251, opened 2022-08-17
Abehjati/release-price-service

ali-bahjati:
Release price service and removes guardian image and uses wormhole published image for "guardian node" and "spy"


jayantk (commented):
i'm going to let @thmzlt review this because i'm not super up to date on all this docker stuff.

.github/workflows/publish-dockerhub.yml
```diff
@@ -0,0 +1,40 @@
...
+  release:
+    types:
+      - created
+env:
+  DOCKER_HUB: docker.io
+  DOCKER_USER: ${{ secrets.DOCKER_IO_USER }}
+jobs:
+  publish-pyth-price-service:
+    name: Publish Pyth price service
+    if: ${{ startsWith(github.ref, 'refs/tags/pyth-price-service-v') }}
```
thmzlt, line 12: If we are conditioning this to a tag prefix, I think it might be easier to just use that as the workflow triggering condition, like we do [here](https://github.com/pyth-network/publisher-exporter/blob/main/.github/workflows/ci.yaml#L6-L7).

thmzlt:
Also, I thought we were going to use the ECR repositories (xc-attest, xc-spy, xc-server) that we created last week?

.github/workflows/publish-dockerhub.yml
ali-bahjati: Oh I didn't know this is possible, I'll change it this way.

thmzlt:
The new ECR repos for xc-attest and xc-server:
- public.ecr.aws/pyth-network/xc-attest
- public.ecr.aws/pyth-network/xc-server

I think the `ecr_login` action doesn't work too well with public ECR repos so I ended up using the regular [`docker-login` action in program_admin](https://github.com/pyth-network/program-admin/blob/main/.github/workflows/ci.yaml#L40-L44). All public ECR repos must be on the us-east-1 region.

You can look at the [program-admin build-and-push-ecr job](https://github.com/pyth-network/program-admin/blob/main/.github/workflows/ci.yaml#L28) for a full example of how to push to a public ECR repo. It's not all that different.

.github/workflows/price-service-image-push.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          registry: public.ecr.aws
+        env:
+          AWS_REGION: us-east-1
+      - run: |
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile.wasm -o type=local,dest=. .
+          DOCKER_BUILDKIT=1 docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f third_party/pyth/price-service/Dockerfile.price_service .
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
+        env:
+          ECR_REGISTRY: public.ecr.aws
```
thmzlt, line 44: I'm not sure if the "pyth-network" is part of the registry name or the repository name. (It makes no difference here because we just concatenate the two to form the URL.)

thmzlt (approved):

.github/workflows/price-service-image-push.yml
ali-bahjati: I think it's none 😓 but I think its fine here.

PR #252, opened 2022-08-17
Abehjati/update-eth-contract-prev-price

ali-bahjati:
Strangely, although `diff` in `AbstractPyth` is private and its children cannot use it, they cannot even define a function with the same name. It's probably because function signature (or its hash) is mapped to its source instruction in Solidity.


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -223,19 +223,22 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
 
         // Check that there is not a significant difference between this chain's time
         // and the price publish time.
-        if (diff(block.timestamp, info.priceFeed.publishTime) > VALID_TIME_PERIOD_SECS) {
+        if (info.priceFeed.status == PythStructs.PriceStatus.TRADING &&
```
jayantk, line 226: maybe leave a comment explaining this logic

jayantk (approved):

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: Done.

PR #257, opened 2022-08-24
Eth contract fix memory + new sdk + refactor

ali-bahjati:
- Change some memory modifiers to improve gas efficiency
- Implement getValidTimePeriod() and remove old staleness logic
- Update the tests

### DIFF
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -226,4 +226,8 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
     function getValidTimePeriod() public override view returns (uint) {
         return validTimePeriodSeconds();
     }
+
+    function version() public pure returns (string memory) {
+        return "0.1.0";
+    }
 }
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -231,7 +231,9 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
         return "0.1.0";
     }
 
-    function deployCommitHash() public pure returns (string memory) {
-        return "DEPLOY_COMMIT_HASH_PLACEHOLDER";
+    function deployCommitHash() public pure returns (bytes20) {
+        // This is a place holder for the commit hash and will be replaced
+        // with the commit hash upon deployment.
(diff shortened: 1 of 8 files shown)


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -211,34 +211,23 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
...
-            // unknown. So, now that status is being set to unknown,
-            // we should move the current price to the previous
-            // price to ensure getLatestAvailablePrice* works
-            // as intended.
-            info.priceFeed.prevPrice = info.priceFeed.price;
-            info.priceFeed.prevConf = info.priceFeed.conf;
-            info.priceFeed.prevPublishTime = info.priceFeed.publishTime;
-        }
-
         return info.priceFeed;
```
tompntn, line 218: I wonder if we should add `getCurrentPriceNoOlderThan(bytes32 id, uint64 maxAge)` which fails if the cached time is older than `maxAge`.

tompntn (approved):

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: We already have that, `getLatestAvailablePriceWithinDuration`:

https://github.com/pyth-network/pyth-sdk-solidity/blob/main/IPyth.sol#L66

Probably your name looks better though 😅 that was what I and Jayant could think of 😬

tompntn (approved):

ali-bahjati:
I have used a placeholder bytes and i'm replacing it within the deployment process I replace the placeholder with the hash in the contract binary.. So it's not part of the contract and `return hex"123123..."` doesn't look quite clean.

@tompntn raised it and suggested to use a storage variable instead, and a method like `setGitCommitHash()` to set as the migration call in the `upgradeProxy`. However, I think it's too much to have it as a migration call. We are likely to use migrate functions and then in each of them we should make sure this hash is set correctly. Also, the nature of this hash is not a variable to be set. In the placeholder version it's within the implementation code and won't change (tied to implementation). In the storage variable it's in the proxy storage and will change every time.

I decided to move on with the existing placeholder approach (although it doesn't look nice), I am not particularly happy with it. We definitely can revisit it. We can change to migration call whenever needed in the future too and we just need to change the underlying value of `deployCommitHash`.

Posted a summary here for future reference when we get back to it.

PR #258, opened 2022-08-24
Start depending on the upstream out-of-tree Wormhole

drozdziak1:
When finished, this will let us drop the in-tree copy of bridge and solitaire sources in favor of verbatim upstream sources downloaded from `certusone/wormhole`


devnet/node.yaml
```diff
@@ -89,18 +87,40 @@ spec:
...
+            # - http://algorand:4001
+            # - --algorandAlgodToken
+            # - "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
             - --solanaContract
             - Bridge1p5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o
-            - --solanaWS
-            - ws://solana-devnet:8900
-            - --solanaRPC
-            - http://solana-devnet:8899
+            # - --solanaWS
```
ali-bahjati, line 120: Why are these commented? 🤔

solana/pyth2wormhole/program/src/migrate.rs
```diff
@@ -64,6 +59,18 @@ pub fn migrate(ctx: &ExecutionContext, accs: &mut Migrate, data: ()) -> SoliResu
         ));
     }
 
+    if *accs.system_program.key != system_program::id() {
```
ali-bahjati, line 62: Where is it used? by `create` method below?

devnet/node.yaml
drozdziak1: I copied most of this file verbatim from upstream, I figured that if I only keep the changes minimal, it should make catching it up to wormhole easier in the future. Not feeling strongly about this, just looking out for future us

solana/pyth2wormhole/program/src/migrate.rs
drozdziak1: correct, we were using a solitaire-client thingy for this previously, which would include the system program automatically for accounts that will possibly need it for creation. Latest solitaire is on-chain focused.

PR #261, opened 2022-08-30
Wormhole message reuse via post_message_unreliable

drozdziak1:
this changeset switches us to a PDA scheme for wormhole messages, accounts for reusing those accounts and conditionally switches between post_message/post_message_unreliable depending on whether a given message account was already initialized.

### DIFF
--- a/solana/pyth2wormhole/client/src/message.rs
+++ b/solana/pyth2wormhole/client/src/message.rs
@@ -1,0 +1,67 @@
+//! Re-usable message scheme for pyth2wormhole
+
+use solana_program::system_instruction;
+use std::time::Instant;
+use std::time::Duration;
+
+use crate::ErrBox;
+
+/// One of the accounts tracked by the attestation client.
+#[derive(Clone, Debug)]
+pub struct P2WMessageAccount {
+    /// Unique ID that lets us derive unique accounts for use on-chain
+    pub id: u64,
+    /// Last time we've posted a message to wormhole with this account
+    pub last_used: Instant,
+}
+
+/// An umbrella data structure for tracking all message accounts in use
+#[derive(Clone, Debug)]
+pub struct P2WMessageIndex {
+    /// The tracked accounts
+    accounts: Vec<P2WMessageAccount>,
+    /// How much time needs to pass between reuses
+    grace_period: Duration,
+}
+
+impl P2WMessageIndex {
+    pub fn new(grace_period: Duration) -> Self {
+        Self {
+            accounts: Vec::new(),
+            grace_period
+        }
+    }
+    /// Finds an account with last_used at least grace_period in the past
+    pub fn get_account(&mut self) -> P2WMessageAccount {
+        let mut ret = None;
+        // Look for the first account available
+        for (idx, acc) in self.accounts.iter().enumerate() {
+            if acc.last_used.elapsed() >= self.grace_period {
+                // Note: Don't confuse idx (the LinkedList position)
+                // with the id field on P2WMessageAccount.
+                ret = Some((idx, acc.clone()));
+                break;
+            }
+        }
+        // Found a good account, no need to add one
+        if let Some((idx,mut acc)) = ret {
+            self.accounts.remove(idx);
+
+            // Update last_used
+            acc.last_used = Instant::now();
+            self.accounts.push(acc.clone());
+
+            return acc;
+        // All accounts were used recently, add a new one
+        } else {
+            let next_id = self.accounts.len() as u64;
+            let new = P2WMessageAccount {
+                id: next_id,
+                last_used: Instant::now()
+            };
+            
+            self.accounts.push(new.clone());
+            return new;
+        }
+    }
+}
--- a/solana/pyth2wormhole/client/src/message.rs
+++ b/solana/pyth2wormhole/client/src/message.rs
@@ -1,5 +1,6 @@
 //! Re-usable message scheme for pyth2wormhole
 
+use log::debug;
 use solana_program::system_instruction;
 use std::{
     collections::VecDeque,
@@ -22,14 +23,14 @@ pub struct P2WMessageAccount {
 
 /// An umbrella data structure for tracking all message accounts in use
 #[derive(Clone, Debug)]
-pub struct P2WMessageIndex {
+pub struct P2WMessageQueue {
     /// The tracked accounts. Sorted from oldest to newest, as guaranteed by get_account()
     accounts: VecDeque<P2WMessageAccount>,
     /// How much time needs to pass between reuses
     grace_period: Duration,
 }
 
-impl P2WMessageIndex {
+impl P2WMessageQueue {
     pub fn new(grace_period: Duration) -> Self {
         Self {
             accounts: VecDeque::new(),
@@ -53,20 +54,71 @@ impl P2WMessageIndex {
                 // and will be old enough eventually.
                 self.accounts.push_front(existing_too_new_acc);
 
+                debug!(
+                    "Increasing message queue size to {}",
+                    self.accounts.len() + 1
+                );
+
                 // Use a new account instead
                 P2WMessageAccount {
                     id: self.accounts.len() as u64,
                     last_used: Instant::now(),
                 }
             }
-            // Base case: Index is empty, use a new account
+            // Base case: Queue is empty, use a new account
             None => P2WMessageAccount {
                 id: self.accounts.len() as u64,
                 last_used: Instant::now(),
             },
         };
-        // The chosen account becomes the newest, push it to the very end. 
+        // The chosen account becomes the newest, push it to the very end.
         self.accounts.push_back(acc.clone());
         return acc;
+    }
+}
+
+pub mod test {
+    use super::*;
+
+    #[test]
+    fn test_empty_grows_only_as_needed() {
+        let mut q = P2WMessageQueue::new(Duration::from_millis(500));
+
+        // Empty -> 1 account
+        let acc = q.get_account();
+
+        assert_eq!(q.accounts.len(), 1);
+        assert_eq!(acc.id, 0);
+
+        // 1 -> 2 accounts, not enough time passes
+        let acc2 = q.get_account();
+
+        assert_eq!(q.accounts.len(), 2);
+        assert_eq!(acc2.id, 1);
+
+        std::thread::sleep(Duration::from_millis(600));
+
+        // Account 0 should be in front, enough time passed 
+        let acc3 = q.get_account();
+
+        assert_eq!(q.accounts.len(), 2);
+        assert_eq!(acc3.id, 0);
+
+        // Account 1 also qualifies
+        let acc4 = q.get_account();
+
+        assert_eq!(q.accounts.len(), 2);
+        assert_eq!(acc4.id, 1);
+
+        // 2 -> 3 accounts, not enough time passes
+        let acc5 = q.get_account();
+
+        assert_eq!(q.accounts.len(), 3);
+        assert_eq!(acc5.id, 2);
+
+        // We should end up with 0, 1 and 2 in order
+        assert_eq!(q.accounts[0].id, 0);
+        assert_eq!(q.accounts[1].id, 1);
+        assert_eq!(q.accounts[2].id, 2);
     }
 }
--- a/solana/pyth2wormhole/client/src/message.rs
+++ b/solana/pyth2wormhole/client/src/message.rs
@@ -10,7 +10,7 @@ use std::{
     },
 };
 
-use crate::ErrBox;
+use crate::ErrBoxSend;
 
 /// One of the accounts tracked by the attestation client.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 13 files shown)


solana/pyth2wormhole/client/src/message.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct P2WMessageAccount {
+    /// Unique ID that lets us derive unique accounts for use on-chain
+    pub id: u64,
+    /// Last time we've posted a message to wormhole with this account
+    pub last_used: Instant,
+}
+
+/// An umbrella data structure for tracking all message accounts in use
+#[derive(Clone, Debug)]
+pub struct P2WMessageIndex {
```
jayantk, line 25: the name of this struct is confusing because `index` could mean a number, and each account has a numeric id. i suggest `P2WMessageAccountQueue` ?

jayantk (approved):
this looks great to me. I'd like to get another pair of eyes on it before merging for security's sake, but as far as I can tell looks really solid.

Agree with your slack comment about making the reuse duration configurable, but that can be done in a separate PR.

solana/pyth2wormhole/client/tests/test_migrate.rs
```diff
@@ -174,16 +170,16 @@ async fn test_migrate_already_migrated() -> Result<(), solitaire::ErrBox> {
...
-        p2w_program_id,
-        p2w_owner,
-        ctx.last_blockhash,
-    )?;
+    let migrate_tx =
+        p2wc::gen_migrate_tx(ctx.payer, p2w_program_id, p2w_owner, ctx.last_blockhash)?;
     info!("Before process_transaction");
 
     // Migration should fail because the new config account is already initialized
-    assert!(ctx.banks_client.process_transaction(migrate_tx).await.is_err());
```
jayantk, line 186: i take it we don't have clippy / pre-commit here which is why there's all this formatting noise?

we should install clippy. it's annoying to review PRs with these format changes mixed in.

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,15 +262,30 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
     );
     solana_program::program::invoke(&transfer_ix, ctx.accounts)?;
 
+    let payload = batch_attestation.serialize().map_err(|e| {
+        trace!(&e.to_string());
+        ProgramError::InvalidAccountData
+    })?;
+
+    // Adjust message account size if necessary
+    if !accs.wh_message.is_initialized() && accs.wh_message.payload.len() != payload.len() {
```
ali-bahjati: does realloc make sure that the account will remain rent exempt? 

I'm thinking that the size might be different if we have batches of different size and it can cause doing resizes more than necessary. 


minor: do we need to handle resizes? Can't we use a version field in the instruction input and `P2WMessageDrvData`? Or maybe a "{size}" as a seed?

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,38 +262,118 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+            // Wormhole message type is constant-size).
+            // positive -> grow,
+            // negative -> shrink,
+            let diff = payload.len() as i64 - accs.wh_message.payload.len() as i64;
+            accs.wh_message.0 .0.realloc(
+                (accs.wh_message.payload.len() as i64 + diff) as usize,
+                false,
+            )?;
+        }
+        trace!("After message size adjustment");
```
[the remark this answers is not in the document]
ali-bahjati: I don't know whether it was for local debug or not. But these `trace!` macros won't do anything on releases (it's only for debug).

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -268,47 +268,112 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
-            AccountMeta::new(*accs.payer.key, true),
-            AccountMeta::new(*accs.wh_fee_collector.key, false),
-            AccountMeta::new_readonly(*accs.clock.info().key, false),
-            AccountMeta::new_readonly(solana_program::sysvar::rent::ID, false),
-            AccountMeta::new_readonly(solana_program::system_program::id(), false),
-        ],
-    );
-
-    trace!("Before cross-call");
+    let ix = if accs.wh_message.is_initialized() {
```
ali-bahjati, line 271: I couldn't understand that why we are using reliable message post when account is not initialized. I think unreliable_post_message also creates an account if needed:
https://github.com/wormhole-foundation/wormhole/blob/dev.v2/solana/bridge/program/src/api/post_message.rs#L160

ali-bahjati (approved):
Very nice work 🔥 

I have one concern about doing more resizes than necessary that I asked, apart from that lgtm.

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,15 +262,30 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+    if !accs.wh_message.is_initialized() && accs.wh_message.payload.len() != payload.len() {
+        // Learn how much we're adjusting (Note: assuming surrounding
+        // Wormhole message type is constant-size).
+        // positive -> grow,
+        // negative -> shrink,
+        let diff = payload.len() as i64 - accs.wh_message.payload.len() as i64;
+        accs.wh_message.0 .0 .0.realloc(
+            (accs.wh_message.payload.len() as i64 + diff) as usize,
+            false,
+        )?;
```
tompntn: Am I missing something (likely!) or is this equivalent to just doing `accs.wh_message.0 .0 .0.realloc(payload.len(), false)`?

solana/pyth2wormhole/program/src/attest.rs
tompntn: Curious - why do we only resize if the account is un-initialized? Does the account ever get reset back to un-initialized, or do we only ever resize once?

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,38 +262,118 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+
+    // Adjust message account size if necessary
+    let ix = if accs.wh_message.is_initialized() {
+        if accs.wh_message.payload.len() != payload.len() {
+            // Learn how much we're adjusting (Note: assuming surrounding
+            // Wormhole message type is constant-size).
+            // positive -> grow,
+            // negative -> shrink,
+            let diff = payload.len() as i64 - accs.wh_message.payload.len() as i64;
+            accs.wh_message.0 .0.realloc(
```
Reisen, line 278:
```suggestion
            accs.wh_message.info().realloc(
```

solana/pyth2wormhole/client/src/message.rs
```diff
@@ -0,0 +1,67 @@
...
+        // Found a good account, no need to add one
+        if let Some((idx,mut acc)) = ret {
+            self.accounts.remove(idx);
+
+            // Update last_used
+            acc.last_used = Instant::now();
+            self.accounts.push(acc.clone());
+
+            return acc;
+        // All accounts were used recently, add a new one
```
tompntn: Would it be a good idea to set an upper bound on the number of accounts we can create, as a safety net in case a bug causes a huge number of accounts to be created?

solana/pyth2wormhole/client/src/message.rs
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+            let next_id = self.accounts.len() as u64;
```
tompntn: Should we check here that the ID hasn't overflowed? Unlikely but paranoid 😉

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,38 +262,118 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+            PostMessageData {
+                nonce: 0, // Superseded by the sequence number
+                payload,
+                consistency_level: data.consistency_level,
+            },
+        );
+
+        trace!("Before unreliable cross-call");
+
+        Instruction::new_with_bytes(
```
Reisen, line 297: Is there a reason `bridge::instructions::post_message_unreliable` can’t be used over manual construction? Same question regarding `post_message` below (though also curious about Ali’s question above, could this just be a call to unreliable?)

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,38 +262,118 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+            accs.config.wh_prog,
+            post_message_unreliable_data.try_to_vec()?.as_slice(),
+            vec![
+                AccountMeta::new(*accs.wh_bridge.key, false),
+                AccountMeta::new(*(accs.wh_message.0).0.key, true),
```
Reisen, line 302:
```suggestion
                AccountMeta::new(*accs.wh_message.info().key, true),
```

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -248,38 +262,118 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+            },
+        );
+
+        trace!("Before reliable cross-call");
+        Instruction::new_with_bytes(
+            accs.config.wh_prog,
+            post_message_data.try_to_vec()?.as_slice(),
+            vec![
+                AccountMeta::new(*accs.wh_bridge.key, false),
+                AccountMeta::new(*(accs.wh_message.0).0.key, true),
```
Reisen, line 328:
```suggestion
                AccountMeta::new(*accs.wh_message.info().key, true),
```

Reisen (changes requested):

solana/pyth2wormhole/program/src/attest.rs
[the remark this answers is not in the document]
drozdziak1: Yeah, The trace!-s evaluate to nothing without the `trace` feature and can speed up debugging

solana/pyth2wormhole/client/tests/test_migrate.rs
drozdziak1: I will try to undo the formatting-related changes. Sorry about that!

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: I don't think it takes care of rent exemptions. nice catch! I think wormhole would start complaining the next time we expand the format/symbol count per batch, which could become a nasty debug session. Thanks!

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: Nice, this simplifies the critical path. Thanks!

solana/pyth2wormhole/client/src/message.rs
drozdziak1: I think we'll go out of memory many orders of magnitude before this counter gets wrecked

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: It could be just unreliable and I probably could use the function. Thanks!

solana/pyth2wormhole/client/src/message.rs
tompntn: I'm more concerned with the downstream effects on the Wormhole guardians running out of space than any sort of memory issues with the attester.

solana/pyth2wormhole/client/src/message.rs
```diff
@@ -53,20 +54,71 @@ impl P2WMessageIndex {
...
             },
         };
-        // The chosen account becomes the newest, push it to the very end. 
+        // The chosen account becomes the newest, push it to the very end.
         self.accounts.push_back(acc.clone());
         return acc;
     }
 }
+
+pub mod test {
```
ali-bahjati, line 80: very nice ser 🙇

solana/pyth2wormhole/client/src/message.rs
drozdziak1: Done

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: Added a routine that makes sure exemption is maintained.

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: Done, this looks so much nicer now!

solana/pyth2wormhole/client/src/message.rs
drozdziak1: Your comment nerdsniped me real good. Ultimately I don't want people waking up on-call over a bullshit human error either. I added a pretty good test and will add the limit.

Pragmatism: 1
Me: 0

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: Done!

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: removed in favor of the upstream instruction generator

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: Done!

solana/pyth2wormhole/client/src/message.rs
drozdziak1: Thanks!

solana/pyth2wormhole/client/src/message.rs
tompntn: Thank-you!

Reisen (approved):

PR #262, opened 2022-08-31
Refactor our CW contract + cleanups

ali-bahjati:
This PR consists of small changes that are mostly mechanical, please review it commit-by-commit.

This PR will:
- Use Pyth CW SDK
- Fixe clippy warnings
- Use Addr instead of strings
- Add `error.rs` to rely on instead of using wormhole errors.
- Format the code
- Remove the cosmwasm pyth contract local dependency to wormhole. Instead uses their repo within the image.
- Removes memmap2 patch (as it's not needed anymore)
- Rename pyth-bridge to pyth-cosmwasm


cosmwasm/contracts/pyth/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Data source does not exists error (on removing data source)
+    #[error("DataSourceDoesNotExists")]
+    DataSourceDoesNotExists,
+    
+    /// Data source already exists error (on adding data source)
+    #[error("DataSourceAlreadyExists")]
+    DataSourceAlreadyExists,
+}
+
+impl ContractError {
```
Reisen, line 31: This error type is much better than before for sure! Instead than the inherent impl though, we can be more idiomatic by implementing `From` here which is used by the `?` operator via `FromResidual`:

```suggestion
impl From<ContractError> for StdError {
  fn from(other: ContractError) -> StdError {
    …
  }
}
```

This will give us a working `?` throughout, anywhere you were using `std_err()` can now just be `?` and anywhere using `std()` can now just be `into()`.

cosmwasm/contracts/pyth/src/state.rs
```diff
@@ -39,20 +36,29 @@ pub const VALID_TIME_PERIOD: Duration = Duration::from_secs(3 * 60);
...
 pub struct ConfigInfo {
-    pub owner: HumanAddr,
-    pub wormhole_contract:  HumanAddr,
-    pub data_sources: HashSet<PythDataSource>,
+    pub owner:             Addr,
+    pub wormhole_contract: Addr,
+    pub data_sources:      HashSet<PythDataSource>,
 }
 
+impl Default for ConfigInfo {
```
Reisen, line 51: If It’s not too late to change this API, I think we would be better using `Option<Addr>` rather than using unchecked empty strings as nullify values, we can also drop the Default impl then.

cosmwasm/contracts/pyth/src/error.rs
ali-bahjati: You are right. will fix it.

Reisen (approved):

PR #263, opened 2022-09-02
remove resource memory limit

cctdaniel:
remove resource memory limit that was causing pyth-price-service to crash on a fresh VM install


ali-bahjati:
More context on this: This is a resource limit for k8s pod of price-service that tilt runs. For this resources to be set cgroup should be loaded. I think it wasn't caught before because it wasn't loaded in working environment of other contributors. So we haven't seen it before.

ali-bahjati (approved):

PR #264, opened 2022-09-02
add new latest_price_info endpoint

cctdaniel:
add new `api/latest_price_info?ids[]=<price_feed_id>&ids[]=<price_feed_id_2>&..` endpoint that augments `latest_price_feeds` with `emitter_chain`, `attestation_time`, and `sequence_number` so that we can store these data in our historical db

sample price info:
```
[
  {
    "price_feed": {
      "conf": "4",
      "ema_conf": "3",
      "ema_price": "1491",
      "expo": -5,
      "id": "af11a9b5e5e14838c1ecab3a0cdec6769c3d6a477da04412bd7d21bfe451ce3d",
      "max_num_publishers": 1,
      "num_publishers": 1,
      "prev_conf": "3",
      "prev_price": "1925",
      "prev_publish_time": 1662140402,
      "price": "1441",
      "product_id": "af11a9b5e5e14838c1ecab3a0cdec6769c3d6a477da04412bd7d21bfe451ce3d",
      "publish_time": 1662140407,
      "status": "Trading"
    },
    "emitter_chain": 1,
    "attestation_time": 1662140410,
    "sequence_number": 14239
  }
]
```

### DIFF
--- a/third_party/pyth/price-service/src/__tests__/rest.test.ts
+++ b/third_party/pyth/price-service/src/__tests__/rest.test.ts
@@ -43,6 +43,7 @@ function dummyPriceInfoPair(
       attestationTime: 0,
       seqNum,
       vaaBytes: Buffer.from(vaa, "hex").toString("binary"),
+      emitterChainId: 0,
     },
   ];
 }
--- a/third_party/pyth/price-service/src/rest.ts
+++ b/third_party/pyth/price-service/src/rest.ts
@@ -1,15 +1,14 @@
-import express, { Express } from "express";
 import cors from "cors";
+import express, { NextFunction, Request, Response } from "express";
+import { Joi, schema, validate, ValidationError } from "express-validation";
+import { Server } from "http";
+import { StatusCodes } from "http-status-codes";
 import morgan from "morgan";
 import responseTime from "response-time";
-import { Request, Response, NextFunction } from "express";
+import { DurationInMs, DurationInSec } from "./helpers";
 import { PriceStore } from "./listen";
 import { logger } from "./logging";
 import { PromClient } from "./promClient";
-import { DurationInMs, DurationInSec } from "./helpers";
-import { StatusCodes } from "http-status-codes";
-import { validate, ValidationError, Joi, schema } from "express-validation";
-import { Server } from "http";
 
 const MORGAN_LOG_FORMAT =
   ':remote-addr - :remote-user ":method :url HTTP/:http-version"' +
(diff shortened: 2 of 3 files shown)


third_party/pyth/price-service/src/rest.ts
```diff
@@ -181,6 +180,51 @@ export class RestAPI {
...
+
+          const freshness: DurationInSec =
+            new Date().getTime() / 1000 - latestPriceInfo.attestationTime;
+          this.promClient?.addApiRequestsPriceFreshness(
+            req.path,
+            id,
+            freshness
+          );
+
+          responseJson.push({"price_feed": latestPriceInfo.priceFeed.toJson(), "emitter_chain": latestPriceInfo.emitterChainId, "attestation_time": latestPriceInfo.attestationTime, "sequence_number": latestPriceInfo.seqNum});
```
jayantk, line 213: it seems like this is the only line of code that changed from the method above ?

Can you make this an additional query parameter on the existing endpoint? I think (a) that would make the overall API less confusing to users, and (b) would reduce code duplication.

cctdaniel: that was the plan initially but I had a call with @ali-bahjati and he suggested that we use a separate endpoint rather than query params  -- but happy to implement query params as well

ali-bahjati: yeah I suggested this...

I want us to have an explicit API for price feeds, unless we discuss we don't want it. We can stop exposing it to users (if it's consuming).

Wherever we expose `price_feed` I think it's better that the returned json to be consistent with our definition of our price feed, with no more info in it (unless it's added everywhere).

Although I'm fine with removing `get_latest_price_feeds` entirely or make it return price feed as a nested value and has the rest of the infos with query param (or they always be there).

cctdaniel:
please ignore the gh actions for now -- this is most likely due to the bug raised by @ali-bahjati here: https://github.com/pyth-network/pyth-crosschain/pull/266 I've rebased off Ali's branch and made sure gh actions was fine

third_party/pyth/price-service/src/rest.ts
```diff
@@ -135,13 +134,16 @@ export class RestAPI {
...
           .required(),
+        verbose: Joi.boolean(),
       }).required(),
     };
     app.get(
       "/api/latest_price_feeds",
       validate(latestPriceFeedsInputSchema),
       (req: Request, res: Response) => {
         let priceIds = req.query.ids as string[];
+        // verbose is optional, default to false
```
ali-bahjati, line 145: it could be default to undefined too 🤔

third_party/pyth/price-service/src/__tests__/rest.test.ts
```diff
@@ -43,6 +43,7 @@ function dummyPriceInfoPair(
       attestationTime: 0,
       seqNum,
       vaaBytes: Buffer.from(vaa, "hex").toString("binary"),
+      emitterChainId: 0,
```
ali-bahjati, line 46: it it's not too much it would be nice to test the verbose request too.

ali-bahjati (approved):
tyvm ser, sorry for being a drag here.

third_party/pyth/price-service/src/rest.ts
cctdaniel: if it's undefined it will default to false since it's doing the check in the next line 🤔

jayantk (approved):

PR #265, opened 2022-09-02
Hack the rent exempt bug

ali-bahjati:
This PR will solves a bug within wormhole that might not create the accounts with correct balance to be rent exempt. 

It will check the account balances in the end of attestation (to avoid breaking wormhole behavior) and transfers money to them if needed.

This PR will also solve a problem on balance transfer when a message account is resized to a larger size. Since the message account is owned by wormhole it is not possible to move funds back in case of resizing to a smaller size.

In order to make sure it works I used the double rent exempt amount in local testing and the transfers succeeded.

### DIFF
--- a/solana/pyth2wormhole/program/src/attest.rs
+++ b/solana/pyth2wormhole/program/src/attest.rs
@@ -358,6 +358,8 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
         .as_slice(),
     )?;
 
+    // NOTE: 2022-09-05
+    //
     // This part is added to avoid rent exemption error that is introduced using
     // a wrong implementation in solitaire
     //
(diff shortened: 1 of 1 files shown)


jayantk (approved):

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -355,5 +358,40 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+        );
+        invoke(&transfer_ix, ctx.accounts)?
+    }
+
+    // Checking the sequence account balance
+    let wg_sequence_balance = accs.wh_sequence.info().lamports();
+    let wh_sequence_rent_exempt = Rent::get()?.minimum_balance(accs.wh_sequence.data_len());
+
+    if wg_sequence_balance < wh_sequence_rent_exempt {
+        let required_deposit = wh_sequence_rent_exempt - wg_sequence_balance;
```
ali-bahjati, line 389: typo

ali-bahjati: thanks @drozdziak1 for raising it. fixed.

drozdziak1 (approved):

PR #267, opened 2022-09-05
p2w-client: add get-emitter for printing emitter address

drozdziak1:
This unblocks speculative target chain contract deployments. Assuming no solana-deployment-level problem occurs, we can add the emitter address to each contract and expect an attester program's messages to be accepted rightaway


ali-bahjati (approved):
tyvm!

PR #268, opened 2022-09-05
Drozdziak1/p2w client error logging and docker caching

drozdziak1:
This changeset improves p2w-client's error logging and helps it build quicker. The logging part is the main focus, addressing two independent issues in pyth2wormhole-client:
* Lack of failed attestation logging due to improper use of `map_err` in the futures-based attestation event loop - Thanks @Reisen :muscle: 
* Upstream Solana SDK changes  to the `to_string` impl on `ClientError`. Even if an error was printed, it would obfuscate program logs to `[N program logs]` which is extremely unhelpful compared to the verbatim program logs. A `Debug`-based struct dump fixes the issue by looking at the type directly. Specifically, `{:#?}` is used for easier human readability.

The back-burner caching fixes entail:
* **not depending on `bridge-client` for `Dockerfile.p2w-attest` anymore, in favor of `certusone/solana:v1.10.31`** - this eradicates `cargo install bridge-client` and `cargo install token-bridge-client` from the container build. The previous usage of the bridge client container was due to compile-time envs `BRIDGE_ADDRESS` and `EMITTER_ADDRESS` being reused. This change redefines the envs instead, as their contents have been stable for at least a year now. On a separate note, the bridge_client and token_bridge_client binaries are not relevant for the p2w-attest container at all.
* **Usage of `--target-dir` on `cargo-install` commands** - This option forces cargo-install to use a specified directory for build artifacts which lets us add a BuildKit cache mount in each case like we do already in `cargo-build` and `cargo-test` cases. This shortens the compilation required on iterative rebuilds of relevant containers.

### DIFF
--- a/Dockerfile.client
+++ b/Dockerfile.client
@@ -12,7 +12,8 @@ COPY solana /usr/src/solana
 WORKDIR /usr/src/solana/pyth2wormhole
 
 RUN --mount=type=cache,target=/root/.cache \
-    cargo install --version =2.0.12 --locked spl-token-cli
+    --mount=type=cache,target=target \
+    cargo install --version =2.0.12 --locked spl-token-cli --target-dir target
 
 
 RUN solana config set --keypair "/usr/src/solana/keys/solana-devnet.json"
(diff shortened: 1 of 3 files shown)


ali-bahjati:
Can't we remove `Dockerfile.client` now that we don't use it? :think

ali-bahjati (approved):

drozdziak1:
solana-devnet continues to use it to bootstrap the bridge contract IIUC

PR #270, opened 2022-09-06
p2w: Make commitment usage consistent

drozdziak1:
This changeset adjusts p2w_autoattest.py commitment to confirmed and
fixes a get_transaction() failure that happened due to the upstream
default finalized commitment setting (RpcClient's set value is not
used). This is fixed by using the chosen commitment explicitly via
get_transaction_with_config()


ali-bahjati (approved):
:shipit:

PR #272, opened 2022-09-07
Abehjati/fix-deployment-bugs

ali-bahjati:
About the deployCommitHash replacement. Replacing the binary was fine but I couldn't verify it. Now we replace the code (which will be used for verification too)

Because for verification we replace the text in the contract (and not in the binary) and due to eth optimizations it can be stored differently, in our case the commit hash was:
`0x65c273fa0afd5872f10972f65ead2cc6fa076da8`. If you look at its binary representation it has 3 trailing zeroes (8 is `0b1000`), then the contract store it shifted 3 bits to the right (and shifts it back later in a separate instruction). This hasn't happened in testnet because probably the previous deploymen commit hash was an odd number!

### DIFF
--- a/ethereum/Deploying.md
+++ b/ethereum/Deploying.md
@@ -12,14 +12,14 @@ rm -f .env; ln -s .env.prod.xyz .env && set -o allexport && source .env set && s
 # The Secret Recovery Phrase for the wallet the contract will be deployed from.
 export MNEMONIC=...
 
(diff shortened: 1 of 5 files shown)


tompntn (approved):

ali-bahjati:
This way of replacement showed that it is not a clean way. Let's replace it with a better solution later.

PR #278, opened 2022-09-14
Move staleness check to new endpoint

jayantk:
I made the staleness parameter an argument to the endpoint also, so we can configure it directly via the alert in datadog without redeploying the service.

I can't figure out how to build / run this code, so would love some help with that.

### DIFF
--- a/third_party/pyth/price-service/src/rest.ts
+++ b/third_party/pyth/price-service/src/rest.ts
@@ -5,10 +5,11 @@ import { Server } from "http";
 import { StatusCodes } from "http-status-codes";
 import morgan from "morgan";
 import responseTime from "response-time";
-import { DurationInMs, DurationInSec } from "./helpers";
+import {DurationInMs, DurationInSec, TimestampInSec} from "./helpers";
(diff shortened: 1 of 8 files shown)


cctdaniel:
you'll have to:
- install docker
- install kubectl
- install tilt
- run `minikube start --driver=docker --kubernetes-version=v1.22.3 --cpus=$(nproc) --memory=16G --disk-size=120g --namespace=wormhole`
- run `tilt up -- --pyth` in `price-service` directory

third_party/pyth/price-service/src/rest.ts
```diff
@@ -5,10 +5,11 @@ import { Server } from "http";
 import { StatusCodes } from "http-status-codes";
 import morgan from "morgan";
 import responseTime from "response-time";
-import { DurationInMs, DurationInSec } from "./helpers";
+import {DurationInMs, DurationInSec, TimestampInSec} from "./helpers";
```
ali-bahjati, line 8: Do you mind formatting this a bit?

ali-bahjati (approved):

PR #279, opened 2022-09-15
Abehjati/add-eth-governance

ali-bahjati:


ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -0,0 +1,115 @@
...
+    function upgradeUpgradableContract(UpgradeContractPayload memory payload) virtual internal;
+
+
+    function setGovernanceDataSource(bytes memory encodedPayload) internal {
+        SetGovernanceDataSourcePayload memory payload = parseSetGovernanceDataSourcePayload(encodedPayload);
+
+        PythInternalStructs.DataSource memory oldGovernanceDatSource = governanceDataSource();
+
+        setGovernanceDataSource(payload.newGovernanceDataSource);
+        setLastExecutedGovernanceSequence(0);
```
tompntn, line 75: Resetting the sequence number here might make us vulnerable to someone submitting an older governance action, if we ever change the data source back to a previously used one. I wonder if it would be safer to store sequence numbers per data source?

This is a very unlikely scenario, but maybe worth thinking about.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,134 @@
...
+ */
+contract PythGovernanceInstructions {
+    // module 2 corresponds to Target Chain contracts
+    uint8 constant MODULE = 2;
+
+    using BytesLib for bytes;
+
+    struct GovernanceInstruction {
+        uint8 module;
+        uint8 action;
```
tompntn, line 21: Could this field have the `GovernanceAction` type, and do the mapping from `uint` to `GovernanceAction` at deserialization time? This would make it less likely to make a mistake later in the code.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,134 @@
...
+ * for Pyth governance instructions.
+ */
+contract PythGovernanceInstructions {
+    // module 2 corresponds to Target Chain contracts
+    uint8 constant MODULE = 2;
+
+    using BytesLib for bytes;
+
+    struct GovernanceInstruction {
+        uint8 module;
```
tompntn, line 20: Could we have a `GovernanceModule` enum, for ease of readability?

tompntn (commented):
Nice! Only minor comments 🎉

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: nice catch, I will add 'initialSequenceNumber` to this command.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -0,0 +1,116 @@
...
+import "./PythGovernanceInstructions.sol";
+import "./PythInternalStructs.sol";
+import "./Pyth.sol";
+
+import "@openzeppelin/contracts/proxy/ERC1967/ERC1967Upgrade.sol";
+
+/**
+ * @dev `Governance` defines a means to enacting changes to the Pyth contract.
+ */
+abstract contract PythGovernance is Pyth, PythGovernanceInstructions {
```
tompntn, line 15: I think `Pyth` should inherit `PythGovernance`, not the other way round.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -0,0 +1,118 @@
...
+        emit GovernanceDataSourceSet(oldGovernanceDatSource, governanceDataSource(), lastExecutedGovernanceSequence());
+    }
+
+    function setDataSources(bytes memory encodedPayload) internal {
+        SetDataSourcesPayload memory payload = parseSetDataSourcesPayload(encodedPayload);
+
+        PythInternalStructs.DataSource[] memory oldDataSources = validDataSources();
+
+        for (uint i = 0; i < oldDataSources.length; i += 1) {
+            _state.isValidDataSource[keccak256(abi.encodePacked(oldDataSources[i].chainId, oldDataSources[i].emitterAddress))] = false;
```
jayantk, line 89: nitpick: extract the code to hash the data source (the `keccak256(...)` bit) into a helper function to reduce code reuse

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        delete _state.validDataSources;
+        for (uint i = 0; i < payload.dataSources.length; i++) {
+            _state.validDataSources.push(payload.dataSources[i]);
```
jayantk, line 94: seems confusing that this variable is called validDataSources when it's only valid if there's a true in the `isValidDataSources` map.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -0,0 +1,118 @@
...
+    function verifyGovernanceVM(bytes memory encodedVM) internal returns (IWormhole.VM memory parsedVM){
+        (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(encodedVM);
+
+        require(valid, reason);
+        require(isValidGovernanceDataSource(vm.emitterChainId, vm.emitterAddress), "VAA is not coming from the governance data source");
+
+        require(vm.sequence > lastExecutedGovernanceSequence(), "VAA is older than the last executed governance VAA");
+
+        setLastExecutedGovernanceSequence(vm.sequence);
+
```
jayantk, line 34: I suggest adding a magic number into the VAA that we can use to verify that this message is intended to be interpreted as a `GovernanceInstruction`. I'm a little worried that someone could trick the emitter to sign a VAA payload that is simultaneously valid for multiple different uses. I think that's pretty unlikely, but a magic number would make it very clear what the possible uses for any given VAA payload are.

jayantk: (e.g., pyth price update VAAs all start with the byte sequence "P2WH" as their magic number)

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,153 @@
...
+ */
+contract PythGovernanceInstructions {
+
+    using BytesLib for bytes;
+
+    enum GovernanceModule {
+        Core, // 0
+        Target, // 1
+        Attest // 2
+    }
```
jayantk, line 21: hmm... I'm a little confused by this. Perhaps this is essentially magic number I was asking for above?

jayantk: If so, it's still probably a good idea to have both a static magic number along with this enum. We do this in the pyth account data, and it's a good practice to prevent serialized messages from accidentally (or maliciously) being misinterpreted.

jayantk (commented):
Looks solid to me, just some minor comments.

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: They are essentially the same, it allows us to iterate over them (otherwise it's impossible).

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    struct SetValidPeriodPayload {
+        uint newValidPeriod;
+    }
+
+    /// @dev Parse a GovernanceInstruction
+    function parseGovernanceInstruction(bytes memory encodedInstruction) public pure returns (GovernanceInstruction memory gi) {
+        uint index = 0;
+
```
guibescos, line 67: Are you checking the magic number here?

ali-bahjati: Ah right, I forgot to add magic here totally. Will do. Thanks.

ethereum/.env.template
```diff
@@ -24,3 +24,6 @@ PYTHNET_EMITTER= # 0xa27839d641b07743c0cb5f68c51f8cd31d2c0762bec00dc6fcd25433ef1
 # valid, after this duration, the price feed is stale and will be invalid.
 # This value should derive from Pyth to wormhole latency, and target chain blocktime and latency.
 VALID_TIME_PERIOD_SECONDS= # 60
+
+GOVERNANCE_CHAIN_ID= # 0x1
```
jayantk, line 28: why are these commented out?

ethereum/contracts/pyth/PythUpgradable.sol
```diff
@@ -65,11 +67,25 @@ contract PythUpgradable is Initializable, OwnableUpgradeable, UUPSUpgradeable, P
         PythSetters.setValidTimePeriodSeconds(newValidTimePeriodSeconds);
     }
 
+    // Privileged function to update the governance emitter
```
jayantk, line 70: this function exists so you can do the migration from the current owner -> governance?

jayantk: if so, maybe leave a comment that this function can be deleted once the migration happens

ethereum/migrations/prod/9_pyth_enable_governance.js
```diff
@@ -0,0 +1,23 @@
...
+ * Version 1.0.0
+ * 
+ * This change:
+ * - Moves away single ownership to Governance coming from the Wormhole
+ */
+module.exports = async function (deployer) {
+    const proxy = await PythUpgradable.deployed();
+    await upgradeProxy(proxy.address, PythUpgradable, { deployer, unsafeSkipStorageCheck: true });
+    await proxy.updateGovernanceDataSource(governanceChainId, governanceEmitter, 0);
+    await proxy.renounceOwnership();
```
jayantk, line 22: i suggest renouncing ownership as a separate migration. Deploy the WH control changes, then run a WH governance instruction to test that it works, then renounce ownership after. It's a bit more forgiving that way in case we screw something up.

third_party/pyth/xc-governance-sdk-js/src/instructions.ts
```diff
@@ -0,0 +1,174 @@
...
+    private module: Module,
+    private action: number,
+    private targetChainId: ChainId,
+  ) { };
+
+  protected abstract serializePayload(): Uint8Array;
+
+  private serializeHeader(): Uint8Array {
+    return SerializeUtils.concat(
+      SerializeUtils.serializeUint32(MAGIC),
```
jayantk, line 79: is there really not an existing library to serialize binary messages like this?

(I know we're in a rush, so don't worry about this for this PR. It's not a big deal right now.)

ethereum/contracts/pyth/PythUpgradable.sol
```diff
@@ -65,11 +67,25 @@ contract PythUpgradable is Initializable, OwnableUpgradeable, UUPSUpgradeable, P
...
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() initializer {}
 
     // Only allow the owner to upgrade the proxy to a new implementation.
     function _authorizeUpgrade(address) internal override onlyOwner {}
 
+    // Execute a UpgradeContract governance message
+    function upgradeUpgradableContract(UpgradeContractPayload memory payload) override internal {
+        address oldImplementation = _getImplementation();
+        _upgradeToAndCallUUPS(payload.newImplementation, new bytes(0), false);
```
jayantk, line 87: does this do any sort of sanity checking to ensure that `payload.newImplementation` points to a new contract? I.e., is there any protection against us accidentally bricking the contract by passing in the wrong value for `newImplementation`?

third_party/pyth/xc-governance-sdk-js/src/instructions.ts
jayantk: also fwiw, for this kind of use case, usually the right pattern is to make some kind of `Builder` object that stores a mutable buffer and has methods to add stuff to the end. Like:

```
let builder =BufferBuilder()
builder.addUint32(...)
builder.addUint64(...)
return builder.build()
```

again though, don't worry about this right now. this is purely an FYI.

ethereum/migrations/test/9_pyth_enable_governance.js
```diff
@@ -0,0 +1,23 @@
...
+ * Version 1.0.0
+ * 
+ * This change:
+ * - Moves away single ownership to Governance coming from the Wormhole
+ */
+module.exports = async function (deployer) {
+    const proxy = await PythUpgradable.deployed();
+    await upgradeProxy(proxy.address, PythUpgradable, { deployer, unsafeSkipStorageCheck: true });
+    await proxy.updateGovernanceDataSource(governanceChainId, governanceEmitter, 0);
+    await proxy.renounceOwnership();
```
jayantk, line 22: same comment about renouncing ownership

jayantk (commented):
dude you are really cranking this out.

I left a couple comments on to make the migration a little less scary. Generally looks good to me though. Of course would be good to get some more eyes on this sort of thing.

tompntn (approved):

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,157 @@
...
+    function parseSetFeePayload(bytes memory encodedPayload) public pure returns (SetFeePayload memory sf) {
+        uint index = 0;
+
+        uint64 val = encodedPayload.toUint64(index);
+        index += 8;
+
+        uint64 expo = encodedPayload.toUint64(index);
+        index += 8;
+
+        sf.newFee = uint256(val) * uint256(10)**uint256(expo);
```
tompntn, line 143: I think we need an overflow check here.

ethereum/contracts/pyth/PythUpgradable.sol
tompntn: One thing we could do is add a public `isPyth` function, which we call here to check that the new implementation is actually the Pyth contract.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
ali-bahjati: From solidity 0.8 onwards all math operations in solidity are checked. (unless you do it unchecked)

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    enum GovernanceModule {
+        Executor, // 0
+        Target // 1
+    }
+
+    GovernanceModule constant MODULE = GovernanceModule.Target;
+
+    enum GovernanceAction {
+        UpgradeContract, // 0
```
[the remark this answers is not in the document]
ali-bahjati: I think the way it is right now is more clear. We have distinction of chainId and action so it won't be confused. And in the js-sdk side, we'll define separate classes.

ethereum/migrations/prod/9_pyth_enable_governance.js
ali-bahjati: Yeah, I was planning to do them separately to test. But It's good to have them in separate migration files (make it easier to test separately)

ethereum/contracts/pyth/PythUpgradable.sol
ali-bahjati: All the privileged functions above are like this. I'm not sure we can remove it with the current truffle deployment process though. (every contract on a new chain is deployed with an owner without governance emitter and then all the migration steps happen)

ethereum/.env.template
ali-bahjati: This one is not used and I forgot to uncomment it.

third_party/pyth/xc-governance-sdk-js/src/instructions.ts
ali-bahjati: There are some utils in `ethers` that worm CLI uses. But I didn't like their approach and tried to stick to the native ways. In a quick search I couldn't find cleaner ways. 

I couldn't find a mutable way to do this and created concat to be more efficient (copy only once). But it's not so important. Initially I thought builder here could be a bit annoying but now it seems good to me. I'll change it.

ethereum/.env.template
ali-bahjati: This is the template file as a description.

third_party/pyth/xc-governance-sdk-js/src/instructions.ts
ali-bahjati: I figured out Buffer has good write methods and I used them and created BufferBuilder. Now it's much clear. I tried to use `mutable-buffer` package but it didn't have 64bit writes. Other packages were too complex for this use case.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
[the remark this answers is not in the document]
tompntn: So we'll have different shaped payloads depending on the combination of chainId and action?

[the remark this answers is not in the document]
ali-bahjati: Yes I like it this way.

third_party/pyth/xc-governance-sdk-js/src/serialize.ts
```diff
@@ -0,0 +1,63 @@
+export interface Serializable {
+  serialize(): Buffer;
+}
+
+export class BufferBuilder {
```
jayantk, line 5: (you really didn't have to do this in this PR. I just mentioned it as an FYI.)

jayantk (approved):

PR #280, opened 2022-09-15
Initialize remote executor program

guibescos:
Context: A generic program for Pythnet that parses PostedVAAs from solana and executes things on Pythnet using the data from the VAAs.

Setup repo for the remote-executor program. 

Looking for feedback on :
- naming
- is this the right place for the program -> it seemed to make sense to have it in the same Cargo workspace as the others.


drozdziak1 (approved):
Lookin' good!

guibescos:
Closing because we don't want to use solitaire.

PR #283, opened 2022-09-16
Guibescos/executor

guibescos:


pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -1,15 +1,61 @@
...
+        assert_or_err(posted_vaa.emitter_chain == CHAIN_ID_SOLANA, err!(ExecutorError::EmitterChainNotSolana))?;
+        assert_or_err(posted_vaa.sequence > claim_record.sequence, err!(ExecutorError::NonIncreasingSequence))?;
+        claim_record.sequence = posted_vaa.sequence;
+
+        let payload = ExecutorPayload::try_from_slice(&posted_vaa.payload)?;
+        payload.check_header()?;
+        
+        for instruction in payload.instructions.iter().map(Instruction::from) {
+            // TO DO: We currently pass `remaining_accounts` down to the CPIs, is there a more efficient way to do it?
+            // TO DO: We don't pass ctx.accounts so executor_key needs to be both in the anchor context and in remaining accounts
```
jayantk, line 31: just so i'm clear on how this works: the `Instruction` already has a list of accounts that it expects to operate on, and this CPI works as long as `remaining_accounts` is a superset of the accounts in the instruction?

jayantk: also i think passing in the executor_key again as part of the remaining_accounts is more explicit anyway. Adding the executor_key by default here could be confusing to the user.

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+
+#[derive(AnchorDeserialize, AnchorSerialize)]
+pub struct GovernanceHeader {
+    pub magic_number : u32,
+    pub module : Module,
+    pub action : Action,
+    pub chain : ReceiverChain
+}
```
jayantk, line 37: i'm a little confused about this bit -- is there a reason we're using the same message header format here as with the target chain stuff? I don't see an issue with reusing the same format, though i also don't see a benefit (?)

jayantk (commented):
Seems about right to me, though i would like to have someone more familiar with wormhole / solana look at it as well.

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
ali-bahjati: I think it's good if we wanted to add extra functionality 🤔

pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -1,15 +1,61 @@
...
     use super::*;
 
     pub fn execute_posted_vaa(ctx: Context<ExecutePostedVaa>) -> Result<()> {
+        let posted_vaa = &ctx.accounts.posted_vaa;
+        let claim_record = &mut ctx.accounts.claim_record;
+        assert_or_err(posted_vaa.emitter_chain == CHAIN_ID_SOLANA, err!(ExecutorError::EmitterChainNotSolana))?;
+        assert_or_err(posted_vaa.sequence > claim_record.sequence, err!(ExecutorError::NonIncreasingSequence))?;
+        claim_record.sequence = posted_vaa.sequence;
+
+        let payload = ExecutorPayload::try_from_slice(&posted_vaa.payload)?;
```
ali-bahjati, line 26: We usually parse these stuff manually. Borsh definition of the structs and their order might be different than our wire format.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
guibescos: Yes that's exact

guibescos: If you're cool with having the user pass executer_key as part of the additional accounts then this should work.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
guibescos: Got it. Indeed borsh was not doing it right. But I fixed it and I'll make sure it works with unit tests.

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
guibescos: I don't have any strong opinions. I think at the very least we should keep the magic number and the module enum.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
ali-bahjati: Isn't executor_key passed to the instruction and parsed by anchor (to check seeds)? Then I'm not sure it's included in remaining accounts (there is a account deduplication logic in solana i think)

ali-bahjati (commented):
in general It looks good to me, I think if it would be nice if we have tests here + a cli to generate the required payload for the wormhole message.

pythnet/remote-executor/programs/remote-executor/src/state/posted_vaa.rs
```diff
@@ -0,0 +1,43 @@
+use std::{io::Write, str::FromStr, ops::Deref};
+use anchor_lang::prelude::*;
+use wormhole_solana::VAA;
+
+impl Owner for AnchorVaa {
+    fn owner() -> Pubkey{
+        Pubkey::from_str("H3fxXJ86ADW2PNuDDmZJg6mzTtPxkYCpNuQUTgmJ7AjU").unwrap() // Pythnet bridge address
```
tompntn, line 7: Could we avoid hardcoding this?

pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -1,15 +1,61 @@
...
+    pub system_program: Program<'info, System>
+}
+
+pub fn assert_or_err(condition : bool, error : Result<()>) -> Result<()>{
+    if !condition {
+        error
+    } else {
+        Result::Ok(())
+    }
+}
```
tompntn, line 61: Do you have `rustfmt` enabled on this repo?

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
```diff
@@ -0,0 +1,125 @@
+use std::{ops::Deref, mem::size_of, io::ErrorKind};
+
+use anchor_lang::{prelude::*, solana_program::instruction::Instruction};
+use wormhole::Chain;
+
+use crate::{assert_or_err, error::ExecutorError};
+
+pub const MAGIC_NUMBER : u32 = 0x4d475450; // Reverse order of the solidity contract because borsh uses little endian numbers
```
tompntn, line 8: Could you add the value from the solidity contract in this comment?

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl ExecutorPayload {
+    const MODULE : Module = Module::Executor;
+    const ACTION : Action = Action::ExecutePostedVaa;
+
+    pub fn check_header(&self) -> Result<()>{
+        assert_or_err(self.header.magic_number == MAGIC_NUMBER, err!(ExecutorError::GovernanceHeaderInvalidMagicNumber))?;
+        assert_or_err(self.header.module == ExecutorPayload::MODULE, err!(ExecutorError::GovernanceHeaderInvalidMagicNumber))?;
+        assert_or_err(self.header.action == ExecutorPayload::ACTION, err!(ExecutorError::GovernanceHeaderInvalidMagicNumber))?;
+        assert_or_err(Chain::from(self.header.chain.value) == Chain::Pythnet, err!(ExecutorError::GovernanceHeaderInvalidMagicNumber))
```
tompntn, line 123: Have you seen the [boolinator](https://crates.io/crates/boolinator) crate? It includes a helpful [`as_result`](https://docs.rs/boolinator/2.4.0/boolinator/trait.Boolinator.html#tymethod.as_result) function that lets you do stuff like
```rust
some_cond.not().as_result((), Err(..))?;
```

which avoids the need for `assert_or_err`.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -1,15 +1,61 @@
...
 #[program]
 pub mod remote_executor {
+    use anchor_lang::solana_program::{program::invoke_signed, instruction::Instruction};
+    use wormhole::Chain::{Solana, self};
+
+    use crate::{state::{governance_payload::ExecutorPayload}, error::ExecutorError};
+
     use super::*;
 
     pub fn execute_posted_vaa(ctx: Context<ExecutePostedVaa>) -> Result<()> {
```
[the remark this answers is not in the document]
jayantk: imo, a struct would be overkill given that there's literally this one function that needs to be implemented. I would agree with you if there were more complexity in this code here (and would suggest refactoring if we did add more complexity).

pythnet/remote-executor/programs/remote-executor/src/state/posted_vaa.rs
jayantk: yeah this seems like it should be configurable at deployment time or something.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
guibescos: I think there's account deduplication at this level. We'll check in the tests.

pythnet/remote-executor/programs/remote-executor/src/lib.rs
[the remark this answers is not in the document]
guibescos: https://github.com/pyth-network/pyth-crosschain/pull/287

guibescos: https://github.com/pyth-network/pyth-crosschain/pull/287

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
guibescos: https://github.com/pyth-network/pyth-crosschain/pull/288

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
guibescos: https://github.com/pyth-network/pyth-crosschain/pull/289

pythnet/remote-executor/programs/remote-executor/src/state/posted_vaa.rs
guibescos: https://github.com/pyth-network/pyth-crosschain/pull/289

PR #291, opened 2022-09-20
Pyth Aptos Target Chain Contract

tompntn:
This PR implements the Pyth target chain contract on Aptos. Please see `aptos/contracts/sources/pyth.move` for documented details of how to interact with this contract.

Will add Aptos to Tilt integration tests in a follow-up PR.

### DIFF
--- a/aptos/contracts/Makefile
+++ b/aptos/contracts/Makefile
@@ -3,7 +3,7 @@ artifacts: build
 
 .PHONY: build
 build:
-	aptos move compile --save-metadata
+	aptos move compile --save-metadata --named-addresses wormhole=0x251011524cd0f76881f16e7c2d822f0c1c9510bfd2430ba24e1b3d52796df204,deployer=0x277fa055b6a73c42c0662d5236c65c864ccbf2d4abd21f174a30c8b786eab84b,pyth=0x277fa055b6a73c42c0662d5236c65c864ccbf2d4abd21f174a30c8b786eab84b
 
 .PHONY: clean
 clean:
--- a/aptos/contracts/Makefile
+++ b/aptos/contracts/Makefile
@@ -1,0 +1,14 @@
+.PHONY: artifacts
+artifacts: build
+
+.PHONY: build
+build:
+	aptos move compile --save-metadata
+
+.PHONY: clean
+clean:
+	aptos move clean --assume-yes
+
+.PHONY: test
+test:
+	aptos move test
--- a/aptos/contracts/sources/deserialize.move
+++ b/aptos/contracts/sources/deserialize.move
@@ -1,0 +1,140 @@
+module pyth::deserialize {
+    use wormhole::deserialize;
+    use wormhole::u16;
+    use wormhole::u32;
+    use wormhole::cursor::{Self, Cursor};
+    use pyth::i64::{Self, I64};
+
+    public fun deserialize_vector(cur: &mut Cursor<u8>, n: u64): vector<u8> {
+        deserialize::deserialize_vector(cur, n)
+    }
+
+    public fun deserialize_u8(cur: &mut Cursor<u8>): u8 {
+        deserialize::deserialize_u8(cur)
+    }
+
+    public fun deserialize_u16(cur: &mut Cursor<u8>): u64 {
+        u16::to_u64(deserialize::deserialize_u16(cur))
+    }
+
+    public fun deserialize_u32(cur: &mut Cursor<u8>): u64 {
+        u32::to_u64(deserialize::deserialize_u32(cur))
+    }
+
+    public fun deserialize_i32(cur: &mut Cursor<u8>): I64 {
+        let deserialized = deserialize_u32(cur);
+        
+        // If negative, pad the value
+        let negative = (deserialized >> 31) == 1;
+        if (negative) {
+            let padded = (0xFFFFFFFF << 32) + deserialized;
+            i64::from_u64(padded)
+        } else {
+            i64::from_u64(deserialized)
+        }
+    }
+
+    public fun deserialize_u64(cur: &mut Cursor<u8>): u64 {
+        deserialize::deserialize_u64(cur)
+    }
+
+    public fun deserialize_i64(cur: &mut Cursor<u8>): I64 {
+        i64::from_u64(deserialize_u64(cur))
+    }
+
+    #[test]
+    fun test_deserialize_u8() {
+        let input = x"48258963";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_u8(&mut cursor);
+        assert!(result == 0x48, 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"258963", 1);
+    }
+
+    #[test]
+    fun test_deserialize_u16() {
+        let input = x"48258963";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_u16(&mut cursor);
+        assert!(result == 0x4825, 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"8963", 1);
+    }
+
+    #[test]
+    fun test_deserialize_u32() {
+        let input = x"4825896349741695";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_u32(&mut cursor);
+        assert!(result == 0x48258963, 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"49741695", 1);
+    }
+
+    #[test]
+    fun test_deserialize_i32_positive() {
+        let input = x"4825896349741695";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_i32(&mut cursor);
+        assert!(result == i64::from_u64(0x48258963), 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"49741695", 1);
+    }
+
+    #[test]
+    fun test_deserialize_i32_negative() {
+        let input = x"FFFFFDC349741695";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_i32(&mut cursor);
+        assert!(result == i64::from_u64(0xFFFFFFFFFFFFFDC3), 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"49741695", 1);
+    }
+
+    #[test]
+    fun test_deserialize_u64() {
+        let input = x"48258963497416957497253486";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_u64(&mut cursor);
+        assert!(result == 0x4825896349741695, 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"7497253486", 1);
+    }
+
+    #[test]
+    fun test_deserialize_i64_positive() {
+        let input = x"48258963497416957497253486";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_i64(&mut cursor);
+        assert!(result == i64::from_u64(0x4825896349741695), 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"7497253486", 1);
+    }
+
+    #[test]
+    fun test_deserialize_i64_negative() {
+        let input = x"FFFFFFFFFFFFFDC37497253486";
+        let cursor = cursor::init(input);
+        
+        let result = deserialize_i64(&mut cursor);
+        assert!(result == i64::from_u64(0xFFFFFFFFFFFFFDC3), 1);
+
+        let rest = cursor::rest(cursor);
+        assert!(rest == x"7497253486", 1);
+    }
+}
--- a/aptos/contracts/sources/event.move
+++ b/aptos/contracts/sources/event.move
@@ -1,0 +1,39 @@
+module pyth::event {
+    use std::event::{Self, EventHandle};
+    use pyth::price_feed::{PriceFeed};
+    use std::account;
+
+    friend pyth::pyth;
+
+    /// Signifies that a price feed has been updated
+    struct PriceFeedUpdate has store, drop {
+        /// Value of the price feed
+        price_feed: PriceFeed,
+        /// Timestamp of the update
+        timestamp: u64,
+    }
+
+    struct PriceFeedUpdateHandle has key, store {
+        event: EventHandle<PriceFeedUpdate>
+    }
+
+    public(friend) fun init(pyth: &signer) {
+        move_to(
+            pyth,
+            PriceFeedUpdateHandle {
+                event: account::new_event_handle<PriceFeedUpdate>(pyth)
+            }
+        );
+    }
+
+    public(friend) fun emit_price_feed_update(price_feed: PriceFeed, timestamp: u64) acquires PriceFeedUpdateHandle {
+        let event_handle = borrow_global_mut<PriceFeedUpdateHandle>(@pyth);
+        event::emit_event<PriceFeedUpdate>(
+            &mut event_handle.event,
+            PriceFeedUpdate {
+                price_feed,
+                timestamp,
+            }
+        );
+    } 
+}
--- a/aptos/contracts/sources/governance/contract_upgrade.move
+++ b/aptos/contracts/sources/governance/contract_upgrade.move
@@ -8,13 +8,15 @@ module pyth::contract_upgrade {
     use aptos_framework::code;
     use pyth::error;
 
+    friend pyth::governance;
+
     const HASH_LENGTH: u64 = 32;
 
     struct AuthorizeContractUpgrade {
         hash: Hash,
     }
 
-    public fun execute(payload: vector<u8>) {
+    public(friend) fun execute(payload: vector<u8>) {
         let AuthorizeContractUpgrade {hash: hash} = from_byte_vec(payload);
         state::set_contract_upgrade_authorized_hash(hash)
     }
--- a/aptos/contracts/sources/governance/contract_upgrade.move
+++ b/aptos/contracts/sources/governance/contract_upgrade.move
@@ -1,0 +1,69 @@
+module pyth::contract_upgrade {
+    use wormhole::cursor;
+    use pyth::deserialize;
+    use pyth::contract_upgrade_hash::{Self, Hash};
+    use pyth::state::{Self};
+    use std::vector;
+    use std::aptos_hash;
+    use aptos_framework::code;
+    use pyth::error;
+
+    const HASH_LENGTH: u64 = 32;
+
+    struct ContractUpgrade {
+        hash: Hash,
+    }
+
+    public fun execute(payload: vector<u8>) {
+        let ContractUpgrade {hash: hash} = from_byte_vec(payload);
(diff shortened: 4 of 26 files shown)


aptos/contracts/sources/price_feed.move
```diff
@@ -0,0 +1,54 @@
...
+        /// The price identifier
+        price_identifier: PriceIdentifier,
+        /// The status of the current aggregate price
+        status: PriceStatus,
+        /// The current aggregate price
+        price: Price,
+        /// The current exponentially moving average aggregate price
+        ema_price: Price,
+        /// The most recent previous price with TRADING status
+        previous_price: Price,
```
jayantk, line 17: should we drop this and make `price` the most recent price with TRADING status (per @ali-bahjati 's changes to the rust / solidity sdk?)

aptos/contracts/sources/deserialize.move
```diff
@@ -0,0 +1,140 @@
...
+
+    public fun deserialize_vector(cur: &mut Cursor<u8>, n: u64): vector<u8> {
+        deserialize::deserialize_vector(cur, n)
+    }
+
+    public fun deserialize_u8(cur: &mut Cursor<u8>): u8 {
+        deserialize::deserialize_u8(cur)
+    }
+
+    public fun deserialize_u16(cur: &mut Cursor<u8>): u64 {
```
jayantk, line 16: why does this return a `u64`?

aptos/contracts/sources/governance/contract_upgrade.move
```diff
@@ -0,0 +1,69 @@
...
+    public entry fun do_contract_upgrade(
+        metadata_serialized: vector<u8>,
+        code: vector<vector<u8>>,
+    ) {
+        // Check to see if the given code matches the authorized hash
+        let hash = state::get_contract_upgrade_authorized_hash();
+        let c = copy code;
+        vector::reverse(&mut c);
+        let a = vector::empty<u8>();
+        while (!vector::is_empty(&c)) vector::append(&mut a, vector::pop_back(&mut c));
```
jayantk, line 41: seems like `a` is a copy of `c` before it was reversed? why is this code necessary? (maybe leave a code comment)

aptos/contracts/sources/governance/governance.move
```diff
@@ -0,0 +1,55 @@
...
+        let action = governance_instruction::get_action(&instruction);
+        if (action == governance_action::destroy(governance_action::new_contract_upgrade())) {
+            contract_upgrade::execute(governance_instruction::destroy(instruction));
+        } else if (action == governance_action::destroy(governance_action::new_set_governance_data_source())) {
+            set_governance_data_source::execute(governance_instruction::destroy(instruction));
+        } else if (action == governance_action::destroy(governance_action::new_set_data_sources())) {
+            set_data_sources::execute(governance_instruction::destroy(instruction));
+        } else if (action == governance_action::destroy(governance_action::new_set_update_fee())) {
+            set_update_fee::execute(governance_instruction::destroy(instruction));
+        } else if (action == governance_action::destroy(governance_action::new_set_stale_price_threshold())) {
```
jayantk, line 29: seems a little annoying that you have to call `::destroy` here to use the `Action` struct. Seems like you're not getting a lot of value from `Action` and it would be simpler / less code to use a `u8`. (or maybe you can use a typedef on u8?)

aptos/contracts/sources/governance/set_update_fee.move
```diff
@@ -0,0 +1,43 @@
...
+    use wormhole::cursor;
+    use pyth::deserialize;
+    use std::math64;
+    use pyth::state;
+
+    const MAX_U64: u128 = (1 << 64) - 1;
+
+    const E_OVERFLOW: u64 = 15;
+
+    struct SetUpdateFee {
```
jayantk, line 11: the ship may have sailed on this already... but it sure seems like it would be simpler to pass a u64 directly instead of using a floating point representation

aptos/contracts/sources/i64.move
```diff
@@ -0,0 +1,99 @@
+module pyth::i64 {
+    use pyth::error;
+        
+    /// As Move does not support negative numbers natively, we use our own internal
+    /// representation.
+    struct I64 has copy, drop, store {
```
jayantk, line 6: this is clearly necessary since move doesn't have i64, but let's be careful with how we use this because the bool/unsigned representation has some weird properties (specifically two zeros). We should be clear in the docs how zero is represented, and then let's try not to implement arithmetic ops on this struct.

jayantk: Let's also explain in the comments on get_price etc. in the file below how we expect people to consume the I64's returned by the various functions. e.g., we expect them to call `get_magnitude_if_positive` for the price

aptos/contracts/sources/i64.move
```diff
@@ -0,0 +1,99 @@
...
+    fun test_get_magnitude_if_negative_negative() {
+        assert!(get_magnitude_if_negative(&new(7686, true)) == 7686, 1);
+    }
+
+    #[test]
+    #[expected_failure(abort_code = 196627)]
+    fun test_get_magnitude_if_negative_positive() {
+        assert!(get_magnitude_if_negative(&new(7686, false)) == 7686, 1);
+    }
+
```
jayantk, line 98: put in a test for 0

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,774 @@
...
+
+    /// Determine if a price feed for the given price_identifier exists
+    public fun price_feed_exists(price_identifier: PriceIdentifier): bool {
+        state::price_info_cached(price_identifier)
+    }
+
+    /// Get the latest cached data for the given price identifier, if that data is 
+    /// no older than the stale price threshold.
+    /// 
+    /// Important: it is recommended to call update_price_feeds() to update the cached data
```
jayantk, line 204: maybe document that the transaction reverts if the price is too old? (and hence why you need to call update_price_feeds)

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,774 @@
...
+        }
+    }
+
+    /// Get the stale price threshold: the amount of time after which a cached price
+    /// is considered stale and no longer returned by get_price().
+    public fun get_stale_price_threshold_secs(): u64 {
+        state::get_stale_price_threshold_secs()
+    }
+
+    /// Get the number of AptosCoin's requried to perform one batch update
```
jayantk, line 262: typo: required

jayantk (commented):
looks pretty reasonable to me. I will let someone who knows move approve though.

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,727 @@
...
+        let price_feed = price_info::get_price_feed(update);
+        let update_timestamp = price::get_timestamp(&price_feed::get_price(price_feed));
+
+        // Get the timestamp of the cached data for the price identifier
+        let price_identifier = price_feed::get_price_identifier(price_feed);
+        if (!price_feed_exists(*price_identifier)) {
+            return true
+        };
+        let cached_timestamp = price::get_timestamp(
+            &price_feed::get_price(price_info::get_price_feed(&state::get_latest_price_info(*price_identifier))));
```
ali-bahjati, line 185: @tompntn shouldn't you use unsafe versions here? Even for the update timestamp (maybe it's old but still newer).

tompntn: This uses `get_latest_price_info`, which is an internal getter function that has no timestamp check. However, using get_price_unsafe would make this code simpler 😄

ali-bahjati: i don't know. i see get_price both as contract interface and method on struct. so probably i'm wrong here.

aptos/contracts/sources/deserialize.move
tompntn: `u64` is the smallest primitive unsigned integer that fits a `u16`: move has `u8`, `u64` and `u128`.

aptos/contracts/sources/i64.move
tompntn: I've put in a check in the constructor to make it impossible to change `(0, true)` to `(0, false)`, so we will never have `(0, true)` values.

aptos/contracts/sources/governance/set_update_fee.move
tompntn: The reasoning behind this was so that we could use the same shaped payload for all the runtimes, while being able to:

- Deserialize integers using the largest common integer size across different runtimes (u64)
- Specify large fees for chains which have a high gas cost

This allows us to use a common format across all chains, while being able to specify a large range of fees.

Not sure whether this is the right call, and it would be better to have differently-shaped payloads. That said, I'm not sure there's much downside to doing it this way either, apart from a small bit of added complexity.

aptos/contracts/Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: artifacts
+artifacts: build
+
+.PHONY: build
```
ali-bahjati, line 4: not for this PR: can we somehow add a github CI check for this?

aptos/contracts/sources/event.move
```diff
@@ -0,0 +1,39 @@
+module pyth::event {
```
ali-bahjati, line 1: Maybe too much for now, can we have events similar to the solidity contract? e.g., have old price, have fee, ...

aptos/contracts/Makefile
tompntn: yep, I will add Aptos to the Tilt environment in a follow-up PR 😄

aptos/contracts/sources/price.move
```diff
@@ -0,0 +1,47 @@
...
+    /// to how this price safely.
+    struct Price has copy, drop, store {
+        price: I64,
+        /// Confidence interval around the price
+        conf: u64,
+        /// The exponent
+        expo: I64,
+        /// Unix timestamp of when this price was computed
+        // can get current timestamp with std::timestamp
+        timestamp: u64,
```
ali-bahjati, line 20: Is't `publish_time` a better name here?

aptos/contracts/sources/price_feed.move
```diff
@@ -0,0 +1,37 @@
+module pyth::price_feed {
+    use pyth::price_identifier::PriceIdentifier;
+    use pyth::price::Price;
+
+    /// PriceFeed represents a current aggregate price for a particular product.
+    struct PriceFeed has copy, drop, store {
+        /// The price identifier
+        price_identifier: PriceIdentifier,
```
ali-bahjati, line 8: `id`? I don't have strong opinion here though.

aptos/contracts/sources/price_status.move
```diff
@@ -0,0 +1,35 @@
...
+    const TRADING: u64 = 1;
+
+    /// PriceStatus represents the availability status of a price feed.
+    /// Prices should only be used if they have a status of trading.
+    struct PriceStatus has copy, drop, store {
+        status: u64,
+    }
+
+    public fun from_u64(status: u64): PriceStatus {
+        PriceStatus {
```
ali-bahjati, line 15: nitpick: I don't know how enum is implemented here nor whether it's important at all. I think here checking `status < 2` could be good.

aptos/contracts/sources/governance/set_update_fee.move
jayantk: ah got it. that makes sense.

aptos/contracts/sources/price_status.move
tompntn: Unfortunately Move doesn't have native enums. You're right though I'll definitely add that check.

aptos/contracts/sources/price_feed.move
jayantk: suggest calling the field `id` and the type `PriceIdentifier`

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,810 @@
...
+            data_sources,
+            update_fee
+        )
+    }
+
+// -----------------------------------------------------------------------------
+// Update the cached prices
+
+    /// Update the cached price feeds with the data in the given VAAs. This is a 
+    /// convienence wrapper around update_price_feeds(), which allows you to update the price feeds
```
jayantk, line 121: typo: convenience

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,810 @@
...
+    /// using an entry function.
+    /// 
+    /// If possible, it is recommended to use update_price_feeds() instead, which avoids the need
+    /// to pass a signer account. update_price_feeds_with_funder() should only be used when
+    /// you need to call an entry function.
+    /// 
+    /// WARNING: this function will charge an update fee, transferring some AptosCoin's
```
jayantk, line 128: nitpick: "WARNING" seems a little scary. I think omitting that and simply saying "This function will charge..." would be better

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,810 @@
...
+                );
+                event::emit_price_feed_update(price_feed, timestamp::now_microseconds());
+            }
+        };
+        vector::destroy_empty(updates);
+    }
+
+    /// Update the cached price feeds with the data in the given VAAs, using
+    /// update_price_feeds(). However, this function will only have an effect if any of the
+    /// prices in the update are fresh. The price_identifiers and publish_times paramaters
```
jayantk, line 188: typo: parameters

aptos/contracts/sources/state.move
```diff
@@ -0,0 +1,194 @@
...
+    struct LatestPriceInfo has key {
+        info: Table<PriceIdentifier, PriceInfo>,
+    }
+
+    /// The allowed data source for governance VAAs
+    struct GovernanceDataSource has key {
+        source: DataSource,
+    }
+
+    /// The ID of the chain this contract is deployed on
```
jayantk, line 50: is ChainID the wormhole chain id? if so, worth mentioning in the comment

jayantk (approved):
Code looks really solid to me. nicely structured and good tests. 👍 well done. 

(Note that I can't review for security because I don't have a good grasp on the security model )

aptos/contracts/sources/governance/set_data_sources.move
```diff
@@ -0,0 +1,42 @@
...
+    use pyth::state;
+    use std::vector;
+
+    friend pyth::governance;
+
+    struct SetDataSources {
+        sources: vector<DataSource>,
+    }
+
+    public(friend) fun execute(payload: vector<u8>) {
```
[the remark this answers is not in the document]
tompntn: This is explained well here https://diem.github.io/move/friends.html (a module declares it's friends)

njkumr (approved):

aptos/contracts/sources/governance/governance_instruction.move
```diff
@@ -0,0 +1,86 @@
...
+
+    public fun get_action(instruction: &GovernanceInstruction): GovernanceAction {
+        instruction.action
+    }
+
+    public fun get_target_chain_id(instruction: &GovernanceInstruction): u64 {
+        instruction.target_chain_id
+    }
+
+    public fun destroy(instruction: GovernanceInstruction): vector<u8> {
```
ali-bahjati, line 56: Just curious here, the term `destroy` is because the struct doesn't have clone and is moved here?

aptos/contracts/sources/governance/set_data_sources.move
```diff
@@ -0,0 +1,42 @@
+module pyth::set_data_sources {
```
ali-bahjati, line 1: It's lack of my knowledge here, couldn't we have nested modules like `pyth::governance::set_data_sources` for all the governance modules (like the folder structure)

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,803 @@
...
+        publish_times: vector<u64>,
+        fee: Coin<AptosCoin>) {
+
+        assert!(vector::length(&price_identifiers) == vector::length(&publish_times),
+            error::invalid_publish_times_length());
+
+        let fresh_data = false;
+        let i = 0;
+        while (i < vector::length(&publish_times)) {
+            let price_identifier = price_identifier::from_byte_vec(
```
ali-bahjati, line 201: I'm not too familiar with move, but looking at `with_funder` method above, shouldn't this be the same? or have a same `with_funder` version? (For an imaginary price pusher service that wants to send update directly)

aptos/contracts/sources/pyth.move
```diff
@@ -0,0 +1,803 @@
...
+    fun update_price_feed_from_single_vaa(vaa: vector<u8>) {
+        // Deserialize the VAA
+        let vaa = vaa::parse_and_verify(vaa);
+
+        // Check that the VAA is from a valid data source (emitter)
+        assert!(
+            state::is_valid_data_source(
+                data_source::new(
+                    u16::to_u64(vaa::get_emitter_chain(&vaa)),
+                    vaa::get_emitter_address(&vaa))),
```
ali-bahjati, line 156: I think it's better to get the update_fee here rather than the provided additional fee (which might be given by mistake)

aptos/contracts/sources/governance/contract_upgrade.move
```diff
@@ -0,0 +1,78 @@
...
+    public entry fun do_contract_upgrade(
+        metadata_serialized: vector<u8>,
+        code: vector<vector<u8>>,
+    ) {
+        // Check to see if the hash of the given code and metadata matches the authorized hash.
+        // The aptos framework does no validation of the metadata, so we include it in the hash.
+        assert!(matches_hash(code, metadata_serialized, state::get_contract_upgrade_authorized_hash()), error::invalid_upgrade_hash());
+        // Perform the upgrade
+        let pyth = state::pyth_signer();
+        code::publish_package_txn(&pyth, metadata_serialized, code);
```
ali-bahjati, line 43: I am not sure how impactful it is but isn't it better to clear up the state at this point? :think:

ali-bahjati (approved):
It's amazing!

I left some comments.
I have one question, is it possible to add a test for an actual upgrade? (It seems to have two phase and only the first one is tested)

aptos/contracts/sources/governance/set_data_sources.move
tompntn: These aren't nested modules, they are `address::module` (`pyth` is a named address, `set_data_sources` is the module).

aptos/contracts/sources/pyth.move
tompntn: Yep, I just didn't have time to add this - I will add one in a follow-up PR.

aptos/contracts/sources/governance/contract_upgrade.move
tompntn: `get_contract_upgrade_authorized_hash` uses `move_from`, so can only be called once.

aptos/contracts/sources/governance/governance_instruction.move
tompntn: Yep, this is so that the instruction is only used once.

PR #292, opened 2022-09-21
Setting up rust tests

guibescos:
The way the tests are organized are the following :
- `ExecutorBench` allows the developer to setup as many accounts as they want pre validator start including VAA accounts
- Calling `start` on `ExecutorBench` starts a validator and returns a client `ExecutorSimulator`
- You can call instructions on the executor by using `ExecutorSimulator::execute_posted_vaa`


pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -13,6 +14,9 @@ use state::{
 mod error;
 mod state;
 
+#[cfg(test)]
+mod tests;
+
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
```
tompntn, line 20: Could you add a comment explaining what this is?

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
```diff
@@ -46,6 +46,19 @@ pub struct GovernanceHeader {
...
 }
 
+impl Default for GovernanceHeader {
+    fn default() -> Self {
+        Self {
+            magic_number: MAGIC_NUMBER,
+            module: Module::Executor,
+            action: Action::ExecutePostedVaa,
+            chain: BigEndianU16 {
+                value: Chain::Pythnet.try_into().unwrap(),
```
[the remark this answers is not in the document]
guibescos: Ok I sort of agree that this is not how the default trait is supposed to be used :)

pythnet/remote-executor/programs/remote-executor/src/lib.rs
guibescos: This is very standard rust. It is just conditionally compiling the tests.

jayantk: i think tom is talking about the declare_id

pythnet/remote-executor/programs/remote-executor/src/state/governance_payload.rs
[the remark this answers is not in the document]
jayantk: I suggest moving this function into the `pub mod test` below so that it can't be accidentally misused.

jayantk (approved):

jayantk:
I did leave a couple comments that are showing as resolved. they're small though so feel free to fix in a follow up PR.

PR #294, opened 2022-09-21
Drozdziak1/p2w client use mapping crawl

drozdziak1:
This change adds and tests the merging of crawled mapping symbols with the existing attestation config. The additional symbols that were not previously seen in other symbol groups  are added to a symbol group called "mapping", This mapping group can contain pre-existing symbols and, more importantly, attestation conditions, allowing to specify the default behavior for symbols found in the mapping

### DIFF
--- a/solana/pyth2wormhole/client/src/main.rs
+++ b/solana/pyth2wormhole/client/src/main.rs
@@ -173,12 +173,13 @@ async fn main() -> Result<(), ErrBox> {
             daemon,
         } => {
             // Load the attestation config yaml
-            let attestation_cfg: AttestationConfig =
+            let mut attestation_cfg: AttestationConfig =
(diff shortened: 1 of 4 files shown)


ali-bahjati (approved):
LGTM! ship it 🚢

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -173,12 +173,13 @@ async fn main() -> Result<(), ErrBox> {
             daemon,
         } => {
             // Load the attestation config yaml
-            let attestation_cfg: AttestationConfig =
+            let mut attestation_cfg: AttestationConfig =
```
ali-bahjati, line 176: maybe this one is not necessary here 🤔

drozdziak1: It's not possible to take a mutable reference to an immutable variable

PR #299, opened 2022-09-22
Fixing some bugs on clap and processing logic

ali-bahjati:
Apparently clap expects the id in `conflicts_with` to be a case-converted value of the field. 

I also realized that our logic was not correct for handling that parameter, so I fixed it.

### DIFF
--- a/solana/pyth2wormhole/client/src/main.rs
+++ b/solana/pyth2wormhole/client/src/main.rs
@@ -136,8 +136,10 @@ async fn main() -> Result<(), ErrBox> {
             
             let new_ops_owner = if remove_ops_owner {
                 None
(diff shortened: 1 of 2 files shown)


solana/pyth2wormhole/client/src/main.rs
```diff
@@ -136,8 +136,10 @@ async fn main() -> Result<(), ErrBox> {
             
             let new_ops_owner = if remove_ops_owner {
                 None
+            } else if let Some(given_ops_owner) = ops_owner_addr {
```
drozdziak1, line 139: Nice catch! I think it would look its cleanest if we followed the `unwrap_or` pattern used by other Option<> values couple lines down

drozdziak1 (approved):
This is a tricky little bit to write, nice work!

PR #307, opened 2022-09-27
Set/Update governance emitter addresses

ali-bahjati:
Please double check the addresses. You can convert base58 to hex in [this website](https://appdevtools.com/base58-encoder-decoder)

Note: I changed the testnet governance because we were using the devnet squads. I should update the bnb testnet governance emitter to the new one separately (because this one was only deployed)


jayantk (commented):
I've checked that the hex addresses map to the base58 addresses that @cctdaniel posted. I want to double check the base58 addresses before hitting approve.

jayantk (approved):
I double checked the emitter via both the squads (it's the vault address converted from base58 -> hex) and also by decoding the emitter from 2 wormhole VAAs.

PR #309, opened 2022-09-28
Guibescos/executor cli

guibescos:


pythnet/remote-executor/cli/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+            let guardian_set_data =
+                GuardianSet::try_from_slice(&rpc_client.get_account_data(&guardian_set)?)?;
+
+            let signature_set_keypair = Keypair::new();
+
+            let vaa = VAA::from_bytes(vaa_bytes.clone())?;
+
+            // RENT HACK STARTS HERE
+            let signature_set_size = 4 + 19 + 32 + 4;
+            let posted_vaa_size = 3 + 1 + 1 + 4 + 32 + 4 + 4 + 8 + 2 + 32 + 4 + vaa.payload.len();
```
jayantk, line 92: suggest documenting where these numbers came from and under what conditions we would have to change them.

pythnet/remote-executor/cli/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+        Action::SendTestVAA { keypair } => {
+            let payer =
+                read_keypair_file(&*shellexpand::tilde(&keypair)).expect("Keypair not found");
+            let rpc_client = RpcClient::new_with_commitment(
+                "https://api.mainnet-beta.solana.com",
+                cli.commitment,
+            );
+
+            let message_keypair = Keypair::new();
+            let wormhole = Pubkey::from_str("worm2ZoG2kUd4vFXhvjh93UUH596ayRfgQ2MgjNMTth")?;
```
jayantk, line 175: (this should also probably be a constant somewhere, and this command also should be configurable per network)

jayantk (approved):
LGTM. left a couple nitpicks but feel free to address later.

pythnet/remote-executor/cli/src/main.rs
tompntn, line 175: It might be nice to store these constants in a struct, and define the CLI actions as methods on the structs.

In general, Rust code should be mainly `impl` definitions on structs, in an OOP style - this helps to organise the code and improve readability/maintainability. Long imperative functions are difficult to parse, understand and maintain.

pythnet/remote-executor/cli/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+
+            let guardian_set = GuardianSet::key(&wormhole, wormhole_config_data.guardian_set_index);
+            let guardian_set_data =
+                GuardianSet::try_from_slice(&rpc_client.get_account_data(&guardian_set)?)?;
+
+            let signature_set_keypair = Keypair::new();
+
+            let vaa = VAA::from_bytes(vaa_bytes.clone())?;
+
+            // RENT HACK STARTS HERE
```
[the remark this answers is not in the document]
ali-bahjati: not important, but i think doing it in the contract after account creation could be easier to understand. (and similar to our hacks everywhere else)

PR #313, opened 2022-09-29
Guibescos/expose attester instructions

guibescos:
Refactor to expose instruction builders for `set_config` and `set_is_active`.

Side effect : pin `pyth-sdk` in `p2w-sdk` to 0.5.0 otherwise I was running into some compilation issues. 
`pyth-solana-sdk` is 0.5.0 in `p2w-sdk` so it should be good


ali-bahjati (approved):
LGTM 🙇 

About the dependency issue. I think using the last version for `pyth-sdk-solana` should fix the problem. But pinning is better.

PR #314, opened 2022-09-29
Guibescos/cli for attester admin

guibescos:
Cli for generating attester governance payloads.


ali-bahjati (approved):
Looks good to me. Thanks for refactoring attester cli to re-use it here.

One question, is there any reason we are not moving this to the remote executor cli? 🤔 (because it does payload generation for some stuff 🤔)

Suggestion for future, it would be better if like attester cli we can pass stuff optionally and fetch the missing ones from on-chain. I know this one doesn't connect to any rpc, so it's hard for now.

guibescos:
I'll try to fix CI first.

ali-bahjati (approved):
Thanks for fixing the CI.

PR #316, opened 2022-09-29
Fix program account not being passed to CPI

guibescos:
I forgot to pass each subinstruction's program_id to the CPI. Tests were passing because I was CPIing into the system program, which is being passed always to the remote-executor.


jayantk (approved):
seems like a test with the non-system program would be nice, but this is fine for now.

PR #320, opened 2022-09-29
Add deployment files

ali-bahjati:
unknown-1337 is for local running or dry-runs (for fork). It should not be there.


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -247,6 +247,6 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
     function deployCommitHash() public pure returns (string memory) {
         // This is a place holder for the commit hash and will be replaced
         // with the commit hash upon deployment.
-        return "9a285cf799ea09eac71704afafc382fc889b0849";
+        return "__DEPLOY_COMMIT_HASH_PLACEHOLER__";
```
jayantk, line 250: why was this a hash previously and a placeholder now?

jayantk (approved):

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: This was supposed to remain the placeholder, which is replaced each time with the hash of current git head. (in Deploying.md). I added it to github by mistake in my last deployment.

PR #329, opened 2022-10-06
Add `update_price_feeds_if_fresh_with_funder` wrapper function

tompntn:
This is a wrapper around `update_price_feeds_if_fresh` which allows users to pass a signer in, allowing it to be used as an entry function.


ali-bahjati (approved):
:shipit:

PR #330, opened 2022-10-06
Drozdziak1/p2w attest cont mapping reload

drozdziak1:
This changeset implements continuous reloads of the mapping account, without interruptions. The complexity bump is there but still appears worth the hassle.


solana/pyth2wormhole/client/src/main.rs
```diff
@@ -301,56 +272,181 @@ async fn handle_attest(
...
+        );
 
-    info!("Spinning up attestation sched jobs");
+        info!(
+            "{} symbol groups read, dividing into batches",
+            attestation_cfg.symbol_groups.len(),
+        );
 
-    let results = futures::future::join_all(attestation_sched_futs).await; // May never finish for daemon mode
+        let mut batches: Vec<_> = attestation_cfg
```
ali-bahjati, line 316: I think we can compute these only when needed (when config has changed below). Performance-wise it's not much different but it will result in more clear logs.

ali-bahjati:
I think the code is technically correct and I don't see any issue (just why tilt is complaining?).

My concern is mostly the complexity of this code, it is very complex now.

Now that we are not using constraints to update prices (and we are doing it as fast as possible) we can let go of the constraint check totally, and then have it as two loop. One loop is always attesting based on the groups. One loop is crawling mapping accounts and updating the groups if needed. This will be reusing everything here and removing many existing code to make it simpler. 

I think we can merge this one and go for simplifying it like ^ in another PR. 

Let's hear what other folks think and if they agree I'll approve this PR.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -301,56 +272,181 @@ async fn handle_attest(
...
-        .enumerate()
-        .filter_map(|(idx, r)| {
-            r.as_ref()
-                .err()
-                .map(|e| format!("Error {}: {:#?}\n", idx + 1, e))
-        })
-        .collect();
+        info!("Spinning up attestation sched jobs");
+        if daemon {
+            let mut hasher = DefaultHasher::new();
```
jayantk, line 369: I think you want a cryptographic hash here -- if you get a hash collision, the code below isn't going to do the right thing.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -301,56 +272,181 @@ async fn handle_attest(
...
+                    old_handle.abort();
+
+                    // The just started futures become the on-going attestation state
+                    old_sched_futs_state = Some((new_sched_futs_handle, new_cfg_hash));
+                }
+            } else {
+                old_sched_futs_state = Some((tokio::spawn(joined_sched_futs), new_cfg_hash));
+            }
+
+        // Sum up elapsed time, wait for next run accurately
```
jayantk, line 394: this code isn't indented properly

jayantk: (needs one more tab from here to the .await)

ali-bahjati (commented):
I talked with Stan about this PR and here is a summary:
- The current code can be further simplified if we can move out the non-deamon logic and it will be much easier to reason about. 
- Also moving some logic in the long functions to separate functions will improve the readability of the code
- There is no way forward to separate the loops with current design. To do it we should step back and change the whole structure of code. Given that it will be a huge refactor it's not clear for me that the outcome is worth the effort.

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -134,9 +144,10 @@ pub struct AttestationConditions {
     #[serde(default = "default_max_batch_jobs")]
     pub max_batch_jobs: usize,
 
-    /// Trigger attestation if price changes by the specified percentage.
+    /// Trigger attestation if price changes by the specified
+    /// percentage, expressed in integer parts per thousand.
     #[serde(default)]
-    pub price_changed_pct: Option<f64>,
+    pub price_changed_ppt: Option<u64>,
```
jayantk, line 150: i really really think you should use basis points here. 1 / 1000th is a weird unit for prices, but 1 / 10000 (1 bp) is very standard. This configuration option is going to confuse people.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -301,56 +272,181 @@ async fn handle_attest(
...
-            retry_interval,
-            daemon,
-            rpc_cfg.clone(),
-            p2w_addr,
-            config.clone(),
-            Keypair::from_bytes(&payer.to_bytes()).unwrap(),
-            message_q_mtx.clone(),
-        )
-    });
+    // This loop governs mapping account reloads in daemon mode. It is
```
jayantk, line 283: this function is long and the logic is hard to understand because it supports all these different options around daemon / no daemon. 

I think it would be good to refactor this by extracting out helper functions and then possibly having two separate top-level functions for daemon / non-daemon mode. Each of those functions could use the helpers to avoid code duplication.

jayantk (commented):
I think the cryptographic hash thing should be fixed now. 

Aside from that, I think @ali-bahjati 's comment above is right. Let's merge this, but would be good to follow up with a refactoring PR to make the logic easier to understand.

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -134,9 +172,10 @@ pub struct AttestationConditions {
     #[serde(default = "default_max_batch_jobs")]
     pub max_batch_jobs: usize,
 
-    /// Trigger attestation if price changes by the specified percentage.
+    /// Trigger attestation if price changes by the specified
+    /// percentage, expressed in integer base points (1bp = 0.01%)
     #[serde(default)]
-    pub price_changed_pct: Option<f64>,
+    pub price_changed_bp: Option<u64>,
```
jayantk, line 178: nitpick: usually you abbreviate basis points to "bps"

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -225,181 +245,307 @@ async fn main() -> Result<(), ErrBox> {
...
+                }
+            }
+        }
+        debug!(
+            "Attestation config (includes mapping accounts):\n{:#?}",
+            attestation_cfg
+        );
+
+        // Hash currently known config
+        let mut hasher = DefaultHasher::new();
```
jayantk, line 301: i think i left a comment last time about using a cryptographic hash here. The code below depends on no collisions for correctness.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -225,181 +245,307 @@ async fn main() -> Result<(), ErrBox> {
...
+                // look up the mapping should not crash the attester
+                Err(e) => {
+                    error!("Could not crawl mapping {}: {:?}", mapping_addr, e);
+                }
+            }
+        }
+        debug!(
+            "Attestation config (includes mapping accounts):\n{:#?}",
+            attestation_cfg
+        );
```
jayantk, line 298: this debug statement needs to be in the `if` i think

jayantk: also the code in the if statement above is duplicated across this method and the one below. consider extracting out a helper function.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -225,181 +245,307 @@ async fn main() -> Result<(), ErrBox> {
...
+
+    let batch_count = batches.len();
+
+    // For enforcing min_msg_reuse_interval_ms, we keep a piece of
+    // state that creates or reuses accounts if enough time had
+    // passed
+    let message_q_mtx = Arc::new(Mutex::new(P2WMessageQueue::new(
+        Duration::from_millis(attestation_cfg.min_msg_reuse_interval_ms),
+        attestation_cfg.max_msg_accounts as usize,
+    )));
```
jayantk, line 495: I think the state of this queue gets reset each time the attestation batches change. Does that mean we can end up reusing accounts more frequently than intended each time the mapping account reloads (and the batches get recreated)?

jayantk: this is not a big deal given that the mapping accounts reload very infrequently. maybe leave a // todo comment if this is the case.

jayantk (approved):
nice! I left you a couple minor comments. I suggest addressing them in a fast follow up PR so that it's easier to review.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: Nice catch

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I left it here to print the config regardless, it's the "includes mapping accounts" part that's new.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: That makes sense. I think the std HashMap has internal collision detection that isn't part of the hasher. I assumed it would be enough, but a nicer 32-byte scheme will fit better here.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: This was an easy fix, just moved the mutex constructor to before the loop so that it and its state is reused. Async runtimes _must_ properly deallocate variables in scope of an aborted task (to avoid counter-intuitive mem leaks), so it should work well now and in the future

PR #338, opened 2022-10-10
pyth2wormhole-client:should_resend(): optimize interval-only batches

drozdziak1:
Most attestation conditions in prod are interval-only. This change ensures that interval-only batches don't waste time on the on-chain state lookup request and data comparisons.

### DIFF
--- a/solana/pyth2wormhole/client/src/attestation_cfg.rs
+++ b/solana/pyth2wormhole/client/src/attestation_cfg.rs
@@ -152,11 +152,11 @@ impl AttestationConditions {
         let AttestationConditions {
             min_interval_secs: _min_interval_secs,
             max_batch_jobs: _max_batch_jobs,
-            price_changed_ppt,
+            price_changed_pct,
             publish_time_min_delta_secs,
         } = self;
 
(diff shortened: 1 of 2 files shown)


solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -144,6 +144,22 @@ pub struct AttestationConditions {
     pub publish_time_min_delta_secs: Option<u64>,
 }
 
+impl AttestationConditions {
+    /// Used by should_resend() to check if it needs to make the expensive RPC request
+    pub fn is_onchain(&self) -> bool {
```
ali-bahjati, line 149: Maybe a more clear name? `need_onchain_lookup` 🤔

ali-bahjati (approved):
Looks good to me! Very clever way of handling it.
Since you have profiled the code, can you also comment how much improvement this change has?

drozdziak1:
Copying from the thread:
should_resend(): 50% -> 2% flamegraph samples
lock_and_make_rpc(): 3% -> 22% flamegrpah samples

profiled using the `tracing` crate and the `tracing_flame` flamegraph compat layer

PR #339, opened 2022-10-10
Start all pyth tilt pods by default

tompntn:
This removes the need to pass in `--pyth` to `tilt up`.


ali-bahjati (approved):

ali-bahjati:
Can you remove the tilt `pyth` variable entirely?

PR #341, opened 2022-10-11
Delete unused imports in Aptos

ranger8888:
There were quite a few unused imports in the Pyth Move contracts. This can be an issue if users are depending on Pyth during unit testing.


tompntn:
Closing in favour of #342

PR #346, opened 2022-10-12
[eth] Add set fee migration step + deployment process improvement

ali-bahjati:
This PR adds a migration step to execute the governance instruction for setting fees on target chains. I have executed them on all of our testnet chains except aurora_devnet (which the rpc seems to busy)

Apart from that, this PR significantly improves the deployment process by:
- Moving cluster variables to a separate `.env.cluster.{mainnet|testnet}` file.
- Adding `deploy.sh` script to deploy the changes.
- Setting an example on how VAAs should be executed (this migration step)

Also it comments out migration 12 that used a bad practice of creating/executing VAAs.

Note: please review this PR commit-by-commit as there are couple of mechanical changes.


ethereum/deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "Adding network metadata to the contract"
+# Merge the network addresses into the artifacts, if some contracts are already deployed.
+npx apply-registry
+
+while [[ $# -ne 0 ]]; do
+    NETWORK=$1
+    shift
+
+    echo "=========== Deploying to ${NETWORK} ==========="
```
tompntn, line 22: Could we add a docstring at the top of this file explaining the possible values of `$NETWORK`, and an example invocation of this script?

ethereum/deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    echo "Migrating..."
+    npx truffle migrate --network $MIGRATIONS_NETWORK
+
+    echo "Deployment to $NETWORK finished successfully"
+done
+
+echo "=========== Cleaning up ==========="
+echo "Reverting back the deploy commit hash..."
+sed -i "s/$(git rev-parse HEAD)/__DEPLOY_COMMIT_HASH_PLACEHOLER__/g" ./contracts/pyth/Pyth.sol
```
tompntn, line 36: I still think this is quite dangerous/brittle.

ethereum/Deploying.md
```diff
@@ -8,35 +8,11 @@ This is the deployment process:
...
-# Then you can use the VAA (in hex) to execute the governance instruction. To do so, run:
-$ npx apply-registry # apparently a deployProxyImpl messes up with correct address of proxy
-$ npx truffle console --network $MIGRATIONS_NETWORK
-> let p = await PythUpgradable.deployed()
-> await p.executeGovernanceInstruction("<VAA in hex like: 0x123002342352>");
+# Perform the migration
+# You might need to repeat this script because of busy RPCs. Repeating would not cause any problem even
+# if the changes are already made. Also, sometimes the gases are not adjusted. Please update them with
+# the network explorer gas tracker. Tips in Troubleshooting section below can help in case of any error.
+./deploy.sh <network_a> <network_b> <...>
```
tompntn, line 15: Could we add some examples of how to call this script with concrete values here (in comments)? A couple of testnet networks, and a couple of mainnet networks.

tompntn (approved):
🙇

ethereum/migrations/prod-receiver/3_deploy_pyth.js
```diff
@@ -1,5 +1,9 @@
-require('dotenv').config({ path: "../.env" });
-const bs58 = require("bs58");
+const dotenv = require("dotenv")
+dotenv.config({ path: "../../.env" });
+
+if (process.env.CLUSTER !== undefined) {
+  dotenv.config({ path: `../../.env.cluster.${process.env.CLUSTER}`});
+}
```
jayantk, line 6: extract this code out to a utility function like `load_dotenv()`

also, fwiw I think using a different config file solution is probably a good idea here. If we used something like HOCON that supports storing dictionaries in the config, then we wouldn't have to have separate config files for each network. `.env` is a rather inflexible standard, even though it is the "standard" for javascript.

ethereum/migrations/prod/12_pyth_set_fee_1_wei.js
```diff
@@ -0,0 +1,21 @@
+const dotenv = require("dotenv")
+dotenv.config({ path: "../../.env" });
+
+if (process.env.CLUSTER !== undefined) {
+  dotenv.config({ path: `../../.env.cluster.${process.env.CLUSTER}`});
+}
+
+const setFeeVaa = process.env.MIGRATION_12_SET_FEE_VAA;
```
jayantk, line 8: is this going to go into the dotenv file and be stored permanently? (I think that it should be)

jayantk (approved):
LGTM, but please do extract out a helper function for that repeated dotenv loading block...

ethereum/deploy.sh
ali-bahjati: Yeah, I'll address this in another PR. Or I might remove this commit hash function entirely (in favor of version)

ethereum/migrations/prod-receiver/3_deploy_pyth.js
ali-bahjati: I looked up HOCON and although it’s interesting, it doesn’t have good adoption. Personally I like it’s features (mostly extending another config) but I don’t like it’s flexibility and it’s frustrating.
It had two parser pkgs in JS. The bigger one does not support include (which we need) and the other one was fairly new and not used much.

ethereum/migrations/prod/12_pyth_set_fee_1_wei.js
ali-bahjati: Yes it's gonna be there forever. In general I like keeping all the changes to contract.

PR #347, opened 2022-10-13
Add link to on-demand and update fee docs

tompntn:
Linking to these directly from the code, as this is most developers first point of entry.

### DIFF
--- a/aptos/contracts/sources/pyth.move
+++ b/aptos/contracts/sources/pyth.move
@@ -288,9 +288,8 @@ module pyth::pyth {
     /// 
     /// Important: Pyth uses an on-demand update model, where consumers need to update the 
     /// cached prices before using them. Please read more about this at https://docs.pyth.network/consume-data/on-demand.
-    /// It is therefore recommended to call update_price_feeds() to update the cached price
-    /// before calling get_price(), as get_price() will abort if the cached price is older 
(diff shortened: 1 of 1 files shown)


aptos/contracts/sources/pyth.move
```diff
@@ -279,8 +286,10 @@ module pyth::pyth {
     /// Please refer to the documentation at https://docs.pyth.network/consumers/best-practices for
     /// how to how this price safely.
     /// 
-    /// Important: it is recommended to call update_price_feeds() to update the cached price
-    /// before calling this function, as get_price() will abort if the cached price is older 
+    /// Important: Pyth uses an on-demand update model, where consumers need to update the 
+    /// cached prices before using them. Please read more about this at https://docs.pyth.network/consume-data/on-demand.
+    /// It is therefore recommended to call update_price_feeds() to update the cached price
+    /// before calling get_price(), as get_price() will abort if the cached price is older
```
jayantk, line 292: "It is therefore recommended" is indirect. How about "

"Pyth uses an on-demand update model where callers must update the cached price feeds before reading them. Please see ... This function is likely to abort unless you call update_price_feeds() to update the cached price feeds beforehand."

jayantk (approved):
just one minor wording suggestion

PR #348, opened 2022-10-14
Abehjati/eth-contract-improvement

ali-bahjati:


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -236,17 +236,7 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
         return validTimePeriodSeconds();
     }
 
-    function isPyth() internal pure returns (bool) {
```
jayantk, line 239: i think you can't delete this until you upgrade the on-chain contracts (because the currently deployed version checks it)

jayantk (commented):
lgtm modulo the comment. I think it would be good to have another review from someone with better solidity knowledge than me as well.

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: This one is very interesting but this check was not working as I wasn't using `this.isPyth` before.

njkumr (commented):
lgtm

ethereum/contracts/pyth/Pyth.sol
ali-bahjati: (I also commented in the `this.pythUpgradableMagic()` method why this is used)

jayantk (approved):

PR #351, opened 2022-10-18
Fix corner case of product with no price

ali-bahjati:


solana/pyth2wormhole/client/src/lib.rs
```diff
@@ -427,6 +427,11 @@ pub async fn crawl_pyth_mapping(
 
             let mut price_addr = prod.px_acc.clone();
 
+            // the product might have no price, can happen in tilt due to race-condition, failed tx to add price, ...
+            if price_addr == Pubkey::default() {
+                continue;
```
drozdziak1, line 432: Nice find and thanks for correcting this after me! Can we also have a `debug!()` log here? priceless products are kind of weird and I think we should know about them if they appear

drozdziak1 (approved):

PR #352, opened 2022-10-18
Run remote exec CI only if it's changed

ali-bahjati:
To save more GH actions time + have faster CI checks (sometimes it takes longer than tilt!)

Afaik this code does not depend on any other codebase here, so running it only when anything within its directory has changed will work as expected.


guibescos (approved):
Sure

PR #355, opened 2022-10-18
[eth] Make governance transfer less error-prone

ali-bahjati:
This PR changes the way we transfer the governance data source. It introduced a new state value `governanceDataSourceIndex` that should be increased each time we transfer the governance to a newer one (extra sanity check).

Consider that we have an `oldGovenrance` and a `newGovernance`. The data wire is that the `newGovernance` creates a `TransferGovernanceDataSourceClaim` and sends to wormhole to create `claimVaa`. Then the `oldGovernance` creates a `TransferGovernanceDataSource` instruction that wraps the `claimVaa` and sends it to wormhole to create the transfer VAA. 

Tom suggested wrapping instead of taking them separately because the result would be one VAA and it fits the `executeGovernanceInstruction` signature to also execute it.

### DIFF
--- a/ethereum/contracts/pyth/PythGovernance.sol
+++ b/ethereum/contracts/pyth/PythGovernance.sol
@@ -77,7 +77,10 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
 
         TransferGovernanceDataSourcePayload memory payload = parseTransferGovernanceDataSourcePayload(encodedPayload);
 
-        // Make sure the VAA is valid
+        // Make sure the claimVaa is a valid VAA with TransferGovernanceDataSourceClaim governance message
+        // If it's valid then its emitter can take over the governance from the current emitter.
+        // The VAA is checked here to ensure that the new governance data source is valid and can send message
+        // through wormhole.
         (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(payload.claimVaa);
         require(valid, reason);
 
@@ -88,6 +91,7 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
 
         TransferGovernanceDataSourceClaimPayload memory claimPayload = parseTransferGovernanceDataSourceClaimPayload(gi.payload);
 
+        // Governance data source index is used to prevent replay attacks, so a claimVaa cannot be used twice.
         require(governanceDataSourceIndex() < claimPayload.governanceDataSourceIndex, 
             "cannot upgrade to an older governance data source");
 
@@ -96,6 +100,8 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
         PythInternalStructs.DataSource memory newGovernanceDS = PythInternalStructs.DataSource(vm.emitterChainId, vm.emitterAddress);
 
         setGovernanceDataSource(newGovernanceDS);
+
+        // Setting the last executed governance to the claimVaa sequence to avoid using older sequences.
         setLastExecutedGovernanceSequence(vm.sequence);
 
         emit GovernanceDataSourceSet(oldGovernanceDatSource, governanceDataSource(), lastExecutedGovernanceSequence());
--- a/ethereum/contracts/pyth/PythGovernance.sol
+++ b/ethereum/contracts/pyth/PythGovernance.sol
@@ -45,16 +45,16 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
         if (gi.action == GovernanceAction.UpgradeContract) {
             require(gi.targetChainId != 0, "upgrade with chain id 0 is not possible");
             upgradeContract(gi.payload);
-        } else if (gi.action == GovernanceAction.TransferGovernanceDataSourceAuthorize) {
-            TransferGovernanceDataSourceAuthorize(gi.payload);
+        } else if (gi.action == GovernanceAction.AuthorizeGovernanceDataSourceTransfer) {
+            AuthorizeGovernanceDataSourceTransfer(gi.payload);
         } else if (gi.action == GovernanceAction.SetDataSources) {
             setDataSources(gi.payload);
         } else if (gi.action == GovernanceAction.SetFee) {
             setFee(gi.payload);
         } else if (gi.action == GovernanceAction.SetValidPeriod) {
             setValidPeriod(gi.payload);
-        } else if (gi.action == GovernanceAction.TransferGovernanceDataSourceRequest) {
-            revert("TransferGovernanceDataSourceRequest can be only part of TransferGovernanceDataSourceAuthorize message");
+        } else if (gi.action == GovernanceAction.RequestGovernanceDataSourceTransfer) {
+            revert("RequestGovernanceDataSourceTransfer can be only part of AuthorizeGovernanceDataSourceTransfer message");
         } else {
             revert("invalid governance action");
         }
@@ -71,12 +71,12 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
 
     // Transfer the governance data source to a new value with sanity checks
     // to ensure the new governance data source can manage the contract.
-    function TransferGovernanceDataSourceAuthorize(bytes memory encodedPayload) internal {
+    function AuthorizeGovernanceDataSourceTransfer(bytes memory encodedPayload) internal {
         PythInternalStructs.DataSource memory oldGovernanceDatSource = governanceDataSource();
 
-        TransferGovernanceDataSourceAuthorizePayload memory payload = parseTransferGovernanceDataSourceAuthorizePayload(encodedPayload);
+        AuthorizeGovernanceDataSourceTransferPayload memory payload = parseAuthorizeGovernanceDataSourceTransferPayload(encodedPayload);
 
-        // Make sure the claimVaa is a valid VAA with TransferGovernanceDataSourceRequest governance message
+        // Make sure the claimVaa is a valid VAA with RequestGovernanceDataSourceTransfer governance message
         // If it's valid then its emitter can take over the governance from the current emitter.
         // The VAA is checked here to ensure that the new governance data source is valid and can send message
         // through wormhole.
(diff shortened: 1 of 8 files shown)


ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -45,14 +45,17 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
         if (gi.action == GovernanceAction.UpgradeContract) {
             require(gi.targetChainId != 0, "upgrade with chain id 0 is not possible");
             upgradeContract(gi.payload);
-        } else if (gi.action == GovernanceAction.SetGovernanceDataSource) {
-            setGovernanceDataSource(gi.payload);
+        } else if (gi.action == GovernanceAction.TransferGovernanceDataSource) {
+            require(gi.targetChainId == 0, "governance data source change should be applied to all chains");
```
tompntn, line 49: This seems like an unnecessary restriction, I'm not sure what this buys us.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -43,9 +44,15 @@ contract PythGovernanceInstructions {
         address newImplementation;
     }
 
-    struct SetGovernanceDataSourcePayload {
-        PythInternalStructs.DataSource newGovernanceDataSource;
-        uint64 initialSequence;
+    struct TransferGovernanceDataSourcePayload {
+        // claimVaa is a VAA with TransferGovernanceDataSourceClaimOwnership
```
tompntn, line 48: Should this be `TransferGovernanceDataSourceClaim`?

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -43,9 +44,15 @@ contract PythGovernanceInstructions {
...
+        // governance instruction (header + payload) that is created from
+        // the second multisig
```
tompntn, line 50: "second multisig" -> "second governance source".

Might be worth adding a comment here about why we are doing this (i.e. to ensure that we can generate valid governance vaas from the new data source).

tompntn: Also would be worth a proper docstring here describing the outline and the purpose of action 1 and 2.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -67,13 +70,33 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
 
     function upgradeUpgradableContract(UpgradeContractPayload memory payload) virtual internal;
 
-    function setGovernanceDataSource(bytes memory encodedPayload) internal {
-        SetGovernanceDataSourcePayload memory payload = parseSetGovernanceDataSourcePayload(encodedPayload);
-
+    // Transfer the governance data source to a new value with sanity checks
+    // to ensure the new governance data source can manage the contract.
+    function transferGovernanceDataSource(bytes memory encodedPayload) internal {
```
tompntn, line 75: Worth re-iterating here the purpose of the second VAA.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -45,14 +45,17 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
         if (gi.action == GovernanceAction.UpgradeContract) {
             require(gi.targetChainId != 0, "upgrade with chain id 0 is not possible");
             upgradeContract(gi.payload);
-        } else if (gi.action == GovernanceAction.SetGovernanceDataSource) {
-            setGovernanceDataSource(gi.payload);
+        } else if (gi.action == GovernanceAction.TransferGovernanceDataSource) {
+            require(gi.targetChainId == 0, "governance data source change should be applied to all chains");
+            transferGovernanceDataSource(gi.payload);
```
tompntn, line 50: Minor point, but we could parse the payload before calling the handlers:
```
transferGovernanceDataSource(parseTransferGovernanceDataSourcePayload(gi.payload));
```
which would then allow us to use meaningful types for the handler function signatures.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -67,13 +70,33 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
...
+    // Transfer the governance data source to a new value with sanity checks
+    // to ensure the new governance data source can manage the contract.
+    function transferGovernanceDataSource(bytes memory encodedPayload) internal {
         PythInternalStructs.DataSource memory oldGovernanceDatSource = governanceDataSource();
 
-        setGovernanceDataSource(payload.newGovernanceDataSource);
-        setLastExecutedGovernanceSequence(payload.initialSequence);
+        TransferGovernanceDataSourcePayload memory payload = parseTransferGovernanceDataSourcePayload(encodedPayload);
+
+        // Make sure the VAA is valid
```
tompntn, line 80: Could we make it clear in this comment that this is a second, claim, VAA? Maybe even extract everything to do with the second VAA into it's own function with a docstring.

tompntn: Worth re-iterating here that the purpose of ensuring the claim VAA is valid is to ensure we can emit governance messages from the new source.

tompntn (commented):
Nice!

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -67,13 +70,33 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
...
+        (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(payload.claimVaa);
+        require(valid, reason);
+
+        GovernanceInstruction memory gi = parseGovernanceInstruction(vm.payload);
+        require(gi.targetChainId == 0, "governance data source change claim vaa should be applied to all chains");
```
jayantk, line 85: why should we care about this targetChainId at all? 

also please document all the requirements for creating a proper set of governance transfer messages above the struct definitions for their payloads.

ethereum/contracts/pyth/PythGovernance.sol
```diff
@@ -67,13 +70,33 @@ abstract contract PythGovernance is PythGetters, PythSetters, PythGovernanceInst
...
+        require(gi.action == GovernanceAction.TransferGovernanceDataSourceClaim,
+            "governance data source change inner vaa is not of claim action type");
+
+        TransferGovernanceDataSourceClaimPayload memory claimPayload = parseTransferGovernanceDataSourceClaimPayload(gi.payload);
+
+        require(governanceDataSourceIndex() < claimPayload.governanceDataSourceIndex, 
+            "cannot upgrade to an older governance data source");
+
+        setGovernanceDataSourceIndex(claimPayload.governanceDataSourceIndex);
```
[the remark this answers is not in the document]
ali-bahjati: You are not missing anything here. As I said in the PR description its an extra sanity check. It's not preventing any obvious replay attacks. We want to reduce the chance of mistake anywhere possible.

Based on my understanding the goal of governance is the be restricted as much as possible. It is better to have an index to be able to point to the version. I believe that we are making it very difficult and I personally don't like it, but I think this is the security decision. 

Now this is the case that a replay attack it can happen which is non-obvious. Imagine we go from A to B and from B to A again.
Then after that one can use the old claim of B (your naming: approval) and wrap it into a new transfer of A (your naming: request) and do transfer the governance to B again, but with the old sequence coming as the result.

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: Based on my understanding the goal of governance is the be restricted as much as possible.

IMO it does not make sense that this action gets applied to only one single chain, and it has overhead of reverting it back. If you see any possible use case that it can happen then I will remove it, otherwise I think it should be there. It's not a security issue but it's avoiding possible mistakes and enforcing the behavior of this action.

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: Nice point, I wish you told it in the earlier PR for governance :D

I'll do it in a separate PR as it touches other places too.

ali-bahjati:
> Also, changing this logic scares me, so let's be sure to thoroughly test the upgrade from the existing contract to this version, plus upgrades with the new logic. (I see the unit test covers the second thing, but the first one seems like we'll have to do on a testnet)

@jayantk I don't get your first point here, it's only a contract upgrade and the old logic is not used at all

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -43,9 +44,15 @@ contract PythGovernanceInstructions {
...
+        bytes claimVaa;
+    }
+
+    struct TransferGovernanceDataSourceClaimPayload {
```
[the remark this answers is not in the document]
ali-bahjati: I don't like this naming, because it implicitly means that these are two separate actions where as one is included in another one.

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: Explained my reasoning in another comment. Will decide based on your answer.

I'll change the js sdk so this is default to 0, so no documentation is needed anymore.

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: I don't like extracting it out to another function. The vaa check is the only logic inside this function and it doesn't make sense to add another one.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -43,9 +44,15 @@ contract PythGovernanceInstructions {
...
     }
 
-    struct SetGovernanceDataSourcePayload {
-        PythInternalStructs.DataSource newGovernanceDataSource;
-        uint64 initialSequence;
+    struct TransferGovernanceDataSourcePayload {
+        // claimVaa is a VAA with TransferGovernanceDataSourceClaimOwnership
+        // governance instruction (header + payload) that is created from
+        // the second multisig
+        bytes claimVaa;
```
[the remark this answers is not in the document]
ali-bahjati: Changed the comment, not feeling good about renaming the claimVaa. It depends on what we decide for the struct naming.

njkumr (approved):
lgtm

ethereum/contracts/pyth/PythGovernance.sol
jayantk: i think there are reasonable cases where we may want to shard governance authority to multiple emitters. e.g., we could make a council that owns the EVM contracts but not the Aptos contracts. I think we

ethereum/contracts/pyth/PythGovernanceInstructions.sol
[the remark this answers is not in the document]
jayantk: There are two separate actions -- one emitter first has to produce the request, and then another emitter has to approve it. The approval points to the request that it's approving. I don't think there's anything weird about that. 

I don't like the current naming because you can't tell which vaa is which. I thought the claim would be the parent VAA when i was reading this initially.

Happy to brainstorm other names with you if you really don't like request/response.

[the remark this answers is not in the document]
tompntn: I think claim is more semantically correct. How about "Authorize"/"Claim"?

ethereum/contracts/pyth/PythGovernance.sol
ali-bahjati: Ah got it, never thought of such a case. Ok then. I'll fix it.

ethereum/contracts/pyth/PythGovernanceInstructions.sol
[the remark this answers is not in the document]
ali-bahjati: I like `Authorize`/`Claim`. Thanks Tom!

jayantk (approved):

PR #356, opened 2022-10-18
[eth] Some contract improvements

ali-bahjati:
This PR:
- Updates contract to use the latest solidity sdk and updates `getUpdateFee` method to get the data instead of size (+ addressing tests + relayer). Keeps the old one for backward compatibility.
- Removes part of the code to transfer back extra paid fee amount (to be consistent with other chains that do not support it. e.g., aptos)

### DIFF
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -57,7 +57,7 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
     }
 
     function updatePriceFeeds(bytes[] calldata updateData) public override payable {
-        uint requiredFee = getUpdateFee(updateData.length);
+        uint requiredFee = getUpdateFee(updateData);
         require(msg.value >= requiredFee, "insufficient paid fee amount");
         payable(msg.sender).transfer(msg.value - requiredFee);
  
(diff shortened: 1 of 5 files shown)


ethereum/contracts/pyth/Pyth.sol
```diff
@@ -68,8 +68,8 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
         emit UpdatePriceFeeds(msg.sender, updateData.length, requiredFee);
     }
 
-    function getUpdateFee(uint updateDataSize) public override view returns (uint feeAmount) {
-        return singleUpdateFeeInWei() * updateDataSize;
+    function getUpdateFee(bytes[] calldata updateData) public override view returns (uint feeAmount) {
```
tompntn, line 70: Isn't this a breaking API change? How are you planning to do the migration?

ali-bahjati: evm != aptos ser. Evm doesn't have upgrades native to have the notion of breaking at all. So in theory nothing is breaking and you can even break the storage if you want to.

Calling a non-existing method results in a revert without any message and that could be a problem if there were protocols already live on top of us, but as there are none, this is fine.

tompntn: Yeah, I was concerned about breaking protocols that are already live using us.

tompntn (approved):

jayantk (changes requested):
I don't think we can make breaking API changes anymore. We don't know for sure that nobody is using us, and we need to be careful.

I think a better solution here would be to mark the old method as deprecated and add a new method that took the full payload.

We could also promise that the old method returns an overestimate of the necessary fee, so everything will continue to work for any users of the old method.

njkumr (approved):
lgtm

tompntn (approved):

jayantk (approved):

PR #361, opened 2022-10-24
Abehjati/simplify-multisig-cli

ali-bahjati:
Refactor the cli specifically to require less arguments (vault and rpc is hard-coded).

It also refactors the code a little.

This code can be divided into multiple files later.


jayantk (commented):
I think @cctdaniel is the right person to approve this -- he's much closer to this code than me

cctdaniel (approved):
thank you sir this is way cleaner

PR #363, opened 2022-10-26
Add verify command for wormhole msg

ali-bahjati:
This command verifies that the given transaction is indeed created by the given payload.

The reason behind this is that it is not possible to verify a wormhole message in the UI. This can do that on the cli.


jayantk (approved):
lgtm. as we discussed on zoom, i think the right goal to strive for is to have a way to take the squads instruction and describe the content of the payload in plain english. this is a step in a good direction though so i'm approving.

cctdaniel (approved):
🙇🏻

PR #366, opened 2022-10-31
xc-tool: Initial commit

drozdziak1:


xc-tool/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+fn main() -> Result<(), ErrBox> {
+    let cli = Cli::parse();
+
+    // Handle interactive mode separately
+    if cli.action == Action::Interactive {
+        start_cli_interactive(cli)?;
+    } else {
+        handle_action_noninteractive(&cli)?;
+    }
+    Ok(())
```
tompntn, line 23: It might be best to keep the interactive stuff stashed in a separate branch, and and merge it in later if we decide to go with Rust for the tool. The code should initially be a very simple sanity-check of using the available Rust SDKs.

xc-tool/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+/// The following code is reused by interactive mode. Interactive mode
+/// is assumed to already be detected at top-level in main, making it an invalid action.
+pub fn handle_action_noninteractive(cli: &Cli) -> Result<(), ErrBox> {
+    match &cli.action {
+        // It makes no sense starting interactive already inside it
+        Action::Interactive => {
+            return Err(format!("Bruh...?").into());
```
jayantk, line 65: lol

jayantk (approved):

PR #371, opened 2022-11-02
[test-publisher]: Set cap on dyn added symbols

ali-bahjati:
Currently the test publisher creates a new symbol every two minutes without any cap and running tilt for a long time will eventually drain all the resources.

This change sets a cap of `PYTH_TEST_SYMBOL_COUNT` on the dynamically added symbols. It means that the test publisher starts with `PYTH_TEST_SYMBOL_COUNT` and will add `PYTH_TEST_SYMBOL_COUNT` new symbols after that over time.

### DIFF
--- a/third_party/pyth/pyth_publisher.py
+++ b/third_party/pyth/pyth_publisher.py
@@ -147,18 +147,22 @@ next_new_symbol_id = PYTH_TEST_SYMBOL_COUNT
 last_new_sym_added_at = time.monotonic()
 
 with ThreadPoolExecutor() as executor: # Used for async adding of products and prices
+    dynamicly_added_symbols = 0
     while True:
         for sym in HTTP_ENDPOINT_DATA["symbols"]:
             publisher_random_update(sym["price"])
 
-        # Add a symbol if new symbol interval configured
-        if PYTH_NEW_SYMBOL_INTERVAL_SECS > 0:
+        # Add a symbol if new symbol interval configured. This will add a new symbol if PYTH_NEW_SYMBOL_INTERVAL_SECS
+        # is passed since adding the previous symbol. The second constraint ensures that
+        # at most PYTH_TEST_SYMBOL_COUNT new price symbols are created.
+        if PYTH_NEW_SYMBOL_INTERVAL_SECS > 0 and dynamicly_added_symbols > PYTH_TEST_SYMBOL_COUNT:
             # Do it if enough time passed
             now = time.monotonic()
             if (now - last_new_sym_added_at) >= PYTH_NEW_SYMBOL_INTERVAL_SECS:
                 executor.submit(add_symbol, next_new_symbol_id) # Returns immediately, runs in background
                 last_sym_added_at = now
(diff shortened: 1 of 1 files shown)


third_party/pyth/pyth_publisher.py
```diff
@@ -147,18 +147,22 @@ def add_symbol(num: int):
 last_new_sym_added_at = time.monotonic()
 
 with ThreadPoolExecutor() as executor: # Used for async adding of products and prices
+    dynamicly_added_symbols = 0
```
drozdziak1, line 150:
```suggestion
    dynamically_added_symbols = 0
```

third_party/pyth/pyth_publisher.py
```diff
@@ -147,18 +147,22 @@ def add_symbol(num: int):
...
     while True:
         for sym in HTTP_ENDPOINT_DATA["symbols"]:
             publisher_random_update(sym["price"])
 
-        # Add a symbol if new symbol interval configured
-        if PYTH_NEW_SYMBOL_INTERVAL_SECS > 0:
+        # Add a symbol if new symbol interval configured. This will add a new symbol if PYTH_NEW_SYMBOL_INTERVAL_SECS
+        # is passed since adding the previous symbol. The second constraint ensures that
+        # at most PYTH_TEST_SYMBOL_COUNT new price symbols are created.
+        if PYTH_NEW_SYMBOL_INTERVAL_SECS > 0 and dynamicly_added_symbols > PYTH_TEST_SYMBOL_COUNT:
```
drozdziak1, line 158:
```suggestion
        if PYTH_NEW_SYMBOL_INTERVAL_SECS > 0 and dynamically_added_symbols > PYTH_TEST_SYMBOL_COUNT:
```

third_party/pyth/pyth_publisher.py
```diff
@@ -147,18 +147,22 @@ def add_symbol(num: int):
...
             # Do it if enough time passed
             now = time.monotonic()
             if (now - last_new_sym_added_at) >= PYTH_NEW_SYMBOL_INTERVAL_SECS:
                 executor.submit(add_symbol, next_new_symbol_id) # Returns immediately, runs in background
                 last_sym_added_at = now
                 next_new_symbol_id += 1
+                dynamicly_added_symbols += 1
```
drozdziak1, line 165:
```suggestion
                dynamically_added_symbols += 1
```

drozdziak1 (approved):
just a typo nit

third_party/pyth/pyth_publisher.py
drozdziak1: I think this should add if it hasn't added enough, instead of acting when there's more than enough.

drozdziak1 (changes requested):

third_party/pyth/pyth_publisher.py
ali-bahjati: yes you are right.

drozdziak1 (approved):

PR #372, opened 2022-11-02
[eth]: Fix gas benchmark to generate useful gas snapshot

ali-bahjati:
This PR changes the benchmark test to generate gas usages that are expected and useful for our usage.

Apparently, Foundry runs all the code of a test in an imaginary transaction which is different than using hardhat/truffle. Because in hardhat/truffle, as their tests are written in JS, each contract call is a transaction. But in Foundry it's different. 

This caused problem in our benchmark because re-writing to a storage slot (other than the first write) in the same transaction costs very little (it is considered a warm access,`G_warmaccess = 100`) compared to a normal write (to a non-zero value, `G_sreset` = 2900). In addition to this, in the first write we are consuming a lot of gas as it is writing to a zero storage slot. This was the reason that fresh and non-fresh calls were not much different from each other because re-writes were really cheap. So to solve it we need to exclude the first write which is very expensive and have each subsequent call to the contract separately. Unfortunately we cannot separate contract calls with each other in a Foundry test..

Hopefully, before giving up on Foundry, I looked at their source code and saw that `setUp` is running separately. So I guessed that can fix the problem and it did!

So, I moved the first price update to the `setUp` method. It is reasonable because we can see it as populating the contract with some data. Also, I removed the benchmark iterations because with what I mentioned above, doing an operation twice will be much cheaper than before and won't be useful in the benchmark. I also moved all non-call logic (such as generating the VAA) to the test setup so they will be excluded.

I also added a benchmark test for the getUpdateFee. Currently it is not doing much, but if it gets complex, it is good that we be able to track its gas usage.

The resulting gas snapshot is listed below. It is reasonable and each benchmark test represent the gas cost of a single call (easier to track). This value also includes an initial transaction cost as well as reading from the contract storage itself. You can get the most accurate result by looking at the gas report (minimum value) or the gas shown in the call trace with `-vvvv` argument to `forge test`.

```
GasBenchmark:testBenchmarkGetPrice() (gas: 44944)
GasBenchmark:testBenchmarkGetUpdateFee() (gas: 133433)
GasBenchmark:testBenchmarkUpdatePriceFeedsFresh() (gas: 517137)
GasBenchmark:testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: 542821)
GasBenchmark:testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 269028)
GasBenchmark:testBenchmarkUpdatePriceFeedsNotFresh() (gas: 454250)
PythTestUtilsTest:testGeneratePriceFeedUpdateVAAWorks() (gas: 8132042)
WormholeTestUtilsTest:testGenerateVaaWorks() (gas: 3112564)
``` 

P.S: Although the gas snapshot works correctly, but still the gas report includes `setUp` calls. So you need to look at the minimum value there.

### DIFF
--- a/ethereum/README.md
+++ b/ethereum/README.md
@@ -59,17 +59,19 @@ A gas report should have a couple of tables like this:
 ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
 │ .............                                                                             ┆ .....           ┆ .....  ┆ .....  ┆ .....   ┆ ..      │
 ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
-│ parseAndVerifyVM                                                                          ┆ 90292           ┆ 91262  ┆ 90292  ┆ 138792  ┆ 50      │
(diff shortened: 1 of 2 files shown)


ethereum/forge-test/GasBenchmark.t.sol
```diff
@@ -79,68 +101,36 @@ contract GasBenchmark is Test, WormholeTestUtils, PythTestUtils {
...
     function testBenchmarkGetPrice() public {
-        (bytes[] memory updateData, uint updateFee) = generateUpdateDataAndFee();
-        pyth.updatePriceFeeds{value: updateFee}(updateData);
+        // Set the block timestamp to 0. As prices have < 10 timestamp and staleness 
+        // is set to 60 seconds, the getPrice should work as expected.
+        vm.warp(0);
 
-        // Set the block timestamp to the publish time, so getPrice work as expected.
-        vm.warp(prices[0].publishTime);
+        pyth.getPrice(priceIds[getRand() % NUM_PRICES]);
```
jayantk, line 130: suggest picking a specific id here instead of using `getRand()`. the current approach will create extra variation between test runs if someone changes the code to add more calls to `getRand` elsewhere.

jayantk (approved):

ethereum/README.md
```diff
@@ -59,17 +59,19 @@ A gas report should have a couple of tables like this:
...
 │ .............                                                                             ┆ .....           ┆ .....  ┆ .....  ┆ .....   ┆ ...     │
 ╰───────────────────────────────────────────────────────────────────────────────────────────┴─────────────────┴────────┴────────┴─────────┴─────────╯
 ```
 
-For most of the methods, the median gas usage is an indication of our desired gas usage. Because the calls that store something in the storage
-for the first time use significantly more gas.
+For most of the methods, the minimum gas usage is an indication of our desired gas usage. Because the calls that store something in the storage
+for the first time in `setUp` use significantly more gas. For example, in the above table, there are two calls to `updatePriceFeeds`. The first
+call has happend in the `setUp` method and costed over a million gas and is not intended for our Benchmark. So our desired value is the
+minimum value which is around 380k gas.
```
tompntn, line 72: Isn't a better approximation the median? Which won't be affected by the single `setUp` outlier.

tompntn: You can also remove the outlier from the mean: ((avg * N) - max)/(N - 1)

tompntn (approved):

ethereum/README.md
ali-bahjati: No we cannot have more than 1 call. as I explained above, calling the benchmark (in the test) for the second time will cost much less than the first time due to repeated write on the same location.

ethereum/forge-test/GasBenchmark.t.sol
ali-bahjati: Yeah you are right.

ethereum/README.md
ali-bahjati: I updated the above description to be more clear.

njkumr (commented):
lgtm

PR #373, opened 2022-11-03
Modify PriceInfo struct for gas optimization

ali-bahjati:
This PR removes unneeded `PriceInfo` fields and rearranges the struct in a way that it takes less storage (only two slots) and be optimized for reads (only one slot). However all the benefits from this change might not be visible now and will come in the next PRs, for example, changing `getPrice` to only read the second half of the price info and accessing timestamp directly. Current change resulted in ~15-25% gas improvement over total transaction gas usage.

The surface logic of the contract has not changed much. The only change is that for the existence of a price feed, instead of ID we are using the `publishTime`.

Snapshot difference:
```
testBenchmarkUpdatePriceFeedsNotFresh() (gas: -66595 (-14.660%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: -104087 (-19.175%)) 
testBenchmarkUpdatePriceFeedsFresh() (gas: -102711 (-19.861%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: -66698 (-24.792%)) 
testBenchmarkGetPrice() (gas: -12442 (-31.470%)) 
```

The snapshot includes initial gas costs too (and is good for measuring price pusher cost). But the actual improvement ratio is different, for example, on updatePriceFeedsFresh it improves the gas usage of the method by 27% percent whereas it is only 20% on the snapshot.

### DIFF
--- a/ethereum/contracts/pyth/PythInternalStructs.sol
+++ b/ethereum/contracts/pyth/PythInternalStructs.sol
@@ -43,11 +43,18 @@ contract PythInternalStructs {
         uint64 prevConf;
     }
 
-    struct PriceInfo {
-        uint256 attestationTime;
-        uint256 arrivalTime;
-        uint256 arrivalBlock;
-        PythStructs.PriceFeed priceFeed;
+ struct PriceInfo {
+        // slot 1
+        int64 price;
+        uint64 conf;
+        uint64 publishTime;
+        int32 expo;
+
+        // slot 2
+        int64 emaPrice;
+        uint64 emaConf;
+        uint64 emaPublishTime;
+        int32 emaExpo;
     }
 
     struct DataSource {
--- a/ethereum/contracts/pyth/PythInternalStructs.sol
+++ b/ethereum/contracts/pyth/PythInternalStructs.sol
@@ -43,18 +43,19 @@ contract PythInternalStructs {
         uint64 prevConf;
     }
 
- struct PriceInfo {
-        // slot 1
+    struct InternalPrice {
         int64 price;
         uint64 conf;
         uint64 publishTime;
         int32 expo;
+    }
+
+    struct PriceInfo {
+        // slot 1
+        InternalPrice price;
 
         // slot 2
-        int64 emaPrice;
-        uint64 emaConf;
-        uint64 emaPublishTime;
-        int32 emaExpo;
+        InternalPrice emaPrice;
     }
 
     struct DataSource {
--- a/ethereum/contracts/pyth/PythState.sol
+++ b/ethereum/contracts/pyth/PythState.sol
@@ -12,7 +12,7 @@ contract PythStorage {
         bytes32 _deprecatedPyth2WormholeEmitter; // Ditto
 
         // After a backward-incompatible change in PriceFeed this mapping got deprecated.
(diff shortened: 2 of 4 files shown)


ethereum/contracts/pyth/PythState.sol
```diff
@@ -38,11 +38,15 @@ contract PythStorage {
 
         // Mapping of cached price information
         // priceId => PriceInfo
-        mapping(bytes32 => PythInternalStructs.PriceInfo) latestPriceInfo;
+        mapping(bytes32 => PythInternalStructs.PriceInfo) _anotherDeprecatedLatestPriceInfo;
```
tompntn, line 41: Shall we rename the PriceInfo structs V1,V2,V3.. etc, which would allow us to call these deprecated mapping fields something more meaningful like `deprecatedLatestPriceInfoV1`, `deprecatedLatestPriceInfoV2` etc?

ethereum/contracts/pyth/PythInternalStructs.sol
```diff
@@ -43,11 +43,18 @@ contract PythInternalStructs {
...
-        uint256 arrivalTime;
-        uint256 arrivalBlock;
-        PythStructs.PriceFeed priceFeed;
+ struct PriceInfo {
+        // slot 1
+        int64 price;
+        uint64 conf;
+        uint64 publishTime;
+        int32 expo;
+
```
tompntn, line 52: Could we abstract out the slot 1 and slot 2 into seperate structs? They both essentially represent a price+conf+timestamp. So we would have 
```
struct Price {
    int64 price;
    uint64 conf;
    uint64 publishTime;
    int32 expo;
}
struct PriceInfo {
    price Price;
    emaPrice Price;
}
```
`Price` isn't the greatest name but you get the idea.

tompntn (approved):
Nice!

ethereum/contracts/pyth/PythInternalStructs.sol
ali-bahjati: I don't know why but after making such a change the gas usage was increased. Its not much 0.37-0.68% change but let's keep it the way it is as I think we can bear it.

```
testBenchmarkUpdatePriceFeedsFresh() (gas: 1558 (0.377%)) 
testBenchmarkUpdatePriceFeedsNotFresh() (gas: 1462 (0.379%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: 1846 (0.423%)) 
testBenchmarkGetPrice() (gas: 142 (0.527%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 1374 (0.684%)) 
```

ali-bahjati: Ok, let me actually keep the clean version. Then I'll revert this change later if needed.

ethereum/contracts/pyth/PythState.sol
ali-bahjati: Done.

njkumr (changes requested):
Think it would be better for the deprecatedLatestPriceInfos to map bytes32 to the earlier PriceInfo structs they were mapping to instead of the new changed PriceInfo struct.

njkumr (approved):

PR #374, opened 2022-11-03
Drozdziak1/xc tool evm bindings

drozdziak1:
This is an initial changeset for automatically generating typesafe EVM bindings for the xc-tool. 

It is a work in progress, because the codegen library acting on EVM ABIs (converting them to Rust code and data structures) encounters a bug caused by our larger structs. This problem is tracked here https://github.com/gakonst/ethers-rs/issues/1829

Other than the mentioned issue, this changeset should provide us with Rust code to interact with all known Solidity contracts within `pyth-crosschain` and Wormhole. I think it is important to invest in this approach, since good codegen can leave much less room for human error and regressions between our APIs.

### DIFF
--- a/xc-tool/build.rs
+++ b/xc-tool/build.rs
@@ -1,0 +1,107 @@
+use std::{env, fs, path::PathBuf, path::Path};
+
+use ethers::contract::Abigen;
+use ethers::solc::{remappings::Remapping, Artifact, Project, ProjectPathsConfig, Solc};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let pyth_evm_src_root =
+        env::var("PYTH_XC_TOOL_EVM_SRC_ROOT").unwrap_or("../ethereum/".to_string());
+    let evm_path = PathBuf::from(&pyth_evm_src_root).canonicalize()?;
+
+    let solc = Solc::default();
+    eprintln!("Got Solc!");
+
+    // Remappings are a primitive that solc uses to facilitate imports
+    // similar to `import "@openzeppelin/contracts/VeryAwesomeStandardContract.sol"`.
+    // In this example, a remapping would tell solc that
+    // `@openzeppelin` means <pkg root>/node_modules/@openzeppelin.
+    let mut remappings = Remapping::find_many(&evm_path);
+
+    // NOTE(2022-11-03): For some reason, the @pythnetwork package
+    // namespace upsets the remapping search, replace the wrong
+    // generated path with the correct value here.
+    remappings.iter_mut().for_each(|r| {
+        if r.name == "@pythnetwork/".to_string() {
+            r.path = evm_path
+                .join("node_modules/@pythnetwork")
+                .to_str()
+                .unwrap()
+                .to_string();
+        }
+    });
+
+    eprintln!("Found {} remappings:", remappings.len());
+    for (idx, r) in remappings.iter().enumerate() {
+        eprintln!("{} - {}", idx + 1, r);
+    }
+    let solc_project = Project::builder()
+        .solc(solc)
+        .paths(
+            ProjectPathsConfig::builder()
+                .root(&evm_path)
+                .artifacts(evm_path.join("out"))
+                .cache(evm_path.join("cache"))
+                .sources(evm_path.join("contracts"))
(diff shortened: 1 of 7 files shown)


xc-tool/build.rs
```diff
@@ -0,0 +1,107 @@
+use std::{env, fs, path::PathBuf, path::Path};
+
+use ethers::contract::Abigen;
+use ethers::solc::{remappings::Remapping, Artifact, Project, ProjectPathsConfig, Solc};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
tompntn, line 6: Can we split this up into runtime-specific methods? Would be good to have all the EVM build stuff in one place, all the Solana stuff in another, etc.

xc-tool/build.rs
```diff
@@ -0,0 +1,107 @@
...
+    let pyth_evm_src_root =
+        env::var("PYTH_XC_TOOL_EVM_SRC_ROOT").unwrap_or("../ethereum/".to_string());
```
tompntn, line 8: nit: not for now, but for later, it's generally easier to use a config crate such as https://crates.io/crates/config instead of ad-hod environment variables

xc-tool/build.rs
```diff
@@ -0,0 +1,107 @@
...
+    let evm_path = PathBuf::from(&pyth_evm_src_root).canonicalize()?;
+
+    let solc = Solc::default();
+    eprintln!("Got Solc!");
```
tompntn, line 12: nit, again for later: using a structured logging library such as https://crates.io/crates/structlog will make things easier as the project grows

xc-tool/build.rs
tompntn, line 6: We should use the `anyhow::Result` type everywhere, it makes dealing with errors from different crates much easier.

tompntn: It would also be good to put the EVM binding code in a function called something like `generate_evm_bindings`, so it's clear what it does.

xc-tool/src/evm.rs
```diff
@@ -0,0 +1,13 @@
+use crate::util::ErrBox;
+
+use ethers::prelude::*;
+
+use ethers::providers::Http;
+
+/// Query the Pyth contract's configuration from an EVM chain
```
tompntn, line 7: Just curious, how are you testing this?

tompntn (commented):
Nice!

xc-tool/build.rs
drozdziak1: That's a good idea. If we get a fix for the codegen, I'll do that.

xc-tool/build.rs
drozdziak1: I am not a fan of elaborate build.rs dependencies, I think we should consider that when we get closer to 10-20 variables for the build.rs

xc-tool/build.rs
drozdziak1: I can't put my finger on it, but I've never clicked with the structured logging pattern. You grep for them anyway in the end. But I know it's very little extra hassle so I think we should do it if you think it's a good fit here.

xc-tool/build.rs
drozdziak1: I can add it, but I have my doubts.

I don't think it's `anyhow`'s fault, I think error crates are still vastly immature in Rust - in general. I can count at least three  U L T I M A T E™ and  D E F I N I T I V E™ crates that were the premier choice over the years. I'm talking about `failure`, then recently `anyhow` and most recently `eyre` seems to be the talk of the town. I'm sure they're awesome, but time and again, the boxed std trait object seemed to be the common denominator, supported by default, working in 80% of the cases. `build.rs` files are meant to be more of a script affair and I'm not convinced of the quality-of-life necessity here.

xc-tool/src/evm.rs
drozdziak1: Unfortunately, it doesn't even build because the codegen tries to `derive(Debug)` on a huge-ass tuple which the standard library does not support (They literally just implement `Debug` statically for a bunch of tuple types between 1 and 9 members that are already `Debug`)

drozdziak1:
Closing in favor of the team effort with the ops team.

PR #375, opened 2022-11-04
Remove middle structs

ali-bahjati:
Before converting VAA payloads to PriceInfo we convert them to PriceAttestation which is the equivalent of the payload wire format. However, this middle conversion has some costs, because (1) we copy stuff twice, and (2) we have some unused fields. 

This PR removes the middle structs and creates PriceInfo directly from the VAA payload. 

The gas improvement is not substantial, but still good (~14k gas save)

The gas snapshot difference:
```
testBenchmarkGetUpdateFee() (gas: -22 (-0.016%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 110 (0.054%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: -22212 (-5.063%)) 
testBenchmarkUpdatePriceFeedsFresh() (gas: -22272 (-5.374%)) 
testBenchmarkUpdatePriceFeedsNotFresh() (gas: -22236 (-5.736%)) 
```

### DIFF
--- a/ethereum/contracts/pyth/PythInternalStructs.sol
+++ b/ethereum/contracts/pyth/PythInternalStructs.sol
@@ -9,11 +9,6 @@ import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
 contract PythInternalStructs {
     using BytesLib for bytes;
 
-    struct PriceAttestation {
-        PriceInfo priceInfo;
-        bytes32 priceId;
-    }
-
     struct InternalPrice {
         int64 price;
         uint64 conf;
--- a/ethereum/contracts/pyth/PythInternalStructs.sol
+++ b/ethereum/contracts/pyth/PythInternalStructs.sol
@@ -9,38 +9,9 @@ import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
 contract PythInternalStructs {
     using BytesLib for bytes;
 
-    struct BatchPriceAttestation {
-        Header header;
-
-        uint16 nAttestations;
-        uint16 attestationSize;
-        PriceAttestation[] attestations;
-    }
-
-    struct Header {
-        uint32 magic;
-        uint16 versionMajor;
-        uint16 versionMinor;
-        uint16 hdrSize;
-        uint8 payloadId;
-    }
-
     struct PriceAttestation {
-        bytes32 productId;
+        PriceInfo priceInfo;
         bytes32 priceId;
-        int64 price;
-        uint64 conf;
-        int32 expo;
-        int64 emaPrice;
-        uint64 emaConf;
-        uint8 status;
-        uint32 numPublishers;
-        uint32 maxNumPublishers;
-        uint64 attestationTime;
-        uint64 publishTime;
-        uint64 prevPublishTime;
-        int64 prevPrice;
-        uint64 prevConf;
     }
 
     struct InternalPrice {
@@ -61,18 +32,5 @@ contract PythInternalStructs {
     struct DataSource {
         uint16 chainId;
         bytes32 emitterAddress;
-    }
-
-    /* PriceAttestationStatus represents the availability status of a price feed passed down in attestation.
-        UNKNOWN: The price feed is not currently updating for an unknown reason.
-        TRADING: The price feed is updating as expected.
-        HALTED: The price feed is not currently updating because trading in the product has been halted.
-        AUCTION: The price feed is not currently updating because an auction is setting the price.
-    */
-    enum PriceAttestationStatus {
-        UNKNOWN,
-        TRADING,
-        HALTED,
-        AUCTION
     }
 }
--- a/ethereum/test/pyth.js
+++ b/ethereum/test/pyth.js
@@ -284,10 +284,6 @@ contract("Pyth", function () {
     }
 
     it("should parse batch price attestation correctly", async function () {
-        const magic = 0x50325748;
-        const versionMajor = 3;
-        const versionMinor = 0;
-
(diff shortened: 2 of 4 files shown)


ethereum/test/pyth.js
```diff
@@ -298,45 +294,27 @@ contract("Pyth", function () {
...
+            assert.equal(parsed[i].priceInfo.price.price, priceVal);
+            assert.equal(parsed[i].priceInfo.price.conf, 101);
+            assert.equal(parsed[i].priceInfo.price.expo, -3);
+            assert.equal(parsed[i].priceInfo.emaPrice.expo, -3);
+
+            assert.equal(parsed[i].priceInfo.emaPrice.price, -42);
+            assert.equal(parsed[i].priceInfo.emaPrice.conf, 42);
+
+            assert.equal(parsed[i].priceInfo.emaPrice.publishTime, publishTime);
+            assert.equal(parsed[i].priceInfo.price.publishTime, publishTime);
```
jayantk: can you also add a test here for the case where the status is unknown?

ethereum/contracts/pyth/PythInternalStructs.sol
```diff
@@ -62,17 +33,4 @@ contract PythInternalStructs {
...
         bytes32 emitterAddress;
     }
-
-    /* PriceAttestationStatus represents the availability status of a price feed passed down in attestation.
-        UNKNOWN: The price feed is not currently updating for an unknown reason.
-        TRADING: The price feed is updating as expected.
-        HALTED: The price feed is not currently updating because trading in the product has been halted.
-        AUCTION: The price feed is not currently updating because an auction is setting the price.
-    */
-    enum PriceAttestationStatus {
```
jayantk, line 72: i take it the enum costs extra gas ? it's a lot nicer to have the enum than a number so would be nice to keep this if possible.

ethereum/contracts/pyth/PythInternalStructs.sol
```diff
@@ -9,38 +9,9 @@ import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
 contract PythInternalStructs {
     using BytesLib for bytes;
 
-    struct BatchPriceAttestation {
```
jayantk, line 12: it's safe to delete these structs because they're never stored anywhere or shared with users outside the contract?

ali-bahjati:
By @njkumr suggestion, I moved the processing part in the parsing method. This will reduce a lot of gas because we won't append to memory any more. 

Difference:

```
testBenchmarkGetUpdateFee() (gas: -22 (-0.016%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 110 (0.054%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: -7455 (-1.758%)) 
testBenchmarkUpdatePriceFeedsFresh() (gas: -7532 (-1.884%)) 
testBenchmarkUpdatePriceFeedsNotFresh() (gas: -7300 (-1.959%)) 
Overall gas change: -22199 (-5.564%)
```

I updated the gas snapshot above as well.

ethereum/test/pyth.js
```diff
@@ -283,61 +286,41 @@ contract("Pyth", function () {
...
+
+        // Then prices should be available because the valid period is now 120 seconds
+        for (var i = 1; i <= RAW_BATCH_ATTESTATION_COUNT; i++) {
+            const price_id =
+                "0x" +
+                (255 - (i % 256)).toString(16).padStart(2, "0").repeat(32);
+
+            const price = await this.pythProxy.getPriceUnsafe(price_id);
+            assert.equal(price.price, priceVal.toString());
+            assert.equal(price.conf, "101"); // Fixed in the fixture.
```
jayantk, line 315: i don't understand the code comment here

ethereum/test/pyth.js
```diff
@@ -283,61 +286,41 @@ contract("Pyth", function () {
...
-            console.debug(
-                `attestation ${i + 1}/${parsed.attestations.length} parsed OK`
-            );
+        const receipt = await updatePriceFeeds(this.pythProxy, [rawBatch]);
+
+        expectEvent(receipt, 'PriceFeedUpdate', {
+            price: "1337",
+        });
+
+        // Then prices should be available because the valid period is now 120 seconds
```
jayantk, line 307: can you add a test for the case where the status is unknown?

jayantk (commented):
generally fine, but would like to see a test for the case where status=unknown. also had some questions about how the safety of the migration (i think this is mostly for my own understanding).

ethereum/test/pyth.js
ali-bahjati: There is already a test for it. It didn't need anychange as it uses end to end just like this one. 

https://github.com/pyth-network/pyth-crosschain/blob/main/ethereum/test/pyth.js#L695
I had added this when I added this logic and it's a good kind of test that needs no change when you change internals  :D

ethereum/contracts/pyth/PythInternalStructs.sol
ali-bahjati: yes

jayantk (approved):

ethereum/contracts/pyth/PythInternalStructs.sol
ali-bahjati: Probably there is a gas cost, maybe not much. But I rather keep it out as we soon will remove status and previous fields from price attestation (it will also save gas + makes logic simpler)

njkumr (approved):
the changes lgtm!

PR #376, opened 2022-11-04
 xc-tool ping-all: Add a manual EVM query listing Pyth data sources

drozdziak1:
This quick and dirty request is successful against the Tilt `eth-devnet` as well as chosen EVM-compatible RPCs (See config_schema.rs for details). Before this goes out of WIP I think  we should settle about:
- [x] Configuration strategy - YAML wins at nesting which I think we'll have lots of, I like TOML otherwise. **Picked YAML for now**
- [x] What should xc-tool do in Tilt? - This is useful as part of integration-testing it and enabling devs to play with it. I think we should just give it a resource yaml, and as a good rule of thumb automate running all known queries. This may become trickier when we approach quirks between similar networks (EVM, pythnet come to mind), we will need to figure out how to test said quirks, possibly without running a dozen mock chain nodes. Hopefully the number of quirks is less than our supported chain count. **I'll build it and run the tests in a following change**
- [x] Are there any other local devnet decisions that may have long-lasting consequences at this stage? **There seems to be more network-agnostic talking points to figure out (e.g. some protocols do not compile into Rust packages cleanly)**

### DIFF
--- a/xc-tool/Cargo.lock
+++ b/xc-tool/Cargo.lock
@@ -3,6 +3,64 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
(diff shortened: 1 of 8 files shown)


drozdziak1:
Closing in favor of the team effort with the ops team

PR #378, opened 2022-11-04
[eth] Wrap parsing method with unchecked

ali-bahjati:
Thanks @njkumr for suggesting this!

There is no overflow possible in the parsing method and we can wrap every line in unchecked which won't be clean. So the whole method is now wrapped in an unchecked scope.

> Most of the math operations below are simple additions. In the places that there is more complex operation there is a comment explaining why it is safe. Also, byteslib operations have proper require checks.

Gas snapshot change is very high:
```
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 0 (0.000%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: -8001 (-1.958%)) 
testBenchmarkUpdatePriceFeedsFresh() (gas: -8001 (-2.076%)) 
testBenchmarkUpdatePriceFeedsNotFresh() (gas: -7737 (-2.207%)) 
```

### DIFF
--- a/ethereum/contracts/pyth/Pyth.sol
+++ b/ethereum/contracts/pyth/Pyth.sol
@@ -37,8 +37,10 @@ abstract contract Pyth is PythGetters, PythSetters, AbstractPyth {
         uint requiredFee = getUpdateFee(updateData);
         require(msg.value >= requiredFee, "insufficient paid fee amount");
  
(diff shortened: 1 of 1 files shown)


ali-bahjati:
@njkumr added that.

njkumr (approved):

PR #380, opened 2022-11-08
[eth] Gas improvement: Use optimized version of BytesLib

ali-bahjati:
We call BytesLib methods for bytes several times during parsing a price update. Those methods always perform a `require` check to ensure they are not accessing out of bound. However, as we are parsing a single bytes data and calling this method couple of times, we can only check for the size once in the end (we already do) and remove those checks.

Gas snapshot difference:

```
testBenchmarkUpdatePriceFeedsIfNecessaryNotFresh() (gas: 0 (0.000%)) 
testBenchmarkUpdatePriceFeedsIfNecessaryFresh() (gas: -7452 (-1.871%)) 
testBenchmarkUpdatePriceFeedsFresh() (gas: -7452 (-1.985%)) 
testBenchmarkUpdatePriceFeedsNotFresh() (gas: -7452 (-2.182%)) 
```

### DIFF
--- a/ethereum/contracts/libraries/external/OptimizedBytesLib.sol
+++ b/ethereum/contracts/libraries/external/OptimizedBytesLib.sol
@@ -1,0 +1,510 @@
+// SPDX-License-Identifier: Unlicense
+/*
+ * @title Solidity Bytes Arrays Utils
+ * @author Gonçalo Sá <goncalo.sa@consensys.net>
+ *
+ * @dev Bytes tightly packed arrays utility library for ethereum contracts written in Solidity.
+ *      The library lets you concatenate, slice and type cast bytes arrays both in memory and storage.
+ */
+pragma solidity >=0.8.0 <0.9.0;
+
+
+library BytesLib {
+    function concat(
+        bytes memory _preBytes,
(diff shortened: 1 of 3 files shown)


njkumr:
@ali-bahjati what's the difference between doing this and instead making the arithmetic unchecked and removing the comment string? For example something like `unchecked{ require(_bytes.length >= _start + 4, ""); }` ?

kcsongor:
Instead of calling it `Optimized` I think a better name would be `Unsafe` since the library no longer has the same safety guarantees, and you have to apply global (whole program) reasoning to ensure correctness.

Also, because of the above, I wonder if the `using ... for ...` syntax should be dropped in favour or direct function calls to make this even more explicit? This is a stylistic choice but being as explicit as possible here is valuable IMO

ali-bahjati:
@njkumr I did what you suggested. It has only 2867 gas improvement and 4585 behind the optimization above.

ali-bahjati:
@kcsongor yeah, it makes sense. I pushed what you suggested. Thank you.

jayantk (approved):

njkumr (approved):

PR #386, opened 2022-11-16
drozdziak1/add xc tool ts

drozdziak1:
- Add xc-tool-ts
- Integrate EVM and Solana


xc-tool-ts/src/config.ts
```diff
@@ -0,0 +1,157 @@
...
+  getEvmConfigs(): EvmConfig[] {
+    return [];
+  }
+
+  getSolConfigs(): SolConfig[] {
+    return [];
+  }
+}
+
+export class LocalDevnetConfig implements NetworkConfig {
```
jayantk, line 28: having separate config classes for mainnet / testnet / etc. seems wrong to me. I think a better solution would be a single config class that was flexible enough to be used across all of them. I think a class with fields like`evmConfigs: Map<string, EvmConfig>` would be sufficient? (the key of the map would be the human-readable name of the network.)

xc-tool-ts/src/config.ts
```diff
@@ -0,0 +1,157 @@
...
+  owner: solWeb3.PublicKey;
+  whProg: solWeb3.PublicKey;
+  pythOwner: solWeb3.PublicKey;
+  maxBatchSize: number;
+  isActive: boolean;
+  opsOwner: solWeb3.PublicKey | null;
+}
+
+// Helper catch-all object to contain the deserialized schema of SOL accounts
+class Assignable {
```
jayantk, line 108: i'm confused about why you need this. maybe explain more in the comment?

xc-tool-ts/src/index.ts
```diff
@@ -0,0 +1,54 @@
+import {MainnetConfig, TestnetConfig, AttesterConfigState, LocalDevnetConfig, getEvmDataSources, getAttesterConfig, Config} from "./config";
+import {PublicKey} from "@solana/web3.js";
+const DEFAULTS: Config = {
```
jayantk, line 3: (suggest moving this to a json or yaml configuration file at some point in the future)

jayantk (commented):
seems pretty reasonable to me, except for the separate config classes for mainnet/testnet etc.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -602,19 +603,25 @@ async fn attestation_sched_job(args: AttestationSchedJobArgs) -> Result<(), ErrB
...
 
         // We never get to error reporting in daemon mode, attach a map_err
-        let job_with_err_msg = job.map_err(move |e| {
-            warn!(
-                "Batch {}/{}, group {:?} ERR: {:#?}",
-                batch_no4err_msg, batch_count4err_msg, group_name4err_msg, e
-            );
-            e
-        });
+        let job_with_err_msg = job.then(move |res| {
```
ali-bahjati, line 610: Is this file part of the PR?

xc-tool-ts/src/config.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  let [cfgAddr, _bumpSeed] = await solWeb3.PublicKey.findProgramAddress([
+    Buffer.from("pyth2wormhole-config-v3"),
+  ], cfg.attesterContract);
+
+  console.log("Looking up account", cfgAddr.toString());
+
+  let rawAccInfo = await conn.getAccountInfo(cfgAddr);
+
+  let schema = new Map([[Assignable, {
```
ali-bahjati, line 129: where this is used?

xc-tool-ts/src/config.ts
ali-bahjati: I agree, if you make it more generalized then we can move the actual config to a config file. Inside config files you can also have a `cluster` field to specify what cluster this is. 

Then, the config file could look like this (just an idea):

```yaml
target_chains:
  - name: bnb_testnet
    chain_type: evm
    cluster: testnet
    rpc: https://....
    contract: 0xabde000
 - ....
```

drozdziak1: I think that using a sacred config file with all the values is a very bad idea. IMO a better approach is to keep up-to-date defaults and only splice in temporary changes, such as "defaults but RPC for Pythnet is different". Such partial snippets would be treated as exceptions instead of the norm. The use of an altered config can then be detected and flagged with the user when the tool is run, instead of guesswork comparing people's configs.

I also think the clear divide between networks is very important.  For example, there's different things on Pythnet than on Solana. SOL has no remote executor, Pythnet has no governance contract. Instead of using a nullable that can mistakenly be filled or left empty, I find it far more helpful to make an invalid choice unrepresentable, by never mentioning remote executor on Solana, and vice versa with governance contract on Pythnet.

xc-tool-ts/src/config.ts
drozdziak1: The Borsh node package is very weird and frustrating. It desperately wants the caller to put a type in first argument of `deserialize`, while also taking a hand-crafted schema object. It does not seem to care about the schema of that aforementioned type. This is mostly copied from Solana's cookbook guide for the JS APIs. I w-was j-just f-following orders.

xc-tool-ts/src/index.ts
drozdziak1: See my response to the mainnet/testnet divide nit.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: No, sorry and thanks!

xc-tool-ts/src/config.ts
drozdziak1: `index.ts`, we call the surrounding function and this schema map is used in `deserialize`. It probably deserves a constant instead.

drozdziak1:
We pivoted to a modular approach in co-op with the ops team, closing

PR #396, opened 2022-11-24
drozdziak1/pyth attester prometheus metrics

drozdziak1:
This changeset adds Prometheus metrics to the attester and configures a Prometheus instance to present them in a dashboard. OOTB it can show basic graphs and discover metrics. Other services also share metrics - we can address which ones we want in a separate change.


solana/pyth2wormhole/client/src/cli.rs
```diff
@@ -83,6 +88,19 @@ pub enum Action {
...
+            help = "Address to use for serving Prometheus metrics.",
+            default_value = "[::]:3000"
+        )]
+        metrics_bind_addr: SocketAddr,
+        #[clap(
+            long,
+            help = "Subpage name for serving Prometheus metrics on http://<metrics_bind_addr>/<metrics_subpage>",
+            default_value = "metrics"
+        )]
+        metrics_subpage: String,
```
ali-bahjati, line 103: I got a little confused by this because this is also hardcoded in some places. I think it's better to have it hard-coded as `metrics` as that's the usual path and we really don't need to configure it (I don't see any usacase)

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -270,12 +309,21 @@ async fn handle_attest_daemon_mode(
     // This loop cranks attestations without interruption. This is
     // achieved by spinning up a new up-to-date symbol set before
     // letting go of the previous one. Additionally, hash of on-chain
-    // and attestation configs is used to prevent needless reloads of
+    // and attestation configs is used to prevent needless reloads of"Ad"Address to use for serving Prometheus metrics. Note: served on ""Address to use for serving Prometheus metrics. Note: served on ""Address to use for serving Prometheus metrics. Note: served on "dress to use for serving Prometheus metrics. Note: served on "
```
ali-bahjati, line 312: I think this was pasted here by mistake.

devnet/p2w-attest.yaml
```diff
@@ -52,3 +54,6 @@ spec:
             - containerPort: 4343
               name: p2w-attest
               protocol: TCP
+            - containerPort: 9090
```
ali-bahjati, line 57: Shouldn't the value here be port 3000?

ali-bahjati (commented):
It's awesome!

I left couple of comments and will approve after you respond to them.

ali-bahjati (approved):
Ship it ser 🚢 🔥

PR #400, opened 2022-11-25
Add clippy to all rust projects

ali-bahjati:
Also enable auto fix option as it solves most of the obvious problems.

I advice at least one person review the mostly automated changes to the files.

Please note that I made some changes manually, for example `pyth_client` was deprecated and I changed it to use `pyth_sdk_solana::state`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
       - id: cargo-clippy-remote-executor
         name: Cargo clippy for remote executor
         language: "rust"
-        entry: cargo +nightly clippy --manifest-path ./pythnet/remote-executor/Cargo.toml --fix --allow-staged -- -D warnings
+        entry: cargo +nightly clippy --manifest-path ./pythnet/remote-executor/Cargo.toml --fix --allow-dirty -- -D warnings
         pass_filenames: false
         files: pythnet/remote-executor/
       # Hooks for the attester
@@ -39,7 +39,9 @@ repos:
       - id: cargo-clippy-attester
         name: Cargo clippy for attester
         language: "rust"
-        entry: cargo +nightly clippy --manifest-path ./solana/pyth2wormhole/Cargo.toml --fix --allow-staged -- -D warnings
+        entry: |
+          bash -c 'EMITTER_ADDRESS=0 BRIDGE_ADDRESS=0 cargo +nightly clippy --manifest-path \
+            ./solana/pyth2wormhole/Cargo.toml --fix --allow-dirty -- -D warnings'
         pass_filenames: false
         files: solana/pyth2wormhole/
       # Hooks for cosmwasm contract
@@ -52,7 +54,7 @@ repos:
       - id: cargo-clippy-cosmwasm
         name: Cargo clippy for cosmwasm contract
         language: "rust"
-        entry: cargo +nightly clippy --manifest-path ./cosmwasm/Cargo.toml --fix --allow-staged -- -D warnings
+        entry: cargo +nightly clippy --manifest-path ./cosmwasm/Cargo.toml --fix --allow-dirty -- -D warnings
         pass_filenames: false
         files: cosmwasm/
       # Hooks for p2w-sdk/rust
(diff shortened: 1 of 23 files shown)


solana/pyth2wormhole/client/Cargo.toml
```diff
@@ -39,6 +39,5 @@ warp = "0.3.3"
 http = "0.2.8"
 
 [dev-dependencies]
-pyth-client = "0.5.0"
```
drozdziak1, line 42: This is wrong! It is used in the standalone tests. I think there should be a way to tell clippy to take `--test` into account.

drozdziak1: I think our Cargo caching may be giving us a false positive about this.

drozdziak1: example: test_attest.rs

solana/pyth2wormhole/client/src/util.rs
```diff
@@ -84,7 +84,7 @@ impl<T> RLMutex<T> {
     pub fn new(val: T, rl_interval: Duration) -> Self {
         Self {
             mtx: Mutex::new(RLMutexState {
-                last_released: Instant::now() - rl_interval,
+                last_released: Instant::now().checked_sub(rl_interval).unwrap(),
```
drozdziak1, line 87: I guess? Underflows should be impossible for any reasonable value (i.e. lower than current unix timestamp value, which is decades worth)

solana/pyth2wormhole/client/tests/fixtures/pyth.rs
```diff
@@ -1,11 +1,10 @@
 //! This module contains test fixtures for instantiating plausible
 //! Pyth accounts for testing purposes.
 use {
-    pyth_client::{
-        AccKey,
+    pyth_sdk_solana::state::{
         AccountType,
-        Price,
-        Product,
+        PriceAccount,
```
drozdziak1, line 6: I'm not sure that I understand clippy's reasoning here. Hos does it make this decision?

solana/pyth2wormhole/program/src/config.rs
```diff
@@ -154,7 +154,7 @@ impl From<Pyth2WormholeConfigV2> for Pyth2WormholeConfigV3 {
             wh_prog,
             pyth_owner,
             max_batch_size,
-            is_active,
+            is_active: _,
```
drozdziak1, line 157: I think it's wrong that we forced is_active to `true` with this, but admittedly this is being used rarely enough that it shouldn't matter

solana/pyth2wormhole/program/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(incomplete_features)]
```
drozdziak1, line 1: I think this is necessary for  the `adt_const_params` just below it in newer nightlies. Nice!

drozdziak1 (changes requested):
Found some nits but I think the useful things are in overwhelming majority.

.pre-commit-config.yaml
```diff
@@ -17,29 +17,56 @@ repos:
...
         entry: cargo +nightly fmt --manifest-path ./pythnet/remote-executor/Cargo.toml --all -- --config-path rustfmt.toml
         pass_filenames: false
         files: pythnet/remote-executor/
-      - id: cargo-clippy-executor
-        name: Cargo clippy executor
+      - id: cargo-clippy-remote-executor
+        name: Cargo clippy for remote executor
         language: "rust"
-        entry: cargo +nightly clippy --manifest-path ./pythnet/remote-executor/Cargo.toml -- -D warnings
+        entry: cargo +nightly clippy --manifest-path ./pythnet/remote-executor/Cargo.toml --fix --allow-dirty -- -D warnings
```
guibescos, line 29: Does this run in the tests as well?

solana/pyth2wormhole/client/Cargo.toml
ali-bahjati: That one I removed manually, sorry that I forgot to mention it myself.
Clippy was complaining that the package is deprecated (we marked all its method deprecated). I replaced it with pyth_sdk_solana

solana/pyth2wormhole/client/src/util.rs
ali-bahjati: yeah I checked at that is nearly impossible to fail. So I think unwrap is fine.

solana/pyth2wormhole/client/tests/fixtures/pyth.rs
drozdziak1: Ohhh, I get it now, these are the same structs but in the changed package. I think that's alright!

drozdziak1 (approved):

.pre-commit-config.yaml
ali-bahjati: added it.

solana/pyth2wormhole/program/src/config.rs
ali-bahjati: yeah. it does not matter now anymore.

jayantk (commented):
seems like you got some reviews here so i'm not going to add one of my own

PR #401, opened 2022-11-28
[price-service] Fine-tune metrics

ali-bahjati:
- Remove nodejs default metrics. We don't use them.
- Remove response time metric.
- Remove freshness metric and add gap metric for attestationTime and publishTime.
  They are similar; however, freshness was measured upon user request but gap is
  measured upon receiving the next update.
- Change receivedVaa to actually represent distinct vaa received. Prior
  to this, the older vaas, or vaas with same attestation time were
  not counted in this metric. This will also improve the performance.
- Refactors the code a little. `vaaBytes` type was not string and was
  Buffer. It is fixed now.

### DIFF
--- a/third_party/pyth/price-service/src/listen.ts
+++ b/third_party/pyth/price-service/src/listen.ts
@@ -25,9 +25,10 @@ import { HexString, PriceFeed } from "@pythnetwork/pyth-sdk-js";
 import { sleep, TimestampInSec } from "./helpers";
 import { logger } from "./logging";
 import { PromClient } from "./promClient";
+import LRUCache from "lru-cache";
 
 export type PriceInfo = {
-  vaaBytes: string;
+  vaa: Buffer;
   seqNum: number;
   attestationTime: TimestampInSec;
   priceFeed: PriceFeed;
@@ -52,6 +53,8 @@ type ListenerConfig = {
   readiness: ListenerReadinessConfig;
 };
 
+type VaaKey = string;
+
 export class Listener implements PriceStore {
   // Mapping of Price Feed Id to Vaa
   private priceFeedVaaMap = new Map<string, PriceInfo>();
@@ -61,6 +64,7 @@ export class Listener implements PriceStore {
   private spyConnectionTime: TimestampInSec | undefined;
   private readinessConfig: ListenerReadinessConfig;
   private updateCallbacks: ((priceInfo: PriceInfo) => any)[];
+  private observedVaas: LRUCache<VaaKey, boolean>;
 
   constructor(config: ListenerConfig, promClient?: PromClient) {
     this.promClient = promClient;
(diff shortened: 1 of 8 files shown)


third_party/pyth/price-service/src/listen.ts
```diff
@@ -61,13 +64,18 @@ export class Listener implements PriceStore {
...
 
   constructor(config: ListenerConfig, promClient?: PromClient) {
     this.promClient = promClient;
     this.spyServiceHost = config.spyServiceHost;
     this.loadFilters(config.filtersRaw);
     this.readinessConfig = config.readiness;
     this.updateCallbacks = [];
+    this.observedVaas = new LRUCache({
+      max: 10000, // At most 10000 items
+      ttl: 60 * 1000, // 60 seconds
```
jayantk, line 77: this cache seems like it could be pretty large? the VAA strings are pretty long. May need to allocate some more memory for the process.

third_party/pyth/price-service/src/listen.ts
```diff
@@ -194,15 +211,21 @@ export class Listener implements PriceStore {
...
+          emitterChainId: parsedVaa.emitter_chain,
           priceServiceReceiveTime: Math.floor(new Date().getTime() / 1000),
         };
         this.priceFeedVaaMap.set(key, priceInfo);
 
+        if (lastAttestationTime !== undefined) {
+          this.promClient?.addPriceUpdatesGap(
+            priceAttestation.attestationTime - lastAttestationTime
+          );
+        }
```
jayantk, line 227: should we also have a similar metric that tracks the gap between the last price update times as well? That would be more of an end-to-end performance metric.

jayantk: Also let's name this histogram something like "attestationGap" to distinguish between the two cases

third_party/pyth/price-service/src/promClient.ts
```diff
@@ -70,31 +72,25 @@ export class PromClient {
     this.receivedVaaCounter.inc();
   }
 
-  addResponseTime(path: string, status: number, duration: DurationInMs) {
+  addPriceUpdatesGap(gap: DurationInSec) {
+    this.priceUpdatesGapHistogram.observe(gap);
+  }
+
+  // We have multiple paths and for the time being it is not important for us
+  // to capture it. We might consider capturing it in the future.
```
jayantk, line 80: this confuses me. Seems like knowing which API method is queried is important? Can you aggregate these series in datadog instead of collapsing the information here?

jayantk (commented):
I'm not sure about the response time thing -- maybe @thmzlt has some thoughts on how to do this? aside from that, generally fine.

third_party/pyth/price-service/src/listen.ts
ali-bahjati: Good idea 👍

third_party/pyth/price-service/src/promClient.ts
ali-bahjati: Ah I think we should not capture it when we don't use it anywhere, we'll add it when we really need it. I can remove the parameter too. I think currently we don't need it, but if we add more endpoints with business logic we probably need to.

And yeah, we can aggregate the metric when we scrape them, but I would like to keep all the logic in the code. It's not blockchain here and YAGNI principle is true 😀

thmzlt:
@ali-bahjati there are better ways to set up response time tracking. You can leave that part out and I'll take care of it.

jayantk (approved):

PR #402, opened 2022-11-28
[WIP] drozdziak1/squads in tilt

drozdziak1:
I'm getting a `could not find module "@certusone/wormhole-sdk"` error when trying to compile the multisig package. It should show up in Tilt


third_party/pyth/multisig-wh-message-builder/tsconfig.json
```diff
@@ -1,15 +1,14 @@
 {
   "compilerOptions": {
     "declaration": true,
-    "target": "es2016",
+    "target": "es6",
     "module": "commonjs",
-    "outDir": "lib",
+    "outDir": "dist",
```
ali-bahjati, line 6: I think `lib` is the common output directory and also in package.json contents of `lib` are specified to be considered as the package.

third_party/pyth/multisig-wh-message-builder/tsconfig.json
```diff
@@ -1,15 +1,14 @@
...
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "skipLibCheck": true,
-    "resolveJsonModule": true,
-    "noErrorTruncation": true
+    "noErrorTruncation": true,
+    "skipLibCheck": true
```
ali-bahjati, line 11: I would keep the old configuration here `resolveJsonModule` is necessary for squads sdk to work I think.

third_party/pyth/prepare_multisig.py
```diff
@@ -0,0 +1,19 @@
+# This script prepares a local Squads multisig deployment for use with
```
ali-bahjati, line 1: Do we really need to have this? 

We can just append `nc -lkp <port> 0.0.0.0` to the pod command to have readiness.

third_party/pyth/multisig-wh-message-builder/tsconfig.json
drozdziak1: oh, I decided to do this because this is not a library package

ali-bahjati: Aha I see, then it makes sense.

third_party/pyth/prepare_multisig.py
drozdziak1: I think it would be nice to be able to orchestrate some scenarios on the multisig - e.g. add some new members and start governance actions in the vault and have mock multisig members approve them. This is different from the `p2w_autoattest` script which is used in prod (which I want to make go away eventually).

ali-bahjati: Got it. make sense then. 👍

PR #403, opened 2022-11-28
Configure symbol groups by name

jayantk:
This PR adds the ability to configure which symbols are in each group using just the name of the symbol. This change will make it easier for us to maintain the configuration file as we customize the groups.

I refactored the code a little to simplify the change. Specifically, the previous code used to read `AttestationConfig` at startup, then mutate it over time to store any additional symbols read from the mapping account. I refactored the code so there's a unidirectional information flow `AttestationConfig` -> merged with mapping data to create list of batches -> batches are published. I then extracted out the logic for merging the config with the mapping data into a function. Check out the first commit to see the general shape of the refactor.

See inline comments for added details on changes.


pythnet/remote-executor/programs/remote-executor/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![deny(warnings)]
-#![allow(clippy::result_large_err)]
+#![allow(clippy::result_unit_err)]
```
jayantk, line 2: clippy changed this for some reason

solana/pyth2wormhole/client/src/lib.rs
```diff
@@ -402,8 +399,8 @@ pub fn gen_attest_tx(
 pub async fn crawl_pyth_mapping(
     rpc_client: &RpcClient,
     first_mapping_addr: &Pubkey,
-) -> Result<HashMap<Pubkey, HashSet<Pubkey>>, ErrBox> {
-    let mut ret = HashMap::new();
+) -> Result<Vec<P2WProductAccount>, ErrBox> {
```
jayantk, line 402: adding the name of each product to the result here (so we can match on it to make the batches)

third_party/pyth/p2w_autoattest.py
```diff
@@ -190,31 +190,32 @@ def find_and_log_seqnos(s):
 mapping_reload_interval_mins: 1 # Very fast for testing purposes
 min_rpc_interval_ms: 0 # RIP RPC
 max_batch_jobs: 1000 # Where we're going there's no oomkiller
-symbol_groups:
+default_attestation_conditions:
+  min_interval_secs: 30
+  price_changed_bps: 500
+name_groups:
```
jayantk, line 196: hacked the existing tests to include name groups

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -41,66 +33,25 @@ pub struct AttestationConfig {
...
+    pub mapping_reload_interval_mins:   u64,
     #[serde(default = "default_min_rpc_interval_ms")]
-    /// Rate-limiting minimum delay between RPC requests in milliseconds"
-    pub min_rpc_interval_ms:          u64,
-    pub symbol_groups:                Vec<SymbolGroup>,
+    /// Rate-limiting minimum delay between RPC requests in milliseconds
+    pub min_rpc_interval_ms:            u64,
+    /// Attestation conditions that will be used for any symbols included in the mapping
+    /// that aren't explicitly in one of the groups below.
+    pub default_attestation_conditions: AttestationConditions,
```
jayantk, line 42: the old code could use `AttestationConditions::default()` for these symbols under some conditions (if the config file didn't have a "mapping" group). That old behavior is likely to lead to configuration bugs, so I fixed it here.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -30,7 +28,10 @@ use {
         attest::P2W_MAX_BATCH_SIZE,
         Pyth2WormholeConfig,
     },
-    pyth2wormhole_client::*,
+    pyth2wormhole_client::{
+        attestation_cfg::SymbolGroup,
+        *,
```
ali-bahjati, line 33: I don't know the import best practices here, personally I rather it to the be either * or explicit. (I'm in favor of explicit imports)

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -510,22 +500,139 @@ async fn handle_attest_non_daemon_mode(
...
+async fn attestation_config_to_batches(
+    rpc_cfg: &Arc<RLMutex<RpcCfg>>,
+    attestation_cfg: &AttestationConfig,
+    max_batch_size: usize,
+) -> Result<Vec<SymbolGroup>, ErrBox> {
+    // Use the mapping if specified
+    if let Some(mapping_addr) = attestation_cfg.mapping_addr.as_ref() {
+        crawl_pyth_mapping(&lock_and_make_rpc(rpc_cfg).await, mapping_addr)
+            .await
+            .map(|additional_accounts| {
```
ali-bahjati, line 512: I think additional_accounts is misleading, all_accounts seems better :D

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -113,15 +64,27 @@ impl AttestationConfig {
                 g.symbols
```
ali-bahjati, line 64: Can't this method also call partition_into_batches?

solana/pyth2wormhole/client/src/attestation_cfg.rs
ali-bahjati: I think the comment is a bit misleading here. It also adds the condition to named groups. 

I think it's even better to add this to the symbol groups as well.

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -41,66 +33,25 @@ pub struct AttestationConfig {
...
-                };
-
-                self.symbol_groups.push(new_group);
-            }
-            None => {}
-        }
-    }
-
-    pub fn as_batches(&self, max_batch_size: usize) -> Vec<BatchState> {
+    pub fn as_batches(&self, max_batch_size: usize) -> Vec<SymbolGroup> {
```
ali-bahjati, line 54: I think this method and attestation_config_to_batches should live in the same place (or maybe even in one function) and probably as `AttestationConfig` implementation.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -510,22 +500,139 @@ async fn handle_attest_non_daemon_mode(
...
+
+                let mut mapping_batches: Vec<SymbolGroup> = vec![];
+                for group in &attestation_cfg.name_groups {
+                    let batch_items: Vec<P2WSymbol> = group
+                        .symbol_names
+                        .iter()
+                        .flat_map(|symbol| {
+                            name_to_symbols
+                                .get(symbol)
+                                .into_iter()
```
ali-bahjati, line 543: Here you are skipping a symbol if it is not found, I think an explicit error is better.

ali-bahjati (commented):
The refactoring and the change set looks great! I left some minor comments about the code.

Just one big concern about testing the code, because afaik we do not add symbol in the `pyth_publisher.py` and the returned name (that you used) is just a debug information. The e2e test is not failing  because the current logic skips the symbols. To add symbol name to a product (and enable testing) you will need to call `pyth_admin upd_price` like [this](https://github.com/pyth-network/pyth-client/blob/main/pyth/tests/conftest.py#L264) test. 

Second comment is that do you see any use case for keeping the symbol_groups approach? I am in favor of removing it entirely (to keep the code simple).

third_party/pyth/p2w_autoattest.py
drozdziak1: I don't think leaving this empty is right. I think Tilt setup should be able to show that the name group settings are respected _by default_. Other than that, we should be able to confirm that this setting is respected in the log, ideally with `warn!` on name groups that were never used.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -438,29 +444,13 @@ async fn handle_attest_non_daemon_mode(
...
-    }
-    debug!(
-        "Attestation config (includes mapping accounts):\n{:#?}",
-        attestation_cfg
-    );
-
-    let batches = attestation_cfg.as_batches(p2w_cfg.max_batch_size as usize);
+    let batch_config =
+        attestation_config_to_batches(&rpc_cfg, &attestation_cfg, p2w_cfg.max_batch_size as usize)
+            .await?;
```
drozdziak1, line 449: I think screaming and trying to attest with what we got is better than failing the whole process here.

drozdziak1 (changes requested):
I value the use case but I am not sold on the scope of this changeset. IIUC the intent is to add the ability to say "Any symbol name containing "BTC" lands here" etc. I think this can be achieved without the need to create a whole new group type (one could slap a `Vec<String>` member on the existing group type instead) and especially without messing with the utility methods on the attestation config type. I do not agree with the unidirectional flow idea because I don't think the new utility functions have a meaningful life outside AttestationConfig. I think that dragging a batch vector outside attestation_config obfuscates the source of truth about available batches. Ultimately, I do not believe this change is complete without unit-testing the new config merging logic, descriptive logging and a convincing reason to widen the scope beyond just sorting the batches into their matching buckets.

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -332,47 +297,4 @@ mod tests {
 
         Ok(())
     }
-
-    #[test]
-    fn test_add_symbols_works() -> Result<(), ErrBox> {
```
drozdziak1, line 337: This was a useful test with no equivalent in the proposed change.

third_party/pyth/p2w_autoattest.py
```diff
@@ -137,6 +137,8 @@
 mapping_reload_interval_mins: 1 # Very fast for testing purposes
 min_rpc_interval_ms: 0 # RIP RPC
 max_batch_jobs: 1000 # Where we're going there's no oomkiller
+default_attestation_conditions:
```
jayantk, line 140: @drozdziak1 and i agreed that i can update the integration tests in a follow-up, since we currently don't have a good way to validate that the things in the config are actually being attested.

ali-bahjati (approved):
Thanks for addressing my comments. Looking forward to adding it to the tests.

solana/pyth2wormhole/client/src/attestation_cfg.rs
```diff
@@ -334,44 +483,109 @@ mod tests {
...
-        let empty_config = AttestationConfig {
-            min_msg_reuse_interval_ms:    1000,
-            max_msg_accounts:             100,
-            min_rpc_interval_ms:          42422,
-            mapping_addr:                 None,
-            mapping_reload_interval_mins: 42,
-            symbol_groups:                vec![],
-        };
+    fn test_instantiate_batches() -> Result<(), ErrBox> {
```
drozdziak1, line 486: This test does not assert desired behavior edge cases:
- Group already has the same symbol
- No name group matches this symbol (default group)
- brand new symbol is added to the right place

third_party/pyth/p2w_autoattest.py
drozdziak1: :+1:

drozdziak1 (approved):

PR #404, opened 2022-11-29
[eth] Use PythErrors everywhere

ali-bahjati:
This PR uses PythErrors structs instead of error strings everywhere and updates the tests (both truffle and foundry). This change reduces contract size from 23.69KB to 20.94KB.

Notes:
- Error structs are only possible with `revert`. So I changed all `require` statements to `revert`.
- Truffle does not support error structs and I adopted a solution from [here](https://github.com/OpenZeppelin/openzeppelin-test-helpers/issues/180) and created `expectRevertCustomError`.

### DIFF
--- a/ethereum/test/pyth.js
+++ b/ethereum/test/pyth.js
@@ -39,7 +39,6 @@ contract("Pyth", function () {
     "0x71f8dcb863d176e2c420ad6610cf687359612b6fb392e0642b0ca6b1f186aa3b";
   const notOwnerError =
     "Ownable: caller is not the owner -- Reason given: Ownable: caller is not the owner.";
-  const insufficientFeeError = "insufficient paid fee amount";
 
   // Place all atomic operations that are done within migrations here.
   beforeEach(async function () {
@@ -433,9 +432,9 @@ contract("Pyth", function () {
     assert.equal(feeInWei, 20);
 
     // When a smaller fee is payed it reverts
-    await expectRevert(
+    await expectRevertCustomError(
       updatePriceFeeds(this.pythProxy, [rawBatch1, rawBatch2], feeInWei - 1),
-      insufficientFeeError
+      "InsufficientFee"
     );
   });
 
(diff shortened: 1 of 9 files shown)


njkumr (approved):

ethereum/test/pyth.js
```diff
@@ -1115,9 +1105,15 @@ contract("Pyth", function () {
       2
     );
 
-    await expectRevert(
-      this.pythProxy.executeGovernanceInstruction(transferBackVaaWrong),
-      "cannot upgrade to an older governance data source"
+    // Strangely this test was failing without the hard coded gas limit.
+    // It will failing randomly on a place that it shouldn't without any message due to
+    // gas limit if the value is not hard-coded.
+    // Gas value 6.7m gas is very high and close to the ganache limit.
```
jayantk, line 1111: nitpick: comments like this should be written in the present tense:

"this test fails without a hard coded gas limit."

ethereum/test/pyth.js
```diff
@@ -1385,3 +1381,19 @@ function expectEventMultipleTimes(receipt, eventName, args, cnt) {
...
   assert(matches === cnt, `Expected ${cnt} event matches, found ${matches}.`);
 }
+
+async function expectRevertCustomError(promise, reason) {
+  try {
+    await promise;
+    expect.fail("Expected promise to throw but it didn't");
+  } catch (revert) {
+    if (reason) {
+      // expect(revert.message).to.include(reason);
```
jayantk, line 1391: delete?

jayantk (approved):

ethereum/test/pyth.js
ali-bahjati: done. thanks.

ethereum/test/pyth.js
ali-bahjati: yeah you are right. thanks.

PR #405, opened 2022-11-29
[eth] Remove ownership entirely in favor of governance

ali-bahjati:
No deployed contract has owner and we renounced ownership for all in favor of our governance. But, the code was still there because deployers using Wormhole receiver might have needed ownership. However, custom deployers usually use the code on some protocol and they can simply swap out the address with a new fresh deployed contract. Also, we are planning to deploy on some new chains using Wormhole receiver and want to manage the contract only by governance (without ownership). Removing ownership codes will also reduce the contract size from 20.94KB to 19.22KB.

This PR removes all ownership code from Pyth contract. Since we want to be compatibile* with the deployed versions it still inherits from `OwnableUpgradeable` but renounces the ownership during initialization.

This also removes all permissioned methods in `PythUpgradeable` as well as all ownership related tests. 

Note: Some tests were testing a contract behaviour and I ensured that there is a sufficient test with governance for the ones that I had removed (e.g., adding or removing a data source), otherwise I modified the test to use governance.

*: `OwnableUpgradeable` uses some storage slots and removing it will cause storage collision.


njkumr (approved):
lgtm

ethereum/test/pyth.js
```diff
@@ -439,17 +303,22 @@ contract("Pyth", function () {
...
-    const defaultAccount = accounts[0];
-    assert.equal(await this.pythProxy.owner(), defaultAccount);
-
     // Check initial fee is zero
     assert.equal(await this.pythProxy.singleUpdateFeeInWei(), 0);
 
-    // Set fee
-    await this.pythProxy.updateSingleUpdateFeeInWei(10);
+    // Set fee to 10
+    await this.pythProxy.executeGovernanceInstruction(
```
jayantk, line 310: make a setFee test helper function?

ethereum/test/pyth.js
```diff
@@ -617,7 +497,18 @@ contract("Pyth", function () {
     await updatePriceFeeds(this.pythProxy, [rawBatch]);
 
     // Setting the validity time to 30 seconds
-    await this.pythProxy.updateValidTimePeriodSeconds(30);
+    await this.pythProxy.executeGovernanceInstruction(
```
jayantk, line 500: this could also be worthwhile to extract into a test helper function to avoid code duplication

jayantk (approved):

ethereum/test/pyth.js
ali-bahjati: yeah I agree, I was a bit lazy :D will do that.

ethereum/test/pyth.js
ali-bahjati: done.

PR #407, opened 2022-11-30
[near] Add receiver contract

Reisen:
This PR adds the contract part for the NEAR receiver. It implements the core functionality but is missing the tooling to run in CI currently, to come in a second PR. Some things that are intentionally not done here (yet) are gas pre-checks, which require fine tuning to get correct but change while developing.

Other things to keep in mind:

- Optionally add NEAR's `assert_one_yota` to methods that should only be called by real wallets.
- Once fixed Gas costs are calculated, adding `deposit_calc()` method will help the calling side.
- The `cargo build --release --target wasm32-unknown-unknown` produces a file that still sees a 30% reduction in size using `wasm-opt` but is not part of the build process here yet.


contracts/near/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[lib]
+name       = "pyth"
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+hex          = { version = "0.4.3" }
+near-sdk     = { version = "4.1.1" }
+p2w-sdk      = { path = "../../third_party/pyth/p2w-sdk/rust" }
+structs      = { git = "https://github.com/wormhole-foundation/wormhole", branch = "pr/jynnantonix/vaa-structs" }
```
Reisen, line 15: This will change before merging.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+            source: UnorderedSet::new(StorageKeys::Source),
+            prices: UnorderedMap::new(StorageKeys::Prices),
+            wormhole,
+            codehash,
+        }
+    }
+
+    /// Add a new (chain, contract) pair to the list of trusted sources for Price Attestations.
+    #[private]
+    pub fn add_source(&mut self, source: Source) {
```
tompntn, line 98: Just curious, why do `add_source`, `del_source`, `set_upgrade_hash` and `attest_prices` need to be public?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
+//#![deny(warnings)]
+
+use {
+    error::Error,
```
tompntn, line 4: Could we use Anyhow errors?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    pub fn get_sources(&self) -> Vec<Source> {
+        self.source.iter().collect()
+    }
+
+    /// Process a Batch of Price Attestations.
+    ///
+    /// This receives price updates, verifies them, and stores the attestations in the contract
+    /// state allowing other contracts to query them.
+    #[private]
```
tompntn, line 121: You can probably steal some docstrings from the Aptos contracts, which include links to the relevant GitBook pages.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas(30_000_000_000_000))
+                    .refund_vaa(env::predecessor_account_id(), env::attached_deposit()),
+            );
+
+        Ok(())
+    }
+
+    /// If submitting the VAA fails then this callback will refund the caller.
+    pub fn refund_vaa(&mut self, account_id: AccountId, amount: u128) {
```
tompntn, line 192: Does this need to be `#[private]`?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+            .then(Self::ext(env::current_account_id()).refund_upgrade(
+                env::predecessor_account_id(),
+                env::attached_deposit(),
+                env::storage_usage(),
+            )))
+    }
+
+    /// This method is called after the upgrade to refund the caller for the storage used by the
+    /// old contract.
+    pub fn refund_upgrade(&mut self, account_id: AccountId, amount: u128, storage: u64) {
```
tompntn, line 265: Does this need to be `#[private]`?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+        Source,
+    },
+    std::io::Cursor,
+};
+
+pub mod error;
+pub mod ext;
+pub mod state;
+
+const GAS_FOR_VERIFY_VAA: Gas = Gas(10_000_000_000_000);
```
tompntn, line 41: Could we add a comment explaining how this value was derived?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+                emitter:            vaa.0.emitter_address.0,
+                pyth_emitter_chain: vaa.0.emitter_chain as u16,
+            }) {
+                return Err(Error::UnknownSource);
+            }
+        }
+
+        // Verify VAA and refund the caller in case of failure.
+        ext::ext_worm_hole::ext(env::current_account_id())
+            .with_static_gas(Gas(30_000_000_000_000))
```
tompntn, line 174: This could be a constant, like `GAS_FOR_VERIFY_VAA`.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        match Action::try_from_slice(rest).unwrap() {
+            Action::AddSource(source) => self.add_source(source),
+            Action::DelSource(source) => self.del_source(source),
+            Action::UpgradeContract(codehash) => self.set_upgrade_hash(codehash),
+            Action::BatchAttest(batch) => {
+                let bytes = &mut Cursor::new(batch);
+                let batch = BatchPriceAttestation::deserialize(bytes).unwrap();
+                let count = self.attest_prices(&batch);
+                log!("Updated {} Price Feeds", count);
```
tompntn, line 229: It might be nice to emit event log messages: https://github.com/near/NEPs/blob/master/specs/Standards/EventsFormat.md

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+                env::predecessor_account_id(),
+                env::attached_deposit(),
+                env::storage_usage(),
+            )))
+    }
+
+    /// This method is called after the upgrade to refund the caller for the storage used by the
+    /// old contract.
+    pub fn refund_upgrade(&mut self, account_id: AccountId, amount: u128, storage: u64) {
+        let refund = env::storage_byte_cost() * (env::storage_usage() - storage) as u128;
```
tompntn, line 266: Could we add brackets so that the order of operations is explicit here?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+        if signature != self.codehash {
+            return Err(Error::UnauthorizedUpgrade);
+        }
+
+        Ok(Promise::new(env::current_account_id())
+            .deploy_contract(new_code)
+            .then(Self::ext(env::current_account_id()).refund_upgrade(
+                env::predecessor_account_id(),
+                env::attached_deposit(),
+                env::storage_usage(),
```
tompntn, line 259: When do the arguments to the callback get executed? Does this happen before or after the new contract is deployed?

tompntn (commented):
This is awesome!! 🎉

contracts/near/src/state.rs
```diff
@@ -0,0 +1,98 @@
...
+    },
+};
+
+/// The PriceFeed structure is stored in the contract under a Price Feed Identifier.
+///
+/// This structure matches the layout of the PriceFeed structure in other Pyth receiver contracts
+/// but uses types that are native to NEAR.
+#[derive(BorshDeserialize, BorshSerialize, Serialize)]
+#[serde(crate = "near_sdk::serde")]
+pub struct PriceFeed {
```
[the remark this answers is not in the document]
ali-bahjati: I'm ok with exporting the whole contract as crate (using a `lib` feature flag) if that's possible.

My concern here is that this PriceFeed definition is outdated. Not for this PR, but can you update PriceService to the updated interface that we are using in Aptos and Ethereum?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+/// The `State` contains all persisted state for the contract. This includes runtime configuration.
+///
+/// There is no valid Default state for this contract, so we derive PanicOnDefault to force
+/// deployment using one of the #[init] functions in the impl below.
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct State {
+    /// The set of `Source`s from which Pyth attestations are allowed to be relayed from.
+    pub source: UnorderedSet<Source>,
```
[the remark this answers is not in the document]
ali-bahjati: I think `data_sources` is better, because we will have `governance_source` too.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+            }
+        }
+
+        // Verify VAA and refund the caller in case of failure.
+        ext::ext_worm_hole::ext(env::current_account_id())
+            .with_static_gas(Gas(30_000_000_000_000))
+            .verify_vaa(vaa.clone())
+            .then(
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas(30_000_000_000_000))
```
ali-bahjati, line 178: I recall that Josh was manually checking gas on each call and did not remember this syntax, is this new to ensure fixed remaining gas?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+                let bytes = &mut Cursor::new(batch);
+                let batch = BatchPriceAttestation::deserialize(bytes).unwrap();
+                let count = self.attest_prices(&batch);
+                log!("Updated {} Price Feeds", count);
+            }
+        }
+
+        // Refund Storage Difference to `account_id` after storage execution.
+        let storage_diff = env::storage_usage() - storage_usage;
+        Promise::new(account_id).transfer(storage_diff as u128 * env::storage_byte_cost());
```
ali-bahjati, line 235: I'm confused here, first, couldn't this value become negative? Second it seems that the caller is paying for the storage difference, not refunded :think:

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+    pub fn process_vaa_callback(
+        &mut self,
+        account_id: AccountId,
+        vaa: String,
+        #[callback_result] _result: Result<u32, near_sdk::PromiseError>,
+    ) {
+        // Get Storage Usage before executed.
+        let storage_usage = env::storage_usage();
+
+        // Deserialize VAA, note that we already deserialized and verified the VAA in `process_vaa`
```
ali-bahjati, line 215: We haven't verified the emitter address of the data yet.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    /// A hash of the current contract code.
+    ///
+    /// If this `hash` does not match the current contract code, then it indicates that a pending
+    /// upgrade has been allowed, calling `upgrade` with code that matches this hash will cause the
+    /// contract to upgrade itself.
+    pub codehash: [u8; 32],
+}
+
+/// The `Action` acts as an enum of possible `Action`'s this contract can receive via VAA.
```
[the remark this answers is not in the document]
ali-bahjati: Also governance actions have a separate source

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum Action {
+    AddSource(Source),
+    DelSource(Source),
```
ali-bahjati, line 78: Our governance message for replacing data sources is different.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+        })
+    }
+
+    /// Instruction for processing VAA's relayed via Wormhole.
+    ///
+    /// Note that VAA verification requires calling Wormhole so processing of the VAA itself is
+    /// done in a callback handler, see `process_vaa_callback`.
+    #[payable]
+    #[handle_result]
+    pub fn process_vaa(&mut self, vaa: String) -> Result<(), Error> {
```
ali-bahjati, line 157: in general we don't want to mention vaa in our public interfaces. can you use the same name as aptos or near?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+        // Verify the VAA is coming from a trusted source chain.
+        {
+            let vaa = hex::decode(&vaa).map_err(|_| Error::InvalidHex)?;
+            let vaa = vaa_payload::from_slice_with_payload::<structs::Vaa<()>>(&vaa)
+                .map_err(|_| Error::InvalidVaa)?;
+
+            if !self.source.contains(&Source {
```
ali-bahjati, line 164: source of governance vs sources of data sources are different

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+    DelSource(Source),
+    BatchAttest(Vec<u8>),
+    UpgradeContract([u8; 32]),
+}
+
+#[near_bindgen]
+impl State {
+    #[init]
+    #[allow(clippy::new_without_default)]
+    pub fn new(wormhole: AccountId, codehash: [u8; 32]) -> Self {
```
ali-bahjati, line 87: I'm not getting how sources could be set when there is no initial source :think: can you explain more?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        // Verify VAA and refund the caller in case of failure.
+        ext::ext_worm_hole::ext(env::current_account_id())
+            .with_static_gas(Gas(30_000_000_000_000))
+            .verify_vaa(vaa.clone())
+            .then(
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas(30_000_000_000_000))
+                    .with_attached_deposit(env::attached_deposit())
+                    .process_vaa_callback(env::predecessor_account_id(), vaa),
```
ali-bahjati, line 180: What is this attached deposit is used for? Also we would need to get fee for our contract too.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+    ///
+    /// If this `hash` does not match the current contract code, then it indicates that a pending
+    /// upgrade has been allowed, calling `upgrade` with code that matches this hash will cause the
+    /// contract to upgrade itself.
+    pub codehash: [u8; 32],
+}
+
+/// The `Action` acts as an enum of possible `Action`'s this contract can receive via VAA.
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum Action {
```
tompntn, line 76: Some governance actions are missing: https://github.com/pyth-network/pyth-crosschain/blob/main/aptos/contracts/sources/governance/governance_action.move#L4-L8

contracts/near/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+    InvalidVaa,
+
+    #[error("Source for attestation is not allowed.")]
+    UnknownSource,
+
+    #[error("Unauthorized Upgrade")]
+    UnauthorizedUpgrade,
+}
+
+impl near_sdk::FunctionError for Error {
```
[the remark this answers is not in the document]
Reisen: Didn't realise it was derivable, nice spot.

contracts/near/src/lib.rs
Reisen: Fixed.

ali-bahjati (commented):
@reisen any reason for not commiting `near/Cargo.lock`?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+                emitter:            vaa.emitter_address,
+                pyth_emitter_chain: vaa.emitter_chain,
+            }) {
+                return Err(Error::UnknownSource);
+            }
+
+            // Verify VAA and refund the caller in case of failure.
+            ext_wormhole::ext(env::current_account_id())
+                .with_static_gas(Gas(30_000_000_000_000))
+                .verify_vaa(vaa_hex.clone())
```
ali-bahjati, line 163: I'm not sure I know WH nor near well here, but it seems that we parse the VAA and wormhole also does that. If there is a gas concern, is it possible to ask wormhole to return the parsed vaa?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    /// Instruction for processing VAA's relayed via Wormhole.
+    ///
+    /// Note that VAA verification requires calling Wormhole so processing of the VAA itself is
+    /// done in a callback handler, see `process_vaa_callback`.
+    #[payable]
+    #[handle_result]
+    pub fn update_price_feeds(&mut self, vaas: Vec<String>) -> Result<(), Error> {
+        for vaa_hex in vaas {
```
ali-bahjati, line 142: Can you explain the behaviour when one of the VAAs fail here? Only the successful ones will succeed, right?
Also since we are in a loop, are gas conditions accurate?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+                .then(
+                    Self::ext(env::current_account_id())
+                        .with_static_gas(Gas(30_000_000_000_000))
+                        .with_attached_deposit(env::attached_deposit())
+                        .verify_vaa_callback(env::predecessor_account_id(), vaa_hex),
+                )
+                .then(
+                    Self::ext(env::current_account_id())
+                        .with_static_gas(Gas(30_000_000_000_000))
+                        .refund_vaa(env::predecessor_account_id(), env::attached_deposit()),
```
ali-bahjati, line 173: Since we are in a loop, are you sure the total refund works correctly? :think: refunding works ok, but refunding extra deposit probably not:
```
            // Otherwise we refund whatever is left over.
            if deposit - cost > 0 {
                Promise::new(refunder).transfer(cost);
            }
```

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Staleness threshold for rejecting price updates.
+    pub stale_threshold: u64,
+
+    /// Fee for updating price.
+    pub update_fee: Balance,
+}
+
+/// The `Action` acts as an enum of possible `Action`'s this contract can receive via VAA.
+#[derive(BorshDeserialize, BorshSerialize)]
```
ali-bahjati, line 92: I see `Action` to be used in the tests for serializing inputs into vaa, since it's mostly for test purposes shouldn't it go there?

Also the governance payloads are different.

ali-bahjati: well actually GovernanceAction has most of it.

contracts/near/src/governance.rs
```diff
@@ -0,0 +1,292 @@
...
+        match action {
+            0 => {
+                let mut hash = [0u8; 32];
+                cursor.read_exact(&mut hash).unwrap();
+                Self::ContractUpgrade(hash)
+            }
+
+            1 => {
+                let mut sources = Vec::new();
+                let count = cursor.read_u32::<LittleEndian>().unwrap();
```
ali-bahjati, line 72: iirc we always use big endian because evm is big endian.

contracts/near/src/governance.rs
```diff
@@ -0,0 +1,292 @@
...
+        env,
+        is_promise_success,
+        near_bindgen,
+        AccountId,
+        Balance,
+        Gas,
+        Promise,
+    },
+    std::io::Read,
+    structs::Chain as WormholeChain,
```
ali-bahjati, line 30: structs being a wormhole thing was really strange @_@

contracts/near/src/governance.rs
```diff
@@ -0,0 +1,292 @@
...
+            }
+
+            3 => {
+                let stale_price_threshold = cursor.read_u64::<LittleEndian>().unwrap();
+                Self::SetStalePriceThreshold(stale_price_threshold)
+            }
+
+            4 => {
+                let update_fee = cursor.read_u128::<LittleEndian>().unwrap();
+                Self::SetUpdateFee(update_fee)
```
ali-bahjati, line 102: SetUpdateFee has to 64 bit numbers, value and exponent => fee = value * 10**expo

contracts/near/src/state.rs
```diff
@@ -0,0 +1,155 @@
...
+)]
+#[serde(crate = "near_sdk::serde")]
+pub struct Source {
+    pub emitter:            [u8; 32],
+    pub pyth_emitter_chain: Chain,
+}
+
+/// A local `Vaa` type converted to from the Wormhole definition, this helps catch any upstream
+/// changes to the Wormhole VAA format.
+pub struct Vaa<P> {
```
ali-bahjati, line 127: It's clever to have it here! On eth we don't have it and I think the assumption is that they will never ever break it unless it becomes a new version with other apis.

ali-bahjati (commented):
It's very nice and everything looks good. I left some comments that needs to be addressed. I think tests are outdated and we need to governance actions are not tested.

There are some functionalities missing:
0. New price feed. 
1. User interface methods such as get_price_no_older_than and update_price_feed_if_fresh (or necessary)
2. Getting update fee from the user

All these are small changes but I think they can be addressed in a separate PR and we can put fixme or todo for them. It would be easier to review them in that sense.

At last, it would be great if you move near out of contracts folder.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub fn get_sources(&self) -> Vec<Source> {
+        self.sources.iter().collect()
+    }
+
+    /// Read the current Price for a given Price Feed.
+    ///
+    /// if the price is past the staleness threshold the trading status is set to Unknown before
+    /// returning to the caller.
+    pub fn get_price(&self, price_id: Identifier) -> Option<PriceFeed> {
+        self.prices.get(&price_id).and_then(|mut price| {
```
tompntn, line 279: The API for reading prices should be equivalent to the other contracts. Instead of the notion of "price status", we simply return `None` here (or maybe an informative `Err`) if a recent-enough price is not available. We then provide other functions to use a different recency threshold if appropiate for them. This is safer, because it forces the user to consider recency.

It would be good if we could have feature parity with https://github.com/pyth-network/pyth-crosschain/blob/main/aptos/contracts/sources/pyth.move:
- price_feed_exists
- get_price (with default staleness threshold)
- get_price_no_older_than (with custom staleness threshold)
- get_price_unsafe (return price regardless of age)

contracts/near/src/state.rs
```diff
@@ -0,0 +1,155 @@
...
+        PriceStatus,
+    },
+    structs::Chain as WormholeChain,
+};
+
+/// The PriceFeed structure is stored in the contract under a Price Feed Identifier.
+///
+/// This structure matches the layout of the PriceFeed structure in other Pyth receiver contracts
+/// but uses types that are native to NEAR.
+#[derive(BorshDeserialize, BorshSerialize, Serialize)]
```
tompntn, line 25: We have simplified the price feed struct, removing "price status" and other obselete fields - see https://github.com/pyth-network/pyth-crosschain/blob/main/aptos/contracts/sources/price_feed.move

contracts/near/src/governance.rs
```diff
@@ -0,0 +1,292 @@
...
+
+const GOVERNANCE_MAGIC: &[u8] = b"5054474d";
+const GOVERNANCE_MODULE: u8 = 0x01;
+
+impl GovernanceAction {
+    fn deserialize(data: &[u8]) -> Self {
+        let mut cursor = std::io::Cursor::new(data);
+        let magic = cursor.read_u32::<LittleEndian>().unwrap();
+        let module: u8 = cursor.read_u8().unwrap();
+        let action: u8 = cursor.read_u8().unwrap();
```
tompntn, line 56: Could we parse this `u8` into the action enum? For increased type-safety.

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+
+/// The `State` contains all persisted state for the contract. This includes runtime configuration.
+///
+/// There is no valid Default state for this contract, so we derive PanicOnDefault to force
+/// deployment using one of the #[init] functions in the impl below.
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Pyth {
+    /// The set of `Source`s from which Pyth attestations are allowed to be relayed from.
+    pub sources: UnorderedSet<Source>,
```
tompntn, line 63: Dumb question, but do you mind explaining how access control works here? I.e. what prevents unauthenticated/unguarded modification of this state?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub wormhole: AccountId,
+
+    /// A hash of the current contract code.
+    ///
+    /// If this `hash` does not match the current contract code, then it indicates that a pending
+    /// upgrade has been allowed, calling `upgrade` with code that matches this hash will cause the
+    /// contract to upgrade itself.
+    pub codehash: [u8; 32],
+
+    /// Staleness threshold for rejecting price updates.
```
tompntn, line 84: Could you specify the units here? Or use a `Duration`?

contracts/near/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    ContractUpgrade([u8; 32]),
+    SetDataSources(Vec<Source>),
+    SetGovernanceSource(Source),
+    SetStalePriceThreshold(u64),
+    SetUpdateFee(Balance),
+}
+
+#[near_bindgen]
+impl Pyth {
+    #[init]
```
tompntn, line 104: Does the account which called `new` have any authority over the contract, after it has been initialized? If so, we should try and make that not the case.

tompntn (commented):
Nice! :fire:

contracts/near/src/lib.rs
Reisen: Unfortunately no. However the VAA parse here ignores the payload and is the serde deserializer is cheap.

contracts/near/src/lib.rs
Reisen: You were right about this, and unfortunately It's not possible to wait on a promise chain in a reasonable way to get the deposit refunds correct, I tried several ways of combining the promises but without threading context through everything into a final handler it just gets lost. I decided It's not worth the complexity risk and callers, should they need it, can simply chain call the update function if they need several. The costs should be similar.

contracts/near/src/lib.rs
Reisen: As above.

contracts/near/src/lib.rs
Reisen: The contract is invoked by the NEAR runtime, which means the only time an `&mut T` is made available is just before the runtime invokes it after a contract call. However, none of these need to be `pub` so I went ahead and removed them all anyway.

contracts/near/src/lib.rs
Reisen: None that I'm aware of / or can find.

contracts/near/src/governance.rs
Reisen: This was because there still wasn't a merged `wormhole` and part of my feedback for that was renaming these things due to them being confusing. Those changes have been merged so this is now `wormhole` and `serde_wormhole` instead of `structs` and `vaa_payload`.

near/receiver/src/governance.rs
```diff
@@ -0,0 +1,385 @@
...
+
+            // Insert before calling Wormhole to prevent re-execution. If we wait until after the
+            // Wormhole call we could end up with multiple VAA's with the same sequence being
+            // executed in parallel.
+            self.executed_gov_sequences.insert(&vaa.sequence);
+        }
+
+        // Verify VAA and refund the caller in case of failure.
+        ext_wormhole::ext(self.wormhole.clone())
+            .with_static_gas(Gas(30_000_000_000_000))
```
jayantk, line 239: how do we know that this quantity is enough? does wormhole guarantee this?

what happens if there isn't enough gas?

jayantk: also extract out the gas number to a constant variable? seems like this number is used in a bunch of places

near/receiver/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas(30_000_000_000_000))
+                    .refund_vaa(env::predecessor_account_id(), env::attached_deposit()),
+            );
+
+        Ok(())
+    }
+
+    /// Return the deposit required to update a price feed.
+    pub fn get_update_fee_estimate(&self) -> u64 {
```
jayantk, line 166: document that this is an upper bound on the fee (I think?) If the PriceFeed is already in storage, then you don't re-incur the update cost?

Also the other contracts now take the list of VAAs as an argument, which would allow us to differentially price VAAs in the future.

near/receiver/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+        state
+    }
+
+    /// Instruction for processing VAA's relayed via Wormhole.
+    ///
+    /// Note that VAA verification requires calling Wormhole so processing of the VAA itself is
+    /// done in a callback handler, see `process_vaa_callback`.
+    #[payable]
+    #[handle_result]
+    pub fn update_price_feed(&mut self, vaa_hex: String) -> Result<(), Error> {
```
jayantk, line 128: FYI the other contracts now accept a list of VAAs here and for the fee estimate function below. Not sure if it's worth it to change the signature here -- if there's another way to craft an atomic tx that does multiple price updates, then this is fine. Especially so because the callbacks for processing multiple vaas will be gnarly.

near/receiver/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /// Read the list of accepted `Source` chains for a price attestation.
+    pub fn get_sources(&self) -> Vec<Source> {
+        self.sources.iter().collect()
+    }
+
+    /// Get the current staleness threshold.
+    pub fn get_stale_threshold(&self) -> u64 {
+        self.stale_threshold
```
jayantk, line 244: suggest naming this variable something like `stale_threshold_secs` so the unit is clear

near/receiver/src/state.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct PriceFeed {
+    /// Unique identifier for this price.
+    pub id:        PriceIdentifier,
+    /// The current aggregation price.
+    pub price:     Price,
+    /// Exponentially moving average price.
+    pub ema_price: Price,
+}
+
+// TODO: Source the Timestamp
```
jayantk, line 63: what does this mean?

near/receiver/tests/workspaces.rs
```diff
@@ -0,0 +1,592 @@
...
+        serde_json::from_slice::<Vec<Source>>(&contract.view("get_sources").await.unwrap().result)
+            .unwrap(),
+        &[
+            Source::default(),
+            Source {
+                emitter:            [1; 32],
+                pyth_emitter_chain: pyth::state::Chain(1),
+            },
+        ]
+    );
```
jayantk, line 131: can you test that we only accept prices from sources in the list?

jayantk (approved):
nice! code overall looks really good to me. I think there are a couple discrepancies between the interface here and the other contracts that will need to be fixed, but feel free to merge and address in later prs.

near/receiver/src/governance.rs
```diff
@@ -0,0 +1,385 @@
...
+                    pyth_emitter_chain: vaa.emitter_chain,
+                })
+            {
+                return Err(Error::UnknownSource);
+            }
+
+            // Insert before calling Wormhole to prevent re-execution. If we wait until after the
+            // Wormhole call we could end up with multiple VAA's with the same sequence being
+            // executed in parallel.
+            self.executed_gov_sequences.insert(&vaa.sequence);
```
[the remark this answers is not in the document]
ali-bahjati: I think with the current implementation someone can just deny our executed governance sequences by calling this with fake vaa. Can you explain why it can cause re-execution if we have it after wormhole call? this contract won't be executed parallel, right?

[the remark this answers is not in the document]
ali-bahjati: Also it's good to use a number instead of set as we do in other chains.. because we do not want out of order vaa execution.

ali-bahjati (approved):
Great. As Jayant said I think it's better to merge this one and address them in subsequent PRs.

I couldn't run the workspace tests and it couldn't deploy the wasm on the sandbox workspace, what should I do to get it working?

near/receiver/src/lib.rs
ali-bahjati: maybe change u64 to `Duration` from near. I'm confused why near has a duration, is the rust Duration not accessible in Near? Also now that `Duration` is near is nanoseconds we should make sure to convert it properly.

near/receiver/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// Get the latest available price cached for the given price identifier.
+    ///
+    /// WARNING:
+    ///
+    /// the returned price can be from arbitrarily far in the past. This function makes no
+    /// guarantees that the returned price is recent or useful for any particular application.
+    /// Users of this function should check the returned timestamp to ensure that the returned
+    /// price is sufficiently recent for their application. The checked get_price_no_older_than()
+    /// function should be used in preference to this.
+    pub fn get_price_unsafe(&self, price_identifier: PriceIdentifier) -> Option<Price> {
```
ali-bahjati, line 274: This is no longer an option

ali-bahjati: ah it is, if price feed doesn't exist

PR #408, opened 2022-11-30
Remove seq output log and py server for that

ali-bahjati:
The python server for serving serving sequence numbers is not used anywhere. The server required attester to output sequence numbers in stdout. This PR removes the server and moves sequence number output in the rust code to the previous info log.

### DIFF
--- a/third_party/pyth/p2w_autoattest.py
+++ b/third_party/pyth/p2w_autoattest.py
@@ -7,9 +7,7 @@ import os
 import re
 import sys
 import threading
-import time
 from http.client import HTTPConnection
-from http.server import BaseHTTPRequestHandler, HTTPServer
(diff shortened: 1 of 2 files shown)


jayantk (approved):

third_party/pyth/p2w_autoattest.py
```diff
@@ -303,31 +244,12 @@ def find_and_log_seqnos(s):
...
-        stdout=PIPE,
-        stderr=STDOUT,
-        text=True,
-        )
-
-    saved_log_lines = []
+        ]
+    )
 
     # Keep listening for seqnos until the program exits
```
drozdziak1, line 250:
```suggestion
    # Wait for an unexpected process exit
```

drozdziak1 (approved):

PR #410, opened 2022-12-02
Initial `pyth aptos` CLI

tompntn:
This is a simple CLI utility, designed to abstract some simple atomic operations on the Pyth Target Chain Aptos contracts.

### DIFF
--- a/aptos/cli/src/commands/aptos.ts
+++ b/aptos/cli/src/commands/aptos.ts
@@ -1,0 +1,499 @@
+import { spawnSync } from "child_process";
+import { BCS, AptosClient, AptosAccount, TxnBuilderTypes } from "aptos";
+import type { CommandBuilder } from "yargs";
+import fs from "fs";
+import sha3 from "js-sha3";
+import { ethers } from "ethers";
+
+const LOCALNET: string = "localnet";
+const TESTNET: string = "testnet";
+const MAINNET: string = "mainnet";
+
+interface Network {
+  // RPC endpoint of the network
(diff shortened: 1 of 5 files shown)


aptos/cli/src/commands/aptos.ts
```diff
@@ -0,0 +1,499 @@
...
+      "Deploy a package",
+      (yargs) => {
+        return yargs
+          .positional("package-dir", { type: "string" })
+          .positional("account", { type: "string" })
+          .option("named-addresses", { type: "string" })
+          .option("network", network);
+      },
+      async (argv) => {
+        const artefact = serializePackage(
```
ali-bahjati, line 91: is there a specific reason it is `artefact`? :D

ali-bahjati (approved):
Looks great to me!

I don't understand everything here, so I do not understand all the commands. I assume the wormhole contract command is useful for tilt setup 🤔 Maybe a description can be helpful.

Also in some places <package-dir> is passed, but maybe if it is only pyth package, hard-coding it as config make the cli easier. But of course it could be used for wormhole for tilt as well.

aptos/cli/src/commands/aptos.ts
tompntn: I took inspiration from the phrase "build artefact", but better name suggestions are welcome 😄

jayantk (approved):
the code itself seems fine,though i don't really understand what it's supposed to do. would suggest adding a README explaining how to use this utility (in a separate PR).

PR #412, opened 2022-12-02
[eth] Improve and automate deployment process

ali-bahjati:
This PR only limits truffle migrations to the first deployments of contracts and adds a `syncPythState` script to handle upgrades and state changes of the contracts. This script is an idempotent scripts that will check state of the off-chain contract vs the on-chain contract and if they are different, creates and executes a multisig transaction to update the on-chain one.

Currently the script only contains the logic for checking the contract version and upgrading it. In the future PRs other cases (data sources, fee, valid time) will be added.

Also, the script is stateful and caches some information about each step in different files to remain idempotent. In the future we will try to make it state-less.


jayantk (approved):
I'm going to approve this since it's temporary and will be superseded by the workflow stuff. there are a bunch of awkward things here that we shouldn't do in the long run (the temp files and the npm cli invocation for the multisig), but it's fine because it's going to get changed anyway.

PR #413, opened 2022-12-05
Integration test for attestations

jayantk:
This PR adds a new job to tilt that checks that all symbols being published are also being attested. I tweaked some things on the publisher http endpoint to ensure that the check only succeeds once all of the dynamically-added symbols are generated. I also tweaked the configuration there to make tilt complete a little faster.

### DIFF
--- a/third_party/pyth/check_attestations.py
+++ b/third_party/pyth/check_attestations.py
@@ -26,7 +26,7 @@ all_prices_attested = False
 while not all_prices_attested:
     publisher_state_map = get_pyth_accounts(PYTH_TEST_ACCOUNTS_HOST, PYTH_TEST_ACCOUNTS_PORT)
     pyth_price_account_ids = sorted([base58_to_hex(x["price"]) for x in publisher_state_map["symbols"]])
-    price_ids = sorted(get_json(PRICE_SERVICE_HOST, "/api/price_feed_ids", PRICE_SERVICE_PORT))
+    price_ids = sorted(get_json(PRICE_SERVICE_HOST, PRICE_SERVICE_PORT, "/api/price_feed_ids"))
 
     if price_ids == pyth_price_account_ids:
         if publisher_state_map["all_symbols_added"]:
--- a/third_party/pyth/check_attestations.py
+++ b/third_party/pyth/check_attestations.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+
+# This script is a CI test in tilt that verifies that prices are flowing through the entire system properly.
+# It checks that all prices being published by the pyth publisher are showing up at the price service.
+import logging
+import time
+from pyth_utils import *
+
+logging.basicConfig(
+    level=logging.DEBUG, format="%(asctime)s | %(module)s | %(levelname)s | %(message)s"
+)
+
+# Where to read the set of accounts from
+PYTH_TEST_ACCOUNTS_HOST = "pyth"
+PYTH_TEST_ACCOUNTS_PORT = 4242
+
+PRICE_SERVICE_HOST = "pyth-price-service" # FIXME
+PRICE_SERVICE_PORT = 4200 # FIXME
+
+all_prices_attested = False
+while not all_prices_attested:
+    publisher_state_map = get_pyth_accounts(PYTH_TEST_ACCOUNTS_HOST, PYTH_TEST_ACCOUNTS_PORT)
+    pyth_price_account_ids = sorted([x["price"] for x in publisher_state_map["symbols"]])
+    price_ids = sorted(get_json(PRICE_SERVICE_HOST, "/api/price_feed_ids", PRICE_SERVICE_PORT))
+
+    if price_ids == pyth_price_account_ids:
+        all_prices_attested = True
+    else:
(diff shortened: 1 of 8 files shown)


third_party/pyth/p2w_autoattest.py
```diff
@@ -167,7 +154,7 @@
     cfg_yaml += f"""
   - group_name: longer_interval_sensitive_changes
     conditions:
-      min_interval_secs: 10
+      min_interval_secs: 3
```
jayantk, line 157: attest faster so we don't have to wait as long for the checking to work

third_party/pyth/pyth_utils.py
```diff
@@ -17,14 +20,15 @@
 # How long to sleep between mock Pyth price updates
 PYTH_PUBLISHER_INTERVAL_SECS = float(os.environ.get("PYTH_PUBLISHER_INTERVAL_SECS", "5"))
 PYTH_TEST_SYMBOL_COUNT = int(os.environ.get("PYTH_TEST_SYMBOL_COUNT", "11"))
+PYTH_DYNAMIC_SYMBOL_COUNT = int(os.environ.get("PYTH_DYNAMIC_SYMBOL_COUNT", "2"))
```
jayantk, line 23: both of the changes here are again to make the attestation checking complete more quickly

third_party/pyth/pyth_utils.py
```diff
@@ -17,14 +20,15 @@
...
+PYTH_DYNAMIC_SYMBOL_COUNT = int(os.environ.get("PYTH_DYNAMIC_SYMBOL_COUNT", "3"))
 
 # If above 0, adds a new test symbol periodically, waiting at least
 # the given number of seconds in between
 #
 # NOTE: the new symbols are added in the HTTP endpoint used by the
 # p2w-attest service in Tilt. You may need to wait to see p2w-attest
 # pick up brand new symbols
-PYTH_NEW_SYMBOL_INTERVAL_SECS = int(os.environ.get("PYTH_NEW_SYMBOL_INTERVAL_SECS", "120"))
+PYTH_NEW_SYMBOL_INTERVAL_SECS = int(os.environ.get("PYTH_NEW_SYMBOL_INTERVAL_SECS", "60"))
```
ali-bahjati, line 31: Do you mind making it 30 seconds? To be even faster?

ali-bahjati (approved):
Very nice!

third_party/pyth/pyth_publisher.py
```diff
@@ -31,7 +31,7 @@ def do_GET(self):
         self.wfile.flush()
 
 # Test publisher state that gets served via the HTTP endpoint. Note: the schema of this dict is extended here and there
-HTTP_ENDPOINT_DATA = {"symbols": [], "mapping_address": None}
+HTTP_ENDPOINT_DATA = {"symbols": [], "mapping_address": None, "all_symbols_added": False}
```
drozdziak1, line 34: If I understand correctly, `all_symbols_added` makes the `check-attestations` dependency on `pyth` a special case. It is special, because an additional decision of the mock publisher  logic permits `check-attestations` to go green. 

I think that's reasonable, but can you note here what this is for?

third_party/pyth/check_attestations.py
```diff
@@ -0,0 +1,45 @@
...
+    price_ids = sorted(get_json(PRICE_SERVICE_HOST, "/api/price_feed_ids", PRICE_SERVICE_PORT))
+
+    if price_ids == pyth_price_account_ids:
+        if publisher_state_map["all_symbols_added"]:
+            logging.info("Price ids match and all symbols added. Enabling readiness probe")
+            all_prices_attested = True
+        else:
+            logging.info("Price ids match but still waiting for more symbols to come online.")
+    else:
+        logging.info("Price ids do not match")
```
drozdziak1, line 38: I think failing the process here could be reasonable. This case fails the assertion of the test and a loud crash+burn is justified. The way this is handled now, it will be just a readiness timeout, which means a delay before any problems are reported.  Failure of the process would be visible as soon as it occurs.

third_party/pyth/pyth_publisher.py
```diff
@@ -164,6 +164,9 @@ def add_symbol(num: int):
                 next_new_symbol_id += 1
                 dynamically_added_symbols += 1
 
+        if dynamically_added_symbols >= PYTH_DYNAMIC_SYMBOL_COUNT:
```
drozdziak1, line 167: I generally like the permissive comparison, but I think the `>` case would be a bug in this script, maybe we could yell about it and only be quiet for `==`?

drozdziak1 (approved):
Lookin' good! I'm not feeling strongly about those comments, but I'd love to see some of them applied.

third_party/pyth/check_attestations.py
jayantk: you can't fail here because there's a race condition between the attester and the HTTP endpoint which can cause this case to trigger (but the next loop iteration succeeds). Looping is more reliable.

drozdziak1: that makes sense

PR #418, opened 2022-12-07
[eth] Add new networks and their configurations

ali-bahjati:
This PR adds 4 new network configurations to the repo (Celo, KCC, Cronos, zkSync). Some of the new networks are using Wormhole receiver. So some bugs in their deployment scripts are fixed.

More importantly, receiver chain ids are added to the governance SDK. As these are not wormhole ids, we start numbering them from 60000 (close to the chain id max value (2^16-1 ~ 65000)

Just a note about zkSync. I have added configurations for both testnet and mainnet. However, only the testnet is open to the public. Unfortunately, as this is not fully EVM compatible I could not deploy it using truffle. I need to use their plugins in hardhat to do that and will work on it in a separate PR.

Note: Files in `networks/` and `.openzeppelin` are added automatically and you can skip them.


third_party/pyth/xc-governance-sdk-js/src/chains.ts
```diff
@@ -0,0 +1,11 @@
+import { CHAINS as WORMHOLE_CHAINS } from "@certusone/wormhole-sdk";
+
+const RECEIVER_CHAINS = {
+  cronos: 60001,
+  kcc: 60002,
+  zksync: 60003,
+};
+
+export const CHAINS = { ...WORMHOLE_CHAINS, ...RECEIVER_CHAINS };
```
jayantk, line 9: maybe check here that there aren't any reused chain ids?

jayantk (approved):

third_party/pyth/xc-governance-sdk-js/src/chains.ts
ali-bahjati: Hmm.. I double checked that the second term replaces the previous one. So I think it is fine as there might be times that wormhole is there and we haven't migrated yet.

ali-bahjati: Added a comment about that.

PR #420, opened 2022-12-08
[multisig] Improve verify, add localnet

guibescos:
- Add localnet to clusters with placeholders that will get replaced when tilt integration happens
- Update verify to use getProposalInstructions


ali-bahjati (approved):
Nice ser.

PR #423, opened 2022-12-09
[cosmos] Implement set fee governance instruction

jayantk:
Add the set fee governance instruction to the cosmos contract. I thought this would be a good warmup but the change sprawled a bit because I needed to add governance instructions, serialization, etc. I'll leave some comments inline on anything nonobvious.

### DIFF
--- a/cosmwasm/contracts/pyth/src/contract.rs
+++ b/cosmwasm/contracts/pyth/src/contract.rs
@@ -43,9 +43,7 @@ use {
         PriceStatus,
         ProductIdentifier,
     },
-    std::{
-        collections::HashSet,
-    },
+    std::collections::HashSet,
     wormhole::{
         msg::QueryMsg as WormholeQueryMsg,
         state::ParsedVAA,
@@ -135,13 +133,32 @@ fn update_price_feeds(
 fn execute_governance_instruction(
     mut deps: DepsMut,
     env: Env,
-    _info: MessageInfo,
+    info: MessageInfo,
     data: &Binary,
 ) -> StdResult<Response> {
-    let state = config_read(deps.storage).load()?;
     let vaa = parse_vaa(deps.branch(), env.block.time.seconds(), data)?;
 
+    execute_governance_instruction_from_vaa(deps, env, info, &vaa)
+}
+
+/// Helper function to improve testability of governance instructions (so we can unit test without wormhole).
+fn execute_governance_instruction_from_vaa(
+    mut deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    vaa: &ParsedVAA,
+) -> StdResult<Response> {
+    let state = config_read(deps.storage).load()?;
+
+    // store updates to the config as a result of this action in here.
+    let mut updated_config: ConfigInfo = state.clone();
     verify_vaa_from_governance_source(&state, &vaa)?;
+
+    if vaa.sequence <= state.governance_sequence_number {
+        return Err(PythContractError::OldGovernanceMessage)?;
+    } else {
+        updated_config.governance_sequence_number = vaa.sequence;
+    }
 
     let data = &vaa.payload;
     let instruction = GovernanceInstruction::deserialize(&data[..])
@@ -151,18 +168,20 @@ fn execute_governance_instruction(
         return Err(PythContractError::InvalidGovernancePayload)?;
     }
 
-    match instruction.action {
+    let response = match instruction.action {
         SetFee { new_fee } => {
-            let mut new_config = state;
-            new_config.fee = new_fee;
-            config(deps.storage).save(&new_config);
+            updated_config.fee = new_fee;
 
-            Ok(Response::new()
+            Response::new()
                 .add_attribute("action", "set_fee")
-                .add_attribute("new_fee", format!("{new_fee}")))
+                .add_attribute("new_fee", format!("{new_fee}"))
         }
         _ => Err(PythContractError::InvalidGovernancePayload)?,
-    }
+    };
+
+    config(deps.storage).save(&updated_config)?;
+
+    Ok(response)
 }
 
 fn add_data_source(
@@ -376,6 +395,10 @@ pub fn query_price_feed(deps: Deps, env: Env, address: &[u8]) -> StdResult<Price
 mod test {
     use {
         super::*,
+        crate::governance::{
+            GovernanceAction,
+            GovernanceModule::Target,
+        },
         cosmwasm_std::{
             testing::{
                 mock_dependencies,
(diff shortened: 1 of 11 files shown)


cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -61,12 +64,21 @@ pub fn instantiate(
...
-        owner:             info.sender,
-        wormhole_contract: deps.api.addr_validate(msg.wormhole_contract.as_ref())?,
-        data_sources:      HashSet::from([PythDataSource {
+        owner:                      info.sender,
+        wormhole_contract:          deps.api.addr_validate(msg.wormhole_contract.as_ref())?,
+        data_sources:               HashSet::from([PythDataSource {
             emitter:            msg.pyth_emitter,
             pyth_emitter_chain: msg.pyth_emitter_chain,
         }]),
+        chain_id:                   msg.chain_id,
```
jayantk, line 73: added governance fields to ConfigInfo and the initialization message. 

Note that `fee_denom` below seems to be needed to determine which coin the fees are paid in. I don't totally understand what this does, but I'm copying from wormhole here https://github.com/wormhole-foundation/wormhole/blob/main/cosmwasm/contracts/wormhole/src/contract.rs#L74

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -397,7 +513,7 @@ mod test {
         vaa.emitter_address = vec![1u8];
         vaa.emitter_chain = 3;
 
-        assert_eq!(verify_vaa_sender(&config_info, &vaa), Ok(()));
+        assert_eq!(verify_vaa_from_data_source(&config_info, &vaa), Ok(()));
```
jayantk, line 516: (this is a rename for clarity, since there are now two different VAA senders)

cosmwasm/contracts/pyth/src/contract.rs
ali-bahjati: We should think a little bit about it and we should ask WH people how they handled it. Basically in cosmos there might be more than 1 native transaction, as in terra there were both LUNA and UST. 

So the question is that do we want to have only one accepted coin per each chain?

cosmwasm/contracts/pyth/src/msg.rs
```diff
@@ -12,17 +12,26 @@ type HumanAddr = String;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 pub struct InstantiateMsg {
-    pub wormhole_contract:  HumanAddr,
-    pub pyth_emitter:       Binary,
-    pub pyth_emitter_chain: u16,
+    pub wormhole_contract:          HumanAddr,
+    pub pyth_emitter:               Binary,
```
ali-bahjati, line 16: Not related to this PR, but can we change the Inst..Msg to get a list of emitter addresses and chains instead of 1?

cosmwasm/contracts/pyth/src/msg.rs
```diff
@@ -12,17 +12,26 @@ type HumanAddr = String;
...
+
+    pub fee:       u128,
+    pub fee_denom: String,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum ExecuteMsg {
-    SubmitVaa { data: Binary },
+    UpdatePriceFeeds { data: Binary },
```
ali-bahjati, line 31: Not related to this PR, but would be nice if we add UpdatePriceFeedsIfNecessary in another PR too.

cosmwasm/contracts/pyth/src/state.rs
```diff
@@ -44,9 +39,17 @@ pub struct PythDataSource {
...
+    pub owner:                      Addr,
+    pub wormhole_contract:          Addr,
+    pub data_sources:               HashSet<PythDataSource>,
+    pub governance_source:          PythDataSource,
+    pub governance_sequence_number: u64,
+    pub chain_id:                   u16,
+    pub valid_time_period:          Duration,
+
+    // The fee to pay, denominated in fee_denom (typically, the chain's native token)
+    pub fee:       u128,
```
ali-bahjati, line 51: Why not storing Coin here?

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -115,6 +134,66 @@ fn submit_vaa(
...
+    if instruction.target_chain_id != state.chain_id && instruction.target_chain_id != 0 {
+        return Err(PythContractError::InvalidGovernancePayload)?;
+    }
+
+    let response = match instruction.action {
+        SetFee { val, expo } => {
+            updated_config.fee = (val as u128)
+                .checked_mul(
+                    10_u128
+                        .checked_pow(expo as u32)
```
ali-bahjati, line 180: Can we do a checked conversion here instead of `expo as u32`? `as` will truncate the value.

ali-bahjati (commented):
This is great! Sorry that the work is more than what you expected. I left couple of comments. 

Also I investigated the tilt problem and that is because of `uint128`. Apparently in cosmwasm you cannot use it at all in Ser/De (more info [here](https://github.com/CosmWasm/cosmwasm/issues/1162)). So you need to use `Uint128` here to fix the initialization issue. I tested it and it worked.

cosmwasm/contracts/pyth/src/state.rs
jayantk: I've been copying the wormhole code for all of this stuff. Let me circle back on this (and the supporting multiple coin types) after implementing the actual payment mechanics, which will help me understand how this works.

cosmwasm/contracts/pyth/src/governance.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        let module_num = bytes.read_u8()?;
+        let module = GovernanceModule::from_u8(module_num)?;
+
+        if module != GovernanceModule::Target {
+            return Err(format!("Invalid governance module {module_num}",).into());
+        }
+
+        // TODO: check endianness
```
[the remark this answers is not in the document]
jayantk: the EVM thing is exactly what i was wondering. thanks.

cosmwasm/contracts/pyth/src/contract.rs
jayantk: wormhole only supports one coin (`uluna`). We can support multiple, but let me add that in a separate pr. (I think we'll need more governance instructions if we go that route.)

ali-bahjati (approved):
Awesome!

PR #427, opened 2022-12-14
Check duplicates

guibescos:
This PR adds a check to stop `create` if the payload is already part of an existing proposal.

### DIFF
--- a/third_party/pyth/multisig-wh-message-builder/src/index.ts
+++ b/third_party/pyth/multisig-wh-message-builder/src/index.ts
@@ -27,7 +27,6 @@ import {
   getManyProposalsInstructions,
   getProposalInstructions,
 } from "./multisig";
-import { option } from "yargs";
 
 setDefaultWasm("node");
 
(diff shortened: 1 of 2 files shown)


third_party/pyth/multisig-wh-message-builder/src/index.ts
```diff
@@ -479,8 +529,7 @@ async function getSquadsClient(
       if (solRpcUrl) {
         return Squads.endpoint(solRpcUrl, wallet);
       } else {
-        console.log("rpc:", solRpcUrl);
-        throw `ERROR: solRpcUrl was not specified for localdevnet!`;
+        return Squads.localnet(wallet);
```
guibescos, line 532: Default RPC node to http://localhost:8899

ali-bahjati (approved):
Awesome!

Did you try it on mainnet/devnet too? I am curious to know how long does it take to fetch everything and do the filters.

guibescos:
> 

It was fine

PR #429, opened 2022-12-14
[wormhole-attester] Add a healthcheck on the metrics port

drozdziak1:
# why a healthcheck?
As a last resort, it makes sense to restart the attester service in production if it consistently fails to produce successful attestations.
# How to decide attester health?
A new attestation config yaml value `healthcheck_window_size` is introduced. This number decides how many last attestations must all fail before the service reports as unhealthy. The default is 100 consecutive failures.
# How to access the status?
A new endpoint is served alongside the metrics under `/healthcheck` (socket binds to 0.0.0.0:3000 by default).
A well-commented usage example is available in the k8s YAML in `devnet/p2w-attest.yaml`
# How was this tested?
1. the k8s tilt service has a new liveness probe for the attester
2. Tilt did not mark the service as green until it saw a 200 OK status code 
3. The ambiguous state before the healthcheck window is filled is represented with the 307 Moved Temporarily status, not seen as a success either in k8s.
4. Manually restarting the solana rpc service produced an unhealthy status after the healthcheck window filled with expected failures.
5. The service recovered some time after the restarted solana RPC recovered.

### DIFF
--- a/devnet/p2w-attest.yaml
+++ b/devnet/p2w-attest.yaml
@@ -45,11 +45,35 @@ spec:
             - name: P2W_EXIT_ON_ERROR
               value: "true"
           tty: true
-          readinessProbe:
-            tcpSocket:
-              port: 2000
-            periodSeconds: 1
-            failureThreshold: 300
+          # Probes, in order of appearance https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+          #
+          # Startup probe - delays other probes until it gets its first success
+          startupProbe:
+            httpGet:
+              path: /healthcheck
+              port: 3000
+            failureThreshold: 100 # up to 100 * 10 seconds to report initial healthy status
+            periodSeconds: 10
+          # Readiness probe - Used to tell load balancers to
+          # start/stop sending traffic to the container, *without*
+          # restarting it. The attester does not accept any traffic as
+          # part of its workflow, which means this isn't very useful.
+          # readinessProbe:
+          # httpGet:
+          #   path: /healthcheck
+          #   port: 3000
+          # failureThreshold: 1
+          # periodSeconds: 10
+          #
+          # Liveness probe - decides restarts for misbehaving
+          # containers
+          livenessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 3000
+            failureThreshold: 1 # If the attester healthcheck fails once,
+            periodSeconds: 10
+
           ports:
             - containerPort: 4343
               name: p2w-attest
--- a/solana/pyth2wormhole/client/src/healthcheck.rs
+++ b/solana/pyth2wormhole/client/src/healthcheck.rs
@@ -1,0 +1,49 @@
+use {
+    std::{
+        collections::VecDeque,
+        sync::Arc,
+    },
+    tokio::sync::Mutex,
+};
+
+lazy_static::lazy_static! {
+    pub static ref HEALTHCHECK_STATE: Arc<Mutex<HealthCheckState>> = Arc::new(Mutex::new(HealthCheckState::new(0)));
+}
+
+/// Helper structure for deciding service health
+pub struct HealthCheckState {
+    /// Sliding LIFO window over last `max_window_size` attestation results (true = ok, false = error)
+    pub window:          VecDeque<bool>,
+    /// Window size; 0 disables the healthcheck, causing it to always return None
+    pub max_window_size: usize,
+}
+
+
+impl HealthCheckState {
+    pub fn new(max_window_size: usize) -> Self {
+        Self {
+            window: VecDeque::with_capacity(max_window_size),
+            max_window_size,
+        }
+    }
+    /// Check service health, return None if not enough data is present
+    pub fn is_healthy(&self) -> Option<bool> {
+        if self.window.len() >= self.max_window_size && self.max_window_size > 0 {
+            // If all results are false, return false (unhealthy).
+            Some(self.window.iter().any(|entry| *entry))
+        } else {
+            // The window isn't big enough yet or the size is 0
+            None
+        }
+    }
+
+    /// Rotate the window
+    pub fn add_result(&mut self, res: bool) {
+        self.window.push_front(res);
+
+        // Trim window back to size if needed. truncate() deletes from
+        // the back and has no effect if new size is greater than
+        // current size.
+        self.window.truncate(self.max_window_size);
+    }
+}
--- a/solana/pyth2wormhole/client/src/healthcheck.rs
+++ b/solana/pyth2wormhole/client/src/healthcheck.rs
@@ -7,32 +7,35 @@ use {
 };
 
 lazy_static::lazy_static! {
-    pub static ref HEALTHCHECK_STATE: Arc<Mutex<HealthCheckState>> = Arc::new(Mutex::new(HealthCheckState::new(0)));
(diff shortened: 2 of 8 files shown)


solana/pyth2wormhole/client/src/main.rs
```diff
@@ -291,6 +294,16 @@ async fn handle_attest_daemon_mode(
     attestation_cfg: AttestationConfig,
     metrics_bind_addr: SocketAddr,
 ) -> Result<(), ErrBox> {
+    // Update healthcheck window size from config
+    {
+        let mut hc = HEALTHCHECK_STATE.lock().await;
+        hc.max_window_size = attestation_cfg.healthcheck_window_size as usize;
+
+        if hc.max_window_size == 0 {
```
jayantk, line 302: I think using None would be better than 0 to disable the check.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -660,12 +673,13 @@ async fn attestation_sched_job(args: AttestationSchedJobArgs) -> Result<(), ErrB
...
         // We never get to error reporting in daemon mode, attach a map_err
-        let job_with_err_msg = job.map_err(move |e| {
+        let job_with_err_msg = job.or_else(move |e| async move {
             warn!(
                 "Batch {}/{}, group {:?} ERR: {:#?}",
                 batch_no4err_msg, batch_count4err_msg, group_name4err_msg, e
             );
-            e
+            HEALTHCHECK_STATE.lock().await.add_result(false); // Report this job as failed to healthcheck
+            Err(e)
```
jayantk, line 682: I think you should extract out a function like `attestation_job_with_healthcheck` that wraps `attestation_job` and adds both cases of the healthcheck logic. That approach reduces the probability of programmer errors compared to splitting the logic out into two different places in the code.

devnet/p2w-attest.yaml
```diff
@@ -45,11 +45,35 @@ spec:
...
-            periodSeconds: 1
-            failureThreshold: 300
+          # Probes, in order of appearance https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+          #
+          # Startup probe - delays other probes until it gets its first success
+          startupProbe:
+            httpGet:
+              path: /healthcheck
+              port: 3000
+            failureThreshold: 100 # up to 100 * 10 seconds to report initial healthy status
```
jayantk, line 55: i don't understand this comment. why the 10 seconds part?

jayantk (commented):
I can't review the actual endpoint logic (i don't know what the behavior there should be), but the rest seems pretty reasonable. left a couple minor comments.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I think an async block inside `attestation_job` could work well for this. As for the separation, it's the non-daemon mode that's making it necessary. I don't think we need the daemon mode anymore. It was only ever used for triggering readiness in the Python script and this changeset makes the Python readiness socket obsolete.

devnet/p2w-attest.yaml
drozdziak1: it will try 100 times, waiting 10 seconds (periodSeconds) in between, for a total of 100 * 10 seconds to become ready and move on to the next probe.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I don't like that because explicit none is hard to represent in serde. Not specifying the value is not a good way to say "off" because then there's no good way to say "use default". I think I'll add an explicit on/off option for the healthcheck and forbid the 0 value (it no longer has a purpose if it doesn't have the special "off" meaning)

solana/pyth2wormhole/client/src/main.rs
drozdziak1: That said, it deserves its own change better IMO

devnet/p2w-attest.yaml
ali-bahjati: I think here `initialDelaySeconds` is cleaner.

solana/pyth2wormhole/client/src/healthcheck.rs
```diff
@@ -0,0 +1,49 @@
...
+    },
+    tokio::sync::Mutex,
+};
+
+lazy_static::lazy_static! {
+    pub static ref HEALTHCHECK_STATE: Arc<Mutex<HealthCheckState>> = Arc::new(Mutex::new(HealthCheckState::new(0)));
+}
+
+/// Helper structure for deciding service health
+pub struct HealthCheckState {
```
ali-bahjati, line 14: Are we planning to make the logic here more complex (e.g.: not healthy if more than 50% failed)? Because I think the current logic could be implemented easier, just with a counter of "how many consecutive attestations have failed", a failure adds one and a success makes it 0.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -82,20 +85,20 @@ use {
...
 
-pub mod cli;
-
 pub const SEQNO_PREFIX: &str = "Program log: Sequence: ";
 
 lazy_static! {
     static ref ATTESTATIONS_OK_CNT: IntCounter =
-        register_int_counter!("attestations_ok", "Number of successful attestations").unwrap();
+        register_int_counter!("attestations_ok", "Number of successful attestations")
+            .expect("FATAL: Could not instantiate ATTESTATIONS_OK_CNT");
```
ali-bahjati, line 93: Good refactor 👍

solana/pyth2wormhole/client/src/main.rs
ali-bahjati: I agree that removing the non-daemon mode is the way to go (and makes code much simpler), specially after this change. I thin it's better to do it in a subsequent PR and not here.

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -826,6 +840,7 @@ async fn attestation_job(args: AttestationJobArgs) -> Result<(), ErrBoxSend> {
         batch_no, batch_count, group_name, seqno
     );
     ATTESTATIONS_OK_CNT.inc();
+    HEALTHCHECK_STATE.lock().await.add_result(true); // Report this job as successful to healthcheck
```
ali-bahjati, line 843: Can you move it the line after? Since it's a blocking operation it's better than there'd be nothing after it. 

In this case the gauge will be more accurate if it is before this statement.

solana/pyth2wormhole/client/src/healthcheck.rs
ali-bahjati: After seeing the rest of the code, I get why you have it. Then it makes sense to have it for reporting.

ali-bahjati (approved):
Awesome, and great PR description 🙇 I would love to see a subsequent PR to remove non-daemon mode after this.

devnet/p2w-attest.yaml
drozdziak1: The initial delay will not poll the service, incurring a fixed delay instead. This check is not expensive, I would not do that.

ali-bahjati: Makes sense. 👍

solana/pyth2wormhole/client/src/healthcheck.rs
drozdziak1: I think it _could_ grow. I think your counter idea is very cool, I didn't think of that. But i think it's reasonable to assume we might want a percentage in the future and only a sliding window will be able to cut it.

PR #430, opened 2022-12-15
[multisig-cli] Add support for json

guibescos:
We need `create` and `verify` to be compatible with program-admin generated json.
This PR adds the `--file` to `create` and `verify`. Only one of `--payload` or `--file` can be provided. 
When `--file` is provided the tool will parse the path provided and try to create a proposal with the instruction in that file.

Welcome to ideas to make it more understandable. I find payload/file somehow confusing.

### DIFF
--- a/third_party/pyth/multisig-wh-message-builder/src/index.ts
+++ b/third_party/pyth/multisig-wh-message-builder/src/index.ts
@@ -165,7 +165,8 @@ program
     "multisig wallet secret key filepath",
     "keys/key.json"
   )
-  .option("-p, --payload <hex-string>", "payload to sign", "0xdeadbeef")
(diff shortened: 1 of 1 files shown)


ali-bahjati (approved):
Nice!

I'm not happy with the structure either. I wish we can change it, or abstract it. Maybe in a way that (1) payload and file both be the same (a json) and one helper method reads from either of them. (2) we have separate command or argument for wormhole.

So it can look like this:
```
./cli create <--payload ""/--file ""> [--solana-instruction] [--wormhole]
./cli verify  <--payload ""/--file ""> [--solana-instruction] [--wormhole]
./cli execute <txKey> [--wormhole]
```

But definitely can go for the next PR.

ali-bahjati (commented):
Oh also I suggest updating the readme too.

PR #433, opened 2022-12-16
[cosmos] Pay fee + mock wormhole for testing

jayantk:
Add fee payment logic to update price feeds. I also added a mock wormhole contract so we can test the contract operations on the raw binary messages (which we needed for the fee test)

### DIFF
--- a/cosmwasm/contracts/pyth/src/contract.rs
+++ b/cosmwasm/contracts/pyth/src/contract.rs
@@ -127,19 +127,14 @@ fn update_price_feeds(
 ) -> StdResult<Response> {
     let state = config_read(deps.storage).load()?;
 
-    println!("Checking fee");
     let fee = Coin::new(state.fee.u128(), state.fee_denom.clone());
     if fee.amount.u128() > 0 && !has_coins(info.funds.as_ref(), &fee) {
         return Err(PythContractError::InsufficientFee.into());
     }
 
-    println!("Parsing VAA");
     let vaa = parse_vaa(deps.branch(), env.block.time.seconds(), data)?;
-
-    println!("checking data source");
     verify_vaa_from_data_source(&state, &vaa)?;
 
-    println!("processing batch attestation");
     let data = &vaa.payload;
     let batch_attestation = BatchPriceAttestation::deserialize(&data[..])
         .map_err(|_| PythContractError::InvalidUpdatePayload)?;
@@ -517,7 +512,7 @@ mod test {
             nonce:              0,
             len_signers:        0,
             emitter_chain:      0,
-            emitter_address:    vec![0; 32],
+            emitter_address:    vec![],
             sequence:           0,
             consistency_level:  0,
             payload:            vec![],
--- a/cosmwasm/contracts/pyth/src/contract.rs
+++ b/cosmwasm/contracts/pyth/src/contract.rs
@@ -127,15 +127,19 @@ fn update_price_feeds(
 ) -> StdResult<Response> {
     let state = config_read(deps.storage).load()?;
 
+    println!("Checking fee");
     let fee = Coin::new(state.fee.u128(), state.fee_denom.clone());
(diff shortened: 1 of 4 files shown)


cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -139,26 +145,15 @@ fn update_price_feeds(
...
+    _info: MessageInfo,
     data: &Binary,
 ) -> StdResult<Response> {
     let vaa = parse_vaa(deps.branch(), env.block.time.seconds(), data)?;
-
-    execute_governance_instruction_from_vaa(deps, env, info, &vaa)
-}
-
-/// Helper function to improve testability of governance instructions (so we can unit test without wormhole).
-fn execute_governance_instruction_from_vaa(
```
jayantk, line 151: helper is not needed anymore because we can test `execute_governance_instruction` directly

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -462,6 +520,20 @@ mod test {
         }
     }
 
+    fn create_price_update_msg(emitter_address: &[u8], emitter_chain: u16) -> Binary {
+        let batch_attestation = BatchPriceAttestation {
+            // TODO: pass these in
+            price_attestations: vec![],
```
jayantk, line 526: we'll be able to use this in the price update tests as well, but i'll do it separately because it'll touch a lot of code.

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -120,13 +122,17 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> S
...
     env: Env,
-    _info: MessageInfo,
+    info: MessageInfo,
     data: &Binary,
 ) -> StdResult<Response> {
     let state = config_read(deps.storage).load()?;
 
-    let vaa = parse_vaa(deps.branch(), env.block.time.seconds(), data)?;
+    let fee = Coin::new(state.fee.u128(), state.fee_denom.clone());
+    if fee.amount.u128() > 0 && !has_coins(info.funds.as_ref(), &fee) {
```
ali-bahjati, line 131: I think it's better to store this as state, iirc Wormhole does the same.

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -446,6 +463,47 @@ mod test {
         (dependencies, mock_env())
     }
 
+    /// Mock handler for wormhole queries.
+    /// Warning: the interface for the `VerifyVAA` action is slightly different than the real wormhole contract.
+    /// In the mock, you pass in a binary-encoded `ParsedVAA`, and that exact vaa will be returned by wormhole.
+    /// The real contract uses a different binary VAA format (see `ParsedVAA::deserialize`) which includes
+    /// the guardian signatures.
+    fn handle_wasm_query(wasm_query: &WasmQuery) -> QuerierResult {
```
ali-bahjati, line 471: I suggest renaming it to `handle_wormhole_wasm_query`

ali-bahjati (approved):
Nice 🔥 I left some minor comments. Also, I fixed the relayer problem (that caused tilt to fail) and pushed it to the repo. Please review it.

For subsequent PRs:
I think for this feature to be complete we need to add a query message to get fee (I left a todo on terra relayer). Please free to ignore the cw sdk for now. But in the future we should update the sdk and also the Terra JS package to add fees in transaction messages.

I have one open question for the future, the interface on CW is not a bit different than EVM and aptos. On those chains we get an array of batches as input to update prices, but here we only get one batch and are creating 1 message per batch. While this works without any problem on the update price feeds method, if we want to add `parsePriceFeedUpdates` then we need to have them together. 
I don't have a strong opinion but slightly prefer to have this the same as other chains.

PR #437, opened 2022-12-20
[multisig-cli] Refactor check that multisig vault already exists

guibescos:
Fixes the check that checks if a vault already exists at the address when trying to create a new vault.


ali-bahjati:
I couldn't understand the relationship of the change and the bug 😅 Maybe Stan knows, but can you explain it a little bit if possible?

drozdziak1 (approved):

PR #444, opened 2022-12-28
add binary option to subscribe to binary vaas

cctdaniel:


third_party/pyth/price-service/src/ws.ts
```diff
@@ -37,6 +39,7 @@ export class WebSocketAPI {
   private wsCounter: number;
   private priceFeedClients: Map<HexString, Set<WebSocket>>;
   private priceFeedClientsVerbosity: Map<HexString, Map<WebSocket, boolean>>;
+  private priceFeedClientsBinary: Map<HexString, Map<WebSocket, boolean>>;
```
jayantk, line 42: Instead of adding more maps like this, would be better to have a single variable of type `Map<HexString, Map<WebSocket, WebsocketConfig>>` where the WebsocketConfig had verbose and boolean fields. That approach makes it impossible for someone to accidentally put the maps in an inconsistent state (e.g., add to one map and not the other).

jayantk (approved):

ali-bahjati (approved):
Nice! Can you add some tests for it before merging it?

PR #447, opened 2022-12-29
Upgrade contract & fix deployment script

jayantk:
Add the code for upgrading the contract and patch up the deployment script so it works on the new Terra network.

I haven't tested the upgrade process yet, as it's hard to do with the current deployment tools. I'll do some more work on the deployment process and then circle back to testing upgrades.

### DIFF
--- a/cosmwasm/contracts/pyth/src/contract.rs
+++ b/cosmwasm/contracts/pyth/src/contract.rs
@@ -29,6 +29,10 @@ use {
             PythDataSource,
         },
     },
+    byteorder::{
+        BigEndian,
+        ReadBytesExt,
+    },
     cosmwasm_std::{
         coin,
         entry_point,
@@ -36,6 +40,7 @@ use {
         to_binary,
         Binary,
         Coin,
+        CosmosMsg,
         Deps,
         DepsMut,
         Env,
@@ -46,6 +51,7 @@ use {
         Response,
         StdResult,
         Timestamp,
+        WasmMsg,
         WasmQuery,
     },
     p2w_sdk::BatchPriceAttestation,
@@ -67,9 +73,21 @@ use {
     },
 };
 
+/// Migration code that runs once when the contract is upgraded. On upgrade, the migrate
+/// function in the *new* code version is run, which allows the new code to update the on-chain
+/// state before any of its other functions are invoked.
+///
+/// After the upgrade is complete, the code in this function can be deleted (and replaced with
+/// different code for the next migration).
+///
+/// Most upgrades won't require any special migration logic. In those cases,
+/// this function can safely be implemented as:
+/// ```ignore
+/// Ok(Response::default())
+/// ```
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
-    Ok(Response::new())
+    Ok(Response::default())
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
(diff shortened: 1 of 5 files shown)


cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -67,9 +74,21 @@ use {
...
+/// state before any of its other functions are invoked.
+///
+/// After the upgrade is complete, the code in this function can be deleted (and replaced with
+/// different code for the next migration).
+///
+/// Most upgrades won't require any special migration logic. In those cases,
+/// this function can safely be implemented as:
+/// ```ignore
+/// Ok(Response::default())
+/// ```
```
jayantk, line 88: (not sure if there's a difference between ::new and ::default, but wormhole is using ::default)

cosmwasm/tools/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "deploy.js",
   "type": "module",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "deploy-pyth": "node deploy-pyth-bridge.js"
```
jayantk, line 8: the README claims this command exists, but it doesn't for whatever reason.

cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -182,9 +201,18 @@ fn execute_governance_instruction(
...
     let response = match instruction.action {
-        UpgradeContract { .. } => {
-            // FIXME: implement this
-            Err(PythContractError::InvalidGovernancePayload)?
+        UpgradeContract { address } => {
+            if instruction.target_chain_id == 0 {
+                Err(PythContractError::InvalidGovernancePayload)?
+            }
+
+            // Hack: We're using the first 8 bytes of the address field as the new code id.
```
ali-bahjati, line 209: Sorry for the inconsistency here.. for this one aptos uses 32 bytes and Eth uses 20 bytes. I think it is fine to define the governance instruction to use only 8 bytes instead of getting a prefix of a longer address. 

Although not used anywhere, you will need to define it here too (for scripts/tools that manage deployments):
https://github.com/pyth-network/pyth-crosschain/blob/main/third_party/pyth/xc-governance-sdk-js/src/instructions.ts#L103-L121

cosmwasm/tools/deploy-pyth-bridge.js
```diff
@@ -64,38 +61,96 @@ const artifact = argv.artifact;
...
+      governance_source: {
+        emitter: Buffer.from(
+          "5635979a221c34931e32620b9293a463065555ea71fe97cd6237ade875b12e9e",
+          "hex"
+        ).toString("base64"),
+        chain_id: 1,
+      },
+      governance_source_index: 0,
+      governance_sequence_number: 0,
+      chain_id: 18,
```
ali-bahjati, line 96: not for this pr, can we query this from wormhole contract instead of manually specifying it?

ali-bahjati (commented):
I left some comments about the upgrade payload (and adding it to the js sdk). I do not fully know general cosmos and terra differences and I have a question before approving this one: Don't we need to change the admin of the contract to itself? If so, can we add that to the deploy script? iirc we cannot know the contract address before instantiation and it should be a separate message.

cosmwasm/tools/deploy-pyth-bridge.js
jayantk: I looked into this as part of the cleanup PR before. Unfortunately wormhole doesn't expose a query for it. We could read it out of their config struct, but that felt a little hacky. wdyt?

jayantk: or we could add support to it to the wormhole contract?

ali-bahjati: ok never mind then 😅 second one is good but it takes time for wh to merge and then deploy it. let's stick to this one 😅

ali-bahjati (approved):
🚢 it

PR #450, opened 2023-01-02
[eth] Add zksync deployment files

ali-bahjati:
This PR adds the requied files and configurations for deploying Pyth to zkSync networks (currently only testnet).

Unfortunately they only support hardhat and I needed to rewrite deployment scripts. The zksync plugin expects deployment scripts to be on `deploy` directory and I couldn't get away from it.

Also as our deployment/sync script is based in truffle contract upgrades won't work automatically and I added a script for the part that needs to be done in hardhat with proper documentation. It is not very clean and I might change it in the future (or the deployment automation might improve it).

Please read Deploying.md for more information on how this works.


jayantk (approved):
thanks. thankfully managing the contracts shouldn't be as bad since upgrades etc. are done via wormhole.

PR #451, opened 2023-01-04
RFC : Add parser

guibescos:
I want some feedback on this. 
The goal is parsing the multisig proposals. This is achieved by parsing the instructions of the proposal one by one. 
Two representations of the instruction are useful :
- A detailed one for the UI (caveat : solana instructions and pythnet remote-instructions will be represented differently than cross-chain governance messages)
- A "summary" (not sure about the word) that will be used to detect conflicts between two proposals. If two instructions have the same summary then they are in conflict with one another. 

I made summary some sort of nested object, but I'm thinking of turning it into a string.

### DIFF
--- a/third_party/pyth/multisig-wh-message-builder/src/parser.ts
+++ b/third_party/pyth/multisig-wh-message-builder/src/parser.ts
@@ -1,0 +1,118 @@
+import { ChainId, CHAIN_ID_SOLANA } from "@certusone/wormhole-sdk";
+import {
+  AccountMeta,
+  PublicKey,
+  TransactionInstruction,
+} from "@solana/web3.js";
+import { pythOracleCoder, pythIdl } from "@pythnetwork/client";
+import { Cluster, CONFIG } from ".";
+import { Idl } from "@project-serum/anchor";
+
+/// A summary of a proposal instruction :
+/// For two proposals A and B in the multisig the set of all the summarized instructions of A and all the summarized instructions of B should have null intersection
+type Summary = {
+  targetChain: ChainId;
+  component: OracleProgram;
+};
+type OracleProgram = {
+  oracleAction:
+    | UpdatePriceFeed
+    | "updatePermissions"
+    | UpdateProductAccount
+    | undefined;
+};
+
+type UpdatePriceFeed = {
+  priceAccount: PublicKey;
+};
+
+type UpdateProductAccount = {
+  productAccount: PublicKey;
+};
+
+type NamedAccounts = Record<string, AccountMeta>;
+
+export function parseProposalInstruction(
+  cluster: Cluster,
+  instruction: TransactionInstruction
+) {
+  switch (instruction.programId.toBase58()) {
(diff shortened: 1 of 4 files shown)


third_party/pyth/multisig-wh-message-builder/src/parser.ts
```diff
@@ -0,0 +1,118 @@
...
+          namedAccounts
+        );
+        return { ...deserializedInstruction, namedAccounts, summary };
+      } else {
+        return null;
+      }
+
+    default:
+      return null;
+  }
```
jayantk, line 61: so then there's going to be a case here for wormhole and then you inspect the VAA itself?

third_party/pyth/multisig-wh-message-builder/src/parser.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export function resolveAccountNames(
```
jayantk, line 64: going through the IDL may be too indirect. you may be better off storing the mapping from account indexes -> keys directly in summarizeOracleInstruction.

jayantk: (i'm not sure about this comment. Key question for you to answer is: what value is the IDL providing?)

third_party/pyth/multisig-wh-message-builder/src/parser.ts
```diff
@@ -0,0 +1,118 @@
...
+        targetChain: CHAIN_ID_SOLANA,
+        component: {
+          oracleAction: { priceAccount: namedAccounts.priceAccount.pubkey },
+        },
+      };
+    case "updPermissions":
+      return {
+        targetChain: CHAIN_ID_SOLANA,
+        component: { oracleAction: "updatePermissions" },
+      };
```
jayantk, line 111: in the future, this bit of code will also decode the arguments to the instruction?

third_party/pyth/multisig-wh-message-builder/src/parser.ts
```diff
@@ -0,0 +1,118 @@
...
+        const namedAccounts = resolveAccountNames(
+          pythIdl as unknown as Idl,
+          deserializedInstruction.name,
+          instruction
+        );
+        const summary: Summary = summarizeOracleInstruction(
+          deserializedInstruction.name,
+          namedAccounts
+        );
+        return { ...deserializedInstruction, namedAccounts, summary };
```
jayantk, line 54: what's in deserialized instruction? is this representation generic enough to work for the other kinds of operations that will be stored in the multisig?

jayantk (commented):
Couple thoughts on this:

1. I think you want the flow to be something like raw instruction -> detailed representation of contained operation -> summary.
2. I think you probably want to control the detailed representation and not depend on whatever the anchor IDL thing does. You're going to need to represent operations that anchor doesn't (e.g., stuff that's in a wormhole VAA), and I doubt you'll want to shoehorn that into anchor's represntation.

PR #453, opened 2023-01-04
[near] Fix governance inconsistencies.

Reisen:
This PR cleans up some of the issues from the previous NEAR merge. In particular:

- Matches the governance actions that other contracts currently have.
- Introduces `strum` for enum safety (compile time only, free).
- Introduces `nom` to clear up the parsing.
- Fixes the fees.

The reason for introducing nom here is it gives parsing errors, parsing exhaustive checks, and generally simplifies the pain around parsing versus the usual byte based approaches using cursors or readers. Much of the common parsing gotchas are taken care of by the combinators it provides, the whole thing has very very little overhead in terms of instructions and binary size, only marginally more than serde `Serialize/Deserialize`.

Test output until in CI:

```
      Starting 10 tests across 2 binaries                                                                                                                                                                                                                           
          PASS [   0.002s]             pyth-near governance::tests::test_governance_serialize_matches_deserialize                                                                                                                                                   
          PASS [   0.002s]             pyth-near governance::tests::test_set_update_fee                                                                                                                                                                             
          PASS [   0.002s]             pyth-near governance::tests::test_set_valid_period                                                                                                                                                                           
          PASS [   0.001s]             pyth-near governance::tests::test_upgrade_fail                                                                                                                                                                               
          PASS [   0.002s]             pyth-near governance::tests::test_upgrade                                                                                                                                                                                    
          PASS [   0.001s]             pyth-near tests::tests::test_contract_init                                                                                                                                                                                   
          PASS [  12.895s] pyth-near::workspaces test_source_add                                                                                                                                                                                                    
          PASS [  16.408s] pyth-near::workspaces test_set_governance_source                                                                                                                                                                                         
          PASS [  16.417s] pyth-near::workspaces test_contract_fees                                                                                                                                                                                                 
          PASS [  19.930s] pyth-near::workspaces test_stale_threshold                                                                                                                                                                                               
  ------------                                                                                                                                                                                                                                                      
       Summary [  19.932s] 10 tests run: 10 passed, 0 skipped
```

Contract size pre and post `wasm-opt`:

```
$ wasm-opt -O4 pyth.wasm -o pyth.O4.wasm # Performance Optimization
$ wasm-opt -Oz pyth.wasm -o pyth.Oz.wasm # Code Size Optimization
$ du -sh *.wasm
452K    pyth.O4.wasm
452K    pyth.Oz.wasm
552K    pyth.wasm
```


target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +29,221 @@ use {
...
+                be_u8,
+            },
+        };
+
+        let input = input.as_ref();
+
+        // Verify Governance header is as expected so we can bail to avoid more parsing.
+        let (input, magic) = take(4usize)(input)?;
+        if magic != GOVERNANCE_MAGIC {
+            return Err(Error::InvalidGovernanceModule);
```
ali-bahjati, line 120: use ensure that you defined above?

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +29,221 @@ use {
...
-
-        Ok(match action {
-            ActionId::ContractUpgrade => {
-                let mut hash = [0u8; 32];
-                cursor.read_exact(&mut hash)?;
-                Self::ContractUpgrade(hash)
-            }
+        // Verify Target is this contract.
+        let (input, target) = be_u8(input)?;
+        if target != GOVERNANCE_MODULE {
```
jayantk, line 125: `GovernanceModule::Target as u8` ? (I think the constant GOVERNANCE_MODULE is redundant now)

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +29,221 @@ use {
...
+        let action = GovernanceActionId::from_u8(action).ok_or(Error::InvalidGovernanceAction)?;
+
+        match action {
+            GovernanceActionId::UpgradeContract => {
+                let (_input, bytes) = all_consuming(take(32usize))(input)?;
+                let mut codehash = [0u8; 32];
+                codehash.copy_from_slice(bytes);
 
-                Self::SetDataSources(sources)
+                // Make sure the buffer has been fully consumed.
```
jayantk, line 139: this comment applies to the all_consuming above?

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +29,221 @@ use {
...
+                    claim_vaa: claim_vaa.to_vec(),
                 })
             }
 
-            ActionId::SetStalePriceThreshold => {
-                let stale_price_threshold = cursor.read_u64::<BigEndian>()?;
-                Self::SetStalePriceThreshold(stale_price_threshold)
+            GovernanceActionId::SetDataSources => {
+                let (_input, data_sources) = many0(|input| {
+                    let (_, (chain, emitter)) = all_consuming(|input| {
```
jayantk, line 152: the nesting of the many_0 and all_consuming here confuses me. shouldn't it be the other way around?

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +29,221 @@ use {
...
+                        .unwrap(),
+                );
+                buf.extend_from_slice(claim_vaa);
+            }
+
+            GovernanceAction::SetDataSources { data_sources } => {
+                buf.push(GovernanceActionId::SetDataSources.to_u8().unwrap());
+                for source in data_sources {
+                    buf.extend_from_slice(
+                        &u16::from(source.pyth_emitter_chain.clone()).to_le_bytes(),
```
jayantk, line 218: i think everything is serialized in big endian

target-chains/near/receiver/src/lib.rs
```diff
@@ -162,11 +163,12 @@ impl Pyth {
         Ok(())
     }
 
-    /// Return the deposit required to update a price feed.
-    pub fn get_update_fee_estimate(&self) -> u64 {
+    /// Return the deposit required to update a price feed. This is the upper limit for an update
+    /// call and any remaining deposit not consumed for storage will be refunded.
+    pub fn get_update_fee_estimate(&self) -> U128 {
```
jayantk, line 168: why did this need to change to U128?

(I think you should use whatever number representation NEAR typically uses for token quantities.)

target-chains/near/receiver/src/governance.rs
Reisen: Yeah some of this parsing needed cleaning up before being reviewed, was fixing up tests. Should be correct now.

target-chains/near/receiver/src/governance.rs
Reisen: Yes though is actually incorrect either way around, fixed.

target-chains/near/receiver/src/governance.rs
```diff
@@ -312,18 +369,64 @@ impl Pyth {
...
+            Unknown
+        );
+
+        // Execute the embedded VAA action.
+        match instruction.action {
+            GovernanceAction::RequestGovernanceDataSourceTransfer {
+                governance_data_source_index,
+            } => {
+                ensure!(
+                    self.executed_governance_change_vaa < governance_data_source_index as u64,
```
jayantk, line 396: i think `executed_governance_change_vaa` needs to be updated by this method (to = governance_data_source_index) after this check

target-chains/near/receiver/src/governance.rs
```diff
@@ -380,6 +480,223 @@ impl Pyth {
...
+        let mut contract = Pyth::new(
+            near_sdk::AccountId::new_unchecked("pyth.near".to_owned()),
+            [0; 32],
+            Source::default(),
+            Source::default(),
+            0.into(),
+            32,
+        );
+
+        contract.set_update_fee(100, 0).expect("Failed to set fee");
```
jayantk, line 589: suggest making the exponent nonzero here (it's a little bit better of a test that way)

jayantk (approved):
lgtm aside from one minor bug

target-chains/near/receiver/src/governance.rs
```diff
@@ -265,26 +321,27 @@ impl Pyth {
...
-        match GovernanceAction::deserialize(rest)? {
-            ContractUpgrade(codehash) => self.set_upgrade_hash(codehash),
-            SetDataSources(sources) => self.set_sources(sources),
-            SetGovernanceSource(source) => self.set_gov_source(source),
-            SetStalePriceThreshold(threshold) => self.set_stale_price_threshold(threshold),
-            SetUpdateFee(fee) => self.set_update_fee(fee),
+            serde_wormhole::from_slice_with_payload(&vaa).map_err(|_| InvalidPayload)?;
+
+        match GovernanceInstruction::deserialize(rest)?.action {
+            UpgradeContract { codehash } => self.set_upgrade_hash(codehash),
```
tompntn, line 337: Since the payload of contract upgrade actions are chain-specific, it might be worth adding a check that the action is destined only for NEAR.

target-chains/near/receiver/src/governance.rs
Reisen: Good catch I'll add a couple of sequencing tests before merging.

target-chains/near/receiver/src/governance.rs
Reisen: Good spot, fun fact all these tests were auto-generated by Copilot! I thought it was an 8 when I checked them over.

target-chains/near/receiver/src/lib.rs
```diff
@@ -162,11 +166,12 @@ impl Pyth {
...
         let byte_cost = env::storage_byte_cost();
```
tompntn, line 172: In the other contracts the fee varies depending on how many price feeds are being updated in a single batch, and the fee estimate functions take a list of the VAAs that will be updated.

(See https://github.com/pyth-network/pyth-crosschain/blob/main/target-chains/aptos/contracts/sources/pyth.move#L375-L380).

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +33,218 @@ use {
...
-        data.extend_from_slice(&0u16.to_le_bytes());
-
-        match self {
-            Self::ContractUpgrade(hash) => {
-                data.extend_from_slice(hash);
+    /// Implements a `serialize` method for the `GovernanceAction` enum. The `nom` library doesn't
+    /// provide serialization but serialization is a safer operation, so we can just use a simple
+    /// push buffer to serialize.
+    pub fn serialize(&self) -> Result<Vec<u8>, Error> {
+        let mut buf = Vec::new();
```
tompntn, line 198: Dumb question, but when do we need to serialize governance actions?

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +33,218 @@ use {
...
+        // Verify Governance header is as expected so we can bail to avoid more parsing.
+        let (input, magic) = take(4usize)(input)?;
+        let (input, module) = be_u8(input)?;
+        let (input, action) = be_u8(input)?;
+        let (input, chain) = be_u16(input)?;
+        let module = GovernanceModule::from_u8(module).ok_or(InvalidGovernanceModule)?;
+        let chain = Chain::from(WormholeChain::from(chain));
+
+        // Safely parse the action ID. [ref:action_discriminants]
+        let action = GovernanceActionId::from_u8(action).ok_or(InvalidGovernanceAction)?;
```
tompntn, line 135: We should also check that the chain id NEAR or "all chains" (0).

target-chains/near/receiver/src/lib.rs
Reisen: Batch VAA's are difficult on NEAR because of the async nature, instead I thought better to allow the caller to simply chain their own promises of this method. Otherwise it needs to thread context through all the VAA verification and updates and calculate how to resolve deposit refunds at the end. I'm not sure It's worth it.

target-chains/near/receiver/src/governance.rs
Reisen: I needed a nice way to get these in tests, I can mark this `#[cfg(test)]` but I figured the utility might be worth leaving it available within the library.

target-chains/near/receiver/src/governance.rs
Reisen: You're right, this check is performed later on but that's a bad place for it, I will lift it up to here.

target-chains/near/receiver/src/lib.rs
ali-bahjati: it's funny how everyone patches u128 for serde_json. i wish they could natively solve it.

target-chains/near/receiver/src/lib.rs
```diff
@@ -162,11 +166,12 @@ impl Pyth {
...
         let data_cost = byte_cost * std::mem::size_of::<PriceFeed>() as u128;
-        4u64 * u64::try_from(data_cost).unwrap() + self.update_fee
+        (4u128 * data_cost + self.update_fee).into()
```
ali-bahjati, line 174: Can you document why is it 4 times the data cost? (why not 5? as batches have 5 price feeds)

target-chains/near/receiver/src/lib.rs
ali-bahjati: that's right but still, we might want to have charge a variable rate fee per vaa. What Tom suggests is that this method should take the update data as input (although we don't use it here)

target-chains/near/receiver/workspace-test.sh
```diff
@@ -0,0 +1,20 @@
...
+cargo build --release --target wasm32-unknown-unknown
+cp target/wasm32-unknown-unknown/release/pyth.wasm .
+
+(
+    cd ../wormhole-stub
+    cargo build --release --target wasm32-unknown-unknown
+    cp target/wasm32-unknown-unknown/release/wormhole_stub.wasm ../receiver
+)
+
+RUST_LOG=info cargo nextest run
```
ali-bahjati, line 20: Can you document that nextest is required?

target-chains/near/receiver/src/governance.rs
ali-bahjati: can you also add some tests to ensure it gets rejected in that case.

target-chains/near/receiver/workspace-test.sh
ali-bahjati: Also it would be awesome if you add this to CI

target-chains/near/receiver/src/governance.rs
```diff
@@ -34,162 +33,218 @@ use {
...
-            ActionId::ContractUpgrade => {
-                let mut hash = [0u8; 32];
-                cursor.read_exact(&mut hash)?;
-                Self::ContractUpgrade(hash)
-            }
+impl GovernanceInstruction {
+    /// Implements a `deserialize` method for the `GovernanceAction` enum using `nom` to
+    /// deserialize the payload. The use of `nom` gives us parser safety, error handling, full
+    /// buffer consumption, and a more readable implementation while staying efficient.
+    pub fn deserialize(input: impl AsRef<[u8]>) -> Result<Self, Error> {
```
ali-bahjati, line 111: I love it!

target-chains/near/receiver/tests/workspaces.rs
```diff
@@ -91,14 +98,21 @@ async fn test_source_add() {
         let mut cur = Cursor::new(Vec::new());
         serde_wormhole::to_writer(&mut cur, &vaa).expect("Failed to serialize VAA");
         cur.write_all(
-            &GovernanceAction::SetDataSources(vec![
-                Source::default(),
-                Source {
-                    emitter:            [1; 32],
-                    pyth_emitter_chain: pyth::state::Chain(1),
+            &GovernanceInstruction {
```
ali-bahjati, line 101: maybe rename this test name from `test_source_add` to `test_set_sources`?

target-chains/near/receiver/tests/workspaces.rs
```diff
@@ -511,24 +580,27 @@ async fn test_contract_fees() {
         .outcome()
         .is_success());
```
ali-bahjati, line 581: Not for this part, rather line 645. Do all transactions in near succeed regardless of their failure in the end? because in some tests I see you commented it should fail and it is still success, then you check the state to figure out.

if so its very weird...

ali-bahjati (approved):

target-chains/near/receiver/workspace-test.sh
ali-bahjati: Also I see the wasm files are not gitignored. it would be great to have them in gitignore too.

target-chains/near/receiver/src/governance.rs
tompntn: Side point - for paranoia, it's probably good to have the test VAAs be generated completely orthogonally to the nom logic, so that if there is a bug in serialize/deserialize it'll get caught. You can use the [`xc-governance-sdk-js`](https://github.com/pyth-network/pyth-crosschain/tree/main/third_party/pyth/xc-governance-sdk-js) package to generate these test VAA's outside of the NEAR code.

PR #457, opened 2023-01-05
Scaffolding for xc-admin (new crosschain admin tool)

guibescos:
Some scaffolding. I see the project as a lerna folder. With:
- a xc-admin-common package that the other packages will share
- a website package
- a cli package
- packages for relayers


cctdaniel (approved):
🚢

PR #461, opened 2023-01-06
Xc admin/add wormhole message header deser

guibescos:
- Port multisig tools to xc-admin common
- Add parser for pyth cross-chain message headers and for the ExecutePostedVaa instruction
- Add units tests for said parser


xc-admin/packages/xc-admin-common/src/multisig.ts
```diff
@@ -0,0 +1,80 @@
...
+  DEFAULT_MULTISIG_PROGRAM_ID,
+  getIxPDA,
+  getTxPDA,
+} from "@sqds/mesh";
+import { InstructionAccount, TransactionAccount } from "@sqds/mesh/lib/types";
+import BN from "bn.js";
+import lodash from "lodash";
+
+export async function getActiveProposals(
+  squad: Squads,
```
jayantk, line 12: minor comment: it may simplify things to put these functions on a class that holds common arguments. (e.g., the class can have a Squads member so you don't need to keep passing it in.)

jayantk: (you can also put config parameters in the class like the multisig program id so the code is more flexible)

xc-admin/packages/xc-admin-common/src/multisig.ts
```diff
@@ -0,0 +1,80 @@
...
+    .filter((x) => lodash.isEqual(x.status, { active: {} }));
+}
+
+export async function getManyProposalsInstructions(
+  squad: Squads,
+  txAccounts: TransactionAccount[]
+): Promise<InstructionAccount[][]> {
+  let allIxsKeys = [];
+  let ownerTransaction = [];
+  for (let [index, txAccount] of txAccounts.entries()) {
```
jayantk, line 34: seems simpler to reuse `getProposalInstructions` here?

i think this whole function is basically `txAccounts.map(a => getProposalInstructions(squads, a)`?

xc-admin/packages/xc-admin-common/src/multisig.ts
```diff
@@ -0,0 +1,80 @@
...
+import Squads, {
+  DEFAULT_MULTISIG_PROGRAM_ID,
+  getIxPDA,
+  getTxPDA,
+} from "@sqds/mesh";
+import { InstructionAccount, TransactionAccount } from "@sqds/mesh/lib/types";
+import BN from "bn.js";
+import lodash from "lodash";
+
+export async function getActiveProposals(
```
jayantk, line 11: please add doc comments for all of these functions

xc-admin/packages/xc-admin-common/src/governance_payload/index.ts
```diff
@@ -0,0 +1,105 @@
...
+export declare const Action: {
+  readonly ExecutePostedVaa: 0;
+  readonly UpgradeContract: 0;
+  readonly AuthorizeGovernanceDataSourceTransfer: 1;
+  readonly SetDataSources: 2;
+  readonly SetFee: 3;
+  readonly SetValidPeriod: 4;
+  readonly RequestGovernanceDataSourceTransfer: 5;
+};
+
```
jayantk, line 13: FYI @Reisen is looking into whether we can use something like protobuf to define the binary format in one place and code gen all of this deserialization stuff.

xc-admin/packages/xc-admin-common/src/governance_payload/index.ts
```diff
@@ -0,0 +1,105 @@
+import { ChainId, ChainName, toChainName } from "@certusone/wormhole-sdk";
+import * as BufferLayout from "@solana/buffer-layout";
+
+export declare const Action: {
+  readonly ExecutePostedVaa: 0;
+  readonly UpgradeContract: 0;
```
jayantk, line 6: it's surprising that there are two 0s here. (I understand why it's the case, but it will confuse the next person who reads this.) Either document why there are two 0s and why its ok, or consider making different versions of Action for each module, or define this in a way where each Action corresponds to a tuple of the moduleId and actionId

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -0,0 +1,112 @@
...
+    super(-1, property);
+    this.element = element;
+  }
+
+  decode(b: Uint8Array, offset?: number | undefined): T[] {
+    const length = BufferLayout.u32().decode(b, offset);
+    return BufferLayout.seq(this.element, length).decode(b, (offset || 0) + 4);
+  }
+  encode(src: T[], b: Uint8Array, offset?: number | undefined): number {
+    return BufferLayout.struct<Readonly<{ lenght: number; src: T[] }>>([
```
jayantk, line 29: typo: length (here and below)

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+export type InstructionData = {
+  programId: Uint8Array;
+  accounts: AccountMetaData[];
+  data: number[];
+};
+
+export type AccountMetaData = {
```
jayantk, line 47: Metadata (no caps on Data). it's one word.

jayantk (approved):

xc-admin/packages/xc-admin-common/src/multisig.ts
guibescos: future PR

xc-admin/packages/xc-admin-common/src/multisig.ts
guibescos: Nah to goal of `getManyProposalsInstructions` is requesting all the instruction accounts for all proposals in one single RPC call.
`txAccounts.map(a => getProposalInstructions(squads, a)` would make 1 RPC call per proposal.

xc-admin/packages/xc-admin-common/src/governance_payload/index.ts
guibescos: perfect

PR #462, opened 2023-01-06
[xc-admin] Add encoder for governance messages

guibescos:
Add encoder for governance header and ExecutePostedVaa

### DIFF
--- a/xc-admin/packages/xc-admin-common/src/__tests__/GovernancePayload.test.ts
+++ b/xc-admin/packages/xc-admin-common/src/__tests__/GovernancePayload.test.ts
@@ -84,14 +84,11 @@ test("GovernancePayload ser/de", (done) => {
     },
     instructions: [] as TransactionInstruction[],
   };
-  buffer = Buffer.alloc(1000);
-  span = encodeExecutePostedVaa(expectedExecuteVaaArgs, buffer);
+  buffer = encodeExecutePostedVaa(expectedExecuteVaaArgs);
   expect(
-    buffer
-      .subarray(0, span)
-      .equals(Buffer.from([80, 84, 71, 77, 0, 0, 0, 26, 0, 0, 0, 0]))
+    buffer.equals(Buffer.from([80, 84, 71, 77, 0, 0, 0, 26, 0, 0, 0, 0]))
   ).toBeTruthy();
-  let executePostedVaaArgs = decodeExecutePostedVaa(buffer.subarray(0, span));
+  let executePostedVaaArgs = decodeExecutePostedVaa(buffer);
   expect(executePostedVaaArgs?.header.targetChainId).toBe("pythnet");
   expect(executePostedVaaArgs?.header.action).toBe("ExecutePostedVaa");
   expect(executePostedVaaArgs?.instructions.length).toBe(0);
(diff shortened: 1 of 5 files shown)


xc-admin/packages/xc-admin-common/src/__tests__/GovernancePayload.test.ts
```diff
@@ -1,7 +1,8 @@
 import { PublicKey, SystemProgram } from "@solana/web3.js";
-import { decodeExecutePostedVaa, decodeHeader } from "..";
+import { decodeExecutePostedVaa, decodeHeader, encodeHeader } from "..";
+import { encodeExecutePostedVaa } from "../governance_payload/ExecutePostedVaa";
 
-test("GovernancePayload", (done) => {
+test("GovernancePayload ser/de", (done) => {
```
ali-bahjati, line 5: I don't quite like this test as changing a payload/structure requires spending time on fixing it here.

specially that you go from "decode" to "encode" again. it means on change either 
(1) someone should hand roll the struct manually (2) that someone should write down another structure and calls encode to get the new value.

I think it's better to encode and decode a struct. That's easier to modify/update.

ali-bahjati:
Can you add the tests to CI?

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -105,3 +110,31 @@ export function decodeExecutePostedVaa(
 
   return { header, instructions };
 }
+
+/** Encode ExecutePostedVaaArgs */
+export function encodeExecutePostedVaa(
```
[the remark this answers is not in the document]
guibescos: Some explanation of why I did this:
- The underlying buffer-layout library uses this signature for encoding (`encode<T>(struct : T, buffer: Buffer) : number`
I think the only place where I use this offset is for the header and the body, so I don't think refactoring into a Writer is worth it right now.

xc-admin/packages/xc-admin-common/src/__tests__/GovernancePayload.test.ts
guibescos: I refactored to to `struct -> encode -> bytes -> decode -> struct`. Should I delete the hardcoded bytes arrays and make it `struct -> encode -> decode -> struct`

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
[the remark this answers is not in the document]
guibescos: I updated `encodeExecutePostedVaa` to abstract away the buffer initialization and offset.

guibescos:
> Can you add the tests to CI?

Done

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -103,5 +109,36 @@ export function decodeExecutePostedVaa(
...
   );
 
-  return { header, instructions };
+  return { targetChainId: header.targetChainId, instructions };
+}
+
+/** Encode ExecutePostedVaaArgs */
+export function encodeExecutePostedVaa(src: ExecutePostedVaaArgs): Buffer {
+  // PACKET_DATA_SIZE is the maximum transactin size of Solana, so our serialized payload will never be bigger than that
+  const buffer = Buffer.alloc(PACKET_DATA_SIZE);
```
jayantk, line 118: this code comment seems wrong because the code doesn't enforce this size constraint. I could theoretically make `src` have a bunch of `instructions` in it such that the serialized version of the message is bigger than PACKET_DATA_SIZE.

jayantk (approved):

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
guibescos: This part is a little awkward, I don't know the length before I encode it.
I'll try to improve the comment.

PR #463, opened 2023-01-06
Xc admin/propose instructions

guibescos:
Scaffolding for a function that proposes an array of instructions as local chain instructions or remote instructions.
I'm hoping to test it once the other pieces are implemented.

The main challenge for this part of the code is batching instructions into transactions. We don't want a ledger user's to have to sign too many transactions. Simultaneously, all instructions might not fit in one transaction.
The current (naive) solution is having 1 instruction per transaction, having the ledger sign all transactions at once and then sending all transactions. This might need to change but don't want to focus on this until we can test it in the UI.

### DIFF
--- a/xc-admin/packages/xc-admin-common/src/propose.ts
+++ b/xc-admin/packages/xc-admin-common/src/propose.ts
@@ -1,6 +1,5 @@
 import Squads, { getIxAuthorityPDA, getTxPDA } from "@sqds/mesh";
 import {
-  PACKET_DATA_SIZE,
   PublicKey,
   Transaction,
   TransactionInstruction,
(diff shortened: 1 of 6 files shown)


xc-admin/packages/xc-admin-common/src/propose.ts
```diff
@@ -0,0 +1,176 @@
...
+ * @param instructions instructions that will be proposed
+ * @param remote whether the instructions should be executed in the chain of the multisig or remotely on Pythnet
+ * @returns the newly created proposal's pubkey
+ */
+export async function proposeInstructions(
+  squad: Squads,
+  vault: PublicKey,
+  instructions: TransactionInstruction[],
+  remote: boolean,
+  wormholeAddress?: PublicKey
```
jayantk, line 35: make a class like `MultisigWrapper` to encapsulate some of the common arguments (squads, wormholeAddress, etc)

jayantk (approved):
I think you may need to refactor this code a bit once you get more stuff written, but I'm going to approve this for now. (I think you'll probably want to develop a systematic organization for the different types of instructions, which will include some representation for remote vs. local instructions. anyway that's something you can do later.)

PR #465, opened 2023-01-09
Injective deployment code

0xfirefist:
The PR add to the work done in this https://github.com/pyth-network/pyth-crosschain/pull/455 by @jayantk 
- All javascript to typescript 
- Injective deployment code
- Refactored 

Q. do we still need deploy.ts? what purpose does it serve?


target-chains/cosmwasm/tools/src/deploy-pyth-bridge.ts
```diff
@@ -0,0 +1,133 @@
...
+  contract,
+} = argv;
+const pythConfig = PythConfig[network];
+const deployer: Deployer = DeployerFactory.create(network, mnemonic);
+
+// checks
+if (inputCodeId === undefined && artifact === undefined)
+  RaiseCLError("Please provide either artifact or code id");
+
+async function deployArtifacts(artifact: string): Promise<number> {
```
jayantk, line 67: these three helper functions seem unnecessary it's more code to have them than to call the function on `deployer` directly in the `run` function below

target-chains/cosmwasm/tools/src/deploy-pyth-bridge.ts
```diff
@@ -0,0 +1,133 @@
...
+  contract: string
+): Promise<void> {
+  await deployer.migrate(contract, codeId);
+}
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function RaiseCLError(message: string) {
```
jayantk, line 95: why is this function name capitalized?

target-chains/cosmwasm/tools/src/deployer/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  instantiate(
+    codeId: number,
+    inst_msg: string | object,
+    label: string
+  ): Promise<string>;
+  migrate(contract: string, codeId: number): Promise<void>;
+}
+
+export class DeployerFactory {
+  static create(network: NETWORKS, mnemonic: string): Deployer {
```
jayantk, line 17: this function would be more flexible if it took the `NetworkConfig` as an argument. That would allow callers to create a custom config instead of forcing them to use a config from `CONFIG`

target-chains/cosmwasm/tools/src/deployer/injective.ts
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    const txResponse = await this.signAndBroadcastMsg(store_code, {
+      amount: [
+        {
+          amount: String(500000000 * 5000000),
+          denom: "inj",
+        },
+      ],
+      gas: "5000000",
```
jayantk, line 87: where did these numbers come from? I suggest leaving a code comment explaining how you determined these and how someone could update them if needed

target-chains/cosmwasm/tools/src/deployer/injective.ts
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    var codeId: number;
+    try {
+      const ci = /"code_id","value":"\\"([^\\"]+)/gm.exec(txResponse.rawLog)![1];
```
jayantk, line 92: maybe extract a utility function to pull out fields from the txResponse log? I think you could do it generically for the code_id here and the contract_address below, plus some similar logic occurs in other places in the PR.

target-chains/cosmwasm/tools/src/deployer/terra.ts
```diff
@@ -0,0 +1,145 @@
...
+    console.log(`Storing WASM: ${artifact} (${contract_bytes.length} bytes)`);
+
+    const store_code = new MsgStoreCode(
+      this.wallet.key.accAddress,
+      contract_bytes.toString("base64")
+    );
+
+    const rs = await this.signAndBroadcastMsg(store_code)
+
+    var codeId: number;
```
jayantk, line 59: another place where the utility function would be useful

target-chains/cosmwasm/tools/src/pyth_config.ts
```diff
@@ -0,0 +1,132 @@
...
+    ],
+    governance_source: {
+      emitter: Buffer.from(
+        "5635979a221c34931e32620b9293a463065555ea71fe97cd6237ade875b12e9e",
+        "hex"
+      ).toString("base64"),
+      chain_id: 1,
+    },
+    governance_source_index: 0,
+    governance_sequence_number: 0,
```
jayantk, line 54: @ali-bahjati question for you: what are we supposed to set as the default values for these two fields? I set them as zero before but we've probably generated a couple governance messages already, right?

target-chains/cosmwasm/tools/src/pyth_config.ts
```diff
@@ -0,0 +1,132 @@
...
+    chain_id: 19,
+    valid_time_period_secs: 60,
+    fee: {
+      amount: "1",
+      // FIXME ??
```
jayantk, line 128: delete this comment? (is the fee denom "inj"?)

jayantk (approved):
generally this looks good to me. I left a couple minor suggestions for how to tidy up the code a little more, but overall nicely done. I'm going to approve this, but please do implement the code cleanup suggestions before merging.

target-chains/cosmwasm/tools/src/pyth_config.ts
ali-bahjati: Yeah but I think using 0 as the beginning is fine. There is no harm executing older ones.

ali-bahjati (approved):
Looks good to me, but please also fix Tilt CI before merging. `deploy.ts` is used for tilt which is our local development enviornment. 

You need to update the `deploy.sh` file in the tools directory that tilt runs in its service (look at `tilt-devnet/k8s/terra-devnet.yaml`)

target-chains/cosmwasm/tools/src/deployer/index.ts
0xfirefist: yeah makes sense

target-chains/cosmwasm/tools/src/pyth_config.ts
0xfirefist: fee denom is inj.
1 inj = 1×10⁻¹⁸ INJ

Do we want to keep the amount 1?

ali-bahjati: Currently we had set the fee to the smallest non-negative value to make sure people integrate with us correctly. In the future we will update the fee everywhere to a proper value.

PR #466, opened 2023-01-09
Xc admin/add executor

guibescos:
Scaffolding for the executor, the crank that runs all the executeReady proposals. 
This is code is untested and inspired by pyth-writer.


xc-admin/packages/crank-executor/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        );
+
+        try {
+          await new AnchorProvider(squad.connection, squad.wallet, {
+            commitment: SOLANA_CONNECTION_COMMITMENT,
+            preflightCommitment: SOLANA_CONNECTION_COMMITMENT,
+          }).sendAndConfirm(transaction, [KEYPAIR]);
+        } catch (error) {
+          // Mark the transaction as cancelled if we failed to run it
+          await squad.cancelTransaction(proposal.publicKey);
```
ali-bahjati, line 67: I think you need to handle the errors better here. For instance you should handle balance errors explicitly. There could be RPC errors and for that it's good to retry a couple of times before giving up on the transaction.

If checking these stuff is difficult I would suggest exponential backoff without cancelling. People will figure it out on the ui and will proceed to cancellation if necessary.

and... hmm... one question, why not letting people execute it on the ui? without us cranking this part? the last signer can also trigger the execute.

xc-admin/packages/xc-admin-common/src/multisig.ts
```diff
@@ -9,16 +9,18 @@ import BN from "bn.js";
...
+ * @param state filter by status
  * @returns All the proposal accounts as `TransactionAccount`
  */
-export async function getActiveProposals(
+export async function getProposals(
   squad: Squads,
   vault: PublicKey,
-  offset: number = 1
+  offset: number = 1,
+  state: "active" | "executeReady" | "executed" | "all" = "all"
```
ali-bahjati, line 23: nice!

ali-bahjati (approved):
I'm approving it. but let's make sure we handle errors properly (maybe on a later PR)

PR #467, opened 2023-01-10
[xc-admin] Start cli and add accept transfer for upgrade authorities

guibescos:
This package should replace the current CLI.

### DIFF
--- a/xc-admin/packages/xc-admin-cli/src/index.ts
+++ b/xc-admin/packages/xc-admin-cli/src/index.ts
@@ -14,16 +14,24 @@ const PROGRAM_AUTHORITY_ESCROW = new PublicKey(
 const BPF_UPGRADABLE_LOADER = new PublicKey(
   "BPFLoaderUpgradeab1e11111111111111111111111"
 );
+
+const mutlisigCommand = (name: string, description: string) =>
+  program
+    .command(name)
+    .description(description)
+    .requiredOption("-c, --cluster <network>", "solana cluster to use")
+    .requiredOption("-w, --wallet <filepath>", "path to the operations key")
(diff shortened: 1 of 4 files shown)


xc-admin/packages/xc-admin-cli/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+);
+program
+  .name("xc-admin-cli")
+  .description("CLI for interacting with Pyth's xc-admin")
+  .version("0.1.0");
+
+program
+  .command("accept-authority")
+  .description("Accept authority from the program authority escrow")
+  .requiredOption("-c, --cluster <network>", "solana cluster to use")
```
ali-bahjati, line 25: This is applicable to our previous code too, sorry that i was lazy: can't we make cluster and wallet and vault a global option and avoid defining it per command?

xc-admin/packages/xc-admin-cli/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  .requiredOption("-w, --wallet <filepath>", "path to the operations key")
+  .requiredOption(
+    "-p, --program-id <pubkey>",
+    "program whose authority we want to transfer"
+  )
+  .requiredOption(
+    "-a, --current <pubkey>",
+    "current authority (before the transfer)"
+  )
+  .requiredOption("-v, --vault <pubkey>", "multisig address")
```
ali-bahjati, line 35: Not necessary as we are not gonna use it manually but i think hard-coding this value is easier (for testing, ...)

ali-bahjati (approved):
Nice! left some minor comments.

PR #469, opened 2023-01-10
Osmosis deployment code

0xfirefist:


target-chains/cosmwasm/tools/src/pyth_config.ts
```diff
@@ -128,4 +128,41 @@ export const CONFIG: Config = {
       denom: "inj",
     },
   },
+  // this details below are invalid and random
```
ali-bahjati, line 131: Can't we abstract common configurations such as data and governance sources to a common place for testnet/mainnet and use them without repeating the logic?

ali-bahjati (approved):
Thanks! Let's go 🚢

jayantk (approved):
🔥

target-chains/cosmwasm/tools/src/pyth_config.ts
0xfirefist: Took the common configurations out.

PR #471, opened 2023-01-11
Xc admin/change error handling

guibescos:
export function `decodeHeader(data: Buffer)` was returning `GovernanceHeader | undefined`.
This made things complicated because when you called this function there were 3 outcomes :
- a) throws error
- b) returns undefined
- c) returns something

So I decided to switch to throwing errors instead of undefined.


ali-bahjati (approved):
nice!

PR #472, opened 2023-01-11
[price-service] Document the spy filters in .env.sample

jayantk:


ali-bahjati (approved):
I think it is not necessary but it's fine. We already explained them in README too and exact values are in the docker files.

PR #473, opened 2023-01-11
[cosmwasm] Deploy tool: Add new commands (update contract admin + get contract info)

0xfirefist:
This PR includes
- add config - for localterra
- add feature - update contract admin 
- add feature - get contract info 
- refactor the yargs command structure
- update the README

### DIFF
--- a/target-chains/cosmwasm/README.md
+++ b/target-chains/cosmwasm/README.md
@@ -12,7 +12,7 @@ Deploying the CosmWasm contract has three steps:
    2. Migrating an existing contract code id to the new code id.
 3. Update contract's admin to itself.
 
-This directory contians the code to perform all the steps. Read below for the details.
+This directory contains the code to perform all the steps. Read below for the details.
 
 ### Uploading the code
 
--- a/target-chains/cosmwasm/README.md
+++ b/target-chains/cosmwasm/README.md
@@ -4,18 +4,19 @@ This directory contains the Pyth contract for CosmWasm and utilities to deploy i
 
 ## Deployment
 
-Deploying the CosmWasm contract has two steps:
+Deploying the CosmWasm contract has three steps:
 
 1. Upload the code. This step will give you a code id.
 2. Either create a new contract or migrate an existing one:
    1. Create a new contract that has an address with a code id as its program.
    2. Migrating an existing contract code id to the new code id.
+3. Update contract's admin to itself.
 
-This directory includes a script to perform both steps. Read below for the details.
+This directory contians the code to perform all the steps. Read below for the details.
 
 ### Uploading the code
 
-First, build the contracts within [the current directory](./):
+First, build the contracts within [the current directory](./). You must have Docker installed.
 
 ```sh
 bash build.sh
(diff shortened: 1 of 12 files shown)


target-chains/cosmwasm/tools/src/deploy-pyth-bridge.ts
```diff
@@ -1,115 +1,181 @@
...
-    );
-  }
+async function updateAdminHandler(argv: any) {
+  const { network, mnemonic, newAdmin, contract } = argv;
+  console.log("Updating contract's admin");
+  await getDeployer(network as NETWORKS, mnemonic as string).updateAdmin(
+    newAdmin,
+    contract
+  );
+  console.log("Contract's admin successfully updates");
```
ali-bahjati, line 168: Is this for removing the admin? can't we have it setAdminToZero? or something like this?

ali-bahjati (approved):
It is really nice. My only comment is maybe only setting admin to zero?

target-chains/cosmwasm/README.md
```diff
@@ -4,18 +4,19 @@ This directory contains the Pyth contract for CosmWasm and utilities to deploy i
...
+Deploying the CosmWasm contract has three steps:
 
 1. Upload the code. This step will give you a code id.
 2. Either create a new contract or migrate an existing one:
    1. Create a new contract that has an address with a code id as its program.
    2. Migrating an existing contract code id to the new code id.
+3. Update contract's admin to itself.
 
-This directory includes a script to perform both steps. Read below for the details.
+This directory contians the code to perform all the steps. Read below for the details.
```
jayantk, line 15: typo contains

target-chains/cosmwasm/tools/src/deployer/injective.ts
```diff
@@ -169,6 +170,38 @@ export class InjectiveDeployer implements Deployer {
...
+    const currAdmin = this.injectiveAddress();
+
+    const updateAdminMsg = new MsgUpdateAdmin({
+      sender: currAdmin,
+      newAdmin,
+      contract,
+    });
+
+    await this.signAndBroadcastMsg(updateAdminMsg);
+    console.log(await this.getContractInfo(contract));
```
jayantk, line 183: nitpick: it would be better to do this logging in the code that calls `updateAdmin` instead of here. In the future, someone else may call this function, and they may not want this log message.

target-chains/cosmwasm/tools/src/deployer/injective.ts
```diff
@@ -169,6 +170,38 @@ export class InjectiveDeployer implements Deployer {
     }
   }
 
+  async updateAdmin(newAdmin: string, contract: string): Promise<void> {
```
jayantk, line 173: I suggest documenting that this function assumes that this deployer is the current admin for the contract, and this will fail if this condition isn't satisfied. (also, does it fail with a reasonable error message?)

target-chains/cosmwasm/tools/src/deployer/injective.ts
jayantk: (i realize you're working off the code that i wrote, which is also probably doing the logging in the wrong place, so that's my fault.)

target-chains/cosmwasm/tools/src/deploy-pyth-bridge.ts
jayantk: on cosmos, you set the contract's admin address to itself, which allows the contract to send itself the migrate message.

jayantk (approved):
nicely done

target-chains/cosmwasm/tools/src/deployer/injective.ts
0xfirefist: it doesn't fail with a reasonable error message. but that is the case for most of the places.
i am gonna raise another pr for error handling.

PR #474, opened 2023-01-11
[price-service] Move to top

ali-bahjati:
This PR moves the price service from `third_party/pyth/price_service` to `price-service` as this is an important component for us. The first commit is just the move and the second commit is the changes required to make everything work.


jayantk (approved):
🙇

PR #477, opened 2023-01-11
[xc admin] Pyth parser

guibescos:
Add PythMultisigInstruction and tests.

### DIFF
--- a/xc-admin/packages/xc-admin-common/src/multisig_transaction/PythMultisigInstruction.ts
+++ b/xc-admin/packages/xc-admin-common/src/multisig_transaction/PythMultisigInstruction.ts
@@ -1,0 +1,44 @@
+import { MultisigInstruction, MultisigInstructionProgram } from ".";
+import { AnchorAccounts, resolveAccountNames } from "./anchor";
+import { pythIdl, pythOracleCoder } from "@pythnetwork/client";
+import { TransactionInstruction } from "@solana/web3.js";
+import { Idl } from "@coral-xyz/anchor";
+
+export class PythMultisigInstruction implements MultisigInstruction {
+  readonly program = MultisigInstructionProgram.PythOracle;
+  readonly name: string;
+  readonly args: { [key: string]: any };
+  readonly accounts: AnchorAccounts;
+
+  constructor(
+    name: string,
+    args: { [key: string]: any },
+    accounts: AnchorAccounts
+  ) {
+    this.name = name;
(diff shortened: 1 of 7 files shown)


xc-admin/packages/xc-admin-common/src/multisig_transaction/PythMultisigInstruction.ts
```diff
@@ -0,0 +1,44 @@
...
+        deserializedData.name,
+        deserializedData.data,
+        resolveAccountNames(pythIdl as Idl, deserializedData.name, instruction)
+      );
+    } else {
+      return new PythMultisigInstruction(
+        "Unrecognized instruction",
+        {},
+        { named: {}, remaining: instruction.keys }
+      );
```
jayantk, line 41: you should think carefully about how you want to handle this case, as the wrong behavior for unknown instructions could cause you a lot of agony downstream.

is the right behavior to return something like `undefined`? Or should there be a specific instruction value that represents something unknown? Or should you have a default wrapper around TransactionInstruction that represents an unknown instruction (but retains all data in that instruction)?

Couple questions to ask yourself:
1. What will the users of PythMultisigInstruction want to do with an unknown instruction?
2. Do you expect MultisigInstruction to be convertible back into TransactionInstruction?

jayantk (approved):
lgtm, though see inline comment and think about it

xc-admin/packages/xc-admin-common/src/multisig_transaction/PythMultisigInstruction.ts
```diff
@@ -0,0 +1,44 @@
+import { MultisigInstruction, MultisigInstructionProgram } from ".";
+import { AnchorAccounts, resolveAccountNames } from "./anchor";
+import { pythIdl, pythOracleCoder } from "@pythnetwork/client";
+import { TransactionInstruction } from "@solana/web3.js";
+import { Idl } from "@coral-xyz/anchor";
+
+export class PythMultisigInstruction implements MultisigInstruction {
```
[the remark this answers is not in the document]
guibescos: Future PR

xc-admin/packages/xc-admin-common/src/multisig_transaction/PythMultisigInstruction.ts
guibescos: Future PR

PR #478, opened 2023-01-11
[xc-admin] Refactor wormhole messages as classes

guibescos:
Because it enables using instanceof like with `MultisigInstruction`
Each type of governanceAction will have it's own class with :
`encode() : Buffer`
`decode(data : Buffer) : Self | undefined`


xc-admin/packages/xc-admin-common/src/multisig_transaction/WormholeMultisigInstruction.ts
```diff
@@ -47,12 +50,12 @@ export class WormholeMultisigInstruction implements MultisigInstruction {
 
       if (result.name === "postMessage") {
         try {
-          const decoded = decodeGovernancePayload(result.args.payload);
-          result.args.governanceName = decoded.name;
-          result.args.governanceArgs = decoded.args;
+          const decoded: PythGovernanceAction = decodeGovernancePayload(
```
guibescos, line 53: I'm gonna try to switch to a paradigm where it doesn't throw errors in a future PR

jayantk (approved):

PR #482, opened 2023-01-12
[price-service] add /get_vaa endpoint that serves cache and db vaas

cctdaniel:
generally:
- store the VAAs in an object with the price feed id as the key and the value is an array of objects that has a `publishTime` and `vaa` (VAAs are stored in an ascending order based on `publishTime`)
- because VAAs are stored in an ascending order based on `publishTime`, we can run binary search on the array to find the publishTime greater than or equal to the target on O(logn) time
- the endpoint first hits the cache and tries to find the VAA, if it doesn't exists it hits the DB
- every minute, the cache will be cleaned up, and `publishTime` that has expired will be removed from the cache

other considerations:
- balanced bst (not trivial + since the array is already sorted we can just use binary search)


third_party/pyth/price-service/src/rest.ts
```diff
@@ -113,6 +120,55 @@ export class RestAPI {
...
+      validate(getVaaInputSchema),
+      (req: Request, res: Response) => {
+        const priceFeedId = req.query.id as string;
+        const publishTime = Number(req.query.publish_time as string);
+        const vaa = this.priceFeedVaaInfo.getVaa(priceFeedId, publishTime);
+        // if publishTime is older than cache ttl or vaa is not found, fetch from db
+        if (
+          publishTime <
+            Math.floor(Date.now() / 1000) -
+              PYTH_WRITER_INSERT_LATENCY_SECONDS ||
```
ali-bahjati, line 143: I don't think injecting pyth writer logic be good here. The check seems unnecessary to me.

third_party/pyth/price-service/src/index.ts
```diff
@@ -61,6 +61,7 @@ async function run() {
   const wsAPI = new WebSocketAPI(listener, promClient);
 
   listener.run();
+  listener.removeExpiredValuesFromVaasCache();
```
ali-bahjati, line 64: The name suggests its a one time action, maybe `runCacheCleanupLoop` or something similar?

third_party/pyth/price-service/src/__tests__/rest.test.ts
```diff
@@ -146,3 +156,17 @@ describe("Latest Vaa Bytes Endpoint", () => {
     expect(resp.body.message).toContain(ids[2]);
   });
 });
+
+// describe("Get Vaa Endpoint", () => {
```
ali-bahjati, line 160: Are you going to add a test later? It's better to have it fully tested on this PR, but if not, i suggest removing this comment block and add it later when you add the tests.

third_party/pyth/price-service/src/rest.ts
```diff
@@ -113,6 +120,55 @@ export class RestAPI {
...
+          // cache miss
+          fetch(
+            `https://web-api.pyth.network/vaa?id=${priceFeedId}&publishTime=${publishTime}&cluster=pythnet`
+          )
+            .then((r: any) => r.json())
+            .then((arr: any) => {
+              if (arr.length > 0 && arr[0]) {
+                res.json(arr[0]);
+              } else {
+                res.status(404).send("VAA not found");
```
ali-bahjati, line 155: Please use `StatusCodes`

third_party/pyth/price-service/src/rest.ts
```diff
@@ -113,6 +120,55 @@ export class RestAPI {
...
+        // if publishTime is older than cache ttl or vaa is not found, fetch from db
+        if (
+          publishTime <
+            Math.floor(Date.now() / 1000) -
+              PYTH_WRITER_INSERT_LATENCY_SECONDS ||
+          !vaa
+        ) {
+          // cache miss
+          fetch(
+            `https://web-api.pyth.network/vaa?id=${priceFeedId}&publishTime=${publishTime}&cluster=pythnet`
```
ali-bahjati, line 148: I would make the endpoint an Env var and if it's not given I won't query it. Because I'm thinking of not exposing it publicly.

ali-bahjati: Also, the cluster is ambiguous here. We should support devnet too. I suggest `testnet` and `mainnet` as that is how we call our price service clusters (and price service works regardless of pythnet/solana)

third_party/pyth/price-service/docker-compose.mainnet.yaml
```diff
@@ -37,6 +37,10 @@ services:
       READINESS_SPY_SYNC_TIME_SECONDS: "20"
       READINESS_NUM_LOADED_SYMBOLS: "50"
       LOG_LEVEL: warning
+      CLUSTER: pythnet
+      REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS: "60"
+      PYTH_WRITER_INSERT_LATENCY_SECONDS: "30"
+      CACHE_TTL_SECONDS: "300"
```
ali-bahjati, line 43: You will also need to add these to `tilt-devnet/k8s/...`.

third_party/pyth/price-service/src/listen.ts
```diff
@@ -105,6 +169,17 @@ export class Listener implements PriceStore {
     logger.info("loaded " + this.filters.length + " filters");
   }
 
+  async removeExpiredValuesFromVaasCache() {
+    const REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS = Number(
+      envOrErr("REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS")
+    );
+    // run this every REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS seconds
+    while (true) {
+      await this.vaasCache.removeExpiredValues();
```
ali-bahjati, line 178: Maybe you can run it by `setInterval` instead of this structure.

third_party/pyth/price-service/src/rest.ts
cctdaniel: `mainnet` actually refers to `pythnet` for web-api endpoint, are you suggesting we should have a mapping instead that maps mainnet -> pythnet, and testnet -> pythtest?

third_party/pyth/price-service/src/listen.ts
```diff
@@ -52,6 +53,67 @@ type ListenerConfig = {
...
+        return arr[middle];
+      } else if (arr[middle].publishTime < publishTime) {
+        left = middle + 1;
+      } else {
+        nextLargest = middle;
+        right = middle - 1;
+      }
+    }
+
+    return nextLargest !== -1 ? arr[nextLargest] : null;
```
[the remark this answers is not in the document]
ali-bahjati: null is at some other places too.

third_party/pyth/price-service/.env.sample
```diff
@@ -17,3 +17,8 @@ PROM_PORT=8081
 
 # The default is to log with level info.
 #LOG_LEVEL=debug
+
+DB_API_CLUSTER=pythnet
+REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS=60
+PYTH_WRITER_INSERT_LATENCY_SECONDS=30
```
ali-bahjati, line 23: Can we remove this one too?

third_party/pyth/price-service/.env.sample
```diff
@@ -17,3 +17,8 @@ PROM_PORT=8081
 
 # The default is to log with level info.
 #LOG_LEVEL=debug
+
+DB_API_CLUSTER=pythnet
```
ali-bahjati, line 21: let's remove this one too. (it's optional right)

third_party/pyth/price-service/src/__tests__/rest.test.ts
```diff
@@ -115,34 +125,48 @@ describe("Latest Price Feed Endpoint", () => {
...
-    expect(resp.status).toBe(StatusCodes.OK);
-    expect(resp.body.length).toBe(2);
-    expect(resp.body).toContain(
-      Buffer.from("a1b2c3d4", "hex").toString("base64")
-    );
-    expect(resp.body).toContain(
-      Buffer.from("bad01bad", "hex").toString("base64")
-    );
-  });
+// describe("Latest Vaa Bytes Endpoint", () => {
```
ali-bahjati, line 128: Can you uncomment this too?

tilt-devnet/k8s/pyth-price-service.yaml
```diff
@@ -72,3 +72,13 @@ spec:
...
               value: debug
+            - name: CLUSTER
+              value: devnet
+            - name: REMOVE_EXPIRED_VALUES_INTERVAL_SECONDS
+              value: "60"
+            - name: PYTH_WRITER_INSERT_LATENCY_SECONDS
+              value: "30"
+            - name: CACHE_TTL_SECONDS
+              value: "300"
+            - name: WEB_API_ENDPOINT
```
ali-bahjati: Let's remove it from here too.

ali-bahjati (approved):
🚢 it ser

PR #484, opened 2023-01-12
[xc-admin] Header becomes class

guibescos:
- Governance Header becomes a class
- Governance Header encoding doesn't require to pass a buffer
- Buffer Layouts become static members of the classes
- Some comments

### DIFF
--- a/xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
+++ b/xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
@@ -1,11 +1,6 @@
 import { ChainId, ChainName } from "@certusone/wormhole-sdk";
 import * as BufferLayout from "@solana/buffer-layout";
-import {
-  encodeHeader,
-  governanceHeaderLayout,
-  PythGovernanceAction,
-  verifyHeader,
-} from ".";
+import { PythGovernanceAction, PythGovernanceHeader } from ".";
 import { Layout } from "@solana/buffer-layout";
 import {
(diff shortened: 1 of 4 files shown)


xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -89,37 +74,36 @@ export class ExecutePostedVaa implements PythGovernanceAction {
...
+      return { programId, keys, data };
+    });
     return new ExecutePostedVaa(header.targetChainId, instructions);
   }
 
-  /** Encode ExecutePostedVaaArgs */
+  /** Encode ExecutePostedVaa */
   encode(): Buffer {
-    // PACKET_DATA_SIZE is the maximum transaction size of Solana, so our serialized payload will never be bigger than that
+    const headerBuffer = new PythGovernanceHeader(
```
jayantk, line 100: These `encode` methods are set up in an inefficient way because you have to copy the data from the returned buffer each time. It would be better if the `encode` function wrote its bytes to the end of a passed in buffer. 

(probably not a big deal in this particular case, but it's the kind of thing that could cause a performance problem downstream.)

jayantk (approved):

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
guibescos: Totally agree but don't feel like performance will be a problem here

PR #486, opened 2023-01-12
[xc-admin] Decoders can't throw

guibescos:
Try to switch to a world where nothing can throw errors.

- Update `decode` to not throw and instead return undefined
- resolveAccountName can't throw

How `undefined` are handled :
- 1. If we don't recognize the program of the instruction, it becomes `UnrecognizedProgram implements MultisigInstruction` which stores the instruction (useful to print the instruction)
- 2. If we don't recognize an instruction in `PythMultisigInstruction` or `WormholeMultisigInstruction` the name of the instruction is set to `unrecognizedInstruction` instead of for example `postMessage` or `addPrice`.
- 3. If we recognize a `postMessage` in `WormholeMultisigInstruction` but we can't deserialize the wormhole payload `wormholeMultisigInstruction.governanceAction` will be `undefined`, but the payload is stored in `wormholeMultisigInstruction.args.payload`

It think it's good to distinguish the 3 cases because the `undefined` is at a different depth. It ends up being a little bit inconsistent, but ultimately it is because we use anchor to deserialize the instructions and that part somehow breaks the types.

### DIFF
--- a/xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
+++ b/xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
@@ -1,6 +1,10 @@
 import { ChainId, ChainName } from "@certusone/wormhole-sdk";
 import * as BufferLayout from "@solana/buffer-layout";
-import { PythGovernanceAction, PythGovernanceHeader } from ".";
+import {
+  PythGovernanceAction,
+  PythGovernanceHeader,
+  safeLayoutDecode,
+} from ".";
 import { Layout } from "@solana/buffer-layout";
 import {
   AccountMeta,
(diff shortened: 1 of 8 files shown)


xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -75,11 +75,21 @@ export class ExecutePostedVaa implements PythGovernanceAction {
...
+
+    let deserialized;
+    try {
+      deserialized = this.layout.decode(
+        data.subarray(PythGovernanceHeader.span)
+      );
+    } catch {
+      return undefined;
+    }
+
```
guibescos, line 92: buffer-layout can throw errors, so we need to handle them here that way the user of ours structs won't have to deal with them

xc-admin/packages/xc-admin-common/src/governance_payload/ExecutePostedVaa.ts
```diff
@@ -75,11 +75,21 @@ export class ExecutePostedVaa implements PythGovernanceAction {
...
+    let deserialized;
+    try {
+      deserialized = this.layout.decode(
+        data.subarray(PythGovernanceHeader.span)
+      );
+    } catch {
+      return undefined;
+    }
```
jayantk, line 91: make a `safeLayoutDecode` helper function so you don't have to repeat this try/catch?

jayantk (approved):
seems like a reasonable approach to me.

One thing I'll note is that I don't think you can necessarily decode -> encode if you have an unrecognized instruction. I'm not sure whether that's important though (maybe not)

guibescos:
> seems like a reasonable approach to me.
> 
> One thing I'll note is that I don't think you can necessarily decode -> encode if you have an unrecognized instruction. I'm not sure whether that's important though (maybe not)

I think you can because the `args` field of unkwownInstruction contains the instruction data.

PR #487, opened 2023-01-12
Cleanup local environment

guibescos:
The goal of this PR is cleaning up some things that I found frustrating about our local environment:
- We don't need to store the private keys of the create key of the multisig and the address of the multisig can be the same as mainnet.
- We don't need to deploy the CPI poster (weird wormhole program)

I also wanted to update the addresses of the Pyth oracle and the attesters but seems like pyth_admin relies on us storing the keypair.


third_party/pyth/prepare_multisig.py
```diff
@@ -11,14 +11,13 @@
...
 MESH_PROGRAM_ADDR = "SMPLVC8MxZ5Bf5EfF7PaMiTCxoBAcmkbM2vkrvMK8ho"
-MESH_CREATE_KEY_PATH = MESH_KEY_DIR + "create_key.json"
 MESH_VAULT_EXT_AUTHORITY_KEY_PATH = MESH_KEY_DIR + "external_authority.json"
 
 ALICE_KEY_PATH = MESH_KEY_DIR + "member_alice.json"
 BOB_KEY_PATH = MESH_KEY_DIR + "member_bob.json"
 CAROL_KEY_PATH = MESH_KEY_DIR + "member_carol.json"
 
-create_key_addr = sol_run_or_die("address", ["--keypair", MESH_CREATE_KEY_PATH], capture_output=True).stdout.strip()
+create_key_addr = "73UuSY2yXat7h7T49MMGg8TiHPqJJKKVc33DmC4b41Hf"
```
ali-bahjati, line 20: What key is this? (what is the keypair for it)

ali-bahjati (approved):
🚢  Let's burn all the unused stuff to a cleaner future 😁

third_party/pyth/prepare_multisig.py
ali-bahjati: If it's not obvious can you add a comment about it?

PR #489, opened 2023-01-13
[governance] move governance related components under this directory

ali-bahjati:
We have many components that are doing governance works, this PR aims to bring them all together in the governance directory. First commit just moves everything and second commit fixes anything affected by the move ci/builds/...


guibescos (approved):
🚢

PR #491, opened 2023-01-13
Fix package json for the crank 

guibescos:
We didn't import xc-admin-common

### DIFF
--- a/governance/xc-admin/packages/crank-executor/package.json
+++ b/governance/xc-admin/packages/crank-executor/package.json
@@ -22,6 +22,7 @@
     "@pythnetwork/client": "^2.9.0",
     "@solana/web3.js": "^1.73.0",
     "@sqds/mesh": "^1.0.6",
-    "ts-node": "^10.9.1"
+    "ts-node": "^10.9.1",
+    "xc-admin-common": "*"
   }
 }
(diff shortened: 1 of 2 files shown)


governance/xc-admin/packages/crank-executor/package.json
```diff
@@ -22,6 +22,7 @@
     "@pythnetwork/client": "^2.9.0",
     "@solana/web3.js": "^1.73.0",
     "@sqds/mesh": "^1.0.6",
-    "ts-node": "^10.9.1"
+    "ts-node": "^10.9.1",
+    "xc-admin-common": "*"
```
ali-bahjati, line 26: anything? 🤔 in rust there is a way for local dependency to be versioned. It helps in not missing the updates if the dependency updates, how would it be here? Of course there is the lock file.

ali-bahjati (approved):

governance/xc-admin/packages/crank-executor/package.json
guibescos: I think this is the way you do local dependencies in lerna

guibescos: It's importing the local package

PR #494, opened 2023-01-16
[xc-admin] permission accounts

cctdaniel:
- add update permissions tab


governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
```diff
@@ -1,13 +1,99 @@
...
+                    }
+                  >
+                    {tab[1].title}
+                  </Tab>
+                ))}
+              </Tab.List>
+            </Tab.Group>
+          </div>
+        </div>
+        {currentTabIndex !== -1 &&
```
guibescos, line 89: When can it be equal to -1

governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
```diff
@@ -1,13 +1,99 @@
...
+import { useRouter } from 'next/router'
+import { useEffect, useState } from 'react'
 import Layout from '../components/layout/Layout'
-import MinPublishers from '../components/MinPublishers'
+import MinPublishers from '../components/tabs/MinPublishers'
+import UpdatePermissions from '../components/tabs/UpdatePermissions'
 import { PythContextProvider } from '../contexts/PythContext'
+import { classNames } from '../utils/classNames'
+
+const tabInfo = {
```
guibescos, line 11: capitalize this so it's easier to know it's a const

governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
cctdaniel: when we visit the site without passing in query string `tab` at line 51

governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
```diff
@@ -1,13 +1,99 @@
...
+    title: 'Update Permissions',
+    description: 'Update the permissions of the program.',
+    queryString: 'update-permissions',
+  },
+}
+
+const DEFAULT_TAB = 'min-publishers'
 
 const Home: NextPage = () => {
+  const [currentTabIndex, setCurrentTabIndex] = useState(0)
```
guibescos, line 28: Does this need to be a number, wouldn't it make more sense as a "updatePermissions" | "setMinPub" or a string

guibescos (approved):
Logic of tabs seems complicated, but there might be a reason why that I don't know

governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
cctdaniel: yes -- we are using Tabs from @headless-ui and they require index to be passed in for their onChange handler

governance/xc-admin/packages/xc-admin-frontend/pages/index.tsx
cctdaniel: good catch, this was for a previous bug that has been fixed, will remove

PR #498, opened 2023-01-17
[xc-admin] add feature to send proposal to update permissions

cctdaniel:
- fix wallet hydration error
- update package-lock.json


governance/xc-admin/packages/xc-admin-frontend/utils/capitalizeFirstLetter.ts
```diff
@@ -0,0 +1,3 @@
+export const capitalizeFirstLetter = (str: string) => {
+  return str.replace(/^\w/, (c: string) => c.toUpperCase())
```
guibescos, line 2: I don't like having an utils folder tbh. And I feel like you could totally get away without this function. It's only called once.

guibescos: Kind of a nitpick though.

governance/xc-admin/packages/xc-admin-frontend/components/EditButton.tsx
```diff
@@ -0,0 +1,18 @@
+const EditButton = ({
+  editable,
+  onClick,
+}: {
+  editable?: boolean | null
```
guibescos, line 5: mmmmm what is the difference between `null` and `false`

governance/xc-admin/packages/xc-admin-frontend/components/tabs/UpdatePermissions.tsx
```diff
@@ -1,88 +1,364 @@
...
+            <CopyIcon className="shrink-0" />
+          </div>
+        </>
+      )
+    },
+    header: () => <span>Public Key</span>,
+  }),
+]
+
+type PubkeyChanges = {
```
guibescos, line 89: Maybe add some types here? There are only three possible keys "MasterAuthority", etc...

guibescos: You could make it a Record<"Master authority"| "Security authority" | "Data Curation", string >

governance/xc-admin/packages/xc-admin-frontend/components/tabs/UpdatePermissions.tsx
```diff
@@ -1,88 +1,364 @@
...
+            UPGRADE_MUTLTISIG[getMultisigCluster(cluster)],
+            1
+          ),
+          programDataAccount,
+        })
+        .instruction()
+        .then(async (instruction) => {
+          instruction.keys.forEach((k) => console.log(k.pubkey.toBase58()))
+          if (!isMultisigLoading && squads) {
+            console.log('setting isSendProposalButtonLoading to true')
```
guibescos, line 238: console log here

guibescos (approved):
I'm calling the 🚒 . A couple of code quality nitpicks.

governance/xc-admin/packages/xc-admin-frontend/utils/capitalizeFirstLetter.ts
cctdaniel: It's only called once for now because we are currently only using this for Update Permissions but once we have more tabs we will be using them almost everywhere we need a toast -- what do you suggest other than having an utils folder?

governance/xc-admin/packages/xc-admin-frontend/components/EditButton.tsx
cctdaniel: good catch will remove

PR #502, opened 2023-01-17
Drozdziak1/refactor p2w autoattest py

drozdziak1:
This change gets rid of non-daemon mode. The Python script still makes sense for Tilt, but removing the explicit daemon mode switch should simplify the prod command line a little.


ali-bahjati (approved):
Nice!

PR #504, opened 2023-01-17
[xc-admin] Add wormhole logic to the crank-executor

guibescos:
The crank that executes the proposals will be aware of wormhole transactions and pay the transfer fee

### DIFF
--- a/governance/xc-admin/package-lock.json
+++ b/governance/xc-admin/package-lock.json
@@ -1948,7 +1948,8 @@
     },
     "node_modules/@certusone/wormhole-sdk": {
       "version": "0.9.9",
-      "license": "Apache-2.0",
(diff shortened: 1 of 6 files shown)


jayantk (approved):
LGTM

this question is somewhat unrelated but it would be nice to have some testing infrastructure for this code. Would be great to have tests to validate that this fee payment worked, etc.  I realize that's not trivial but it's worth thinking about how to set it up.

guibescos:
> LGTM
> 
> this question is somewhat unrelated but it would be nice to have some testing infrastructure for this code. Would be great to have tests to validate that this fee payment worked, etc. I realize that's not trivial but it's worth thinking about how to set it up.

I've been testing it on tilt

ali-bahjati:
@guibescos I think its good to spend some time and set it up in tilt, both the backend and the frontend.

PR #506, opened 2023-01-18
[tilt] RFC add pythnet

guibescos:
Wanna hear thoughts on this, I feel like it might be useful to have Pythnet here to test remote governance locally + when solana becomes a receiver chain.


ali-bahjati:
Hmmm... I might do them in a single solana instance to use less resources but I get that having the exact same structure is better. I might just make some of these components optional and enable them in tilt ci but not on default development mode.

ali-bahjati (approved):

PR #509, opened 2023-01-18
[cosmwasm] update contract

0xfirefist:
- removed dependency pyth-sdk-cw
- add dependency pyth-sdk
- update contract logic to use latest price feed information and structs

### DIFF
--- a/target-chains/cosmwasm/contracts/pyth/src/contract.rs
+++ b/target-chains/cosmwasm/contracts/pyth/src/contract.rs
@@ -56,6 +56,7 @@ use {
     },
     p2w_sdk::{
         BatchPriceAttestation,
+        PriceAttestation,
         PriceStatus,
     },
     std::{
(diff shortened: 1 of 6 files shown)


target-chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -423,42 +439,19 @@ fn update_price_feed_if_new(
...
-            // - If a price is coming from future it's tolerated up to VALID_TIME_PERIOD seconds
-            //   (using abs diff) but more than that is set to unknown, the reason could be the
-            //   clock time drift between the source and target chains.
-            let time_abs_diff = if env_time_sec > price_pub_time_sec {
-                env_time_sec - price_pub_time_sec
-            } else {
-                price_pub_time_sec - env_time_sec
-            };
-
-            if time_abs_diff > config.valid_time_period.as_secs() {
```
jayantk, line 454: so is the intended use here that if people want to use the default valid time period, they can query the contract to retrieve the configured value, then pass that to `get_price_no_older_than`?

target-chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -348,22 +348,38 @@ fn process_batch_attestation(
...
-            ProductIdentifier::new(price_attestation.product_id.to_bytes()),
-            price_attestation.price,
-            price_attestation.conf,
-            price_attestation.ema_price,
-            price_attestation.ema_conf,
-            price_attestation.prev_price,
-            price_attestation.prev_conf,
-            price_attestation.prev_publish_time,
-        );
+        let price_feed = match price_attestation.status {
```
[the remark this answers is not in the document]
0xfirefist: added for this logic and then some

target-chains/cosmwasm/contracts/pyth/src/contract.rs
0xfirefist: Yes, it is upto them how they want to validate the price. They can fetch the default one. 
@ali-bahjati, can you please confirm this?

ali-bahjati (approved):

target-chains/cosmwasm/contracts/pyth/src/contract.rs
ali-bahjati: That's right. To my experience people usually don't use default time and rather use their own desired time. However, the problem about this was that the pyth-sdk has no clue about the `valid_time_period`. We are planning to remove the `pyth-sdk` entirely in the future and have per chain PriceFeed definition.

Also in other platforms we have more query methods, like `get_price` `get_price_no_older_than` `get_price_unsafe`. We can implement them in the future here if we realize the query of PriceFeed is expensive.

PR #510, opened 2023-01-18
[remote-executor] Add idl

guibescos:
Normally idl is automatically generated but I needed to touch it because the autogeneration wasn't detecting all our user-defined types, so I figured it is good to have this one handy.


ali-bahjati (approved):
Do we have any way to check idls are updated? a CI action or sth similar? Not for this program though, in general.

guibescos:
> Do we have any way to check idls are updated? a CI action or sth similar? Not for this program though, in general.

No, but I see this program as almost immutable.

PR #511, opened 2023-01-19
[xc-admin] Add pythnet relayer

guibescos:


governance/xc-admin/packages/xc-admin-common/src/wormhole.ts
```diff
@@ -9,3 +9,12 @@ export const WORMHOLE_ADDRESS: Record<PythCluster, PublicKey | undefined> = {
   localnet: new PublicKey("Bridge1p5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o"),
   testnet: undefined,
 };
+
```
ali-bahjati, line 12: Let's just leave a comment here to this link if anyone wants to change it in the future:
https://book.wormhole.com/reference/rpcnodes.html

ali-bahjati (approved):
Looks good to me ser, 🚢 it

PR #515, opened 2023-01-19
[cosmwasm] generate schema

0xfirefist:


target-chains/cosmwasm/contracts/pyth/schema/pyth-cosmwasm.json
```diff
@@ -0,0 +1,319 @@
+{
```
ali-bahjati, line 1: Can we make sure that this always gets updated? Maybe add it to the pre-commit?

ali-bahjati (approved):

PR #521, opened 2023-01-23
[price-service] Improve and update tests

ali-bahjati:
This PR improves the existing tests and adds test for the newly added `get_vaa` endpoint. Also adds tests to the tilt so we never miss them.

`listen.ts` is changed slightly to make testing easier.


jayantk (approved):
very nicely done. and thank you for adding tests!

cctdaniel (approved):
thanks for adding test ser 🫡

PR #537, opened 2023-01-26
[wormhole-attester] Remove wasm in js sdk

ali-bahjati:
This PR contains initial works of removing wasm usage in wormhole attester js sdk. Asking for comment before continuing the work.

Currently, the wasm dependency is our setup is very annoying. It gets built using a Dockerfile that does not work on M1 cpus and is not part of the npm workflow. I actually came across doing it while I wanted to setup monorepo for all the projects. However after finishing the code I looked at the governance repo and saw they included wasm build as part of their npm workflow, but it still requires rust to be installed to work. This is much better.

The benefit of using wasm is less reduntant code (not much) and in our case it uses the exact code that produced the attested messages. However, there is a big downside that I realized while doing this. Since wire format is designed to be forward-compatible, we cannot deploy a new version of the price-service unless the attester is deployed first (otherwise it will crash). We will have more flexibility if we have a separate code. Of course we can make the rust deserialize less strict but less strict can cause us to forget to change it in the future.

Given the arugments above, I like to know your opinion.

Remaining work:
- [] Clean up rust code & js dependencies
- [] Add tests


wormhole_attester/sdk/js/src/index.ts
```diff
@@ -38,13 +22,149 @@ export type BatchPriceAttestation = {
   priceAttestations: PriceAttestation[];
 };
 
-export async function parseBatchPriceAttestation(
```
guibescos, line 41: Name change?

wormhole_attester/sdk/js/src/index.ts
```diff
@@ -75,27 +195,6 @@ export function getBatchSummary(batch: BatchPriceAttestation): string {
   return JSON.stringify(abstractRepresentation);
 }
 
-export async function getSignedAttestation(
```
guibescos, line 78: Is this function not used anywhere?

ali-bahjati: ah no, i checked before deleting it. a removed component was using it.

wormhole_attester/sdk/js/src/index.ts
ali-bahjati: Hmm no the signature is the same, only the arg name has changed.

guibescos:
Ship it

wormhole_attester/sdk/js/src/index.ts
```diff
@@ -1,19 +1,11 @@
...
-    } else {
-      _P2W_WASM = await import("./wasm/bundler/pyth_wormhole_attester_sdk");
-    }
-  }
-  return _P2W_WASM;
+export enum PriceAttestationStatus {
+  Unknown = 0,
+  Trading = 1,
+  Halted = 2,
+  Auction = 3,
```
jayantk, line 8: I think i added `Ignored = 4` here

jayantk (approved):
I think this is simpler than the wasm for sure, and the wasm doesn't save much code, so this seems better.

wormhole_attester/sdk/js/src/index.ts
ali-bahjati: ah yeah, i'll add it.

PR #541, opened 2023-01-27
[Xc admin] upd permissions remote

guibescos:
This enables upd_permission for pythtest

TODO: Move multisigAuthority, isRemote, WORMHOLE_ADDRESS[getMultisigCluster(cluster)] and UPGRADE_MULTISIG[getMultisigCluster(cluster)] to the multisig hook? The remote, not-remote logic will be used in all tabs.
WDYT @cctdaniel


cctdaniel (approved):
sounds good

PR #548, opened 2023-02-01
[price-service] Remove wasm from readme and workflow

guibescos:
Ok so I found some Dockerfile.wasm mentions in the repo. @ali-bahjati you're probably already aware, since I think these command will need to be updated to build lerna.


ali-bahjati:
I think we need to build lerna base image first. Let me update it.

ali-bahjati (approved):
I updated the readme with lerna builds and checked it locally to make sure it works.

PR #559, opened 2023-02-02
[price-service/client] Add e2 test

ali-bahjati:
The tests are very basic to ensure the api is correct and works as expected. Will improve it when we need to touch the clients.


third_party/pyth/p2w_autoattest.py
```diff
@@ -114,7 +114,7 @@
 min_rpc_interval_ms: 0 # RIP RPC
 max_batch_jobs: 1000 # Where we're going there's no oomkiller
 default_attestation_conditions:
-  min_interval_secs: 60
+  min_interval_secs: 10
```
ali-bahjati, line 117: 10 seconds is still slow compared to the rest of groups but will help to make e2e tests run faster.

jayantk (approved):

PR #560, opened 2023-02-02
[price-service] Make the README better

guibescos:
I tried running the price-service by myself and these things would have helped me.


ali-bahjati (approved):

ali-bahjati:
thank you ser

PR #562, opened 2023-02-02
[prettier] Prettier sync

guibescos:
We have two different version of prettier :
- one from the precommit hooks
- one from the packages.json
This PR pins the lerna prettier to match the precommit prettier
The two versions don't always agree.

@ali-bahjati let me known if better ideas of how to solve this


ali-bahjati (approved):
Nice! thank you ser

PR #567, opened 2023-02-06
wormhole_attester: per-symbol PDAs for attestation state

drozdziak1:
This change replaces the big attestation state PDA with individual per-symbol ones. This ensures more predictable tx costs and is generally cheaper. The change is achieved by reusing the product accounts on the attest() endpoint and putting the state PDAs in their place.


wormhole_attester/program/src/attest.rs
```diff
@@ -66,42 +63,41 @@ pub const P2W_MAX_BATCH_SIZE: u16 = 5;
...
 
-    pub pyth_product4: Option<Info<'b>>,
-    pub pyth_price4:   Option<Info<'b>>,
+    pub pyth_state4: Option<AttestationStatePDA<'b>>,
+    pub pyth_price4: Option<Info<'b>>,
 
-    pub pyth_product5: Option<Info<'b>>,
-    pub pyth_price5:   Option<Info<'b>>,
+    pub pyth_state5: Option<AttestationStatePDA<'b>>,
+    pub pyth_price5: Option<Info<'b>>,
```
ali-bahjati, line 86: I guess you need to make the states mutable here

ali-bahjati (commented):
lgtm, i'll approve once the CI works. It seems like a mutability issue that I think is related to considering the state accounts mutable (on the contract)

wormhole_attester/program/src/attest.rs
```diff
@@ -224,53 +218,70 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
-            ));
-            return Err(ProgramError::InvalidAccountData.into());
+        // Update the on-chain value using publish_time or
+        // prev_publish_time if the price is not currently trading
+        state.0.last_attested_trading_publish_time = Some(new_last_attested_trading_publish_time);
+
+        // Serialize the state to calculate rent/account size adjustments
+        let state_serialized = state.0 .1.try_to_vec()?;
+
+        if state.0.is_initialized() {
```
guibescos, line 245: I think you probably don't need this resize logic. Can't you make solitaire take care of this?

wormhole_attester/program/src/attest.rs
```diff
@@ -224,53 +218,70 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
+
+            trace!("Attestation state rent transfer OK");
+        } else {
+            let seeds = state.self_bumped_seeds(price.key, ctx.program_id);
+            solitaire::create_account(
+                ctx,
+                state.0.info(),
+                accs.payer.key,
+                solitaire::CreationLamports::Exempt,
+                state_serialized.len(),
```
guibescos, line 273: Instead of use `state_serialized`, create a constant `AttestationState::LEN` which is equal to the maximum size the account can have. This will save you resizing.

wormhole_attester/program/src/attestation_state.rs
```diff
@@ -14,33 +14,47 @@ use {
...
+            symbol_id.to_bytes().to_vec(),
+        ]
     }
 }
 
-pub type AttestationStateMapPDA<'b> = Derive<
-    Data<'b, AttestationStateMap, { AccountState::MaybeInitialized }>,
-    "p2w-attestation-state-v1",
->;
+impl<'a, 'b: 'a> Peel<'a, 'b> for AttestationStatePDA<'b> {
```
guibescos, line 49: Surprised you have to derive peel and persist manually. Why can't solitaire handle this?

guibescos (commented):
Some comments

wormhole_attester/program/src/attest.rs
guibescos: Or turn `last_attested_trading_publish_time` into a `Timestamp` instead of an Option and use the `AccountSize` solitaire trait.

wormhole_attester/program/src/attestation_state.rs
drozdziak1: Oh, that's because automating this in a derive macro is surprisingly difficult and Solitaire cannot do this currently

Going on a bit of a framework design tangent, it's actually a very interesting topic. I think a solid dedicated `derive(Peel)` could make this go away. The thing with Solana account deserialization is that aimost any attempt at a framework will focus on adding behavior modifiers on some sort of user accounts struct, one way or another. Anchor chose to go with macro attributes as the first class citizen, where you just specify requirements in the spirit of ETH modifiers, onto struct members as part of the derive macro. Solitaire went a different direction to do as little simple codegen as necessary. The idea being to contain as much of the validation logic in regular Rust code instead of the derive macro. So the way to enforce specific rules is abstracted away from the derive macro (Notice how it says FromAccounts and not Peel). The macro only knows that all members should implement Peel and it just calls that on each account in the incoming slice on-chain. Thus most automated validation comes from stacking types like `Mut<Derive<Data<MyDataStruct, Initialized>>>`, because they implement Peel and try to cover most simple cases. There are several unanticipated problems with our approach.First,t Solana's Rust safety choices make it very hard to  come up with easy to use generics (we have to use lifetimes like 'a, 'b and so on). Also, our struct-based strategy that makes very little info available to the derive macro means that relationships between account struct members are almost impossible to express statically in struct definition

ali-bahjati (approved):
🔥

wormhole_attester/program/src/attestation_state.rs
guibescos: AHHHH ok I think I get what's the problem

guibescos: You can get away with this by making : 
`pub type AttestationStatePDA<'b> =Data<'b, AttestationState, { AccountState::MaybeInitialized }>
`
instead of :
```
pub struct AttestationStatePDA<'b>(
    pub Data<'b, AttestationState, { AccountState::MaybeInitialized }>,
);
```

guibescos: I think this is just a strict simplification of the code

wormhole_attester/program/src/attest.rs
```diff
@@ -224,53 +218,84 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
-            return Err(ProgramError::InvalidAccountData.into());
+        // Save the new value for the next attestation of this symbol
+        state.0 .0.last_attested_trading_publish_time = new_last_attested_trading_publish_time;
+
+        // handling of last_attested_trading_publish_time ends here
+
+        // Serialize the state to calculate rent/account size adjustments
+        let state_serialized = state.0 .0 .1.try_to_vec()?;
+
+        if state.0 .0.is_initialized() {
```
jayantk, line 259: extract a helper function for all of this account allocation logic

jayantk (commented):
i'm happy if @ali-bahjati and @guibescos are

wormhole_attester/program/src/attest.rs
guibescos: I think this realloc logic needs to go away. This account is fixed sized and this code is too complex.

wormhole_attester/program/src/attestation_state.rs
drozdziak1: That's exactly what I tried to do before I needed to add the `Seeded` impl. However, It is not possible to implement foreign traits for foreign types in Rust :( In this case, the typedef is considered equivalent to the foreign type we assign to it, so the compiler yells, because the trait is also foreign.

wormhole_attester/program/src/attest.rs
drozdziak1: You're right, this made sense only for the large PDA. Thanks.

wormhole_attester/program/src/attest.rs
drozdziak1: I think I prefer spending the extra gas in exchange for not maintaining a constant for this. I think without the realloc logic it can live in the uninitialized if branch, so this is only done once in a PDA's lifetime ( that's the case in the change that removes the unneeded realloc()). Fundamentally, I think Solitaire could benefit from a flag that triggers this call for uninitialized data accounts. On persist() it serializes the struct anyway.

guibescos (approved):

PR #573, opened 2023-02-09
[price_service/server] Improve vaa counter metric

ali-bahjati:
Currently our vaa counter metric seems to be showing more vaas than what we actually receive. This PR will try to address that issue by increase observedVaaCache and moving some checks around to ensure no VAA is counted twice.


cctdaniel (approved):
🙇🏻‍♂️

PR #580, opened 2023-02-10
[xc-admin] Crank : Improve error logging

guibescos:


governance/xc_admin/packages/crank_executor/src/index.ts
```diff
@@ -213,5 +213,9 @@ async function run() {
 }
 
 (async () => {
-  await run();
+  try {
+    await run();
+  } catch (err) {
+    console.error(err);
+  }
```
jayantk, line 220: this no longer throws on errors -- is that intentional?

jayantk (approved):

governance/xc_admin/packages/crank_executor/src/index.ts
jayantk: i mean it's not just improving logging, it's also suppressing errors

guibescos: I'll make it throw...

guibescos: Without this trick the error wasn't getting printed

PR #584, opened 2023-02-13
[xc-admin] add verified/unverified tags for individual proposals

cctdaniel:
- fix urlsIndex not being reset to 0 when changing cluster
- fix proposals accept/reject buttons not centered when on mobile view
- add verified tag to individual proposals

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
@@ -239,7 +239,7 @@ const Proposal = ({
           </div>
         </div>
         {proposalStatus === 'active' ? (
-          <div className="flex items-center justify-between px-8 pt-3">
+          <div className="flex items-center justify-center space-x-8 pt-3">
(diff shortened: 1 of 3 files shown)


governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
```diff
@@ -141,6 +142,9 @@ const Proposal = ({
           })
           proposalIxs.push(parsedInstruction)
         }
+        setIsVerified(
+          !proposalIxs.some((ix) => ix instanceof UnrecognizedProgram)
```
guibescos, line 146: I think the definition of verified is more something like :
`ix instanceof PythMultisigInstruction || (ix instanceof WormholeMultisigInstruction && ix.name === "postMessage" && ix.governanceAction instanceOf ExecutePostedVaa && ix.governanceAction.instructions.every(remote_ix => parse(remote_ix) instanceOf PythMultisigInstruction`

guibescos (approved):

PR #585, opened 2023-02-13
[cosmwasm] add testing module

0xfirefist:
### Completed
- [x] init package
- [x] move all the data structures to the package and consume them in the contract 
- [x] add MockPyth to facilitate testing for pyth contract consumers
### Pending (to cover in other PRs)
- [ ] any helper function if required
- [ ] generate schema
- [ ] setup release pipeline
- [ ] update contract to consume pyth-sdk-cw instead of pyth-cosmwasm.

### DIFF
--- a/target_chains/cosmwasm/pyth-sdk-cw/src/testing.rs
+++ b/target_chains/cosmwasm/pyth-sdk-cw/src/testing.rs
@@ -1,0 +1,87 @@
+use {
+    crate::{
+        error::PythContractError,
+        PriceFeed,
+        PriceFeedResponse,
+        PriceIdentifier,
+        QueryMsg,
+    },
+    cosmwasm_std::{
+        from_binary,
+        to_binary,
+        Binary,
+        Coin,
+        ContractResult,
+        QuerierResult,
+        SystemError,
+        SystemResult,
+    },
+    std::{
+        collections::HashMap,
+        time::Duration,
+        u128,
+    },
+};
+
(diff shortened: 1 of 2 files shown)


target_chains/cosmwasm/pyth-sdk-cw/src/testing.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /// Add a price feed that will be returned on queries.
+    pub fn add_feed(&mut self, feed: PriceFeed) {
+        self.feeds.insert(feed.id, feed);
+    }
+
+    /// TODO: Update this with example contracts -> Handler for processing query messages. See the tests in `contract.rs` for how to use this
+    /// handler within your tests.
```
jayantk, line 57: i note that contract.rs isn't included here at the moment. if you're going to add that in a future PR, great. If not, we should document how to use this somewhere.

jayantk (approved):

target_chains/cosmwasm/pyth-sdk-cw/src/testing.rs
0xfirefist: Yes, the example contract here - https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/cosmwasm/examples/cw-contract/src/contract.rs 

Will update the example in future PRs. This Todo will be resolved then.

PR #593, opened 2023-02-14
[cosmwasm] Update example contract

0xfirefist:
### Completed
- [x] init package
- [x] move all the data structures to the package and consume them in the contract 
- [x] add MockPyth to facilitate testing for pyth contract consumers
- [x] query helper function added
- [x] generated schema - no diff
- [x] update contract to consume pyth-sdk-cw instead of pyth-cosmwasm.
### Pending (to cover in other PRs)
- [ ] setup release pipeline

### DIFF
--- a/target_chains/cosmwasm/examples/cw-contract/Cargo.toml
+++ b/target_chains/cosmwasm/examples/cw-contract/Cargo.toml
@@ -34,5 +34,5 @@ cosmwasm-storage = { version = "1.0.0" }
 cw-storage-plus = "0.13.4"
 schemars = "0.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-pyth-cosmwasm = { git="https://github.com/pyth-network/pyth-crosschain", tag="pyth-cosmwasm-v0.1.0", features=["library"] }
+pyth-sdk-cw = { path="../../pyth-sdk-cw"}
 cosmwasm-schema = "1.1.9"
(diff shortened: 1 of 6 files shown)


target_chains/cosmwasm/examples/cw-contract/Cargo.toml
```diff
@@ -34,5 +34,5 @@ cosmwasm-storage = { version = "1.0.0" }
 cw-storage-plus = "0.13.4"
 schemars = "0.8"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-pyth-cosmwasm = { git="https://github.com/pyth-network/pyth-crosschain", tag="pyth-cosmwasm-v0.1.0", features=["library"] }
+pyth-sdk-cw = { path="../../pyth-sdk-cw"}
```
ali-bahjati, line 37: It would be good if you also add a version here so we get error if they are not a match and also add a comment that people need to use the deployed version when they copy/paste the example.

```
pyth-sdk-cw = { path="..", version="x.y.z" }
```

ali-bahjati (approved):
This is great!

In the future, it would be great to have a method on the `execute` side along with some typescript code to update the price at first.

target_chains/cosmwasm/examples/cw-contract/Cargo.toml
0xfirefist: Yes, in the next PR. I am going to release the module and will update this part to consume that.

PR #595, opened 2023-02-14
[cosmwasm] pyth-sdk-cw update version v1.0.0

0xfirefist:


ali-bahjati (approved):
Can you also update the actual contract version to v1.0.0 too?

PR #600, opened 2023-02-14
[xc-admin] Show rejected instead of cancelled

guibescos:


cctdaniel (approved):
thanks for catching this bug

PR #601, opened 2023-02-15
Bump wormhole rust crates to v2.14.8, remove rent adjustment

drozdziak1:
This change upgrades our wormhole cargo dependency to the current latest release. It introduces a fix in rent calculation logic, making Solitaire compatible with PythNet. This lets us get rid of dedicated rent adjustment logic, which we also remove in this change. Additionally, the Dockerfile.client Wormhole CLI's are bumped to match on top of Guillermo's recent version bump

### DIFF
--- a/target_chains/cosmwasm/Cargo.lock
+++ b/target_chains/cosmwasm/Cargo.lock
@@ -549,7 +549,7 @@ dependencies = [
 [[package]]
 name = "cw20-wrapped-2"
 version = "0.1.0"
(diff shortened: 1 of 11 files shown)


drozdziak1:
Note: I tested this manually in Pythnet under a separate address, we are producing seqnos as expected

guibescos (approved):
This is awesome. I'm hoping Tilt can run on this tho before merging.

drozdziak1:
My local Tilt is green for this. I hope we get the CI back soon

PR #603, opened 2023-02-15
[xc-admin] Make button green

guibescos:


cctdaniel:
is it just me or does the color look difficult to differentiate from rejected as well -- maybe try dark green?

cctdaniel (approved):

PR #604, opened 2023-02-15
Add USDC in xc-admin-cli

guibescos:


governance/xc_admin/packages/xc_admin_cli/src/index.ts
```diff
@@ -299,4 +305,59 @@ mutlisigCommand("approve", "Approve a transaction sitting in the multisig")
     await squad.approveTransaction(transaction);
   });
 
+mutlisigCommand("propose-token-transfer", "Propose token transfer")
+  .requiredOption("-a, --amount <number>", "amount in dollars")
+  .requiredOption("-d, --destination <pubkey>", "destination address")
+  .option(
+    "-m --mint <pubkey>",
+    "mint to transfer",
+    "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v" // default value is USDC SPL
```
ali-bahjati, line 314: Is it the same on both networks?

guibescos: I think this will only be used on mainnet

PR #607, opened 2023-02-16
[xc-admin] Forbid more than 32 pubs

guibescos:


cctdaniel (commented):
nitpick: add a period at the end of the error message

cctdaniel (approved):

PR #611, opened 2023-02-17
[sdk] Move JS sdks into this repo

jayantk:
This PR moves the contents of https://github.com/pyth-network/pyth-js into this repository. I did the following:

* Moved most of `target_chains/ethereum/` to `target_chains/ethereum/contracts` to match the organization of aptos etc. exception is the `examples` folder which stayed in `ethereum/examples`
* Copied each javascript sdk from pyth-js to `chain_name/js_sdk` 
* Copied the price pusher to the top-level `price_pusher` directory. (I think we will probably generalize this to support more chains, so top level made sense)

I have tried to keep this to moving/copying files, so I haven't changed any dependencies, package.jsons, or docs beyond what was necessary to get ci to pass. I will follow this up with a PR fixing all of that, plus the build & release process for the js libraries.


ali-bahjati (approved):
Thank you very much. I ran tilt locally to make sure it works. 

I think the next steps are adding the new packages to lerna, having publish action for them and image push action for the price pusher.

jayantk:
tested with tilt locally.

PR #612, opened 2023-02-20
[xc-admin] Batch instructions

guibescos:
This PR reduces the number of instruction that it takes to propose a change by batching instructions together.
It works great on devnet and testnet, but pythnet, pythtest will take some more work.


cctdaniel (approved):
🙇🏻‍♂️

PR #620, opened 2023-02-22
[xc-admin] add permission/depermission publisher keys for all asset types

cctdaniel:


governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
```diff
@@ -0,0 +1,250 @@
...
+        setIsSubmitButtonLoading(false)
+      }
+    }
+  }
+
+  useEffect(() => {
+    if (!dataIsLoading) {
+      const res: PublicKey[] = []
+      rawConfig.mappingAccounts[0].products.map((product: ProductRawConfig) => {
+        if (product.metadata.asset_type === selectedAssetType) {
```
guibescos, line 122: This not handling the all case properly

governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
```diff
@@ -0,0 +1,250 @@
...
+      const isRemote: boolean = isRemoteCluster(cluster)
+      const multisigCluster: Cluster | 'localnet' = getMultisigCluster(cluster)
+      const wormholeAddress = WORMHOLE_ADDRESS[multisigCluster]
+      const fundingAccount = isRemote
+        ? mapKey(multisigAuthority)
+        : multisigAuthority
+      priceAccounts.map((priceAccount) => {
+        isPermission
+          ? pythProgramClient.methods
+              .addPublisher(new PublicKey(publisherKey))
```
guibescos, line 83: How do you plan on handling the case where the product already has 32 pubs

guibescos (changes requested):
Need to handle case where there are 32 pubs

governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
cctdaniel: ah good catch

governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
```diff
@@ -120,15 +120,20 @@ const PermissionDepermissionKey = ({
       const res: PublicKey[] = []
       rawConfig.mappingAccounts[0].products.map((product: ProductRawConfig) => {
         if (
-          selectedAssetType === 'All' ||
-          product.metadata.asset_type === selectedAssetType
+          (isPermission && product.priceAccounts[0].publishers.length < 32) ||
+          !isPermission
         ) {
-          res.push(product.priceAccounts[0].address)
+          if (
```
guibescos, line 126: Nit pick make it && statement instead of a double if statement

guibescos (changes requested):
I think you need to handle the case where the publisher already exists for some price feeds. Similarly, only delete it from those price feeds where it is permissioned.

guibescos (approved):

PR #622, opened 2023-02-22
attester: Add an on-chain last attestation timestamp and rate limit arg

drozdziak1:
This new value is bumped on chain with each new attestation, on a per-symbol basis. The timestamp is common for  all cranks. This enables attester cranks to rate limit their attestations of a symbol together with all other attesters. Ultimately, our tx expenses  should drop while still fulfilling our preferred attestation rates.

## Rate limiting logic in depth
The rate-limiting works by letting clients specify a desired rate limiting interval in instruction args. If all of the batch's symbols were already attested up to that amount of time ago, the TX fails and no SOL is spent. Note: this rate limit may be reached because of a different attester working in parallel with the client that sees this soft error. The rate limiting error does not contribute to metrics nor healthcheck.

## Review areas of interest
* Comments and naming - This feature is the first of its kind, affecting concurrent attester behavior. I think it's important to make sure it's easy to understand the results coming from using this new feature in the crank service. Hopefully the comments are enough and easy to understand
* New logic in `attest.rs` code path - there's new logic evaluating the rate limit on chain and updating the value afterwards. 
* New attestation job logic in `main.rs` - We use tx simulation to detect the exact rate limit error and avoid counting it in monitoring ok/error counters. This is done by returning early from a rather dense code block. Style/readability suggestions are very welcome there. 

## Testing
The mock attester in Tilt gets a 2-second rate limit using the config value on its 1-second min interval attestation group. The rate limiting code can be seen as non-error INFO level messages about the rate limit. This happens only on 50-66% of resend attempts, as expected for the fast resends on a slower rate limit.

### DIFF
--- a/third_party/pyth/p2w_autoattest.py
+++ b/third_party/pyth/p2w_autoattest.py
@@ -116,9 +116,10 @@ max_batch_jobs: 1000 # Where we're going there's no oomkiller
 default_attestation_conditions:
   min_interval_secs: 10
 symbol_groups:
-  - group_name: fast_interval_only
+  - group_name: fast_interval_rate_limited
     conditions:
       min_interval_secs: 1
+      rate_limit_interval_secs: 2
     symbols:
 """
 
--- a/wormhole_attester/client/src/attestation_cfg.rs
+++ b/wormhole_attester/client/src/attestation_cfg.rs
@@ -318,6 +318,10 @@ pub const fn default_min_interval_secs() -> u64 {
     60
 }
 
+pub const fn default_rate_limit_interval_secs() -> Option<u32> {
+    Some(1)
+}
+
 pub const fn default_max_batch_jobs() -> usize {
     20
 }
@@ -340,7 +344,7 @@ pub struct AttestationConditions {
     /// redundant batch resends and tx expenses. NOTE: The client
     /// logic does not include rate limit failures in monitoring error
     /// counts.
-    #[serde(default)]
+    #[serde(default = "default_rate_limit_interval_secs")]
     pub rate_limit_interval_secs: Option<u32>,
 
     /// Limit concurrent attestation attempts per batch. This setting
@@ -384,7 +388,7 @@ impl Default for AttestationConditions {
             max_batch_jobs:              default_max_batch_jobs(),
             price_changed_bps:           None,
             publish_time_min_delta_secs: None,
-            rate_limit_interval_secs:    Some(1),
+            rate_limit_interval_secs:    default_rate_limit_interval_secs(),
         }
     }
 }
--- a/wormhole_attester/client/src/attestation_cfg.rs
+++ b/wormhole_attester/client/src/attestation_cfg.rs
@@ -66,6 +66,7 @@ pub struct AttestationConfig {
     /// Attestation conditions that will be used for any symbols included in the mapping
     /// that aren't explicitly in one of the groups below, and any groups without explicitly
     /// configured attestation conditions.
+    #[serde(default)]
     pub default_attestation_conditions: AttestationConditions,
 
     /// Groups of symbols to publish.
@@ -383,7 +384,7 @@ impl Default for AttestationConditions {
             max_batch_jobs:              default_max_batch_jobs(),
             price_changed_bps:           None,
             publish_time_min_delta_secs: None,
-            rate_limit_interval_secs:    None,
+            rate_limit_interval_secs:    Some(1),
         }
     }
 }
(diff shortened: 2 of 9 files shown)


third_party/pyth/p2w_autoattest.py
```diff
@@ -116,9 +116,10 @@
 default_attestation_conditions:
   min_interval_secs: 10
 symbol_groups:
-  - group_name: fast_interval_only
+  - group_name: fast_interval_rate_limited
     conditions:
       min_interval_secs: 1
+      rate_limit_interval_secs: 2
```
jayantk, line 122: do we want to have a version of this under `default_attestation_conditions` as well? I think in practice we're going to set this to 1 for most batches, and that would save us some configuration.

jayantk (approved):
looks great to me

guibescos (approved):
This made sense to me. Although, I don't know the intricacies of the attester crank so that's hard to review for me.

third_party/pyth/p2w_autoattest.py
drozdziak1: I think that makes sense, will do

wormhole_attester/client/src/main.rs
```diff
@@ -662,19 +667,42 @@ async fn attestation_job(args: AttestationJobArgs) -> Result<(), ErrBoxSend> {
...
             &config,
             &payer,
             wh_msg_id,
             symbols.as_slice(),
             latest_blockhash,
-        );
+            rate_limit_interval_secs,
+        )?;
+
+        // Detect rate limiting error early
```
ali-bahjati, line 680: I am a bit hesitant to this change because the RPC latency (which seems to be around 1s) might actually make this check worthless and even reduce the actual attestation rate hitting the network (regardless of the result). 

So 2 unfortunate consequences are:
1. You decide not to attest because the simulation fails, but if you continue the instruction and it actually sends the transaction it might not get rate limited.
2. You decide to attest and then it gets rate limited when it's sent for real.

I think it's better to send the tx and only capture the log of its result instead.

The slightly more nuance here is that rpc performs simulation before the actual tx send. I assume that when the contract returns error then the transaction will get reverted, right?
So they do the simulation and the above scenario's can still happen that actually make our performance a bit un-reliable. 
My suggestion is that you do not fail return an error when it is rate limited and just keep sending the msg to the log (either it is attesting or not). Then you can also have 2 metrics instead of 1 for sent messages. one for sent tx and one for tx es that really attest.

ali-bahjati (commented):
Amazing code and PR description 🔥 

I just have one comment that I like to know your opinion about. I might be wrong about what i think.

wormhole_attester/client/src/main.rs
drozdziak1: Thanks for noticing this. I thought about this and for some reason I couldn't find a way to get the logs from an error in `solana_client` :(

drozdziak1: Because in that case, We would only need to do a `map_err` and fish the expected log line out of the error type.

drozdziak1: Come to think of it, the right thing to do might be using a custom error code instead of a log line, I think I'll do that instead.

drozdziak1: Hmm, Maybe, but I don't like _not failing_ the transaction very much, because preflight failures are guaranteed to make this free

ali-bahjati: But our concern is not much the fees, they are essentially free right?

guibescos: I think you can disable preflight simulation (I know how to do it on javascript).

drozdziak1: I think Guillermo is right. I'll keep thinking about this for a while and settle for a preflight-less tx. This is probably the only way to get 100% precise rate limiting of the wormhole cross-txs

drozdziak1: Okay, I think I came up with the right compromise. We use an error code, look for it in failing tx, finish attestation job if it's detected, pass the error through otherwise. This means the attester will **never** breach the desired rate limit, and at absolute worst pay for the failing tx if  the rare race condition is met.

drozdziak1: Tilt just went green locally.

ali-bahjati (approved):

wormhole_attester/client/src/attestation_cfg.rs
```diff
@@ -318,6 +318,10 @@ pub const fn default_min_interval_secs() -> u64 {
     60
 }
 
+pub const fn default_rate_limit_interval_secs() -> Option<u32> {
+    Some(1)
```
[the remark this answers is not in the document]
drozdziak1: uuugh, now we can't pass none :1st_place_medal:

ali-bahjati (approved):

PR #624, opened 2023-02-23
[price-pusher] Refactor Price Pusher to support other chains 

0xfirefist:
moved shouldUpdate to price-config as it is closely related to it
pusher changed to controller + moved the code to push updates into evm.ts
evm.ts contains everything related to evm. PriceGetter, PricePusher, Contract

one change i do think should have been in the separate PR in this file price_pusher/src/pyth-price-listener.ts


0xfirefist:
I will send other PRs for above changes!

price_pusher/src/pyth-price-listener.ts
```diff
@@ -1,21 +1,15 @@
-import {
-  EvmPriceServiceConnection,
-  HexString,
-  PriceFeed,
-} from "@pythnetwork/pyth-evm-js";
+import { HexString, PriceFeed } from "@pythnetwork/pyth-evm-js";
```
ali-bahjati, line 1: You can remove this too.

price_pusher/src/price-config.ts
```diff
@@ -47,3 +48,84 @@ export function readPriceConfigFile(path: string): PriceConfig[] {
     return priceConfig;
   });
 }
+
```
ali-bahjati, line 51: Why did you move it here? I think maybe it's better to be on `controller.ts`

0xfirefist: it is very closely related to price config. 
if config changes we will need to update the should Update logic too. 
controller doesn't need to know anything related to priceConfig. 
I am thinking in this direction.

price_pusher/src/pyth-price-listener.ts
0xfirefist: will update it in other prs

PR #627, opened 2023-02-24
[price-pusher] add injective price listener

0xfirefist:


price_pusher/src/index.ts
```diff
@@ -11,13 +11,21 @@ import { EvmPriceListener, EvmPricePusher, PythContractFactory } from "./evm";
 import { PythPriceListener } from "./pyth-price-listener";
 import fs from "fs";
 import { readPriceConfigFile } from "./price-config";
+import { PriceServiceConnection } from "@pythnetwork/pyth-common-js";
+import { InjectivePriceListener, InjectivePricePusher } from "./injective";
 
 const argv = yargs(hideBin(process.argv))
-  .option("evm-endpoint", {
+  .option("network", {
```
ali-bahjati, line 18: Can you also update the readme with this new configs and values? You would need to update the compose files too.

price_pusher/package.json
```diff
@@ -43,6 +43,7 @@
     "typescript": "^4.6.3"
```
ali-bahjati, line 43: I think we need to bump version as major for this change.

price_pusher/src/index.ts
```diff
@@ -79,13 +87,45 @@ if (CONTRACT_ADDR[argv.pythContract] !== undefined) {
 
 const priceConfigs = readPriceConfigFile(argv.priceConfigFile);
 
-async function run() {
+async function injectiveRun() {
```
ali-bahjati, line 90: minor: I think `injective` and `evm` can have the same run function (or use a common function for running controller, ...) as many things are similar in your code.

price_pusher/src/injective.ts
```diff
@@ -0,0 +1,111 @@
...
+    console.log(`Polling the prices every ${this.pollingFrequency} seconds...`);
+    setInterval(this.pollPrices.bind(this), this.pollingFrequency * 1000);
+
+    await this.pollPrices();
+  }
+
+  private async pollPrices() {
+    console.log("Polling injective prices...");
+    for (const priceId of this.priceIds) {
+      const currentPriceInfo = await this.getOnChainPriceInfo(priceId);
```
ali-bahjati, line 50: What happens if a priceId does not exist?

price_pusher/src/injective.ts
```diff
@@ -0,0 +1,111 @@
...
+      price: string;
+      conf: string;
+      expo: number;
+      publish_time: number;
+    };
+  };
+};
+
+// this use price without leading 0x
+export class InjectivePriceListener implements PriceListener {
```
ali-bahjati, line 20: I see that this is very similar to EVM. Maybe it's good to extract common identical methods to `PriceListener`.

ali-bahjati (commented):
This is awesome. I really appreciate that you are improving the quality of this code. I left some comments and will approve once you address them.

price_pusher/src/index.ts
0xfirefist: I am not sure if this file is final yet. 
I have added the code to it just so that injective pieces are working fine. 
I will refactor this file too and then will update the README

price_pusher/package.json
0xfirefist: will send a pr with readme and version changes. which will be the final one after refactoring.

price_pusher/src/index.ts
0xfirefist: as mentioned above
will clean this file in separate PR

price_pusher/src/index.ts
jayantk: command line options will be annoying as the configuration becomes more complicated. I suggest moving this stuff to a config file. I note that there's already a price config file, so maybe it goes in there.

price_pusher/src/injective.ts
jayantk: agree with this. i think the latestPriceInfo / update / get methods could be moved up to priceListener

I wouldn't move the pollPrices method even though it is very similar to the evm method, as you want to give implementations the flexibility to implement the reading logic in different ways.

price_pusher/src/index.ts
jayantk: FYI if you want to do this stuff in a separate PR, i suggest leaving a code comment like 

`// FIXME: unify with evmRun() below once the interfaces are figured out.`

then people will understand what you're trying to do when reading the PR and won't leave comments about it

jayantk (commented):
i think extracting common methods like @ali-bahjati says is a good thing to do. If ali is happy, then I am; no need to wait for me to approve.

ali-bahjati (approved):

PR #630, opened 2023-02-24
[sdk] Add new SDK packages to lerna

jayantk:
Add the packages to the lerna build and fix dependencies. Note that this change also makes CI work: adding the packages to the root package.json means the existing lerna build / test CI step runs on these packages as well.


target_chains/ethereum/sdk/js/src/examples/EvmRelay.ts
```diff
@@ -55,6 +55,7 @@ async function run() {
     providerOrUrl: network,
   });
 
+  // @ts-ignore
```
jayantk, line 58: this is a weird one: the package versions changed, and in the process somehow the `HDWalletProvider` is no longer type-compatible with the argument to `Web3`. I tried changing the versions a bit and couldn't get this to work out. I validated that the script still runs properly though.

.github/workflows/lerna.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
           node-version: 18
           cache: "npm"
       - name: Install deps
-        run: npm ci
+        run: npm install
```
ali-bahjati, line 16: Please don't change this to install. We want to make sure the lock is updated.

jayantk: yes, will fix. (i'm just trying to debug the ci failure, which works fine locally, so not sure what's going on)

ali-bahjati (approved):

PR #633, opened 2023-02-27
[pice-pusher] add injective pusher

0xfirefist:


price_pusher/src/injective.ts
```diff
@@ -103,10 +122,126 @@ export class InjectivePriceListener implements PriceListener {
...
+    mnemonic: string
+  ) {
+    this.wallet = PrivateKey.fromMnemonic(mnemonic);
+  }
+
+  private injectiveAddress(): string {
+    return this.wallet.toBech32();
+  }
+
+  private async signAndBroadcastMsg(
```
ali-bahjati, line 139: It is a bit surprising to me that injective does not have a proper method for doing it.

price_pusher/src/injective.ts
```diff
@@ -103,10 +122,126 @@ export class InjectivePriceListener implements PriceListener {
...
+  }
+}
+
+// FIXME: a better place for it while refactoring
+export class CwPriceServiceConnection extends PriceServiceConnection {
+  constructor(endpoint: string, config?: PriceServiceConnectionConfig) {
+    super(endpoint, config);
+  }
+
+  async getPriceFeedUpdateObject(priceIds: string[]): Promise<any> {
```
ali-bahjati, line 238: If the only functionality is this we might just move it to the pusher 🤔

0xfirefist: Yeah. Actually, the version of pyth-common-js being used was an old one. getLatestVaas method was not public in the old one. 
Updated the version to latest. Now we can use getLatestVaas in this class too. 

Removed cwPriceServiceConnection in the latest commit.

PR #636, opened 2023-02-27
[price-pusher] refactor index page

0xfirefist:
- remove dependency on pyth-evm-js. we are only using a single method from it and contract addresses. 
- refactored the script to run this whole thing


price_pusher/src/index.ts
```diff
@@ -79,102 +76,99 @@ const argv = yargs(hideBin(process.argv))
...
+    targetPricePusher,
     {
       cooldownDuration: argv.cooldownDuration,
     }
   );
 
-  await injectivePriceListener.start();
-  await pythPriceListener.start();
+  await sourcePriceListener.start();
+  await targetPriceListener.start();
```
jayantk, line 103: maybe handler.start() should start the inner listeners too? that may simplify the usage.

price_pusher/src/interface.ts
```diff
@@ -8,6 +8,7 @@ export type PriceInfo = {
 };
 
 export interface IPriceListener {
+  start(): Promise<void>;
```
jayantk, line 11: can you document the assumptions around this method? I think `start` must fetch the current prices before the promise returns, right?

jayantk: Also what happens if someone starts the class twice?

jayantk (approved):
nice

price_pusher/src/index.ts
0xfirefist: yeah, it does

price_pusher/src/interface.ts
0xfirefist: it will setup another interval to poll prices
and another web socket connection 

Should we disallow it? I can add a static variable to check and return if it was already started

0xfirefist: we will discuss it later. this can be done in different pr

PR #639, opened 2023-02-28
[price-pusher] add a network helper

0xfirefist:


price_pusher/src/evm.ts
```diff
@@ -115,15 +118,39 @@ export class EvmPriceListener extends ChainPriceListener {
       publishTime: Number(priceRaw.publishTime),
     };
   }
+
+  static create(
+    endpoint: string,
+    pythContractAddr: string,
+    priceItems: PriceItem[],
+    pollingFrequency: DurationInSeconds
```
ali-bahjati, line 126: You are not passing the custom gas station here.

price_pusher/src/evm.ts
```diff
@@ -215,39 +242,49 @@ export class EvmPricePusher implements ChainPricePusher {
...
-  constructor(
-    private endpoint: string,
-    private mnemonic: string,
-    private pythContractAddr: string
-  ) {}
+  static create(
+    endpoint: string,
+    pythContractAddr: string,
+    mnemonic: string,
+    priceServiceConnection: PriceServiceConnection
```
ali-bahjati, line 250: You also need to pass the custom gas station

PR #645, opened 2023-02-28
Add documentation for audit

jayantk:
Add some docs to each function to make like easier for our auditors. 

I moved one `if` statement as well because I couldn't find the check and got scared for a second. Turns out we just had it in a weird place. There's a test case for this check which passes before/after the change.


target_chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -193,10 +210,18 @@ fn execute_governance_instruction(
...
+    // chain_id = 0 means the instruction applies to all chains
     if instruction.target_chain_id != state.chain_id && instruction.target_chain_id != 0 {
         return Err(PythContractError::InvalidGovernancePayload)?;
     }
 
+    // Check that the instruction is intended for this target chain contract (as opposed to
+    // other Pyth contracts that may live on the same chain).
+    if instruction.module != GovernanceModule::Target {
+        return Err(PythContractError::InvalidGovernancePayload)?;
+    }
```
jayantk, line 223: this is the one check that moved. we used to fail to deserialize governance instructions if they pointed to the other module, which doesn't really make sense.

0xfirefist: what do you mean by `other Pyth contracts that may live on the same chain`?

is the intention here the generated governance vaa should only run in target chain contract? Was there a problem we faced or is it something we can face in future.

0xfirefist (approved):

target_chains/cosmwasm/contracts/pyth/src/contract.rs
jayantk: good question. The governance message format encompasses all Pyth contracts, and this sort of target chain receiver contract is only one such type. At the moment, we only have one other module (an executor contract on pythnet), but we could add more in the future.

PR #647, opened 2023-03-01
Add an example script for benchmarks

jayantk:
This example uses the get_vaa endpoint and passes the result to parsePriceFeedUpdates. I validated that this works locally.

### DIFF
--- a/target_chains/ethereum/sdk/js/src/examples/EvmBenchmark.ts
+++ b/target_chains/ethereum/sdk/js/src/examples/EvmBenchmark.ts
@@ -1,0 +1,117 @@
+import Web3 from "web3";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+
+import { EvmPriceServiceConnection } from "../index";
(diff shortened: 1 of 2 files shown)


target_chains/ethereum/sdk/js/src/examples/EvmBenchmark.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  const [priceFeedUpdateVaa, updateTimestamp] = await connection.getVaa(
+    priceId,
+    unixTimestamp
+  );
+  console.log(`Next pyth update was at: ${updateTimestamp}`);
+  console.log(priceFeedUpdateVaa);
+
+  const priceFeedUpdate =
+    "0x" + Buffer.from(priceFeedUpdateVaa, "base64").toString("hex");
```
jayantk, line 78: it's kind of janky that the user has to convert the data format. We should fix this as part of our overall client js refactoring.

ali-bahjati (approved):

PR #648, opened 2023-03-01
[price-pusher] refactor

0xfirefist:
probably the last major refactor
after this I will raise a cleanup PR for minor changes and doc update

this pr implements 
separate command structure for separate chains

I have implemented a different approach at - https://github.com/pyth-network/pyth-crosschain/pull/639
With the pr to add custom gas station we saw how it affected the structure of commands. Options were added for injective too even thought they don't make any sense for injective. there was coupling between the constructors for different chain which I thought to be a valid assumption then.

This approach is better.
- good isolation
  - change in command of one won't affect another
- flexibility
  - implementation can vary wildly as long as they implement the priceListener and pricePusher interface. 
  
Will close the other PR after this.


price_pusher/src/index.ts
```diff
@@ -1,191 +1,9 @@
...
-    return new CustomGasStation(customGasStation, txSpeed);
-  }
-}
-
-start({
-  sourcePriceListener: pythPriceListener,
-  targetPriceListener: getNetworkPriceListener(argv.network),
-  targetPricePusher: getNetworkPricePusher(argv.network),
-});
+yargs(hideBin(process.argv)).command(evm).command(injective).help().argv;
```
jayantk, line 9: can you update the documentation / README to show the usage of the new command now

jayantk (approved):
lgtm, but please do update the documentation before merging. 

(one more comment here is that I would have considered going with a configuration file given the number of cli arguments. it's generally a little easier to manage configs and it gives you an easy way to save different versions of the system that you want to run. But this is ok too.)

price_pusher/src/index.ts
0xfirefist: yep, that's the next PR. cleanup and docs updates

0xfirefist:
> lgtm, but please do update the documentation before merging.
> 
> (one more comment here is that I would have considered going with a configuration file given the number of cli arguments. it's generally a little easier to manage configs and it gives you an easy way to save different versions of the system that you want to run. But this is ok too.)

I was thinking about this. I discovered the solution. Pushed a commit for the same. 
Both command options and a config file is supported

0xfirefist:
docs updated. merging it!

PR #649, opened 2023-03-02
[pyth-cosmwasm] sec review update

0xfirefist:
address the feedback for the initial sec review

following in particular:

1) all other chain contracts compare publish_time while this one checks for the [attestation time](https://github.com/pyth-network/pyth-crosschain/blob/fc08ec277e0c148883535515b7026fc90afd7571/target_chains/cosmwasm/contracts/pyth/src/contract.rs#L422) before updating. I don't see much impact to this except in the case of an asset that isn't trading anymore, it would be an extra storage call. Might make sense to make it uniform with the other contracts however. 
2) the owner field in the state doesn't seem to be used anywhere,

### DIFF
--- a/target_chains/cosmwasm/contracts/pyth/src/contract.rs
+++ b/target_chains/cosmwasm/contracts/pyth/src/contract.rs
@@ -384,9 +384,9 @@ fn process_batch_attestation(
     for price_attestation in batch_attestation.price_attestations.iter() {
         let price_feed = create_price_feed_from_price_attestation(price_attestation);
 
-        let attestation_time = Timestamp::from_seconds(price_attestation.attestation_time as u64);
+        let publish_time = Timestamp::from_seconds(price_attestation.publish_time as u64);
 
-        if update_price_feed_if_new(deps, env, price_feed, attestation_time)? {
+        if update_price_feed_if_new(deps, env, price_feed, publish_time)? {
             new_attestations_cnt += 1;
         }
     }
@@ -441,7 +441,7 @@ fn update_price_feed_if_new(
     deps: &mut DepsMut,
     env: &Env,
     price_feed: PriceFeed,
-    attestation_time: Timestamp,
+    publish_time: Timestamp,
 ) -> StdResult<bool> {
     let mut is_new_price = true;
     price_info(deps.storage).update(
@@ -450,14 +450,14 @@ fn update_price_feed_if_new(
             match maybe_price_info {
                 Some(price_info) => {
                     // This check ensures that a price won't be updated with the same or older
-                    // message. Attestation_time is guaranteed increasing in
+                    // message. Publish_TIme is guaranteed increasing in
                     // solana
-                    if price_info.attestation_time < attestation_time {
+                    if price_info.publish_time < publish_time {
                         Ok(PriceInfo {
                             arrival_time: env.block.time,
                             arrival_block: env.block.height,
                             price_feed,
-                            attestation_time,
+                            publish_time,
                         })
                     } else {
                         is_new_price = false;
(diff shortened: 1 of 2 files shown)


target_chains/cosmwasm/contracts/pyth/src/state.rs
```diff
@@ -41,7 +41,7 @@ pub struct PythDataSource {
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct ConfigInfo {
-    pub owner:                      Addr,
+    // TODO: ASK do we need to update config key?
```
ali-bahjati, line 44: What do you mean here?

target_chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -469,7 +468,7 @@ fn update_price_feed_if_new(
                     arrival_time: env.block.time,
                     arrival_block: env.block.height,
                     price_feed,
-                    attestation_time,
+                    publish_time,
```
ali-bahjati, line 471: I think this should remain the attestation_time, we are storing publish time inside the pricefeed right?

target_chains/cosmwasm/contracts/pyth/src/state.rs
```diff
@@ -68,10 +68,10 @@ pub struct ConfigInfo {
 #[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub struct PriceInfo {
```
ali-bahjati, line 70: I think we can remove the PriceInfo entirely. We had it before but these extra information here are not useful for people and add storage cost. We can only store PriceFeed.

target_chains/cosmwasm/contracts/pyth/src/contract.rs
0xfirefist: storing price feed instead of price info. no need of this anymore

target_chains/cosmwasm/contracts/pyth/src/state.rs
ali-bahjati: I don't know to what format Cosmwasm serialized though but I think it will break. We can update the config and have proper migration or rename it to `deprecated_owner`  and keep it. @njk-64 what do you think?

jayantk: Yeah this will probably break. 

Actually, this brings up a good question, which is how are we going to manage changes to this config class going forward?

I think we probably want to (1) stick a version number field in ConfigInfo so we can implement hard migration logic, and (2) add `#repr[(C)]` so we can extend it with fields at the end without having rust be too smart about packing the fields of the struct and potentially reordering things.

target_chains/cosmwasm/contracts/pyth/src/state.rs
```diff
@@ -65,15 +65,6 @@ pub struct ConfigInfo {
     pub fee: Coin,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq, JsonSchema)]
-#[serde(rename_all = "snake_case")]
-pub struct PriceInfo {
```
jayantk, line 70: you deleted this because we don't use any of the info here for anything anymore?

jayantk (approved):
LGTM. we should deal with the config issue also. will slack you.

target_chains/cosmwasm/contracts/pyth/src/state.rs
0xfirefist: yeah.

PR #653, opened 2023-03-02
[wormhole-attester] Increase accuracy and improve logging

drozdziak1:
- Tiltfile: TILT_DOCKER_REGISTRY def behavior, re-add namespace_create
  This helped me with k8s setup details that caused me issues locally:
    * The unspecified registry would make tilt fail to start by default
    * In minikube/k3s/<your development cluster> use cases, recreating the cluster may confuse Tilt because it doesn't create the desired namespace by default. It looks like tilt is starting but short after all services show that it cannot find the namespace.
-  wormhole_attester/client v5.0.0: accuracy and logging opimisations
    
    * [BREAKING CHANGE] min_interval_secs switches to milliseconds under
      min_interval_ms
    * attestation jobs no longer use preflight checks - this includes a
    custom variant of send_and_confirm_transaction(), see util.rs for details
    * attestation error logging no longer pretty-prints the error
      structs ({:#?} became {:?})

### DIFF
--- a/wormhole_attester/client/src/util.rs
+++ b/wormhole_attester/client/src/util.rs
@@ -6,6 +6,20 @@ use {
         trace,
     },
     prometheus::TextEncoder,
+    solana_client::{
+        client_error::Result as SolClientResult,
+        nonblocking::rpc_client::RpcClient,
+        rpc_config::RpcSendTransactionConfig,
+        rpc_request::RpcError,
+    },
+    solana_sdk::{
+        commitment_config::CommitmentConfig,
+        signature::Signature,
+        transaction::{
+            uses_durable_nonce,
+            Transaction,
+        },
+    },
     std::{
         net::SocketAddr,
         ops::{
@@ -17,9 +31,12 @@ use {
             Instant,
         },
     },
-    tokio::sync::{
-        Mutex,
-        MutexGuard,
+    tokio::{
+        sync::{
+            Mutex,
+            MutexGuard,
+        },
+        time::sleep,
     },
     warp::{
         reply,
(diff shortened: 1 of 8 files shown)


ali-bahjati (commented):
Awesome! Looks good to me. Let's get more eyes on this before merging.

wormhole_attester/client/src/util.rs
```diff
@@ -179,3 +196,71 @@ pub async fn start_metrics_server(addr: impl Into<SocketAddr> + 'static) {
...
+/// TODO(2023-03-02): Use an upstream method when
+/// it's available.
+///
+/// This method is almost identical to
+/// RpcClient::send_and_confirm_transaction(). The only difference is
+/// that we let the user specify the config and replace
+/// send_transaction() inside with
+/// send_transaction_with_config(). This variant is currently missing
+/// from solana_client.
+pub async fn send_and_confirm_transaction_with_config(
```
guibescos, line 212: I think something we might consider is dropping the confirmation as well

guibescos: Haven't thought very deeply but are we getting anything from confirmation

guibescos (approved):
I don't understand the titlfile change

wormhole_attester/client/src/util.rs
ali-bahjati: Yeah it makes it better but will make us blind on how much we are attesting successfully.

drozdziak1: Yeah, I think these results are critical to monitoring the service

PR #662, opened 2023-03-04
add metrics for each price id

orca-x:
we'd like to know the performance of each price pair and get notified when data has lags.

### DIFF
--- a/price_service/server/src/promClient.ts
+++ b/price_service/server/src/promClient.ts
@@ -51,6 +51,49 @@ export class PromClient {
     }
   });
 
+  private priceIdToPriceUpdatePublishTimeGapHistogram = new Map<
+    string,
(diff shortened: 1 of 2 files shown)


price_service/server/src/promClient.ts
```diff
@@ -51,6 +51,49 @@ export class PromClient {
     }
   });
 
+  private priceIdToPriceUpdatePublishTimeGapHistogram = new Map<
+    string,
+    client.Histogram<string>
+  >();
```
ali-bahjati, line 57: Is there any reason that you haven't used the `priceId` as a label and are one metric per each?

ali-bahjati:
Unfortunately this implementation is not suitable for our use case and we would rather use labels for price ids instead of a separate metric. We will address it later in the future.

PR #670, opened 2023-03-07
[xc-admin-frontend] Stop reloading proposals when you connect a wallet

guibescos:
Proposals would get refreshed when u connect disconnect your wallet. I refactored squads into `readonlySquads` and `squads` so that this behavior doesn't happen.


governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
```diff
@@ -60,7 +62,6 @@ const getSortedProposals = async (
 }
 
 export const useMultisig = (wallet: Wallet): MultisigHookData => {
-  const connectionRef = useRef<Connection>()
```
guibescos, line 63: Drive by cleanup

cctdaniel (approved):

PR #679, opened 2023-03-09
[price-pusher] Add nonce for evm + refactor

ali-bahjati:
This PR changes the EVM pusher to rely on the account nonce to increase reliability and cost-efficiency of the pusher.

On EVM networks nonce is the transaction index of the user transaction. it increments 1 by 1 and you cannot send two transactions with the same nonce; if there are multiple transactions with the same nonce only one gets chosen on-chain.

Sometimes our transactions do not land on chain because the gas is lower than the necessary amount to be included on-chain. By setting nonce manually we can replace the older transaction with a higher gas to increase  likelihood of the price landing on-chain without paying any extra cost. The `override-gas-price-multiplier` option with the default of 1.1 is added to handle this situation. Also we can try updating the prices on a faster frequency or even run multiple pushers with the same sender without adding any extra cost because they will be excluded (without reverting on-chain). Since we get nonce from the RPC, if the tx is landed with the RPC knowledge, then the quick update will get rejected on simulation because the RPC state is up to date.  However, in our experiments due to RPCs being behind a load-balancer and not being completely in-sync there were rare occurances of on-chain reverted transaction under extreme conditions (pushing every second). We can reduce the chances of this happening by reducing the pushing frequency.

Because of the above change and better handlings of errors, now it is safe to run multiple instances of the pusher using the same account without additional cost overhead.

In this PR there are some refactorings:
- The price service connection logger is changed to only log warning/error to reduce the log noise.
- The cooldown-duration is renamed to pushing-frequency to be more understandable. As this is a breaking change, the version is bumped to 4.0.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ bigtable-writer.json
 .aptos
 tsconfig.tsbuildinfo
 *~
+*mnemonic*
--- a/package-lock.json
+++ b/package-lock.json
@@ -47819,7 +47819,7 @@
     },
     "price_pusher": {
       "name": "@pythnetwork/pyth-price-pusher",
-      "version": "3.1.0",
+      "version": "4.0.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@injectivelabs/sdk-ts": "^1.0.457",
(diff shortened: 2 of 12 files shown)


0xfirefist (commented):
LGTM

jayantk (approved):

PR #680, opened 2023-03-09
[cosmwasm] Injective cosmwasm

0xfirefist:


target_chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -167,16 +180,30 @@ fn update_price_feeds(
...
+        }
     }
 
-    Ok(Response::new()
+    let num_total_new_attestations = total_new_attestations.len();
+
+    let response = Response::new();
+
+    #[cfg(feature = "injective")]
+    {
```
ali-bahjati, line 196: I think here you can have it in the code:
```
if cfg!(feature = "injective") {
 ...
}
```

ali-bahjati (approved):
Your change looks good to me. I think it is worth to add your injective build trick to the Readme (or other docs that we have) so people be aware of that. 

I suggest that we change the injective struct in such a way that it only contains the fields that PriceFeed has and tell them to update their contract. We do not have any guarantee over additional fields.

target_chains/cosmwasm/contracts/pyth/src/contract.rs
0xfirefist:
```
cannot find function `create_relay_pyth_prices_msg` in this scope
not found in this scope
```

I get this error doing it like this

ali-bahjati (approved):

PR #686, opened 2023-03-14
[eth] add evmos mainnet

cctdaniel:


target_chains/ethereum/sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/pyth-evm-js",
-  "version": "1.4.0",
+  "version": "1.5.0",
```
guibescos, line 3: Why do you need to bump the version

ali-bahjati: This is because the index file will change. @cctdaniel I think you missed to add the contract address here.

cctdaniel: great catch ser

ali-bahjati (approved):
Awesome!

PR #688, opened 2023-03-14
[price-pusher] correct option name

0xfirefist:


ali-bahjati (approved):
Thanks a lot, can you bump the version (as a patch) too?

PR #689, opened 2023-03-14
[price-pusher] Simulate injective tx

0xfirefist:
Prev we were passing hardcoded gas due to which transactions were failing as it was not enough
Simulating transaction to calculate it.

### DIFF
--- a/price_pusher/src/injective/injective.ts
+++ b/price_pusher/src/injective/injective.ts
@@ -85,15 +85,29 @@ export class InjectivePriceListener extends ChainPriceListener {
   }
 }
 
+type InjectiveConfig = {
+  chainId: string;
+  gasMultiplier: number;
+  gasPrice: number;
+};
 export class InjectivePricePusher implements IPricePusher {
   private wallet: PrivateKey;
+  private chainConfig: InjectiveConfig;
+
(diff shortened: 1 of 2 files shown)


price_pusher/src/injective/injective.ts
```diff
@@ -99,14 +100,44 @@ export class InjectivePricePusher implements IPricePusher {
...
+  private async signAndBroadcastMsg(msg: Msgs): Promise<TxResponse> {
     const chainGrpcAuthApi = new ChainGrpcAuthApi(this.grpcEndpoint);
     const account = await chainGrpcAuthApi.fetchAccount(
       this.injectiveAddress()
     );
+    const { txRaw: simulateTxRaw } = createTransactionFromMsg({
+      sequence: account.baseAccount.sequence,
+      accountNumber: account.baseAccount.accountNumber,
+      message: msg,
+      chainId: "injective-888",
```
jayantk, line 112: you may want to make the chainId, the gas price, and the extra_gas_multiplier members of the class (which you can initialize to the defaults in the constructor). That would make the code more flexible in case someone wants to use different values later.

jayantk (approved):

PR #694, opened 2023-03-15
[price-service] Rust implementation of a P2P price service.

Reisen:
This PR is based on the PoC that was written originally as a Geyser plugin for Accumulator account updates. Most code related to that has been removed and instead what is kept are two components:

- An initial implementation of an RPC using `Axum` to serve price service requests.
- A P2P implementation to join the Wormhole cluster, which we can leverage later as our own P2P cluster.

There's some things here that aren't complete which don't make sense to fill in until work on the price service API itself begins, so for example the `Vaa` type is not deserialised and the generic message handler is included (but a no-op, but kept for structure).

The P2P component is implemented as a Go library due to the fact that Rust's libp2p implementation doesn't support QUIC+TLS yet, which is Wormhole's current protocol requirement. This doesn't seem to be far off so once it's implemented we can drop all the Go pieces and replace it with Rust.


price_service/server-rust/build.rs
```diff
@@ -0,0 +1,73 @@
...
+    env,
+    path::PathBuf,
+    process::Command,
+};
+
+fn main() {
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out_var = env::var("OUT_DIR").unwrap();
+
+    // Clone the Wormhole repository, which we need to access the protobuf definitions for Wormhole
```
ali-bahjati, line 11: I think it's better to pull a versioned release of wormhole (instead of git) and pin the version. We can do this in another PR, but please leave a todo here.

price_service/server-rust/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name                           = "price-service"
+version                        = "0.1.0"
+edition                        = "2021"
+
+[dependencies]
+axum                           = { version = "0.6.9", features = ["json", "ws"] }
+axum-macros                    = { version = "0.3.4" }
+# yerpc                          = { version = "0.4.0", features = ["anyhow", "support-axum"] }
```
ali-bahjati, line 9: remove this one? :think:

price_service/server-rust/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
ali-bahjati, line 1: Can you add ignores for the build files?

price_service/server-rust/build.rs
```diff
@@ -0,0 +1,73 @@
...
+        .args(["+x", "tools/bin/*"])
+        .output()
+        .expect("failed to execute process");
+
+    // Generate the protobuf definitions. See buf.gen.yaml to see how we rename the module for our
+    // particular use case.
+    let _ = Command::new("./tools/bin/buf")
+        .args(["generate", "--path", "src"])
+        .output()
+        .expect("failed to execute process");
```
ali-bahjati, line 56: Alternatively couldn't we just copy the proto definition and keep it in our repo and build the go file ourself without relying on wh tools?

price_service/server-rust/src/main.rs
```diff
@@ -0,0 +1,138 @@
+#![feature(never_type)]
```
ali-bahjati, line 1: where this is used?

price_service/server-rust/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+
+//! This application implements an RPC and P2P gossip layer for PythNet that operates under
```
ali-bahjati, line 3: I think the doc here might be old now

price_service/server-rust/src/main.rs
Reisen: As the return type in main due to the process exit: `Result<!>` which reads "This program never returns a value, but if it was to return, it would have to be an error. Never types have been in Rust since the beginning:

```
fn main() -> ! {
    loop {}
}
```

which has always been valid, but using them in types such as `Result` is unstable but is already used through the standard library. I added this to remove the unreachable end of function warning, which happens due to the fact that `main` returns `Result<>` which rusts directly supports and will catch any errors that somehow manage to bubble out of the program).

It let's the compiler trust you more, and you can write code that is (reasonably) unreachable, write code without `Ok()` when it knows you can never write `Ok`, and stops forcing you to write conditionals that can never execute.

price_service/server-rust/src/network.rs
```diff
@@ -0,0 +1,6 @@
+pub mod p2p;
+pub mod rpc;
+pub use {
+    p2p::*,
+    rpc::*,
+};
```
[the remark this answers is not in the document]
Reisen: Agreed, fixed.

price_service/server-rust/build.rs
Reisen: Yes It's an option, I preferred this for a couple of reasons:

- It matches the `Makefile` in Wormhole
- The Wormhole repo doesn't have the compiled proto files, so to include them we'd need to do a few manual steps that aren't reflected in the repo. There's not a lot but they're not intuitive IMO:
  - Fetch upstream Wormhole.
  - Build the tools in our repo to capture the right hardcoded paths (ugly Go compiler thing).
  - Create a patch `buf.gen.yaml` file (which differs from Wormhole's) and generate p2p.pb.go. 
- I think we're on the way to wanting Proto definitions of our own types, and we'd have a similar process.

Really the wormhole clone is the ugliest part of this I agree, I think instead of pinning a version as you said above we should just make building the tools part of our own repo and copy the `.proto` file. I'll do it in a second PR.

.pre-commit-config.yaml
```diff
@@ -58,3 +58,9 @@ repos:
         entry: cargo +nightly clippy --manifest-path ./target_chains/cosmwasm/Cargo.toml --tests --fix --allow-dirty --allow-staged -- -D warnings
         pass_filenames: false
         files: target_chains/cosmwasm
+      - id: cargo-fmt-price-service
+        name: Cargo format for Rust Price Service
+        language: "rust"
+        entry: cargo +nightly fmt --manifest-path ./price_service/server-rust/Cargo.toml --all -- --config-path rustfmt.toml
```
jayantk, line 64: suggest fixing a version on this -- we've had issues before where CI fails because the nightly changes

price_service/server-rust/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+serde_json                     = { version = "1.0.93" }
+sha256                         = { version = "1.1.2" }
+structopt                      = { version = "0.3.26" }
+tokio                          = { version = "1.26.0", features = ["full"] }
+typescript-type-def            = { version = "0.5.5" }
+log                            = { version = "0.4.17" }
+
+# Setup LibP2P. Unfortunately the dependencies required by libp2p are shared
+# with the dependencies required by solana's geyser plugin. This means that we
+# would have to use the same version of libp2p as solana. Luckily we don't need
```
jayantk, line 35: this shouldn't be an issue if we have the two separate processes right?

price_service/server-rust/build.rs
jayantk: yeah i think the plan ^ makes sense.

price_service/server-rust/src/config.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/// StructOpt definitions that provides the following arguments and commands:
+///
+/// Some of these arguments are not currently used, but are included for future use to guide the
+/// structure of the application.
+#[derive(StructOpt, Debug)]
+#[structopt(name = "pythnet", about = "PythNet")]
+pub enum Options {
+    /// Run the PythNet P2P service.
```
jayantk, line 17: price service ?

price_service/server-rust/src/config.rs
```diff
@@ -0,0 +1,55 @@
...
+    Run {
+        /// A Path to a protobuf encoded ed25519 private key.
+        #[structopt(short, long)]
+        id: Option<PathBuf>,
+
+        /// A Path to a protobuf encoded secp256k1 private key.
+        #[structopt(long)]
+        id_secp256k1: Option<PathBuf>,
```
jayantk, line 25: what are these keys for?

price_service/server-rust/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Handler for LibP2P messages. Currently these consist only of Wormhole Observations.
+fn handle_message(_observation: network::p2p::Observation) -> Result<()> {
+    println!("Rust: Received Observation");
+    Ok(())
+}
+
+/// Initialize the Application. This can be invoked either by real main, or by the Geyser plugin.
+async fn init(_update_channel: Receiver<AccountUpdate>) -> Result<()> {
+    log::info!("Initializing PythNet...");
```
jayantk, line 44: price service -- couple more instances of "PythNet" in comments here, suggest find/replace.

price_service/server-rust/src/config.rs
jayantk: possible leftover from when this started an actual validator?

price_service/server-rust/src/network/rpc.rs
```diff
@@ -0,0 +1,118 @@
...
+            .unwrap_or_else(|e| e);
+        entry.insert(key, (timestamp, vaa));
+        Ok(())
+    }
+
+    /// Remove expired VAA's from the cache.
+    fn remove_expired(&mut self) -> Result<()> {
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)?
+            .as_secs() as i64;
```
jayantk, line 36: i think you need to subtract a fixed duration here (right now this is going to clear the cache entirely every time it runs)

price_service/server-rust/src/network/rpc.rs
```diff
@@ -0,0 +1,118 @@
...
+pub struct VaaCache(Arc<DashMap<String, Vec<(i64, Vaa)>>>);
+
+impl VaaCache {
+    /// Add a VAA to the cache. Keeps the cache sorted by timestamp.
+    fn add(&mut self, key: String, timestamp: i64, vaa: Vaa) -> Result<()> {
+        self.remove_expired()?;
+        let mut entry = self.0.entry(key).or_default();
+        let key = entry
+            .binary_search_by(|(t, _)| t.cmp(&timestamp))
+            .unwrap_or_else(|e| e);
```
jayantk, line 27: (nitpick: it's a little confusing to call the index into the vec `key` -- i spent a moment trying to figure out why you were redefining the existing map key here. i think `index` would be better)

Also suggest leaving a comment that this is potentially a performance problem -- i think this is doing O(n) work on each insert, and we could do better with a binary search tree. but no need to change it unless there's actually a problem.

price_service/server-rust/src/network/rpc.rs
```diff
@@ -0,0 +1,118 @@
...
+            .iter()
+            .filter_map(|item| {
+                if !ids.contains(item.key()) {
+                    return None;
+                }
+
+                let (_, latest_vaa) = item.value().last()?;
+                Some((item.key().clone(), latest_vaa.clone()))
+            })
+            .collect()
```
jayantk, line 62: do you want to return an Err here if you can't find one of the ids? It might be surprising for callers of `latest_vaas` to get back an update that contains only a subset of the data they requested. I'm not sure if that logic makes sense here or in the endpoint itself.

jayantk (approved):
nice! left you a couple minor comments, but should be easy to address.

price_service/server-rust/Cargo.toml
Reisen: Correct, just wanted to leave the comment just in case someone runs into it down the line.

price_service/server-rust/src/config.rs
Reisen: I actually think we should move away from thinking of this software as the price service (though adding it as price_service doesn't help that effort, I will outline why in a README I'll add).

price_service/server-rust/src/config.rs
Reisen: These were used when p2p was implemented in Rust, but passing them to the Go layer requires some additional Go wrappers that I haven't done yet. This PR is meant to be an initial structure and I think It's good to keep these to show how the software should be run even though they aren't forwarded through to Go yet. Instead Go is generating random keys/using hardcoded constants temporarily.

price_service/server-rust/src/main.rs
Reisen: Ditto to above comment, will clarify soon.

price_service/server-rust/src/network/rpc.rs
Reisen: That's correct, actually all the code in this is just temporary and bad and I just wanted something that fulfilled the interface, I wrote this with Ali quite quickly just so I could understand the current price service. I'll update this whole API to be make more sense once I understand what Ali is doing with the API.

price_service/server-rust/src/network/rpc.rs
Reisen: Yes this wasn't meant to be correct yet, and more demonstrating the structure. I just wrote something that worked, it will improve with the price service implementation.

ali-bahjati (approved):

PR #702, opened 2023-03-16
Pin rust version for formatting CI checks

jayantk:
The current nightly version of rust has a bug in rustc, which is causing our CI checks to fail. This happens periodically, so i'd like to pin the rust version used to something specific.

One downside of this approach is that it doesn't pin the version used on your local machine, so you can still get pre-commit differences between your local build and CI. We could fix that by pinning the toolchain using a toolchain file instead https://rust-lang.github.io/rustup/overrides.html?highlight=rust-toolchain#the-toolchain-file .

definitely open to the toolchain file approach if that seems better to you. It's a little more configuration work, but it standardizes everything everywhere.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.69
+          toolchain: nightly-2023-03-01
           components: rustfmt, clippy
           default: true
       - uses: pre-commit/action@v2.0.3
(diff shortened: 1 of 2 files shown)


guibescos (changes requested):
I don't think breaking the local precommit is worth it. This would require everytime I choose to work on this repo for me to change my default toolchain to nightly.

The global rust-toolchain file also doesn't work, since remote-executor tests don't work on nightly.
Our repo is cursed because there isn't a single toolchain that makes everything work.

jayantk:
discussed with @guibescos and it seems like there isn't a great way to pin this version, so nevermind. closing this pr.

PR #703, opened 2023-03-17
[evm] Verifying shimmer_testnet and neon_devnet with hardhat

guibescos:


ali-bahjati (approved):
Thank you very much for fixing this 🙇

PR #707, opened 2023-03-21
Add CI workflow to publish javascript packages

jayantk:
I think this should work, but we'll have to try it live to find out. See the diff of README to understand how it works.


price_pusher/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@pythnetwork/pyth-price-pusher",
+  "name": "@pythnetwork/price-pusher",
```
jayantk, line 2: this hasn't been published yet so fix the name while i'm at it

target_chains/aptos/sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/pyth-aptos-js",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
jayantk, line 3: nothing has changed code-wise, but i want to make sure this one and the one below get published

ali-bahjati (approved):

PR #710, opened 2023-03-22
sui: state, batch_price_attestation, set

optke3:
### Dependencies & Misc
- Change WH dependency from `sui/wormhole-cleanup` to `sui/integration_v2`
- Add TODO comments.
 
### `set.move`
- The Wormhole Set does not have the empty function, so we instead migrate the Aptos-Pyth set.move and adapt it for Sui
- Give the set values the store ability

### batch_price_attestation.move
- Use Wormhole bytes serialization/deserialization library in `batch_price_attestation.move`.
- Cast values in batch_price_attestation.move to `u64` as necessary
- Use `tx_context::epoch` for putting a timestamp on batch price attestation arrival time, since the Sui Clock feature is not available yet. The Sui clock would allow us to generate timestamps to 1 millisecond of precision and not use a `ctx` parameter.

state.move
- LatestPriceInfo (Mapping of cached price information) is no longer a thing, because we no longer maintain a centralized mapping of PriceInfos. This is because we want to be able to parallelize the updating of PriceInfo (which contains the PriceFeeds).
- Remove `state:: price_info_cached`, `state:: get_latest_price_info`. In the future, we may want to think about maintaining a lookup table for `price_identifier => price_info` without compromising on parallel updates.

### DIFF
--- a/target_chains/sui/contracts/sources/state.move
+++ b/target_chains/sui/contracts/sources/state.move
@@ -2,15 +2,10 @@ module pyth::state {
     use std::vector;
     use sui::object::{Self, UID};
     use sui::transfer::{Self};
-    use sui::tx_context::{Self, TxContext};
-    use sui::test_scenario::{Self};
+    use sui::tx_context::{TxContext};
 
-    //use pyth::price_identifier::PriceIdentifier;
-    use pyth::data_source::{Self, DataSource};
-    //use pyth::price_info::PriceInfo;
-
-    use wormhole::set::{Self, Set};
-    use wormhole::external_address::{Self};
+    use pyth::data_source::{DataSource};
+    use pyth::set::{Self, Set};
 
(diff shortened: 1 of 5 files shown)


target_chains/sui/contracts/sources/state.move
```diff
@@ -0,0 +1,117 @@
...
+        let DeployerCap { id } = deployer;
+        object::delete(id);
+
+        // Convert the vector of DataSource objects into a set
+        // of DataSource objects
+        let data_sources = set::new<DataSource>(ctx);
+        while (!vector::is_empty(&sources)) {
+            set::add(&mut data_sources, vector::pop_back(&mut sources));
+        };
+            transfer::share_object(
```
Reisen, line 58: Nitpick: indentation.

Reisen (approved):

PR #711, opened 2023-03-22
[evm] Polygon ZKEVM deployment

jayantk:
Deploy on zkevm testnet. I also added a polling interval option to the deployment to make rate limiting errors less of an issue.


cctdaniel (approved):
🙇🏻‍♂️ could you also update the js-sdk (for ref: https://github.com/pyth-network/pyth-crosschain/pull/683/files)

PR #716, opened 2023-03-23
[xc-admin] Introduce the new UPGRADE_OPS_KEY

guibescos:
I need this PR because the logic in `propose.ts` was wrong since it was hardcoded to use the price feeds ops key.
Now that the upgrade multisig has a different ops key we need to fix that.

- `OPS_KEY` becomes `PRICE_FEED_OPS_KEY` 
- `UPGRADE_OPS_KEY` introduced
- `UPGRADE_MULTISIG` and `PRICE_FEED_MULTISIG` moved to `xc-admin-common`
- `getOpsKey` gives you the right ops key for each multisig


cctdaniel (approved):
🙇🏻‍♂️

PR #717, opened 2023-03-23
[hermes] Add basic price store and rpc

ali-bahjati:
This PR adds the skeleton for the flow of a VAA until serving it to the VAA. The focus of this PR is the structure and flow of data between component. There are a lot of readmes that outline the needed future work and we need to add proper tests. I will add them in later PRs. 

Db is now a trait to allow multiple type of Dbs. Currently only the local cache is implemented and later we will implement a local_cache_with_backup_db to be on par with the node js one and we can add other things.


hermes/src/db.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum RequestTime {
+    Latest,
+    FirstAfter(UnixTimestamp),
+}
+
+/// This trait defines the interface for a database. The design can support both current
+/// batch-based attestation and future attestation based on accumulators. A accumulated
+/// data contains all the state at a single timestamp; but as we will probably have different
+/// accumulators, we store each data on a separate key. The data is defined as a binary
```
jayantk, line 22: i don't understand the sentence starting with "a accumulated data". I think a better way to say this would be something like: "If an item should appear under multiple keys (e.g., a batch of price updates), then ..."

hermes/src/db.rs
```diff
@@ -0,0 +1,27 @@
...
+/// to abstract away the details of the proof data.
+pub trait Db: Clone + Sync + Send {
+    fn insert(&mut self, key: &[u8], record: DbRecord) -> Result<()>;
```
jayantk, line 25: make a typedef for the keys?

hermes/src/network/p2p.go
```diff
@@ -46,122 +46,123 @@ import (
 
 //export RegisterObservationCallback
 func RegisterObservationCallback(f C.callback_t) {
-    go func() {
```
jayantk, line 49: seems like this is mostly formatting? can we add this to pre-commit to keep the diffs clean?

hermes/src/network/rpc/rest.rs
```diff
@@ -1,35 +1,102 @@
...
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct LatestVaaQueryParams {
     ids: Vec<String>,
 }
 
-/// REST endpoint /latest_price_feeds?ids[]=...&ids[]=...&ids[]=...
-/// TODO: Replace Infallible with an actual error return type instead of unwrap() crashing the RPC.
+/// REST endpoint /latest_vaas?ids[]=...&ids[]=...&ids[]=...
+///
+/// TODO: This endpoint returns proof as a byte array. We should probably return a base64 or hex string.
```
jayantk, line 57: suggest making it configurable with a get parameter

hermes/src/proof_store.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub price_feed:   PriceFeed,
+    pub vaa_bytes:    Vec<u8>,
+    pub publish_time: UnixTimestamp,
+}
+
+#[derive(Clone, Default, Serialize, Deserialize)]
+pub struct PriceFeedsWithProof {
+    pub price_feeds: HashMap<PriceIdentifier, PriceFeed>,
+    pub proof:       Proof,
+}
```
jayantk, line 93: this struct and Proof are a little confusing to me for two reasons:
1. It looks like the `Proof` contains the update data itself in addition to the proof itself, which makes the name "proof" a little confusing. It's really the entire verifiable update. Probably also worth documenting that the price_feeds are extracted from the proof and don't need to be recombined with the proof later.
2. Proof containing a `Vec<Vec<u8>>` seems like an odd mix between generic and specific. `Vec<u8>` would be the most generic form imo. However, I think you have `Vec<Vec<u8>>` because the contracts take an array of byte arrays? Maybe worth documenting why this is the format in that case.

hermes/src/proof_store.rs
```diff
@@ -0,0 +1,207 @@
...
+                        vaa_bytes: vaa_bytes.clone(),
+                        publish_time,
+                    };
+
+                    let key = price_feed.id.to_bytes().to_vec();
+                    let record = DbRecord {
+                        time:  publish_time,
+                        value: price_info.try_to_vec()?,
+                    };
+                    self.db.insert(&key, record)?;
```
jayantk, line 133: duplicating the proofs for each key (e.g., cloning the vaa) is going to be really memory inefficient once we get the merkle trees. It's going to clone each merkle tree O(price feeds) times. Is there a way you can make all the db entries point to the same piece of memory?

hermes/src/proof_store.rs
```diff
@@ -0,0 +1,207 @@
...
+                    let price_info = PriceInfo {
+                        price_feed,
+                        vaa_bytes: vaa_bytes.clone(),
+                        publish_time,
+                    };
+
+                    let key = price_feed.id.to_bytes().to_vec();
+                    let record = DbRecord {
+                        time:  publish_time,
+                        value: price_info.try_to_vec()?,
```
jayantk, line 131: i find it weird that you have to serialize the values into bytes and deserialize it below. Seems to me like you should make the Db trait generic instead

hermes/src/db/local_cache.rs
```diff
@@ -0,0 +1,88 @@
...
+            max_size_per_key,
+        }
+    }
+}
+
+impl Db for LocalCache {
+    /// Add a new db record to the deque for the key while keeping it sorted.
+    /// This function also removes the oldest record in the cache if the max_size is reached.
+    /// Records are usually added in increasing order and the function is optimized for that case.
+    fn insert(&mut self, key: &[u8], record: DbRecord) -> Result<()> {
```
jayantk, line 34: imo you should keep `DbRecord` internal to the DB implementation. You have to query with a unix timestamp below anyway, so make insert take a timestamp and the value as arguments. Then the `Db` trait is something like a `TemporalKeyValueStore` or something.

jayantk (commented):
I left you some implementation suggestions. I'm going to let @Reisen approve this though because I don't totally understand where this is going

hermes/src/network/p2p.go
ali-bahjati: Well. it seems that go strongly suggests formatting using tabs and it is enforced by gofmt (and it's not easily customisable)

hermes/src/proof_store.rs
ali-bahjati: Your suggestion makes sense. The reason I didn't make the Db trait generic was allowing us to handle multiple kinds of data (from different proofs) at once. I can have separate Db instances for different proof types.

hermes/src/network/rpc/rest.rs
ali-bahjati: Will do in another PR that works on this API more specifically.

hermes/src/proof_store.rs
ali-bahjati: About 1. You are right. Do you have a better name for it? I'll use UpdateData to be more clear.
About 2. You are right and David also mentioned it. The contracts will take it and also with the current batch vaa approach the proof is an array of byte arrays.

hermes/src/db.rs
```diff
@@ -0,0 +1,27 @@
...
+pub enum RequestTime {
+    Latest,
+    FirstAfter(UnixTimestamp),
+}
+
+/// This trait defines the interface for a database. The design can support both current
+/// batch-based attestation and future attestation based on accumulators. A accumulated
+/// data contains all the state at a single timestamp; but as we will probably have different
+/// accumulators, we store each data on a separate key. The data is defined as a binary
+/// to abstract away the details of the proof data.
```
Reisen, line 23:
```suggestion
/// This trait defines the interface for a Proof Storage
///
/// Price proofs for Pyth can come in multiple formats, for example VAA's and
/// Merkle proofs. The abstraction therefore allows storing these as binary
/// data to abstract the details of the proof data, and so each proof is stored
/// under a separate key. The caller is responsible for specifying the right
/// key for the proof data they wish to access.
```

hermes/src/db/local_cache.rs
```diff
@@ -0,0 +1,88 @@
...
+    dashmap::DashMap,
+    std::{
+        collections::VecDeque,
+        sync::Arc,
+    },
+};
+
+#[derive(Clone)]
+pub struct LocalCache {
+    cache:            Arc<DashMap<Vec<u8>, VecDeque<DbRecord>>>,
```
Reisen, line 17: In order to avoid the Sync problem we're relying on the fact the fields of these database objects are themselves Sync. This is fine but we should make sure to document that this is our design at the top of this file. If we were to introduce `sqlite` for example it would also require us to Arc<Mutex> the field and this should be clear.

hermes/src/db/local_cache.rs
```diff
@@ -0,0 +1,88 @@
...
+            cache: Arc::new(DashMap::new()),
+            max_size_per_key,
+        }
+    }
+}
+
+impl Db for LocalCache {
+    /// Add a new db record to the deque for the key while keeping it sorted.
+    /// This function also removes the oldest record in the cache if the max_size is reached.
+    /// Records are usually added in increasing order and the function is optimized for that case.
```
Reisen, line 33:
```suggestion
    /// Add a new db entry to the cache.
    /// 
    /// This method keeps the backed store sorted for efficiency, and removes
    /// the oldest record in the cache if the max_size is reached. Entries are
    /// usually added in increasing order and likely to be inserted near the
    /// end of the deque. The function is optimized for this specific case.
```

Added a single space between the one-line description and the body, we should maintain this style for all doc-style comments because it more closely reflects what Rustdoc looks like.

hermes/src/db/local_cache.rs
Reisen: Agree, just to expand. It would be nice if once `Vec<u8>` is typed, that that is the return and insert type. That way we keep a typed interface to the database.

hermes/src/proof_store.rs
Reisen: I think having different databases for different proofs is going to overcomplicate the interface a lot. The database as much as possible should be a single thing we have access to through the codebase. We should revisit the interface to either type the values being inserted or to define Db for `Vec<&dyn Db>` if we are going to have multiple databases so that we can write to multiple backends which is the natural expected behaviour for a program user rather than having to think about where things are being stored.

Reisen (changes requested):

hermes/src/main.rs
```diff
@@ -15,7 +16,9 @@ use {
 };
 
 mod config;
+mod db;
 mod network;
+mod proof_store;
```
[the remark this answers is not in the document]
ali-bahjati: I like it

hermes/src/db.rs
```diff
@@ -0,0 +1,27 @@
...
+pub trait Db: Clone + Sync + Send {
+    fn insert(&mut self, key: &[u8], record: DbRecord) -> Result<()>;
+    fn get(&self, key: &[u8], request_time: RequestTime) -> Result<Option<Vec<u8>>>;
```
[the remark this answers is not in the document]
ali-bahjati: There is no assumption on the trait level. Assumptions are on the actual databases.

hermes/src/db.rs
ali-bahjati: Thanks. Will do

hermes/src/proof_store.rs
ali-bahjati: I had a conversation with David and we decided to change the structure in a way that it knows the value explicitly. It will require us to implement specific things per DB but it is an expected behaviour to some extent. With this we won't have ser/de for local cache and will decide per DB

hermes/src/proof_store.rs
ali-bahjati: per David's suggestion in another thread, I changed the proof type to have all proofs and now vec<vec<u8>> is specific to batch_vaa.

hermes/src/db/local_cache.rs
ali-bahjati: The Storage trait has the bounds on `Sync + Send` to ensure all implementations follow it.

hermes/src/proof_store.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub db: D,
+}
+
+impl<D: Db> ProofStore<D> {
+    pub fn new(db: D) -> Self {
+        Self { db }
+    }
+
+    // TODO: This should return the updated feeds so the subscribers can be notified.
+    pub fn process_update(&mut self, update: ProofUpdate) -> Result<()> {
```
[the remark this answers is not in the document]
ali-bahjati: 👍

hermes/src/proof_store.rs
```diff
@@ -0,0 +1,207 @@
...
+                        value: price_info.try_to_vec()?,
+                    };
+                    self.db.insert(&key, record)?;
+                }
+            }
+        };
+        Ok(())
+    }
+
+    pub fn get_price_feeds_with_proof(
```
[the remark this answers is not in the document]
ali-bahjati: So I made this change, however it introduced another problem which only exists between batch_vaa and other proof types. The problem is that we cannot be certain that the latest price feed has both merkle and batch_vaa proof because they are generated differently. We can handle it to have different price feeds for each but that makes the api ugly. What should we do?

hermes/src/store/storage/local_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+                    RequestTime::FirstAfter(time) => {
+                        // If the requested time is before the first element in the vector, we are
+                        // not sure that the first element is the closest one.
+                        if let Some(oldest_record) = key_cache.front() {
+                            if time < oldest_record.time {
+                                return Ok(None);
+                            }
+                        }
+
+                        // Binary search returns Ok(idx) if the element is found at index idx or Err(idx) if it's not found which idx
```
Reisen, line 83: Tiny nitpick but very long comment, especially for terminal based editors. Might be worth adding comment formatting length to rustfmt.toml but not too important.

Reisen (changes requested):

hermes/src/store/proof/batch_vaa.rs
```diff
@@ -0,0 +1,140 @@
...
+        .map_err(|_| anyhow!("Failed to deserialize VAA"))?;
+
+    for price_attestation in batch_price_attestation.price_attestations {
+        let price_feed = price_attestation_to_price_feed(price_attestation);
+
+        let publish_time = price_feed.get_price_unchecked().publish_time.try_into()?;
+
+        let price_info = PriceInfo {
+            price_feed,
+            vaa_bytes: vaa_bytes.clone(),
```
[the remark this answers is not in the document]
ali-bahjati: This is inside a for loop and it needs to clone

Reisen (approved):

PR #721, opened 2023-03-24
[wormhole-attester] Fix rate_limit logic

ali-bahjati:
With the current logic, we attest every second and have different attestation times. However, users use price publish time and we need to perform the logic on the publish time. This PR fixes this issue. 

We need to deploy our contracts after this change. Since only the internal logic has changes, we don't need to touch attesters.

### DIFF
--- a/wormhole_attester/program/src/attest.rs
+++ b/wormhole_attester/program/src/attest.rs
@@ -255,25 +255,20 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
             price_struct,
         );
 
-        // don't re-evaluate if at least one symbol was found to be under limit
-        if over_rate_limit {
-            // Evaluate rate limit - should be smaller than duration from last attestation
-            let trading_publish_time_diff = new_last_attested_trading_publish_time
-                - state.0 .0.last_attested_trading_publish_time;
-            let attestation_publish_time_diff =
-                this_attestation_time - state.0 .0.last_attestation_time;
+        // Evaluate rate limit - should be smaller than duration from last attestation
+        let trading_publish_time_diff =
+            new_last_attested_trading_publish_time - state.0 .0.last_attested_trading_publish_time;
+        let attestation_time_diff = this_attestation_time - state.0 .0.last_attestation_time;
 
-            // We like to have the rate_limit for trading publish_time because that is the field that
-            // the users consume. However, when the price is not trading we still want to send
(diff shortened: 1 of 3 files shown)


wormhole_attester/program/src/attest.rs
```diff
@@ -255,21 +255,30 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
...
-                >= data.rate_limit_interval_secs as i64
+            let trading_publish_time_diff = new_last_attested_trading_publish_time
+                - state.0 .0.last_attested_trading_publish_time;
+            let attestation_publish_time_diff =
+                this_attestation_time - state.0 .0.last_attestation_time;
+
+            // We like to have the rate_limit for trading publish_time because that is the field that
+            // the users consume. However, when the price is not trading we still want to send
+            // the prices. We do not use the same frequency because that might cause two
+            // attestations within the rate_limit which is not a desired behavior.
```
drozdziak1, line 269: I'm not sure if I get this comment, can you help me understand? it is not actually possible to cause two attestations at the same, because one must be violating the rate limit, and we fail such transactions.

On a different note, can we change  `attestation_publish_time_diff` to `attestation_time_diff`? The naming is very hard to get right in this code path and I think `trading_publish_time_diff` is pretty good. However for `attestation_publish_time_diff`, saying `publish`when talking about `(last_)attestation_time` suggests it is related to the publish time coming from the oracle upstream, which is not the case. Most other uses just say `*_attestation_time` and I would stick with that. Let me know if there's something I am missing

wormhole_attester/program/src/attest.rs
```diff
@@ -255,21 +255,30 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
             price_struct,
         );
 
-        // Save the new value for the next attestation of this symbol
-        state.0 .0.last_attested_trading_publish_time = new_last_attested_trading_publish_time;
-
         // don't re-evaluate if at least one symbol was found to be under limit
         if over_rate_limit {
             // Evaluate rate limit - should be smaller than duration from last attestation
```
drozdziak1, line 260: My old comment suggests this is simpler than it really is, I'd get rid of it

drozdziak1 (approved):
Hey, this is pretty cool! Left a couple nits.

drozdziak1 (approved):
Lookin' good!

PR #722, opened 2023-03-24
Update .dockerignore

ali-bahjati:
The main purpose of this PR is to ignore as many things as possible for docker build to reduce the build transferred context size. Also, this PR fixes some bugs in our images to make Tilt working.


thmzlt:
CI failed with a different error. Maybe some of those paths needs to appear in the context?

image

PR #725, opened 2023-03-29
[evm] Deploy on Canto Testnet

cinquemb:


ali-bahjati (changes requested):
When testing the configurations using `bash deploy.sh canto_testnet` I realized that the wormhole chainId is not the one you mentioned (it is 5). Do you mind removing 7701.json file and then run these commands to redeploy to a new instance with the right chain Id?

```
npx lerna run build
export MNEMONIC=...
bash deploy.sh canto_testnet
```

Also the CI failed in pre-commit too. Can you install pre-commit and run `pre-commit install` on this repo before the commit? Then it will fix the formatting before the commit (or you can run `pre-commit run --all-files`

target_chains/ethereum/contracts/truffle-config.js
```diff
@@ -17,6 +17,8 @@ function payerProvider(url) {
     });
 }
 
+console.log(process.env.MIGRATIONS_DIR);
```
jayantk, line 20: delete this?

jayantk (approved):

ali-bahjati (approved):

PR #728, opened 2023-03-29
[price_pusher] Price pusher docker image workflow

jayantk:
Moving over the workflow to publish the price pusher from the pyth-js repo


ali-bahjati:
We have already moved it here ser.

PR #729, opened 2023-03-29
[accumulator-updater 2/x] Manual Serialization & Zero-copy for Mock-cpi-program

swimricky:
### Summary
Use zero-copy and implement manual serialization for mock-cpi-program schemas

### Updates

- add `schema` module in `mock-cpi-caller` that holds the various structs to transform the `PythAccount` into
- add `AccumulatorSerializer` trait for manual implementation of serialization (using big endian)
- add typescript types for above mentioned schemas
- add simple authorization mechanism for updating the `whitelist` in the `accumulator-updater` program

### TODOs

- [x] update mock-cpi-caller to use native solana + zero-copy as much as possible to more closely mimic pyth
- [ ] implement update/delete ixs and other types of accounts (product & mapping)
- [ ] handle(account_type, schema) mapping

### Testing
cd accumulator_updater && anchor test


accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -18,27 +18,34 @@ declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 pub mod accumulator_updater {
     use super::*;
 
-    pub fn initialize(ctx: Context<Initialize>) -> Result<()> {
+    pub fn initialize(ctx: Context<Initialize>, authority: Pubkey) -> Result<()> {
```
jayantk, line 21: does this need to be permissioned?

accumulator_updater/tests/accumulator_updater.ts
```diff
@@ -180,16 +228,54 @@ function parseAccumulatorInput({
...
   } else {
     console.log(`[compact]data: ${data.toString("hex")}`);
-    return mockCpiProg.coder.types.decode("PriceOnly", data);
+    return parsePriceCompact(data);
   }
 }
+
+//TODO: follow wormhole sdk parsing structure?
+// - https://github.com/wormhole-foundation/wormhole/blob/main/sdk/js/src/vaa/generic.ts
+type AccumulatorPriceAccount = PriceFull | PriceCompact;
```
jayantk, line 245: fwiw i find this name confusing. shouldn't this be something very generic like `Message`? the accumulator basically stores a bunch of arbitrary messages (each with their own schema)?

jayantk: actually i think this comment applies more generally, as the names currently make it unclear what things are generic and what things are specific to price accounts. I think the code would be easier to follow if eg. `PythSchema` was named something generic like `MessageSchema`, and then the stuff in `price.rs` was `message.rs`. The point here is to be clear via the naming and code structure what is generic data in the accumulator vs. what is pyth specific.

jayantk (commented):
cool glad to see you're making progress. I left you some comments, but I'm going to let @Reisen approve because he has all the context.

accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -105,10 +112,28 @@ pub mod accumulator_updater {
...
+        require!(
+            !self.allowed_programs.contains(&allowed_program),
+            AccumulatorUpdaterError::DuplicateAllowedProgram
+        );
+        Ok(())
+    }
+
+    pub fn validate_new_authority(&self, new_authority: Pubkey) -> Result<()> {
+        require_keys_neq!(new_authority, Pubkey::default());
+        require_keys_neq!(self.authority, new_authority);
```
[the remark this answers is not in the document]
swimricky: My thinking was that there would be no reason to need to call this ix if the caller was updating the authority to the current existing authority but if that seems unintuitive I can remove this

accumulator_updater/programs/mock-cpi-caller/src/lib.rs
```diff
@@ -209,14 +206,21 @@ pub struct AddPrice<'info> {
 
 
 //Note: this will use anchor's default borsh serialization schema with the header
-#[account]
+//  #[account(zero_copy)] is alias for
+//
+// #[derive(Copy, Clone)]
+// #[derive(bytemuck::Zeroable)]
+// #[derive(bytemuck::Pod)]
+// #[repr(C)]
```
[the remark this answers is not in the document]
swimricky: got it will clean this up thanks

accumulator_updater/programs/accumulator_updater/src/lib.rs
swimricky: is there a way to have this permissioned other than hard-coding the initial authority? I wasn't planning on doing the initialization call through CPI. this could theoretically be a problem  since the current implementation doesn't support multiple whitelists (that's a future TODO) on the off-chance that someone randomly calls this initialize ix before we do but even in this scenario we would be able to re-deploy the program to a new address.

accumulator_updater/tests/accumulator_updater.ts
swimricky: I agree, the naming should definitely be cleared up. There is the generic `AccumulatorInput` account that is 
```
AccumulatorInput: {
   header: AccumulatorHeader,
   data: Vec<u8> 
}
```

There is a little bit a blurry area though since I'm assuming we may want to be flexible and support being able to add `Product/Mapping Accounts` into the accumulator. Even though everything going into the `AccumulatorInput` PDAs are generic "messages", we do need to eventually map the data in these back into concrete `(PythAccountType, PythSchema)` structs ("PriceFull", "ProductCompact", etc).

accumulator_updater/programs/accumulator_updater/src/lib.rs
Reisen: I think It's ok to do it this way, unlikely someone initializes before us. Wormhole was setup this way also.

accumulator_updater/programs/accumulator_updater/src/lib.rs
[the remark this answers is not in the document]
Reisen: It's fine to do it this way, Wormhole was initialized the same way. No opinion on the second check however, I don't think it hurts to have necessarily but It's also not a harmful action.

accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -153,17 +181,36 @@ pub struct Initialize<'info> {
     pub system_program: Program<'info, System>,
```
Reisen, line 181: It's a nitpick but it would be good to format comments to a good 80-100 char width. Rustfmt can do it automatically but it tends to destroy intentionally formatted comments which is why we don't have it in rustfmt.toml right now, but it would still be good for hygiene to keep things formatted.

accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -153,17 +181,19 @@ pub struct Initialize<'info> {
...
+
 #[derive(Accounts)]
-pub struct AddAllowedProgram<'info> {
+pub struct UpdateWhitelist<'info> {
     #[account(mut)]
-    pub payer:          Signer<'info>,
+    pub payer:     Signer<'info>,
+    pub authority: Signer<'info>,
     #[account(
     mut,
```
Reisen, line 191: Nitpick: I don't know if rustfmt can/should be doing this but a single extra indent here makes sense.

Reisen (changes requested):

accumulator_updater/programs/accumulator_updater/src/lib.rs
swimricky: it seems like based on the below github issues/PRs,  rustfmt should've handled this automatically but not sure why it didn't. I manually indented it though. thanks for the catch

https://github.com/rust-lang/rustfmt/issues/2423
https://github.com/rust-lang/rustfmt/pull/2809/files#diff-e47dc5ac75eba9b921cbcd39a55310f28c152f14f38e5b496b3b8c3f805ac0e5

Reisen (approved):

accumulator_updater/programs/accumulator_updater/src/lib.rs
Reisen: Nah that's ok, intentionally not formatting comments because when you actually want to add structure it just craps over everything. There's a comment above I mentioned but It's good to manually keep formatting in comments good because we can't rely on rustfmt

PR #733, opened 2023-03-30
[cosmwasm] implement a build script

0xfirefist:
The way to build injective's code is by adding a default feature to `Cargo.toml` which is a manual process. 

This PR implements a script to automate that. 
I am going to use this script in our release pipeline but that is for another PR to release both cosmwasm and injective specific contract.

Additionally, I deleted some files './build.sh' we don't need it anymore. 
Docker files were there for no purpose.


jayantk (approved):
very nice

PR #736, opened 2023-03-31
[solana receiver] Try solana signature verification builtin

jayantk:
The wormhole program on solana has some downsides for Pyth. Specifically, it uses the secp256k1 instruction, which means a price update has to use multiple transactions. It also creates a VAA account which significantly increases the transaction cost. 

Another approach would be to check the wormhole signatures ourselves using the `secp256k1_recover`. This approach may let us do everything in one instruction & without the added tx cost. This PR is messing around with this function.

I also added some test infrastructure that deploys & runs the program in a simulated solana environment. 

Note there are some weird formatting things here. I'll fix those in the next PR by adding pre-commit for this directory, but I held off in this PR to avoid creating diff noise.


target_chains/solana/programs/solana-receiver/src/lib.rs
```diff
@@ -64,34 +95,8 @@ impl crate::accounts::DecodePostedVaa {
...
 }
 
-#[cfg(test)]
-mod tests {
-    use pyth_sdk::Identifier;
-    use pyth_wormhole_attester_sdk::PriceStatus;
-    use pyth_wormhole_attester_sdk::PriceAttestation;
-
-    #[test]
-    fn mock_attestation() {
```
jayantk, line 74: (deleted this because it doesn't do anything.)

target_chains/solana/programs/solana-receiver/src/tests/simulator.rs
```diff
@@ -0,0 +1,163 @@
...
+    },
+    std::{
+        mem::size_of,
+        path::Path,
+    },
+};
+
+/// Simulator for the state of the target chain program on Solana. You can run solana transactions against
+/// this struct to test how pyth instructions execute in the Solana runtime.
+pub struct ProgramSimulator {
```
jayantk, line 44: this whole file is pretty much a copy of https://github.com/pyth-network/pyth-client/blob/main/program/rust/src/tests/pyth_simulator.rs .

target_chains/solana/programs/solana-receiver/src/tests/simulator.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+impl ProgramSimulator {
+
+    /// Deploys the target chain contract as upgradable
+    pub async fn new() -> ProgramSimulator {
+        let mut bpf_data = read_file(
+            std::env::current_dir()
+                .unwrap()
+                .join(Path::new("../../target/deploy/pyth_solana_receiver.so")),
```
jayantk, line 63: this is a little janky but it's the only way that @guibescos found to deploy the program as upgradeable (so you can test program upgrades).

guibescos: For this one you don't need to deploy the program as upgradable, I recommend trying something like what I did here https://github.com/pyth-network/program-authority-escrow/blob/main/programs/program-authority-escrow/src/tests/simulator.rs#L50

guibescos (approved):
Great. Left one comment to cleanup some tests.

target_chains/solana/programs/solana-receiver/src/tests/simulator.rs
jayantk: agree it's not needed right now, but we'll want this as we go forward with the target chain contract (because we'll want to test program upgrades via governance).

target_chains/solana/programs/solana-receiver/src/tests/test_update_price.rs
```diff
@@ -0,0 +1,47 @@
...
+ToAccountMetas,
+InstructionData,
+AnchorDeserialize,
+};
+use rand::rngs::OsRng;
+
+use crate::tests::simulator::ProgramSimulator;
+
+#[tokio::test]
+async fn test_update_price() {
```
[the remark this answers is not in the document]
jayantk: yeah... this doesn't actually test anything right now, it just prints stuff out. i'll make it do something useful in a bit.

PR #739, opened 2023-04-03
[cosmwasm] Refactor deployer

0xfirefist:
Work still in progress. In order to keep this PR small. Other things will be done in further PRs

This PR adds the generic code to the repo.

### DIFF
--- a/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
+++ b/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
@@ -2,7 +2,7 @@ import { CosmwasmExecutor } from "./cosmwasm";
 import { InjectiveExecutor } from "./injective";
 
 import { Coin } from "@cosmjs/stargate";
-
+// TODO: expose these executors and consume them in price pusher
 /**
  * Interface for classes that implements contract management for a chain.
  *
--- a/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
+++ b/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
@@ -173,7 +173,7 @@ export type CosmwasmConfig = {
   // FIXME: probably have to pass the hd path too. not sure
 };
 
-export function ChainExecutorFactory({
+export function executorFromConfig({
   type,
   value,
 }: ChainExecutorFactoryArgs): ChainExecutor {
--- a/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
+++ b/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
@@ -1,6 +1,3 @@
-import { CosmwasmExecutor } from "./cosmwasm";
-import { InjectiveExecutor } from "./injective";
-
 import { Coin } from "@cosmjs/stargate";
 // TODO: expose these executors and consume them in price pusher
 /**
@@ -144,47 +141,3 @@ export type UpdateContractAdminRequest = {
 export type UpdateContractAdminResponse = {
   txHash: string;
 };
-
-export enum ChainExecutorType {
-  INJECTIVE = "injective",
-  COSMWASM = "cosmwasm",
-}
-
-export type ChainExecutorFactoryArgs =
-  | {
-      type: ChainExecutorType.INJECTIVE;
-      value: InjectiveConfig;
-    }
-  | {
-      type: ChainExecutorType.COSMWASM;
-      value: CosmwasmConfig;
-    };
-
-export type InjectiveConfig = {
-  grpcEndpoint: string;
-  mnemonic: string;
-};
-
-export type CosmwasmConfig = {
-  endpoint: string;
-  mnemonic: string;
-  prefix: string;
-  gasPrice: string;
-  // FIXME: probably have to pass the hd path too. not sure
-};
-
-export function executorFromConfig({
-  type,
-  value,
-}: ChainExecutorFactoryArgs): ChainExecutor {
-  if (type === ChainExecutorType.INJECTIVE) {
-    return new InjectiveExecutor(value.grpcEndpoint, value.mnemonic);
-  }
-
-  return new CosmwasmExecutor(
-    value.endpoint,
-    value.mnemonic,
-    value.prefix,
-    value.gasPrice
-  );
-}
--- a/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
+++ b/target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
@@ -1,0 +1,68 @@
+export type Fund = {
+  denom: string;
+  amount: string;
+};
+export type StoreCodeRequest = {
+  contractBytes: Buffer;
+};
+export type StoreCodeResponse = {
+  codeId: number;
+  txHash: string;
+};
+
+export type InstantiateContractRequest = {
+  codeId: number;
+  instMsg: object;
+  label: string;
+};
+export type InstantiateContractResponse = {
+  contractAddr: string;
+  txHash: string;
+};
+
+export type ExecuteContractRequest = {
+  contractAddr: string;
+  msg: object;
+  funds?: Fund[];
+};
+export type ExecuteContractResponse = {
+  txHash: string;
+};
+
+export type MigrateContractRequest = {
+  contractAddr: string;
+  newCodeId: number;
+  migrateMsg?: object;
+};
+export type MigrateContractResponse = {
+  txHash: string;
+};
+
+export type UpdateContractAdminRequest = {
+  newAdmin: string;
+  contractAddr: string;
+};
+export type UpdateContractAdminResponse = {
+  txHash: string;
+};
+
+// if a transaction fails, methods will raise an error
+export interface ChainExecutor {
+  storeCode(req: StoreCodeRequest): Promise<StoreCodeResponse>;
+  instantiateContract(
+    req: InstantiateContractRequest
+  ): Promise<InstantiateContractResponse>;
+  executeContract(
+    req: ExecuteContractRequest
+  ): Promise<ExecuteContractResponse>;
+  // this method assumes that the key used to execute this method is
+  // the admin of the given contract or else it will throw an error
+  migrateContract(
+    req: MigrateContractRequest
+  ): Promise<MigrateContractResponse>;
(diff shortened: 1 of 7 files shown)


target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type UpdateContractAdminRequest = {
+  newAdmin: string;
+  contractAddr: string;
+};
+export type UpdateContractAdminResponse = {
+  txHash: string;
+};
+
+// if a transaction fails, methods will raise an error
```
jayantk, line 52: I think it would be good to document this class more extensively. Explain what each method does, how to invoke them, and any special cases or concerns the caller needs to worry about. (you've done some of the last thing by documenting assumptions)

you probably don't have to document the types ^ as long as these methods are documented.

target_chains/cosmwasm/tools/src/chains-manager/chain-executor.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type CosmwasmConfig = {
+  endpoint: string;
+  mnemonic: string;
+  prefix: string;
+  gasPrice: string;
+  // FIXME: probably have to pass the hd path too. not sure
+};
+
+export function ChainExecutorFactory({
```
jayantk, line 101: this is function so should have a lower-cased name. Also I think it's not great to call it a "factory" because that suggests that it's a class. How about something like `executorFromConfig`?

target_chains/cosmwasm/tools/src/chains-manager/chain-querier.ts
```diff
@@ -0,0 +1,102 @@
...
+export type RawContractStateRequest = {
+  contractAddr: string;
+  key: Buffer;
+};
+
+export type AllContractStateRequest = {
+  contractAddr: string;
+};
+
+export interface ChainQuerier {
```
jayantk, line 30: (this is a nitpick) I think this code would be easier to understand if you defined this interface first in the file, then had the request types after. This interface is the main thing the reader needs to understand, and once they see this, the request types will make sense. (similar organizational comment applies to the file above)

also please document these methods, especially since they return `Object` which doesn't say a lot about what's in them.

target_chains/cosmwasm/tools/src/chains-manager/chain-querier.ts
```diff
@@ -0,0 +1,102 @@
...
+  getContractInfo(req: ContractInfoRequest): Promise<ContractInfoResponse>;
+  getSmartContractState(req: SmartContractRequest): Promise<Object>;
+  getRawContractState(req: RawContractStateRequest): Promise<Object>;
+  getAllContractState(req: AllContractStateRequest): Promise<Object>;
+}
+
+export class GenericQuerier implements ChainQuerier {
```
jayantk, line 37: a doc comment at the top of this class would be helpful too, e.g., to explain why this is a "generic" querier

target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
```diff
@@ -0,0 +1,240 @@
...
+  MsgStoreCodeEncodeObject,
+  SigningCosmWasmClient,
+  MsgMigrateContractEncodeObject,
+  MsgUpdateAdminEncodeObject,
+} from "@cosmjs/cosmwasm-stargate";
+import { GasPrice } from "@cosmjs/stargate";
+import Long from "long";
+import assert from "assert";
+
+export class CosmwasmExecutor implements ChainExecutor {
```
jayantk, line 28: is this the same "generic" version of the executor? if so, suggest naming both classes along the same lines so the reader understands the relationship.

target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
```diff
@@ -0,0 +1,240 @@
...
+    );
+
+    return {
+      txHash: txResponse.transactionHash,
+    };
+  }
+}
+
+// enter key of what to extract
+function extractFromRawLog(rawLog: string, key: string): string {
```
jayantk, line 229: is the response json? would we be better off parsing it than searching it with a regex?

you don't have to implement that now, but could be worth a TODO or a code comment for future reference.

target_chains/cosmwasm/tools/src/chains-manager/injective.ts
```diff
@@ -0,0 +1,231 @@
...
+    readonly mnemonic: string
+  ) {
+    this.wallet = PrivateKey.fromMnemonic(mnemonic);
+  }
+
+  private getAddress(): string {
+    return this.wallet.toBech32();
+  }
+
+  private async signAndBroadcastMsg(msg: Msgs): Promise<TxResponse> {
```
jayantk, line 47: this method pretty much exists in the price pusher already, right? is there a way that we can share code between the deployer here and the pusher? Seems like the generic cosmwasm capabilities will be useful in the pusher as well.

(again, feel free to leave a TODO)

target_chains/cosmwasm/tools/src/chains-manager/injective.ts
```diff
@@ -0,0 +1,231 @@
...
+    const txResponse = await this.signAndBroadcastMsg(updateAdminMsg);
+
+    return {
+      txHash: txResponse.txHash,
+    };
+  }
+}
+
+// enter key of what to extract
+function extractFromRawLog(rawLog: string, key: string): string {
```
jayantk, line 220: this is function exists twice in this PR (and also exists in the pusher code i presume). I think you should reuse a single implementation in this PR at least, and then we should think about deduplicating across the pusher as well.

target_chains/cosmwasm/tools/src/pyth-wrapper.ts
```diff
@@ -0,0 +1,118 @@
...
+    } catch (e) {
+      throw new Error("Error fetching update fee");
+    }
+  }
+}
+
+export type Fund = {
+  denom: string;
+  amount: string;
+};
```
jayantk, line 74: I think the combination of an amount and denom is usually called a `Coin` in cosmos?

target_chains/cosmwasm/tools/src/pyth-wrapper.ts
```diff
@@ -0,0 +1,118 @@
...
+  publish_time: number;
+};
+
+export type PriceFeedResponse = {
+  id: string;
+  price: Price;
+  ema_price: Price;
+};
+
+export class PythWrapperQuerier {
```
jayantk, line 25: document these classes too. doesn't have to be long, but needs something.

target_chains/cosmwasm/tools/src/pyth-wrapper.ts
```diff
@@ -0,0 +1,118 @@
...
+  constructor(private chainQuerier: ChainQuerier) {}
+
+  // get update fee
```
jayantk, line 28: this comment isn't very helpful -- i get the same information from the method name itself. You should document what it does and how to call it. E.g.:

"Get the fee required to update the Pyth on-chain contract with the given VAAs. This function will invoke the corresponding function on the contract itself and return the result. The result is represented as both a quantity of tokens and a denomination which may vary depending on the chain being queried."

target_chains/cosmwasm/tools/src/pyth-wrapper.ts
jayantk: also (1) this type seems unused and (2) it's a duplicate of UpdateFeeResponse

jayantk (commented):
I think the code here makes sense. My main ask is to document things so it's easier for people to understand in the future. I also note there's some code duplication between this and the pusher, and it would be good to have a plan to eliminate that. You don't have to eliminate the duplication in this PR, but we should in the future.

I would like to review this again after you've added documentation.

target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
0xfirefist: When parsing we would need to know the complete structure of the object which I don't think we need to know. 
For regex parsing we just need to know that particular part how it looks like.

target_chains/cosmwasm/tools/src/chains-manager/injective.ts
0xfirefist: it doesn't exist in price pusher. 

In this pr, this was a bug. both implementations are different at a very small point.

target_chains/cosmwasm/tools/src/pyth-wrapper.ts
0xfirefist: removing it

target_chains/cosmwasm/tools/src/chains-manager/chain-querier.ts
```diff
@@ -0,0 +1,162 @@
...
+  getRawContractState(req: RawContractStateRequest): Promise<Object>;
+
+  /**
+   * Contracts local storage on chain is structured as key-value pairs.
+   * `getAllContractState` query the local storage of a contract for all such pairs.
+   * @param {AllContractStateRequest} req
+   * @param {string} req.contractAddr
+   * @returns {Object} - It returns an object. As the response is contract dependent we can't have a structure for it.
+   *
+   * @throws an error if it fails.
```
jayantk, line 63: this `@throws` text isn't very helpful -- it doesn't tell me the conditions under which it would fail, or how I should handle errors. I would either omit it entirely (if the error cases aren't particularly important and shouldn't happen), or describe the errors that can happen in a little more detail. If all of the methods on this interface can fail for roughly the same conditions (e.g., your network connection is out), you can document the error cases once in the class docstring and omit it on the functions.

target_chains/cosmwasm/tools/src/chains-manager/injective.ts
jayantk: ooh they are different... ok nevermind. 

(I think you could extract out a helper function with some of the shared logic (and just call that helper with a different regexp here), but not a big deal)

jayantk (approved):
nice! I generally like the documentation you've written. I left you some comments on the docs themselves. Feel free to ignore them -- what you have now is pretty good, and there's no need to spend a lot of time on this if you don't want to. Take them more as tips & tricks for writing documentation in the future.

PR #741, opened 2023-04-03
[accumulator-updater 3/x] Update Inputs Ix

swimricky:
### Summary
Implement `mock-cpi-program::update_price` & `accumuulator-updater::update_inputs` ixs

### Updates

- refactor state & ixs in both programs into their own modules
- implement update_price & update_inputs ixs

### TODOs
- [x] combine `create_inputs` & `update_inputs` ix into one (see Notes below)
- [ ] Rust client sdk/examples 
- [ ] delete_inputs (See Notes below)
- [ ]  update `invoke_cpi_solana` to use `instruction::new_with_borsh` for `instruction_data` serialization

### Testing
cd accumulator_updater && anchor test

### Notes
- for `delete_inputs` do we need to handle deleting inputs outside of a CPI call? one use-case would be if an accumulator message schema was being deprecated/no longer used


accumulator_updater/programs/accumulator_updater/src/instructions/create_inputs.rs
swimricky, line 1: refactor only

accumulator_updater/programs/accumulator_updater/src/instructions/update_inputs.rs
swimricky, line 1: new update_input logic

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
swimricky, line 1: refactor of AccumulatorInput & AccumulatorHeader account state

accumulator_updater/programs/accumulator_updater/src/state/whitelist.rs
swimricky, line 1: refactor of whitelist related logic

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
swimricky, line 1: refactor of mock add_price

accumulator_updater/programs/mock-cpi-caller/src/instructions/mod.rs
swimricky, line 1: refactor

accumulator_updater/programs/mock-cpi-caller/src/state/price.rs
swimricky, line 1: refactor besides `PriceAccount.update()`

accumulator_updater/programs/accumulator_updater/src/instructions/create_inputs.rs
```diff
@@ -0,0 +1,135 @@
...
+            self,
+            CreateAccount,
+        },
+    },
+};
+
+
+pub fn create_inputs<'info>(
+    ctx: Context<'_, '_, '_, 'info, CreateInputs<'info>>,
+    base_account: Pubkey,
```
guibescos, line 20: Is this base account field necesary?

guibescos: Why not use the caller program as the base

swimricky: the caller program is included as part of the seeds for the PDAs that are created. We need the `base_account` to distinguish between the PDAs generated for the same account type/schema e.g.  `PriceFullMessage` for BTC/USD vs ETH/USD. This could be something other than the `base_account` pubkey but this seemed to be the most natural fit.

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+    },
+};
+
+
+pub fn emit_inputs<'info>(
+    ctx: Context<'_, '_, '_, 'info, EmitInputs<'info>>,
+    base_account: Pubkey,
+    data: Vec<Vec<u8>>,
+    account_type: u32,
+    account_schemas: Vec<u8>,
```
jayantk, line 23: I think it would be nice if the interface of this function resembled a commonly-used data structure so that callers had a natural reference point for what to expect. I think the accumulator is most similar to a `Map` where the keys are `(base_account, account_schema)` and the values are arbitrary data. Given that, I would try to make the arguments map on to the function signature for inserting multiple items in a Map. E.g., I think a function signature like

```
/// Insert elements into the Map identified by the calling program and map_id.
/// Each element is written into a separate account (explain how to derive the PDA). 
/// The caller of this instruction must pass those PDAs while calling this function in the same order as elements. 
put_all(ctx, map_id: Pubkey, elements: <Vec<(u8, Vec<u8>)>>)
```

would make more sense. (map_id is base_account)

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+        },
+    },
+};
+
+
+pub fn emit_inputs<'info>(
+    ctx: Context<'_, '_, '_, 'info, EmitInputs<'info>>,
+    base_account: Pubkey,
+    data: Vec<Vec<u8>>,
+    account_type: u32,
```
jayantk, line 22: do we need account_type at all? I don't remember exactly but i'm looking back at our previous convo and it's not mentioned there:

https://github.com/pyth-network/pyth-crosschain/pull/712#discussion_r1146392083

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+            let seeds = accumulator_acc_seeds!(cpi_caller, base_account, account_schema);
+            let (pda, bump) = Pubkey::find_program_address(seeds, &crate::ID);
+            require_keys_eq!(ai.key(), pda);
+
+            //TODO: Update this with serialization logic
+            let accumulator_size = 8 + AccumulatorInput::get_initial_size(&account_data);
+            let accumulator_input = AccumulatorInput::new(
+                AccumulatorHeader::new(
+                    bump,
+                    1, //from CPI caller?
```
jayantk, line 46: I think this program controls the header format (and therefore gets to set the version)

I would suggest extracting this value out to a constant like `AccumulatorHeader::CURRENT_VERSION` though so it's easy to bump all the places where it may be used.

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+                    cpi_caller,
+                    base_account,
+                    account_type,
+                    account_schema,
+                )?;
+            }
+
+            //TODO: update header?
+            accumulator_input.data = account_data;
+            accumulator_input.persist(ai)?;
```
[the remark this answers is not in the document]
jayantk: fwiw i'm pretty sure we have code somewhere for doing the create/update w/ resizing if needed in some solana program somewhere. @drozdziak1 or @guibescos will be able to point you to it.

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+    let cpi_caller = ctx.accounts.whitelist_verifier.is_allowed()?;
+    let accts = ctx.remaining_accounts;
+    require_eq!(accts.len(), data.len());
+    require_eq!(data.len(), account_schemas.len());
+    let mut zip = data.into_iter().zip(account_schemas.into_iter());
+
+    let rent = Rent::get()?;
+    let (mut initialized, mut updated) = (vec![], vec![]);
+    for ai in accts {
+        let (account_data, account_schema) = zip.next().unwrap();
```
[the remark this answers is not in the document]
jayantk: also kind of seems like it would be better to zip all three arrays ?

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        if is_uninitialized_account(ai) {
+            let seeds = accumulator_acc_seeds!(cpi_caller, base_account, account_schema);
+            let (pda, bump) = Pubkey::find_program_address(seeds, &crate::ID);
+            require_keys_eq!(ai.key(), pda);
+
+            //TODO: Update this with serialization logic
```
jayantk, line 41: i don't know what this todo means

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -0,0 +1,84 @@
...
+impl AccumulatorInput {
+    pub fn get_initial_size(data: &Vec<u8>) -> usize {
+        AccumulatorHeader::SIZE + 4 + data.len()
+    }
+
+    pub fn new(header: AccumulatorHeader, data: Vec<u8>) -> Self {
+        Self { header, data }
+    }
+
+    pub fn validate_account_info(
```
jayantk, line 25: this doesn't seem like the right encapsulation of functionality -- it's a very specific function that can only be used in one place. I think a better function would be something like `derive_pda_for(cpi_caller, base_account, account_schema)` ?

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+                    account_schema,
+                ),
+                account_data,
+            );
+            EmitInputs::init_accumulator_input(
+                ai,
+                accumulator_input,
+                accumulator_size,
+                &ctx.accounts.payer,
+                &[accumulator_acc_seeds_with_bump!(
```
jayantk, line 57: i'm a little confused about the use of this function w/ the bump vs. the one w/o the bump. It seems like you want there to be a single PDA for each key, right? if so, isn't the bump deterministically generated, so there shouldn't be a reason to save it? (is this an efficiency improvement? if so, does it not introduce security holes because there are now multiple PDAs that can pass the checks for the same key?)

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
```diff
@@ -0,0 +1,129 @@
...
+        account_data.push(price_compact_data);
+    }
+
+
+    let account_schemas = schemas.into_iter().map(|s| s.to_u8()).collect::<Vec<u8>>();
+
+    // 44444 compute units
+    // AddPrice::invoke_cpi_anchor(ctx, account_data, PythAccountType::Price, account_schemas)
+    // 44045 compute units
+    AddPrice::emit_accumulator_inputs(ctx, account_data, PythAccountType::Price, account_schemas)
```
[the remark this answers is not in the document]
jayantk: oh, I think this question may come down to a misunderstanding of how the oracle program works. In the oracle program, when you add a price account, it doesn't actually contain any price information. it simply allocates the space for putting prices in later. So there wouldn't even be any data to emit. (whereas this version of add price actually puts some data in the price account).

anyway, since this is just a test program, so I would scope it down to the minimum necessary code to demonstrate how to invoke the accumulators & test all the needed functionality. pretty sure you don't need both add price and update price to do that.

jayantk (approved):
nice! I left you a bunch of minor comments and questions, but overall seems like a step in the right direction.

accumulator_updater/programs/accumulator_updater/src/instructions/create_inputs.rs
guibescos: ah got it

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
guibescos: I second this

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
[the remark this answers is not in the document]
guibescos: I removed the resizing logic from the oracle but it's here https://github.com/pyth-network/pyth-client/commit/90ba9972e5f61fb094e1bcc609c3f358c51bca74#diff-6b3dac646ee82aa19e2da0df982aa54c40e2005422fa4fd5805d1b5e840d3ca5L86 . It's pretty straightforward with realloc

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub fn add_price<'info>(
+    ctx: Context<'_, '_, '_, 'info, AddPrice<'info>>,
+    params: AddPriceParams,
+) -> Result<()> {
+    let mut account_data: Vec<Vec<u8>> = vec![];
+    let schemas = get_schemas(PythAccountType::Price);
+
+    {
+        let pyth_price_acct = &mut ctx.accounts.pyth_price_account.load_init()?;
```
guibescos, line 32: I don't think you have to use load_init, the account got initialized in the context already

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
swimricky: this was for flexibility if we ever wanted to include different types of accounts into the accumulator aside from the price accounts - maybe even if it was something outside of pyth (i don't have any specific use-case in mind and not sure if we would ever want product or mapping accounts but I can remove this if we really don't need this)

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
swimricky: there's the "canonical bump" which is the first bump counting down from 255 that provides a valid PDA address (off the ed25519 curve). This is the bump that's returned with `find_program_address`. But there could technically be multiple bumps that create valid PDA addresses. 

There's 2 ways to verify a PDA address:
1. assume you're always using the canonical bump and check that `Pubkey::find_program_address(&[seeds]).0 == account.key()`
2. store the bump in the account and check`account.key() == Pubkey::create_program_address(&[seeds, &[account.bump]])` 

2 is more efficient (marginally) since 1 could theoretically have to do a guess-and-check multiple times to find the first valid seed. 

Regarding your concern about security holes - since only a program can create PDAs it owns, it shouldn't be possible to have multiple PDAs using the same seeds but different bumps. The accumulator-updater program will be the only program to create these PDAs and it's using the canonical bumps when the PDAs are first initialized.

Having 2 macros is a bit confusing so i'll remove the non-bump one since that's only ever used once

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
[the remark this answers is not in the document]
swimricky: oh got it thanks for clarifying that. i'll update this and the relevant tests/code and scope it down

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
swimricky: for PDAs that use  `zero-copy`/`AccountLoader` it still needs to be called with `load_init()` otherwise the discriminator isn't set. 
This part of the code isn't that important though since it's just mocking pyth price accounts

I used the anchor test code for `CreateBar` from here as a template - https://github.com/coral-xyz/anchor/blob/master/tests/zero-copy/programs/zero-copy/src/lib.rs#L95

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
swimricky: will remove for now

accumulator_updater/programs/accumulator_updater/src/instructions/emit_inputs.rs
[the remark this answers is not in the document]
swimricky: will handle the re-sizing in a separate PR

accumulator_updater/programs/mock-cpi-caller/src/instructions/add_price.rs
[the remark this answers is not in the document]
swimricky: I left the `add_price` & cpi call in for now but left a comment describing the actual behavior in the oracle program.

PR #745, opened 2023-04-05
[multisig_wh_message_builder] separate command tool and reusable fns

0xfirefist:
`index.ts` had the command line code, configs and a lot of methods. 
If one imports anything from `index.ts`, it runs the commandline code too.

Hence we need to separate commandline code and reusable code. So that anyone can consume the configs or methods without triggering the commandline code. 

I moved all the commandline specific code to `commander.ts` file. No file should import anything from this file. I made sure of that by removing all the `export` from it. Added a note to the file too.

The reusable methods, configs are moved to `helper.ts` for now.


governance/multisig_wh_message_builder/src/commander.ts
```diff
@@ -1,36 +1,39 @@
-import { ixFromRust, setDefaultWasm } from "@certusone/wormhole-sdk";
```
guibescos, line 1: Should this file keep the name index.ts

0xfirefist: I was planning to add two method for now in index.ts, createMultisigTx and executeMultisigTx [not final names]. and export them.

Anyone consuming them would do so using 
`import {createMultisigTx, executeMultisigTx} from "@pythnet/package_name";`

and not something like
`import {createMultisigTx, executeMultisigTx} from "@pythnet/package_name/file_name";`

we can leave commander as it is for no one will be using that file.
Does this make sense?

jayantk: I think `cli.ts` would be a good name for this file

governance/multisig_wh_message_builder/src/helper.ts
```diff
@@ -0,0 +1,545 @@
+import { ixFromRust } from "@certusone/wormhole-sdk";
```
jayantk, line 1: very minor nitpick: ^ i think this filename should be plural if it has lots of different helper functions in it (vs. a single thing called a helper). `helpers.ts` ?

jayantk (approved):
code itself looks good to me. I think you should probably catch up with @guibescos on the state of contract governance & clis though and make sure that modifying this code is the right thing to do. It's possible we should be making the changes to the other CLI that guillermo mentioned.

0xfirefist:
Closing this PR. Will add the required code to xc-admin instead of this package.

PR #746, opened 2023-04-06
[hermes] Add more rest api methods

ali-bahjati:
Add many of the price service apis. Per David suggestion, we do validation in parsing instead of doing it later. I didn't find any suitable library to deserialize our hex format so I created a macro to implement it because we use it in a couple of places. I tried making a generic HexInput but couldn't make it working (and I need other crates like generic_array for it which makes the code more complex)

### DIFF
--- a/hermes/src/network/rpc/rest.rs
+++ b/hermes/src/network/rpc/rest.rs
@@ -31,6 +31,7 @@ use {
 
 #[derive(Debug, Clone, Deref, DerefMut)]
 pub struct PriceIdInput([u8; 32]);
+// TODO: Use const generics instead of macro.
 impl_deserialize_for_hex_string_wrapper!(PriceIdInput, 32);
 
 impl From<PriceIdInput> for PriceIdentifier {
@@ -159,7 +160,7 @@ pub struct GetVaaCcipQueryParams {
 
 #[derive(Debug, serde::Serialize)]
 pub struct GetVaaCcipResponse {
-    data: String,
+    data: String, // TODO: Use a typed wrapper for the hex output with leading 0x.
 }
 
 pub async fn get_vaa_ccip(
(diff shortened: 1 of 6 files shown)


hermes/src/network/rpc/rest.rs
```diff
@@ -107,6 +102,91 @@ pub async fn latest_price_feeds(
...
+impl_deserialize_for_hex_string_wrapper!(GetVaaCcipInput, 40);
+
+#[derive(Debug, serde::Deserialize)]
+pub struct GetVaaCcipQueryParams {
+    data: GetVaaCcipInput,
+}
+
+#[derive(Debug, serde::Serialize)]
+pub struct GetVaaCcipResponse {
+    data: String,
```
Reisen, line 162: TODO: Add type.

hermes/src/network/rpc/rest.rs
```diff
@@ -24,47 +15,56 @@ use {
...
+    },
+    pyth_sdk::{
+        PriceFeed,
+        PriceIdentifier,
+    },
 };
 
+#[derive(Debug, Clone, Deref, DerefMut)]
+pub struct PriceIdInput([u8; 32]);
+impl_deserialize_for_hex_string_wrapper!(PriceIdInput, 32);
```
Reisen, line 34: May be better to replace with `struct Hex` with const generics.

hermes/src/network/rpc/rest.rs
```diff
@@ -107,6 +102,91 @@ pub async fn latest_price_feeds(
...
+
+    let price_feeds_with_update_data = state
+        .store
+        .get_price_feeds_with_update_data(vec![price_id], RequestTime::FirstAfter(publish_time))
+        .map_err(|_| RestError::UpdateDataNotFound)?;
+
+    let vaa = price_feeds_with_update_data
+        .update_data
+        .batch_vaa
+        .get(0)
```
Reisen, line 180: Add comment explaining why `get(0)` is safe, and the semantics behind the batch structure.

Reisen (changes requested):

Reisen (approved):

hermes/src/network/rpc/rest.rs
ali-bahjati: I will address this in a later PR. Added a todo for it.

PR #747, opened 2023-04-06
[hermes] add get price feed ids + refactor

ali-bahjati:
This change touches many of our abstractions and I like to know your opinion on how we can do it better. Essentially we won't have a single way (across accumulation ways) to say what are the price ids and the solution seems to be taking the union of all accumulation states. The reason is that the prices are derived from the accumulations not the other way. 

This PR also adds a CcipUpdateNotFound error to handle the special status code situation.

### DIFF
--- a/hermes/src/store/proof/batch_vaa.rs
+++ b/hermes/src/store/proof/batch_vaa.rs
@@ -58,7 +58,7 @@ pub fn store_vaa_update(state: State, vaa_bytes: Vec<u8>) -> Result<()> {
             publish_time,
         };
 
-        let key = Key::new(price_feed.id.to_bytes().to_vec());
+        let key = Key::BatchVaa(price_feed.id);
         state.insert(key, publish_time, StorageData::BatchVaa(price_info))?;
     }
     Ok(())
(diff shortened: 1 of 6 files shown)


hermes/src/store/proof/batch_vaa.rs
```diff
@@ -97,6 +96,18 @@ pub fn get_price_feeds_with_update_data(
 }
 
 
+pub fn get_price_feed_ids(state: State) -> Vec<PriceIdentifier> {
```
Reisen, line 99: This could be cleaner with `try_map` which can collapse errors thrown during iteration.

Reisen (approved):

hermes/src/store/proof/batch_vaa.rs
```diff
@@ -97,6 +96,18 @@ pub fn get_price_feeds_with_update_data(
...
+    let mut price_ids = Vec::new();
+    for key in state.keys() {
+        let maybe_32_bytes: Result<[u8; 32], _> = ((*key).clone()).try_into();
```
Reisen, line 102: This really implies that the key should be an `enum`, matching on length to distinguish is an implicit/unsafe version of the same thing, so we me as well type it.

PR #751, opened 2023-04-07
[eth] Add Canto mainnet

ali-bahjati:
This deployment also enables contract rewards on canto. Since the changed code is only specific to the Canto network and is used only in initialization, only the diff is added as a file.

This deployment also registers the Wormhole Receiver contract to collect fees (to the Pyth contract address). Since Wormhole registration happens in Pyth init after its own registration, the init fails if it cannot register Wormhole. Also no one can reregister Wormhole to any other receiver because the `assign/register` method can be called only once per contract.


target_chains/ethereum/contracts/canto-deployment-patch.diff
```diff
@@ -0,0 +1,64 @@
...
+diff --git a/target_chains/ethereum/contracts/contracts/wormhole/interfaces/IWormhole.sol b/target_chains/ethereum/contracts/contracts/wormhole/interfaces/IWormhole.sol
+index 69d60a62..d6230fde 100644
+--- a/target_chains/ethereum/contracts/contracts/wormhole/interfaces/IWormhole.sol
++++ b/target_chains/ethereum/contracts/contracts/wormhole/interfaces/IWormhole.sol
+@@ -62,4 +62,6 @@ interface IWormhole is Structs {
+     function governanceContract() external view returns (bytes32);
+
+     function messageFee() external view returns (uint256);
++
++    function assignCsr(uint256 tokenId) external;
```
jayantk, line 63: add the .DS_Store files below to .gitignore (and remove from this pr)

jayantk (commented):
let me make sure I understand how this is going to work. You've deployed it once with this registration logic, so the CSR rewards for both contracts are going to get stored in the PythUpgradable. That's the same address that holds the existing fees, so we should be able to handle the rewards in a uniform way. 

Furthermore, future upgrades to the canto contract do not need to apply this diff, as we've already registered once, and that registration will still apply post-upgrade?

Fwiw I think it would be good to document all of these network-specific assumptions and logic somewhere in this repo. (e.g., we also had some for zksync right?)

ali-bahjati:
Yeah your understanding is correct. We don’t need them in the upgrades of Pyth because the address won’t change. However, we might need it if we redeploy another Wormhole to a new address (which is very unlikely)

I have zksync stuff in the Deploying.md already. Added this one. Can you take a look? @jayantk

jayantk (approved):
🙇

PR #756, opened 2023-04-10
Gas usage experiments

jayantk:
We wanted to benchmark the gas usage of different ways of verifying prices in the ethereum contract. I tried out a bunch of different things and here's what I get:

| Update Type                                   | Gas Cost |
| --------------------------------------------- | ------- |
| Native update (use tx signature of 1 price update)   | 48k     |
| Threshold update (sign 1 price update directly)                              | 54.5k   |
| Threshold update for 1 price w/ 8-deep merkle tree         | 58k     |
| WH update for 1 price w/ 8-deep merkle tree                | 201k    |
| WH batch update w/ 5 prices                    | 222k    |

The two things that I think are worth pointing out are:
1. The merkle tree doesn't add that much overhead over the native update, which represents the lower bound of what's achievable. That's good, because it means our current path will end up in a near-optimal state once everything is done.
2. We do get a minor gas improvement from the merkle tree update itself.

I think our current path is reasonable given these results, so that's good. No need to change anything.

### DIFF
--- a/target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
+++ b/target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
@@ -15,7 +15,7 @@ import "./utils/WormholeTestUtils.t.sol";
 import "./utils/PythTestUtils.t.sol";
 
 // Experiments to measure the gas usage of different ways of verifying prices in the EVM contract.
-contract GasUsageExperiments is Test, WormholeTestUtils, PythTestUtils {
+contract VerificationExperiments is Test, WormholeTestUtils, PythTestUtils {
     // 19, current mainnet number of guardians, is used to have gas estimates
     // close to our mainnet transactions.
     uint8 constant NUM_GUARDIANS = 19;
--- a/target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
+++ b/target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
@@ -13,9 +13,15 @@ import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
 import "../contracts/pyth/PythInternalStructs.sol";
 import "./utils/WormholeTestUtils.t.sol";
 import "./utils/PythTestUtils.t.sol";
+import "./utils/RandTestUtils.t.sol";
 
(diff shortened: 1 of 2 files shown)


target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
```diff
@@ -0,0 +1,600 @@
...
+            freshPrices[0],
+            8
+        );
+
+        thresholdUpdate = generateThresholdUpdate(priceIds[0], freshPrices[0]);
+
+        nativeUpdate = generateAttestationPayload(priceIds[0], freshPrices[0]);
+    }
+
+    function getRand() internal returns (uint val) {
```
ali-bahjati, line 179: We have `RandTestUtils` for this.

target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
```diff
@@ -0,0 +1,600 @@
...
+            emit PriceFeedUpdate(
+                priceId,
+                info.publishTime,
+                info.price,
+                info.conf
+            );
+        }
+    }
+
+    // Update a single price feed via a "native" price update (i.e., using the default ethereum tx signature for authentication).
```
ali-bahjati, line 499: nitpik: i think for this one you need to check the message sender to have a fixed address.

target_chains/ethereum/contracts/forge-test/VerificationExperiments.t.sol
```diff
@@ -0,0 +1,600 @@
...
+        address signer
+    ) public pure returns (bool) {
+        bytes32 r;
+        bytes32 s;
+        uint8 v;
+        assembly {
+            r := mload(add(signature, 32))
+            s := mload(add(signature, 64))
+            v := byte(0, mload(add(signature, 96)))
+        }
```
ali-bahjati, line 536: Nitpik: I think we can do it easily without this using the byteslib that we have. I prefer not to have assembly code here.

ali-bahjati (approved):
This is great! I left some minor comments that would be good to be addressed before merging it. 

The result seems reasonable to me. Just the thing to keep in mind is that it doesn't have the initialization cost which has a variable gas cost of ~16*inputDataLength and this might put more gas difference between Merkle and without Merkle.

PR #760, opened 2023-04-11
[accumulator-updater 5/x] Single PDA

swimricky:
### Summary
Write all schemas/messages for an account into one PDA

### Updates
- update `AccumulatorInput` and related structs to `zero_copy`
- add `InputIndex` for indexing into `AccumulatorInput.data`
### TODOs
- funding account
- realloc (need wrapper struct in increments of 10240 bytes?)

### DIFF
--- a/accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
+++ b/accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
@@ -12,17 +12,19 @@ use {
     },
 };
 
+// NOte: get rid of this. not using schema.
 #[derive(AnchorSerialize, AnchorDeserialize)]
 pub struct InputSchemaAndData {
     pub schema: u8,
     pub data:   Vec<u8>,
 }
 
+pub const ACCUMULATOR: &[u8; 11] = b"accumulator";
 
 pub fn put_all<'info>(
     ctx: Context<'_, '_, '_, 'info, PutAll<'info>>,
     base_account_key: Pubkey,
-    values: Vec<InputSchemaAndData>,
+    values: Vec<Vec<u8>>,
 ) -> Result<()> {
     let cpi_caller = ctx.accounts.whitelist_verifier.is_allowed()?;
     let account_infos = ctx.remaining_accounts;
@@ -39,7 +41,7 @@ pub fn put_all<'info>(
             let (pda, bump) = Pubkey::find_program_address(
                 &[
                     cpi_caller.as_ref(),
-                    b"accumulator".as_ref(),
+                    ACCUMULATOR.as_ref(),
                     base_account_key.as_ref(),
                 ],
                 &crate::ID,
@@ -47,7 +49,7 @@ pub fn put_all<'info>(
             require_keys_eq!(accumulator_input_ai.key(), pda);
             let signer_seeds = &[
                 cpi_caller.as_ref(),
-                b"accumulator".as_ref(),
+                ACCUMULATOR.as_ref(),
                 base_account_key.as_ref(),
                 &[bump],
             ];
@@ -68,10 +70,12 @@ pub fn put_all<'info>(
         } else {
             loader = AccountLoader::<AccumulatorInput>::try_from(accumulator_input_ai)?;
             let mut accumulator_input = loader.load_mut()?;
-            accumulator_input.validate(accumulator_input_ai.key(), cpi_caller, base_account_key)?;
             accumulator_input.header.set_version();
             accumulator_input
         });
+        // note: redundant for uninitialized code path but safer to check here.
+        // compute budget cost should be minimal
+        accumulator_input.validate(accumulator_input_ai.key(), cpi_caller, base_account_key)?;
         accumulator_input.put_all(values)?;
     }
 
--- a/accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
+++ b/accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
@@ -26,85 +26,58 @@ pub fn put_all<'info>(
 ) -> Result<()> {
     let cpi_caller = ctx.accounts.whitelist_verifier.is_allowed()?;
     let account_infos = ctx.remaining_accounts;
-    require_eq!(account_infos.len(), values.len());
+    let accumulator_input_ai = account_infos
+        .first()
+        .ok_or(AccumulatorUpdaterError::AccumulatorInputNotProvided)?;
 
-    let rent = Rent::get()?;
-    let (mut initialized, mut updated) = (vec![], vec![]);
+    let _rent = Rent::get()?;
 
-    for (
-        ai,
-        InputSchemaAndData {
-            schema: account_schema,
-            data: account_data,
-        },
-    ) in account_infos.iter().zip(values)
+    let loader;
+
     {
-        let bump = if is_uninitialized_account(ai) {
-            let seeds = &[
-                cpi_caller.as_ref(),
-                b"accumulator".as_ref(),
-                base_account_key.as_ref(),
-                &account_schema.to_le_bytes(),
-            ];
-            let (pda, bump) = Pubkey::find_program_address(seeds, &crate::ID);
-            require_keys_eq!(ai.key(), pda);
-
-
-            //TODO: Update this with serialization logic
-            // 8 for anchor discriminator
-            let accumulator_size = 8 + AccumulatorInput::size(&account_data);
-            PutAll::create_account(
-                ai,
-                accumulator_size,
-                &ctx.accounts.payer,
+        let accumulator_input = &mut (if is_uninitialized_account(accumulator_input_ai) {
+            let (pda, bump) = Pubkey::find_program_address(
                 &[
                     cpi_caller.as_ref(),
                     b"accumulator".as_ref(),
                     base_account_key.as_ref(),
-                    &account_schema.to_le_bytes(),
-                    &[bump],
                 ],
-                &rent,
+                &crate::ID,
+            );
+            require_keys_eq!(accumulator_input_ai.key(), pda);
+            let signer_seeds = &[
+                cpi_caller.as_ref(),
+                b"accumulator".as_ref(),
+                base_account_key.as_ref(),
+                &[bump],
+            ];
+            PutAll::create_account(
+                accumulator_input_ai,
+                8 + AccumulatorInput::INIT_SPACE,
+                &ctx.accounts.payer,
+                signer_seeds,
                 &ctx.accounts.system_program,
             )?;
-            initialized.push(ai.key());
-
-            bump
-        } else {
-            let accumulator_input = <AccumulatorInput as AccountDeserialize>::try_deserialize(
-                &mut &**ai.try_borrow_mut_data()?,
+            loader = AccountLoader::<AccumulatorInput>::try_from_unchecked(
+                &crate::ID,
+                accumulator_input_ai,
             )?;
-            {
-                // TODO: allow re-sizing?
-                require_gte!(
-                    accumulator_input.data.len(),
-                    account_data.len(),
-                    AccumulatorUpdaterError::CurrentDataLengthExceeded
-                );
-
-                accumulator_input.validate(
-                    ai.key(),
-                    cpi_caller,
-                    base_account_key,
-                    account_schema,
-                )?;
-            }
-
-
-            updated.push(ai.key());
-            accumulator_input.header.bump
-        };
-
-        let accumulator_input =
-            AccumulatorInput::new(AccumulatorHeader::new(bump, account_schema), account_data);
-        accumulator_input.persist(ai)?;
+            let mut accumulator_input = loader.load_init()?;
+            accumulator_input.header = AccumulatorHeader::new(bump);
+            accumulator_input
+        } else {
+            loader = AccountLoader::<AccumulatorInput>::try_from(accumulator_input_ai)?;
+            let mut accumulator_input = loader.load_mut()?;
+            accumulator_input.validate(accumulator_input_ai.key(), cpi_caller, base_account_key)?;
+            accumulator_input.header.set_version();
+            accumulator_input
+        });
+        accumulator_input.put_all(values)?;
     }
 
-    msg!(
-        "[emit-updates]: initialized: {:?}, updated: {:?}",
-        initialized,
-        updated
-    );
+
+    loader.exit(&crate::ID)?;
+
     Ok(())
 }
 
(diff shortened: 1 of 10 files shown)


accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
```diff
@@ -26,85 +26,58 @@ pub fn put_all<'info>(
...
-        },
-    ) in account_infos.iter().zip(values)
     {
-        let bump = if is_uninitialized_account(ai) {
-            let seeds = &[
+        let accumulator_input = &mut (if is_uninitialized_account(accumulator_input_ai) {
+            let (pda, bump) = Pubkey::find_program_address(
+                &[
+                    cpi_caller.as_ref(),
+                    b"accumulator".as_ref(),
```
jayantk, line 42: suggest extracting a string constant for this

accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
```diff
@@ -26,85 +26,58 @@ pub fn put_all<'info>(
...
-                )?;
-            }
-
+            let mut accumulator_input = loader.load_init()?;
+            accumulator_input.header = AccumulatorHeader::new(bump);
+            accumulator_input
+        } else {
+            loader = AccountLoader::<AccumulatorInput>::try_from(accumulator_input_ai)?;
+            let mut accumulator_input = loader.load_mut()?;
+            accumulator_input.validate(accumulator_input_ai.key(), cpi_caller, base_account_key)?;
```
jayantk, line 71: I would suggest putting this `validate` call outside the if/else right above `put_all`, as I think that reduces the chance for bugs. (It's redundant for the first codepath, but it makes it easier to reason that the code performs the necessary security checks.)

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +15,105 @@ use {
...
+        self.version = Self::CURRENT_VERSION;
+    }
+}
+
+#[zero_copy]
+#[derive(Debug, InitSpace)]
+pub struct InputIndex {
+    pub offset:   u32,
+    pub len:      u16,
+    pub unused_0: u16,
```
jayantk, line 64: having both offset and len seems redundant? I think an `[u16]` would suffice for the `indexes` field above, where the assumptions are that the indexes occur in sorted order so you can subtract adjacent ones to get the length. e.g., if the array was like [0, 10, 12] you would have a 10 byte message at 0, then a 2 byte message at 10. 

(this would also reserve more space for actual messages, which would be good. Not great that 20% of the space goes to indexes right now)

accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
```diff
@@ -26,85 +26,58 @@ pub fn put_all<'info>(
 ) -> Result<()> {
```
jayantk, line 26: this method needs some very clear documentation about what it does. Does it overwrite the data in the accumulator input account? What happens if two callers invoke it on the same slot? 

also we had that conversation about being tolerant of an insufficient account size, which also should be documented here. I think callers who want to add new values to an existing list *must do so at the end of the list*

accumulator_updater/tests/accumulator_updater.ts
```diff
@@ -322,16 +304,32 @@ function parseAccumulatorInput({
...
+    if (inputIndex.len == 0) {
+      continue;
+    }
+
+    const inputData = dataBuffer.subarray(
+      inputIndex.offset,
+      inputIndex.offset + inputIndex.len
+    );
+    if (i == 0) {
+      messages.push(parseFullPriceMessage(inputData));
```
jayantk, line 324: is this temporary (presumably the messages should themselves encode what type of data they are)? maybe worth a comment

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +15,105 @@ use {
...
+
+    pub fn new(bump: u8) -> Self {
+        let header = AccumulatorHeader::new(bump);
+        Self {
+            header,
+            data: [0u8; 8_176],
+        }
+    }
+
+    pub fn put_all(&mut self, values: Vec<InputSchemaAndData>) -> Result<()> {
```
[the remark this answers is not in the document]
jayantk: (this also suggests that your test suite doesn't have enough cases)

jayantk (commented):
cool. I think you're on the right track here, though I think there are a couple bugs in the implementation

accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
swimricky: just to be clear, adding new values to the end of the list should be the responsibility of the caller but you just want to make sure it's documented here correct?

I added more documentation on `put_all` in `lib.rs`

also i've added an implementation for now which will write as much as possible and then do a log if it wasn't able to write everything.

accumulator_updater/tests/accumulator_updater.ts
swimricky: i've updated this thanks!

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +15,105 @@ use {
...
-    pub fn new(header: AccumulatorHeader, data: Vec<u8>) -> Self {
-        Self { header, data }
+    pub fn size(&self) -> usize {
+        AccumulatorHeader::INIT_SPACE + 4 + self.data.len()
     }
 
-    pub fn update(&mut self, new_data: Vec<u8>) {
-        self.header = AccumulatorHeader::new(self.header.bump, self.header.account_schema);
-        self.data = new_data;
+    pub fn offset(&self) -> u32 {
```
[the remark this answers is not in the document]
swimricky: Removed this since no longer used with the new end_offset implementation

accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -52,24 +52,35 @@ pub mod accumulator_updater {
...
+    /// * `messages`            - Vec of vec of bytes, each representing a message
+    ///                           to be hashed and accumulated
+    ///
+    /// This ix will write as many of the messages up to the length
+    /// of the `accumulator_input.data`.
+    /// If `accumulator_input.data.len() < messages.map(|x| x.len()).sum()`
+    /// then the remaining messages will be ignored.
+    ///
+    /// The current implementation assumes that each invocation of this
+    /// ix is independent of any previous invocations.
```
jayantk, line 72: also be clear that this overwrites all existing contents

accumulator_updater/programs/accumulator_updater/src/instructions/put_all.rs
jayantk: ah I didn't realize the user-facing docs were on the method that calls this. makes sense.

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
...
+pub struct AccumulatorHeader {
+    pub bump:        u8, // 1
+    pub version:     u8, // 1
+    // byte offset of accounts where data starts
+    // e.g. account_info.data[offset + header_len]
+    pub header_len:  u16, // 2
+    /// endpoints of every message.
+    /// ex: [10, 14]
+    /// => msg1 = data[(header_len + 0)..(header_len + 10)]
+    /// => msg2 = data[(header_len + 10)..(header_len + 14)]
```
jayantk, line 41: 👍 really like the examples here. they add a lot of clarity as to how this works

one note here is that you may want to rename "data" on AccumulatorInput because it can be confusing right now as to whether you're talking about the account_info.data or the accumulator_input.data

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
...
+}
+
+impl AccumulatorHeader {
+    // header_size. Keeping same structure as `BatchPriceAttestation` header
+    // HEADER_LEN allows for append-only forward-compatibility for the header.
+    // this is the number of bytes from the beginning of the account_info.data
+    // to the start of the `AccumulatorInput` data.
+    //
+    // *NOTE* this implementation is slightly different
+    // than the `BatchPriceAttestation` header_size
```
jayantk, line 52: (fwiw i think this comment would be better if it didn't reference BatchPriceAttestation. The reader probably won't understand how that format works deeply, so the comparison likely won't help)

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
...
+            header,
+            data: [0u8; 9_718],
+        }
     }
 
-    pub fn update(&mut self, new_data: Vec<u8>) {
-        self.header = AccumulatorHeader::new(self.header.bump, self.header.account_schema);
-        self.data = new_data;
+    // note: this does not handle if multiple CPI calls
+    // need to be made to write all the messages
```
jayantk, line 84: This comment is talking about CPIs and stuff which are at a different level of the stack. This comment should describe what this method does literally -- in particular, it overwrites all of the existing data, so you can't append data via multiple calls.

jayantk: also probably good to describe what it returns because it's not self-explanatory

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
...
+    // TODO: add a end_offsets index parameter for "continuation"
+    // TODO: test max size of parameters that can be passed into CPI call
+    pub fn put_all(&mut self, values: &Vec<Vec<u8>>) -> (usize, u16) {
+        let mut offset = 0u16;
+
+        for (i, v) in values.iter().enumerate() {
+            let start = offset;
+            let end = offset + (v.len() as u16);
```
jayantk, line 92: do a try_from safe conversion here just in case someone gives you a value that's greater than 2^16 in length?

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
...
+            if end > self.data.len() as u16 {
+                // need to return number of messages written & length?
```
jayantk, line 94: you did this already

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -43,48 +111,138 @@ impl AccumulatorInput {
...
+            r"
+
+            header
+                size: {header_idx_size:?}
+                align: {header_idx_align:?}
+            input
+                size: {input_size:?}
+                align: {input_align:?}
+        "
+        );
```
jayantk, line 165: delete the println

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -43,48 +111,138 @@ impl AccumulatorInput {
...
+
+
+        assert_eq!(accumulator_input.header.end_offsets[0], 2);
+        assert_eq!(accumulator_input.header.end_offsets[1], 5);
+
+
+        let accumulator_input_bytes = bytes_of(accumulator_input);
+
+        // *note* minus 8 here since no account discriminator when using
+        // `bytes_of`directly on accumulator_input
```
jayantk, line 191: general note on wording your documentation: you tend to write comments in a way that assumes the reader has lots of context for the code. It's usually more helpful to explain things in a way that assumes the reader *doesn't* have context. For example, this could say: "The header_len field represents the size of all data prior to the message bytes. This includes the account discriminator, which is not part of the header struct. Subtract the size of the discriminator (8 bytes) to compensate."

(this is like exactly the kind of place where you want to put a comment, so my comment is about the wording.)

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -14,26 +14,94 @@ use {
 /// the CPI calling program (e.g. Pyth Oracle)
 ///
 /// TODO: implement custom serialization & set alignment
-#[account]
+#[account(zero_copy)]
+#[derive(Debug, InitSpace)]
```
jayantk, line 18: nitpick for a future PR: can we rename this thing? It's more like a `MessageBuffer` or something right? It's a thing that stores messages. Those messages happen to be accumulated by the validator, but that doesn't have to play a role in the name of this struct.

(I think we're using the term "accumulator" way too much throughout our code / conversations right now. Everything is an accumulator X or accumulator Y, which makes it hard to distinguish the different concepts)

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
```diff
@@ -43,48 +111,138 @@ impl AccumulatorInput {
...
+        for offset in iter {
+            let end_offset = header_len + *offset as usize;
+            let accumulator_input_data = &accumulator_input_bytes[start..end_offset];
+            let expected_data = data_iter.next().unwrap();
+            assert_eq!(accumulator_input_data, expected_data.as_slice());
+            start = end_offset;
+        }
+
+        assert_eq!(accumulator_input.header.end_offsets[2], 0);
+    }
```
jayantk, line 247: can you add a test where you submit a Vec of data that's longer than a u16 can represent?

jayantk (approved):
nice! I left you a bunch of minor comments (mostly around documentation and naming). Please address and then feel free to merge without another review.

(in general, if i leave comments and approve, it means I trust you to address the feedback without another review)

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
swimricky: renamed to messages

accumulator_updater/programs/accumulator_updater/src/state/accumulator_input.rs
swimricky: this actually helped catch a bug thanks

PR #764, opened 2023-04-13
Get price feed endpoint

jayantk:
This PR adds an endpoint that's the equivalent of get_vaa (so price id & publish time), but that returns the parsed price feed data. All of the other methods have both vaa & price feed options except this one, so this makes the API uniform.

### DIFF
--- a/tilt_devnet/docker_images/Dockerfile.lerna
+++ b/tilt_devnet/docker_images/Dockerfile.lerna
@@ -18,6 +18,8 @@ COPY ./tsconfig.base.json ./
 
 FROM node:18.13.0@sha256:d9061fd0205c20cd47f70bdc879a7a84fb472b822d3ad3158aeef40698d2ce36 as lerna
 
+RUN apt-get update && apt-get install -y libusb-dev
+
 # 1000 is the uid and gid of the node user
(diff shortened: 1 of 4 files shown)


tilt_devnet/docker_images/Dockerfile.lerna
```diff
@@ -18,6 +18,8 @@ COPY ./tsconfig.base.json ./
 
 FROM node:18.13.0@sha256:d9061fd0205c20cd47f70bdc879a7a84fb472b822d3ad3158aeef40698d2ce36 as lerna
 
+RUN apt-get update && apt-get install -y libusb-dev
```
jayantk, line 21: this docker image failed to build for me locally because it's missing this dependency. not sure what changed here 🤷

cctdaniel (approved):
so if I understand correctly this is essentially get_vaa but returns the parsed vaa to price feed?

jayantk:
Yup exactly

tilt_devnet/docker_images/Dockerfile.lerna
ali-bahjati: Probably because you are in M1

PR #766, opened 2023-04-14
[price_service] Let callers specify VAA encoding

jayantk:
We get a lot of questions about how to convert the base64 VAAs into hex and whatnot. This PR adds an "encoding" parameter to all of the endpoints that return a VAA which lets you pick the thing you want. 

Note that I added a "0x" option which is hex, but with a 0x prepended. The EVM APIs seem to need this in various places.

**UPDATE**

I changed my mind about how this should work. The encoding parameter is itself hard for people to figure out, so I replaced it with a `target_chain` argument which is self-explanatory. The server will map the target chains to the desired encoding.

### DIFF
--- a/price_service/server/package.json
+++ b/price_service/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-service-server",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "description": "Webservice for retrieving prices from the Pyth oracle.",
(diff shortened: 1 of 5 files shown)


cctdaniel (approved):
🙇🏻‍♂️

cctdaniel (commented):
nice! this seems to be more user friendly 🙇🏻‍♂️

PR #770, opened 2023-04-19
[cosmwasm] wormhole stub 

0xfirefist:


target_chains/cosmwasm/wormhole-stub/Dockerfile.wormhole
```diff
@@ -0,0 +1,22 @@
+# Run with:
+# docker build -f Dockerfile.wormhole -o artifacts .
+
+FROM cosmwasm/workspace-optimizer-arm64:0.12.11@sha256:a00c01c2fdae8c74d6a9db95fa7b0cbb8e81101ce23a78487012e363d632fee9 AS builder
+# FROM cosmwasm/workspace-optimizer:0.12.11@sha256:b6b0b1d5ed76dafd0ec0d3812a1e7a275316cf862fd6d0a918f14d71c4cb187f AS builder
```
jayantk, line 5: this seems a little janky -- do you have to comment / uncomment this depending on the architecture? (you may be able to fix this by passing the platform flag to docker, or asking users to change their shell to use emulation)

jayantk (approved):
Can you add documentation to the README explaining what this does, when you will need it, and how to run it?

target_chains/cosmwasm/wormhole-stub/Dockerfile.wormhole
0xfirefist: Yes, this is the case. 
I don't think passing platform flag will work. As both of these are separate images and not one image supporting multiple architecture.

0xfirefist: I have added comment for it. 
Arm64 ones are only to be used while testing and not for production usage

PR #772, opened 2023-04-19
[accumulator-updater 7/x] - cpi max limit

swimricky:
### Summary
Add instruction to mock-cpi-program to allow for testing max cpi instruction size limits

### DIFF
--- a/accumulator_updater/Anchor.toml
+++ b/accumulator_updater/Anchor.toml
@@ -4,8 +4,6 @@ skip-lint = false
 [programs.localnet]
 accumulator_updater = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
 mock_cpi_caller = "Dg5PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
-#accumulator_updater = "3xzsrbvWEM4SMM2W7wbVofpDi3zrdhb2Z2xLwMEGJQiZ"
-#mock_cpi_caller = "EKwG38D6Sd67NqsAVc4R7Hqr6mZFPzFotzBchoQJRRDe"
 
 [registry]
 url = "https://api.apr.dev"
--- a/accumulator_updater/Anchor.toml
+++ b/accumulator_updater/Anchor.toml
@@ -4,6 +4,8 @@ skip-lint = false
 [programs.localnet]
 accumulator_updater = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
 mock_cpi_caller = "Dg5PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
+#accumulator_updater = "3xzsrbvWEM4SMM2W7wbVofpDi3zrdhb2Z2xLwMEGJQiZ"
+#mock_cpi_caller = "EKwG38D6Sd67NqsAVc4R7Hqr6mZFPzFotzBchoQJRRDe"
 
 [registry]
 url = "https://api.apr.dev"
--- a/accumulator_updater/NOTES.md
+++ b/accumulator_updater/NOTES.md
@@ -1,12 +1,29 @@
+## Testing
+
+- run `anchor test` if no special customization for the test-validator is needed
+- `anchor test` will run `solana-test-validator` will all features activated.
+  One of the features activated on the test-validator which is not currently activated on pythnet is
+
+```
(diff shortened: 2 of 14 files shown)


accumulator_updater/Anchor.toml
```diff
@@ -4,6 +4,8 @@ skip-lint = false
 [programs.localnet]
 accumulator_updater = "Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
 mock_cpi_caller = "Dg5PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS"
+#accumulator_updater = "3xzsrbvWEM4SMM2W7wbVofpDi3zrdhb2Z2xLwMEGJQiZ"
+#mock_cpi_caller = "EKwG38D6Sd67NqsAVc4R7Hqr6mZFPzFotzBchoQJRRDe"
```
jayantk, line 8: why is this commented out? (delete or leave a comment as to why you would want to uncomment)

accumulator_updater/NOTES.md
```diff
@@ -1,12 +1,29 @@
...
+- `anchor test` will run `solana-test-validator` will all features activated.
+  One of the features activated on the test-validator which is not currently activated on pythnet is
+
+```
+"GDH5TVdbTPUpRnXaRyQqiKUa7uZAbZ28Q2N9bhbKoMLm  loosen cpi size restrictions #26641"
+```
+
+In order to run `solana-test-validator` with this feature deactivated, do the following:
+
+1. open a terminal and run `solana-test-validator --reset --deactivate-feature GDH5TVdbTPUpRnXaRyQqiKUa7uZAbZ28Q2N9bhbKoMLm`
```
jayantk, line 13: let's remember to set this flag properly in the integration testing framework

accumulator_updater/programs/accumulator_updater/src/lib.rs
```diff
@@ -10,6 +10,7 @@ use {
 };
 
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
+// declare_id!("3xzsrbvWEM4SMM2W7wbVofpDi3zrdhb2Z2xLwMEGJQiZ");
```
jayantk, line 13: same comment about leaving commented out code

accumulator_updater/programs/accumulator_updater/src/state/whitelist.rs
```diff
@@ -47,7 +47,7 @@ pub struct WhitelistVerifier<'info> {
         seeds = [b"accumulator".as_ref(), b"whitelist".as_ref()],
         bump = whitelist.bump,
     )]
-    pub whitelist:  Account<'info, Whitelist>,
+    pub whitelist:  Box<Account<'info, Whitelist>>,
```
jayantk, line 50: why did this change?

accumulator_updater/programs/mock-cpi-caller/src/instructions/cpi_max_test.rs
```diff
@@ -0,0 +1,54 @@
...
+        pyth_price_acct.update(params)?;
+
+        let price_full_data = FullPriceMessage::from(&**pyth_price_acct).accumulator_serialize()?;
+
+        inputs.push(price_full_data);
+
+
+        let price_compact_data =
+            CompactPriceMessage::from(&**pyth_price_acct).accumulator_serialize()?;
+        inputs.push(price_compact_data);
```
jayantk, line 40: fwiw i think this would be a cleaner test if this pushed only the dummy messages. You really want to be able to precisely control the amount of data being written into the accumulator, and having multiple message types makes it harder to do that.

In fact, I think the optimal way to write this test would have been to like write an array of bytes where the caller chooses the length of the array.

(not a big deal though)

jayantk (approved):

accumulator_updater/programs/accumulator_updater/src/state/whitelist.rs
swimricky: using a `Box` will move the account to the heap which preserves more space for the stack . Heap gets allocated 32KB by default - https://github.com/solana-developers/anchor-zero-copy-example/blob/main/README.md#explanation-of-solana-memory-and-zero-copy

accumulator_updater/programs/mock-cpi-caller/src/instructions/cpi_max_test.rs
swimricky: updated test following your suggestion

accumulator_updater/programs/accumulator_updater/src/state/whitelist.rs
swimricky: added comment to clarify reasoning for this

PR #773, opened 2023-04-19
[hermes] Add WS

ali-bahjati:
This PR adds Websocket implementation with price service interface to Hermes. Each websocket connection is wrapped in a Subscriber struct which acts as an actor. 

As discussed before, after merging this I am planning to restructure everything to only support accumulator and things gets changed. The current structure is not great and will be changed when the refactor is done.


hermes/src/network/rpc/types.rs
```diff
@@ -0,0 +1,74 @@
+use {
```
ali-bahjati, line 1: This file is mostly the rename of the deleted rpc_price_feed.rs part and moved some other types from rest.rs here.

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+pub type SubscriberId = usize;
+
+/// Subscriber is an actor that handles a single websocket connection.
+/// It listens to the store for updates and sends them to the client.
+pub struct Subscriber {
+    id:                      SubscriberId,
+    closed:                  bool,
+    store:                   Store,
```
ali-bahjati, line 77: This kind of dependency is not great. We should rethink about the structure in the coming refactors.

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,290 @@
...
+        Ok(())
+    }
+}
+
+pub async fn dispatch_updates(update_feed_ids: Vec<PriceIdentifier>, state: super::State) {
+    let ws_state = state.ws.clone();
+    let mut closed_subscribers: Vec<SubscriberId> = Vec::new();
+
+    for subscriber in ws_state.subscribers.iter_mut() {
+        if let Err(e) = subscriber.send(update_feed_ids.clone()).await {
```
jayantk, line 226: I think you want these sends to run in parallel right? 

consider doing a join! like this https://rust-lang.github.io/async-book/06_multiple_futures/02_join.html

jayantk: actually you probably need this https://docs.rs/futures/0.3.5/futures/future/fn.join_all.html

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+
+    pub async fn run(&mut self) {
+        loop {
+            if self.closed {
+                break;
+            }
+            if let Err(e) = self.handle_next().await {
+                log::error!("Subscriber {}: Error handling next message: {}", self.id, e);
```
jayantk, line 109: do you really want to kill the subscriber here ? seems like it might be simpler to have one way to terminate the connection (setting self.closed) and then move all error logging here. I note you have some code below that transforms errs into Oks which could go away if you logged here instead.

jayantk (approved):
I don't totally understand how this works but it seems generally reasonable. I think some tests would be nice (though understand why you wouldn't want to do that given you're about to refactor everything)

hermes/src/network/rpc/ws.rs
ali-bahjati: There are locks to prevent misuse (Rust doesn't allow otherwise :D) but I also created the direction of dependency in a way that it doesn't introduce a deadlock.

ali-bahjati: I think generally it's fine. I don't like it much because it's half actor and half part of a global component.

I could store the Subscriber as a global component and let the global dispatcher directly call the methods and I tried. The result wasn't great because the Subscriber has a stream coming from the client and when it gets closed on destruction it should remove itself from the global reference to it (which it has access to the reference) and ...

I think the actor model offers better decoupling but that's a bigger design decision that I didn't want to make on this PR.

hermes/src/network/rpc/types.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub attestation_time:           UnixTimestamp,
+    pub sequence_number:            u64,
+    pub price_service_receive_time: UnixTimestamp,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct RpcPriceFeed {
+    pub id:        PriceIdentifier,
+    pub price:     Price,
+    pub ema_price: Price,
```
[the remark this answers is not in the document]
ali-bahjati: The publish time is inside Price

hermes/src/network/rpc/types.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub metadata:  Option<RpcPriceFeedMetadata>,
+    /// Vaa binary represented in base64.
+    pub vaa:       Option<Base64String>,
+}
+
+impl RpcPriceFeed {
+    pub fn from_price_info(price_info: PriceInfo, verbose: bool, binary: bool) -> Self {
```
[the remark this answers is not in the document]
ali-bahjati: Well usually the From and Into interface get the object because the caller mostly is aiming for a conversion (and the values gets moved instead of being created). I'm following that.

hermes/src/network/rpc/ws.rs
ali-bahjati: Oh yeah I totally forgot it. I'll do it.

hermes/src/network/rpc/ws.rs
ali-bahjati: The reason I did that was that close is not an error and the log there is on debug level and this one has error log level. Usually this should not happen and we need a higher log level for it.

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,303 @@
...
+            store,
+            update_rx,
+            receiver,
+            sender,
+            price_feeds_with_config: HashMap::new(),
+        }
+    }
+
+    pub async fn run(&mut self) {
+        loop {
```
Reisen, line 105:
```suggestion
        while !self.closed {
```

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+    }
+
+    async fn handle_next(&mut self) -> Result<()> {
+        tokio::select! {
+            Some(update_feed_ids) = self.update_rx.recv() => {
+                self.handle_price_feeds_update(update_feed_ids).await?;
+            },
+            Some(message_or_err) = self.receiver.next() => {
```
Reisen, line 121: What circumstances does this return `None`? My understanding is that this will only occur when the stream ends but that likely means you want to return an error also?

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,303 @@
...
+            if let Some(config) = self.price_feeds_with_config.get(&price_feed_id) {
+                let price_feeds_with_update_data = self.store.get_price_feeds_with_update_data(
+                    vec![price_feed_id],
+                    crate::store::RequestTime::Latest,
+                )?;
+                let price_info = price_feeds_with_update_data
+                    .batch_vaa
+                    .price_infos
+                    .get(&price_feed_id)
+                    .unwrap()
```
Reisen, line 144: Can return an error here as in a `Result<()>` right?

hermes/src/network/rpc/ws.rs
```diff
@@ -0,0 +1,303 @@
...
+                    let price_id: PriceIdentifier = id.into();
+                    self.price_feeds_with_config.remove(&price_id);
+                }
+            }
+        }
+
+        self.sender
+            .send(
+                serde_json::to_string(&ServerMessage::Response(ServerResponseMessage::Ok))
+                    .unwrap()
```
Reisen, line 213: Also can return an error.

Reisen (approved):

hermes/src/network/rpc/types.rs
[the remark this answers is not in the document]
ali-bahjati: You are right. But as things are gonna change a lot. I added a TODO about it to address later when I refactor this.

hermes/src/network/rpc/ws.rs
ali-bahjati: I am handling it better now.

PR #775, opened 2023-04-20
accumulator-updater 8/x] - whitelist verification pda auth

swimricky:
### Summary
Update whitelist verification mechanism to use a signed PDA from the CPI caller


accumulator_updater/programs/mock-cpi-caller/src/instructions/mod.rs
```diff
@@ -24,3 +24,5 @@ pub fn sighash(namespace: &str, name: &str) -> [u8; 8] {
 }
 
 pub const ACCUMULATOR_UPDATER_IX_NAME: &str = "put_all";
+pub const CPI: &str = "cpi";
+pub const AUTH: &str = "auth";
```
jayantk, line 28: nitpick: having a path like `accumulator_program_key/cpi` is enough right? why two separate seeds after the program key?

jayantk (approved):
nice

PR #779, opened 2023-04-25
[message-buffer 10/X] - Message Buffer Admin IXs & variable length

swimricky:
### Summary
Add `create/resize/delete_buffer` ixs, allow for `MessageBuffer` to be variable length

### Changes
- Remove `fund`
- Rename `whitelist.authority` to `whitelist.admin` and check that admin is signer & payer for all admin ixs.
- `MessageBuffer` struct now only defines the header fields and allows for variable length for the actual messages (see `MessageBuffer::put_all_in_buffer`)

PR #784, opened 2023-04-26
fix(message-buffer): address PR feedback from 779, update tests

swimricky:
Add new unit test for reading with cursor, update test setup

Addresses feedback from #779

### DIFF
--- a/message_buffer/programs/message_buffer/src/instructions/resize_buffer.rs
+++ b/message_buffer/programs/message_buffer/src/instructions/resize_buffer.rs
@@ -1,6 +1,5 @@
 use {
     crate::{
-        instructions::verify_message_buffer,
         state::*,
         MessageBufferError,
         MESSAGE,
@@ -30,7 +29,7 @@ pub fn resize_buffer<'info>(
     ctx.accounts
         .whitelist
         .is_allowed_program_auth(&allowed_program_auth)?;
-    verify_message_buffer(message_buffer_account_info)?;
+    MessageBuffer::check_discriminator(message_buffer_account_info)?;
 
     require_gte!(
         target_size,
@@ -62,7 +61,10 @@ pub fn resize_buffer<'info>(
         MessageBufferError::InvalidPDA
     );
 
-    if target_size_delta > 0 {
+    // allow for delta == 0 in case Rent requirements have changed
+    // and additional lamports need to be transferred.
+    // the realloc step will be a no-op in this case.
+    if target_size_delta >= 0 {
         let target_rent = Rent::get()?.minimum_balance(target_size);
         if message_buffer_account_info.lamports() < target_rent {
             system_program::transfer(
--- a/message_buffer/programs/message_buffer/src/state/message_buffer.rs
+++ b/message_buffer/programs/message_buffer/src/state/message_buffer.rs
@@ -1,9 +1,13 @@
 use {
     crate::{
         accumulator_input_seeds,
+        instructions,
         MessageBufferError,
     },
-    anchor_lang::prelude::*,
+    anchor_lang::{
+        prelude::*,
+        Discriminator,
+    },
 };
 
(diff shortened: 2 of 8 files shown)


message_buffer/programs/message_buffer/src/instructions/resize_buffer.rs
```diff
@@ -62,7 +61,10 @@ pub fn resize_buffer<'info>(
         MessageBufferError::InvalidPDA
     );
 
-    if target_size_delta > 0 {
+    // allow for delta == 0 in case Rent requirements have changed
+    // and additional lamports need to be transferred.
+    // the realloc step will be a no-op in this case.
```
jayantk, line 66: smart

message_buffer/programs/message_buffer/src/state/message_buffer.rs
```diff
@@ -325,4 +357,69 @@ mod test {
...
+
+        let message_buffer: &MessageBuffer =
+            bytemuck::from_bytes(&account_info_data.as_slice()[8..header_len]);
+
+        assert_eq!(message_buffer.end_offsets[0], 2);
+        assert_eq!(message_buffer.end_offsets[1], 5);
+
+
+        let mut cursor = std::io::Cursor::new(&account_info_data[10..]);
+        let header_len = cursor.read_u16::<LittleEndian>().unwrap();
```
jayantk, line 402: i think we agreed that everything would be big endian. it's going to be very confusing if this is little endian while everything else is big endian.

message_buffer/programs/message_buffer/src/state/message_buffer.rs
```diff
@@ -325,4 +357,69 @@ mod test {
...
+        let mut header_begin = header_len;
+        let mut header_end = cursor.read_u16::<LittleEndian>().unwrap();
+        let mut data_iter = data_bytes.iter();
+        println!("init header_end: {}", header_end);
+        let read_data = &mut vec![];
+        while header_end != 0 {
+            let end_offset = header_len + header_end;
+            let accumulator_input_data =
+                &account_info_data[header_begin as usize..end_offset as usize];
+            header_end = cursor.read_u16::<LittleEndian>().unwrap();
```
jayantk, line 413: header_end and header_begin are bad names for these two variables because they move to other locations in this loop

message_buffer/programs/message_buffer/src/state/message_buffer.rs
```diff
@@ -227,23 +253,27 @@ mod test {
...
         let header_len = message_buffer.header_len as usize;
-        let message_buffer_bytes = bytes_of_mut(message_buffer);
-        // assuming account_info.data.len() == 10KB
-        let messages = &mut vec![0u8; 10_240 - header_len];
 
         let account_info_data = &mut vec![];
         let discriminator = &mut sighash("accounts", "MessageBuffer");
+        let destination = &mut vec![0u8; 10_240 - header_len];
+
+
```
jayantk, line 261: you've copy/pasted a lot of code to initialize message buffers in these tests. I recommend extracting out a helper method that sets everything up for you.

jayantk: actually these tests seem to have lots of repeated code blocks in them. I think this isn't ideal -- you should think about how to factor things into helper methods to reduce duplication.

jayantk (approved):
I'm going to approve this but we should definitely decide on the endianness issue

message_buffer/programs/message_buffer/src/state/message_buffer.rs
swimricky: The `MessageBuffer` fields are only ever going to be read by the Pythnet validator & Hermes. I'm not sure off the top of my head how to combine BigEndian with zero_copy but i'll leave that as something to look into later on

PR #786, opened 2023-04-27
[Price pusher client] Fix axios bug by downgrading, bump all dependents

guibescos:
This bug was found by Brett from telegram. Basically EvmConnection wasn't working in a React app because of some axios weirdness.
Minimal example here : 
https://github.com/guibescos/test-connection

I want to make this change and make a release ASAP.

### DIFF
--- a/price_service/client/js/package.json
+++ b/price_service/client/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-service-client",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "description": "Pyth price service client",
   "author": {
     "name": "Pyth Data Association"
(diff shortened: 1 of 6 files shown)


price_service/client/js/package.json
```diff
@@ -50,7 +50,7 @@
   "dependencies": {
     "@pythnetwork/price-service-sdk": "*",
     "@types/ws": "^8.5.3",
-    "axios": "^1.2.5",
```
guibescos, line 53: I downgraded axios and it seemed to fix the problem

cctdaniel: lol what... nice find though

cctdaniel (approved):

ali-bahjati (approved):

PR #789, opened 2023-05-01
Wormhole stub

0xfirefist:
This PR contains
- the compiled wormhole code. we don't have to re comply it. 
- the script to deploy wormhole stubs to cosmos chains
- the deployed result for sei, neutron, and osmosis 5 testnets. 
- to help managing the results and everything, i have implemented a pipeline.


target_chains/cosmwasm/wormhole-stub/wormhole.testnet.json
ali-bahjati, line 1: I'm a bit confused about this file. How is it generated? I couldn't find it in the current PR and maybe I'm missing the context.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,216 @@
...
+    };
+
+// A stage will contain an identifier and a step method
+type Stage = {
+  stageId: string;
+  step: Step;
+};
+
+// A pipeline is conists of multiple stages
+// A stage contains of multiple steps
```
ali-bahjati, line 104: I think having stages having multiple "equal" steps makes things very complex. I think it is because in first part you want to deploy every contract, then instantiate and so on?

I think having a pipeline executing a sequence of steps is easier and enough for us.

ali-bahjati (commented):
I think @jayantk is more familiar with your progress on this and is more suitable to review this PR.

I like the pipeline concept but the implementation seems very complex and I cannot understand the rationale behind it. I think we can make it simpler while preserving the good behaviours that it has.

We can use the same thing for ETH too.

p.s: fun fact: pipeline here (with a single step) reminds me of Migrations in truffle :D The difference is that they store the last executed step on-chain instead of a local file :))

target_chains/cosmwasm/wormhole-stub/artifacts/wormhole.wasm
ali-bahjati, line 1: I recommend adding version (wormhole repo tag) to the file name so we know which wormhole version it corresponds to.

target_chains/cosmwasm/wormhole-stub/wormhole.testnet.json
0xfirefist: It is not generated. I wrote this.

target_chains/cosmwasm/wormhole-stub/artifacts/wormhole.wasm
0xfirefist: oh yeah, this makes sense. Thanks !

target_chains/cosmwasm/wormhole-stub/artifacts/checksums.txt
```diff
@@ -0,0 +1 @@
+ca4dead04cffc986bb05b9fa792327ac4ea2609ec928ef44dd7f48d3e16995f5  wormhole.wasm
```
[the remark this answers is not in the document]
0xfirefist: We don't use it. It's just for if someone wants to see. I think we can remove it though. Should I just remove it?

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,216 @@
...
+
+// This function lets you write a question to the terminal
+// And returns you the response of the user
+function readLineAsync(msg: string) {
+  const readline = createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
0xfirefist: checked. can't

target_chains/cosmwasm/tools/src/wormhole-stub.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (argv.mnemonic === undefined) {
+    console.log("Please provide the mnemonic");
+    return;
+  }
+
+  // IMPORTANT: IN ORDER TO RUN THIS SCRIPT FOR OTHER CHAINS
+  // WE NEED SOME METADATA
+  // HERE IS WHERE WE WILL ADDING THAT
+  // REPLACE THIS PART OF THE CODE FOR NEW CHAINS
+  const chainIds = [ChainId.NEUTRON_TESTNET_PION_1];
```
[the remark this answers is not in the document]
0xfirefist: We can automate this part too. Thinking about it though. When a new chain will be added we will have to add that chain metadata. This can be done with that step.

target_chains/cosmwasm/tools/src/wormhole-stub.ts
```diff
@@ -0,0 +1,173 @@
...
+  mainnet?: boolean
+): Step {
+  return (chainId, getResultOfPastStage) => {
+    // @ts-ignore
+    const wormholeChainId = CHAINS[wormholeChainInfo[chainId].chainId];
+    if (wormholeChainId === undefined)
+      throw new Error(`wormhole chain id undefined for chain: ${chainId}`);
+    const feeDenom = wormholeChainInfo[chainId].feeDenom;
+    if (feeDenom === undefined)
+      throw new Error(`fee denom undefined for chain: ${chainId}`);
```
jayantk, line 113: shouldn't these errors be checked when the step is instantiated, not when it's run? (i.e., move outside the returned function?)

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,216 @@
...
+  });
+}
+
+type StateStore<T = any> = {
+  [stepId: string]: T;
+};
+// StateManager helps in getting and setting the state locally
+// It manipulates data in-memory and once the consumer has finished manipulating it
+// They need to commit the data to permanent storage using the commit method
+class StateManager<T> {
```
jayantk, line 25: the type parameter T here seems inaccurate because the result types of the various steps are not uniform (?) IMO it would be better to drop the type parameter and have the requirement on results be something like "the result of a step must be serializable as JSON". The users of the results can cast them to a type (which is effectively what's happening now anyway).

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,216 @@
...
+        let result = stateManager.getStepState(stepId);
+        // pipeline will only proceed to the next stage if the previous one is fulfilled
+        if (result !== undefined && result.status === "rejected") {
+          throw new Error("previous stage was not fulfilled");
+        }
+        if (result === undefined) {
+          throw new Error(
+            `either the previous stage is not fully processed or a future stage is being referred to: ${stageId}`
+          );
+        }
```
jayantk, line 140: why is this error logic in here? seems like if a stage failed, the pipeline shouldn't even be trying to run the next stage?

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,216 @@
...
+  constructor(
+    private readonly stageId: string,
+    private readonly version: string
+  ) {
+    // dir check
+    if (!existsSync(this.dirPath)) {
+      mkdirSync(this.dirPath);
+    }
+
+    this.filePath = `${this.dirPath}/${this.stageId}-${this.version}${this.fileExt}`;
```
jayantk, line 41: i think this path structure should be defined at a higher level in the stack (like in the CLI, not in the framework itself)

jayantk (commented):
I read through this and I think there are many ways you can simplify this. It's probably better discussed live, but here's a quick summary:

1. Instead of parallelizing across steps (which are chains) within the Stages of the Pipeline, it would be simpler to instantiate a separate Pipeline for each chain, then run those pipelines asynchronously in parallel. If you make this change, it will simplify a lot of things, and it will let you move a lot of application-specific logic out of the generic pipeline code.
2. pipeline.run() should be idempotent, so if it fails, you can rerun it and it automatically picks up where it left off. You're already storing the state it produces in the filesystem, so you should build the processing logic around that (read the filesystem, decide what to do, run a step, write out, then loop). You can name the steps something like 1_deploy_wormhole to make it easy to figure out what's next. (also, if you do this, you should commit the migration results to the repo, so that anyone can pick up the code and resume migrating from where you left off)

This would actually be a good thing to pair on if we can find some time later today or tomorrow.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (currentResult !== undefined && currentResult.status === "fulfilled")
+      return true;
+
+    // Else we will process the new stage and store the result
+    const newResult = await this.stageExecutorWrapper(stage.executor)();
+    this.pipelineStore.setStageState(stage.id, newResult);
+
+    if (newResult.status === "fulfilled") return true;
+
+    // Some steps can fail due to some one time errors like API issues
```
ali-bahjati, line 119: we don't have step anymore.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,181 @@
...
+    for (let stage of this.stages) {
+      console.log(`${this.id}: Running stage with id: ${stage.id}`);
+
+      // This method is only going to process stage if all the past ones have been fulfilled
+      let fulfilled = await this.processStage(stage);
+      if (fulfilled === false) break;
+    }
+
+    // store the whole processing locally
+    this.pipelineStore.commit();
```
jayantk, line 141: you probably want to do this after processing every `stage`, just in case this code crashes for some reason.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+type StoreStructure = {
+  [stageId: string]: any;
+};
```
jayantk, line 147: this type seems unnecessary -- only used once, could just define it inline.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,181 @@
...
+      return;
+    }
+
+    this.store = JSON.parse(readFileSync(this.filePath).toString());
+  }
+
+  // It gets the latest state for the given stage
+  // the state after the last operation
+  // if there is no stage stage, in case it was no process it will return undefined.
+  getStageState<T = any>(stageId: string): T | undefined {
```
jayantk, line 166: maybe comment that the caller can basically pick the T and it'll cast it. It's the callers responsibility to know what the expected type is.

target_chains/cosmwasm/tools/src/pipeline.ts
```diff
@@ -0,0 +1,181 @@
...
+    readline.question(msg, (userRes) => {
+      resolve(userRes);
+    });
+  });
+}
+
+// The stage executor is where the stage functionality is defined
+// Optionally it can take in a method `getResultOfPastStage` as a parameter
+// if it wants to access the result of the previous stages
+export type StageExecutor =
```
jayantk, line 23: probably should document that each stage should have exactly one atomic operation (like sending a transaction), because this code doesn't enforce atomicity. So if you had 2 atomic operations in one stage, then you could end up fulfilling one and failing the other.

jayantk (approved):
awesome this is much clearer this way. I left you some minor comments but feel free to merge whenever you're ready.

PR #799, opened 2023-05-04
[cosmwasm] instantiate contracts on multiple chains

0xfirefist:
This PR contains:
- code that adds stages to the wormhole pipeline. Stages to upgrade wormhole guardian set.
- script for instantiating pyth
- deployment results for chains


target_chains/cosmwasm/tools/src/helper.ts
```diff
@@ -0,0 +1,113 @@
...
+  // clear tmp directory before downloading contracts
+  rimrafSync(tmpCodeStorageDir);
+
+  if (existsSync(tmpCodeStorageDir) === false) {
+    mkdirSync(tmpCodeStorageDir, { recursive: true });
+  }
+
+  const uniqueArtifactsZipName = Array.from(new Set(artifactZipFileNames));
+
+  // download zip files
```
ali-bahjati, line 27: I might be a bit paranoid, but can we also verify the checksum? If one of our gh accounts get compromised they can swap the binary without us knowing it.

target_chains/cosmwasm/tools/src/wormhole-stub.ts
```diff
@@ -123,19 +108,36 @@ async function run() {
       },
     });
 
+    // 4 vaa upgrades for guardian set
+    Object.keys(VAA_UPGRADES).forEach((id) => {
```
ali-bahjati, line 112: For mainnet only right?

target_chains/cosmwasm/tools/src/chains-manager/chains.ts
```diff
@@ -96,16 +97,23 @@ export const ChainsConfigTestnet: Record<ChainIdTestnet, ChainConfig> = {
...
 };
 
 /**
- * This method will return an executor for that corresponding chainType for given chainId.
+ * This method will return an executor for given chainConfig.
  */
-export function createExecutorForChain(
-  chainId: ChainIdTestnet,
+export function createExecutorForChain<T extends ChainConfig>(
+  chainConfig: T,
```
jayantk, line 114: shouldn't this be `chainConfig: ChainConfig`? You don't use the `T` beyond that

target_chains/cosmwasm/tools/src/helper.ts
jayantk: yeah agree with this

target_chains/cosmwasm/tools/src/helper.ts
```diff
@@ -0,0 +1,113 @@
...
+// We have released the compile contacts on github. If a chain needs some specific
+// feature in a contract, a version of the contract with that specific features is
+// also released. For example, "injective.zip" for injective.
+// Given a contract version (eg 1.2.0) and  zipFileNames (eg ["injective", "osmosis"])
+// this method will return an object with key = zipFileName and value = compiledCode
+export async function getContractBytesDict(
+  artifactZipFileNames: string[],
+  version: string
+) {
+  const githubArtifactsLink = `https://github.com/pyth-network/pyth-crosschain/releases/download/pyth-cosmwasm-contract-v${version}/`;
```
jayantk, line 16: i think it's a little questionable to hardcode this URL here. This ties the implementation to our git repository url, which is weird. It seems like this should be configurable and passed in. I don't feel strongly about this comment in this instance though -- no one else is going to run this -- so feel free to ignore.

I think it's also worthwhile clarifying exactly what expectations this function has on the zip files (e.g., it needs the wasm artifact inside to have a specific name, etc)

target_chains/cosmwasm/tools/src/helper.ts
```diff
@@ -0,0 +1,113 @@
...
+// This method returns the file name which stores the deployment result for
+// wormhole stub
+export function getWormholeFileName(
+  chainId: string,
+  version: string,
+  mainnet: boolean
+): string {
+  const WORMHOLE_STORAGE_DIR = mainnet
+    ? "../wormhole-stub/mainnet"
+    : "../wormhole-stub/testnet";
```
jayantk, line 85: extract constant variables to hold these string constants. that way you can rename the directories in code without chasing down all of the instances of the string "wormhole-stub"

jayantk: also why is this filepath "wormhole-stub/testnet" and the pyth one is "testnet/instantiate-pyth"? seems like those should be in the same order

target_chains/cosmwasm/tools/src/instantiate-pyth.ts
```diff
@@ -0,0 +1,164 @@
...
+    governance_source_index: 0,
+    governance_sequence_number: 0,
+    chain_id: wormholeChainId,
+    valid_time_period_secs: 60,
+    fee: {
+      amount: "1",
+      denom: feeDenom,
+    },
+    data_sources: [
+      {
```
jayantk, line 139: maybe comment here that all chains (including testnets) are using the mainnet data sources

target_chains/cosmwasm/tools/src/wormhole-stub.ts
jayantk: dev is doing the "testnets read mainnet feeds" thing we discussed for cosmwasm

target_chains/cosmwasm/tools/src/helper.ts
jayantk: also, might be nice to write down what filepaths this generates and what gets stored there. Could be a good code comment on wherever you put the string constants.

jayantk (commented):
nice. I have mostly minor comments. The one thing I would like to see is the checksum checking, because there is a potential security hole there (and we should be paranoid about this stuff). Let's write down the expected checksums in this repo as a config file and then we can check the downloads against that.

target_chains/cosmwasm/tools/src/wormhole-stub.ts
0xfirefist: testnet will read from mainnet feeds

target_chains/cosmwasm/tools/src/helper.ts
0xfirefist: Extracting constant variables. This is the only place it is being used and hence the only place we would need to make change.

target_chains/cosmwasm/tools/src/helper.ts
0xfirefist: hardcoded checksum

target_chains/cosmwasm/tools/src/helper.ts
0xfirefist: Clarified the comment.

ali-bahjati (approved):

ali-bahjati (approved):

PR #803, opened 2023-05-04
[pythnet] Replace pythnet_sdk with the right files

Reisen:
This fixes #802 which accidentally included the old Pyth files from the wrong commit.


swimricky (approved):
LGTM

jayantk (approved):

PR #810, opened 2023-05-09
docker-compose: update guardiand@2.17 & xc-server@3.0.3

drptbl:


ali-bahjati:
@drptbl thanks for your PR. #814 includes the same changes too. I'll update guardian version there too.

PR #821, opened 2023-05-17
[hermes] Rewrite with Accumulator

ali-bahjati:
This PR is a rewrite of hermes that drops batch vaa support and adds support for accumulators. Many things are hardcoded or not handled well with the purpose of releasing an MVP as soon as possible. It lacks some features from price service such as db connection, encoding, metrics, some endpoints, and more.

The design of the hermes (without api) will remain more or less the same but small things will be refactored. Also the terminology is not great and we need to thing more on that.

Lastly, there are no tests as we are still restructuring everything. We will write tests soon after in separate PRs.


pythnet/pythnet_sdk/examples/generate_pyth_data.rs
```diff
@@ -8,8 +8,10 @@ use {
     solana_client::rpc_client::RpcClient,
     solana_pyth::PYTH_PID,
     solana_sdk::pubkey::Pubkey,
-    std::str::FromStr,
-    std::io::Write,
+    std::{
+        io::Write,
+        str::FromStr,
+    },
```
ali-bahjati, line 14: came from formatter

hermes/src/store/storage.rs
```diff
@@ -27,8 +19,15 @@ pub enum Key {
 /// data to abstract the details of the update data, and so each update data is stored
 /// under a separate key. The caller is responsible for specifying the right
 /// key for the update data they wish to access.
-pub trait Storage: Sync + Send {
-    fn insert(&self, key: Key, time: UnixTimestamp, value: StorageData) -> Result<()>;
-    fn get(&self, key: Key, request_time: RequestTime) -> Result<Option<StorageData>>;
-    fn keys(&self) -> Vec<Key>;
+pub trait Storage: Send + Sync {
+    fn store_message_states(&self, message_states: Vec<MessageState>) -> Result<()>;
```
jayantk, line 23: I think the trait should be more like put(key, value) than storing an object that internally has a key in it. These two function signatures are confusing -- it's hard for the reader to tell how to retrieve something they inserted.

hermes/src/store/proof/wormhole_merkle.rs
```diff
@@ -0,0 +1,148 @@
...
+                    .prove(m.as_ref())
+                    .ok_or(anyhow!("Failed to prove message"))?,
+            })
+        })
+        .collect::<Result<Vec<WormholeMerkleMessageProof>>>()
+}
+
+pub fn construct_update_data(mut message_states: Vec<MessageState>) -> Result<Vec<Vec<u8>>> {
+    message_states.sort_by_key(
+        |m| m.proof_set.wormhole_merkle_proof.vaa.clone(), // FIXME: This is not efficient
```
jayantk, line 104: there must be a way to do this without cloning

hermes/src/store/storage/local_storage.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn retrieve_message_states(
+        &self,
+        ids: Vec<MessageIdentifier>,
+        request_type: RequestType,
```
jayantk, line 114: why do you need RequestType at all? seems like you just wrap and unwrap MessageType

hermes/src/store/types.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl MessageState {
+    pub fn time(&self) -> MessageTime {
+        MessageTime {
+            publish_time: self.publish_time,
+            slot:         self.slot,
+        }
+    }
+
+    pub fn key(&self) -> MessageKey {
```
jayantk, line 109: this method doesn't belong on MessageState. This is an implementation detail of the storage system (and you could imagine other storage systems choosing a different key)

hermes/src/store/storage.rs
jayantk: after reading through the rest of the code, I think this abstraction should be called something like `TemporalCache<K, V>` and the function signatures should be `put(key: K, timestamp, value: V)` and `get(key: K, request_time): Option<V>`.

hermes/src/store/types.rs
```diff
@@ -0,0 +1,303 @@
...
+#[derive(Clone, PartialEq, Debug)]
+pub struct WormholeMerkleState {
+    pub digest_proof: Vec<u8>,
+    pub tree:         Option<Vec<Vec<u8>>>,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct MessageKey {
+    // -> this is the real message id
+    pub id:    MessageIdentifier, // -> this is price feed id
```
jayantk, line 74: I don't understand these comments here

hermes/src/store/types.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub type_: MessageType,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub struct MessageTime {
+    pub publish_time: UnixTimestamp,
+    pub slot:         Slot,
```
jayantk, line 81: this is a bad abstraction because you can construct MessageTimes where the publish_time and the slot don't correspond (so they would be sorted differently depending on which field you chose). I don't think this is getting you a lot of value also, so I would just delete it.

hermes/src/store/types.rs
```diff
@@ -0,0 +1,303 @@
...
+}
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct ProofSet {
+    pub wormhole_merkle_proof: WormholeMerkleMessageProof,
+}
+
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct MessageState {
```
jayantk, line 91: nitpick: this struct could be named better. MessageState is weird to me because it sounds like it's the state of the message, but messages don't have different states. It's more like a Message plus its proof and metadata. How about `ProvenMessage`?

hermes/src/store/types.rs
jayantk: if MessageIdentifier is PriceFeedId, why not call it PriceFeedId?

hermes/src/store/types.rs
```diff
@@ -0,0 +1,303 @@
+use {
```
jayantk, line 1: I think you should split up where these types are defined such that they are close to the code that uses them. This class mixes up code from many different layers of the system (the rest API, the storage system, etc). You typically want to keep those layers separated so they have crisp boundaries.

jayantk (approved):
cool. I'm going to approve to unblock you, as I know this is work in progress. I also left you a bunch of comments on how this code could be improved.

also would be great to get some tests. (I know this is work in progress so no worries)

hermes/src/network/pythnet.rs
```diff
@@ -0,0 +1,92 @@
...
+            let ring_index = accumulator_slot.rem(RING_SIZE as u64) as u32;
+
+            let (candidate, _) = Pubkey::find_program_address(
+                &[
+                    b"AccumulatorState",
+                    &pyth.to_bytes(),
+                    &ring_index.to_be_bytes(),
+                ],
+                &system_program::id(),
+            );
```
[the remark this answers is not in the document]
ali-bahjati: yeah, many of this logic will go to sdk as I mentioned in the PR description too.

hermes/src/store.rs
```diff
@@ -1,73 +1,212 @@
...
+                self.pending_accumulations
+                    .insert(
+                        ring_index,
+                        pending_acc.accumulator_messages(accumulator_messages),
+                    )
+                    .await;
+
+                ring_index
+            }
+        };
```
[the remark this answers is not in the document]
ali-bahjati: I should think more about it. Your idea is not scalable when we add other types of messages.

hermes/src/store.rs
```diff
@@ -1,73 +1,212 @@
...
+
+        let state = match pending_state.build() {
+            Ok(state) => state,
+            Err(_) => return Ok(()),
+        };
+
+        log::info!("State: {:?}", state);
+
+        let wormhole_merkle_message_states_proofs = construct_message_states_proofs(state.clone())?;
```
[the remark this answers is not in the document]
ali-bahjati: I thought about it. There are some tradeoffs here. Most importantly, I'll like the storage to have this proof per message completely because it will probably gets stored off-chain and will be easier for reconstruction. Making it lazy will make the storage api more complex.

hermes/src/store/proof/wormhole_merkle.rs
```diff
@@ -0,0 +1,148 @@
...
+    let merkle_acc = match MerkleAccumulator::<Keccak160>::from_set(
+        state
+            .accumulator_messages
+            .messages
+            .iter()
+            .map(|m| m.as_ref()),
+    ) {
+        Some(merkle_acc) => merkle_acc,
+        None => return Ok(vec![]), // It only happens when the message set is empty
+    };
```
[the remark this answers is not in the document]
ali-bahjati: Good question. we should test it but anyway it doesn't belong to here. We sort the accounts before calling merkle tree and we store them we the same sorted order.

hermes/src/store/storage.rs
ali-bahjati: yeah i agree with your suggestion.

hermes/src/store/proof/wormhole_merkle.rs
ali-bahjati: yeah i will revise all the clones + other traits later as I mentioned in the PR description.

hermes/src/store/types.rs
ali-bahjati: I disagree, this is the key identifier of a message. Also MessageState is also part of the storage system interface. But let's discuss it later.

hermes/src/store/types.rs
ali-bahjati: I am going to deprecate price feed id later, that's there because of the past.

hermes/src/store/types.rs
ali-bahjati: There are some edge cases here that require slot to be present at the key. It is also important for the ordering of the messages. Mostly you should think that the slot is kind of the attestation time. It's also future looking to the sub second accuracy.

hermes/src/store/types.rs
ali-bahjati: agreed. the names should be improved in general.

hermes/src/store/types.rs
ali-bahjati: yeah i noted this in the PR description too.

hermes/src/api.rs
```diff
@@ -55,25 +49,25 @@ pub async fn spawn(rpc_addr: String, store: Store) -> Result<()> {
...
+    tokio::spawn(async move {
+        // FIXME handle errors properly
+        axum::Server::bind(&rpc_addr.parse().unwrap())
+            .serve(app.into_make_service())
+            .await
+            .unwrap();
+    });
+
+    // Call dispatch updates to websocket every 1 seconds
+    // FIXME use a channel to get updates from the store
```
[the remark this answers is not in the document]
ali-bahjati: Channels are very fast in rust and shouldn't introduce much latency.

hermes/src/store.rs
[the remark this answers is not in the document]
Reisen: Cache was removed in favor of storing the incomplete state, which is better for recovery if the process is restarted or for debugging why we were unable to complete some specific pair.

hermes/src/network/ethereum.rs
```diff
@@ -0,0 +1,187 @@
...
+        let current_index = match current_index.as_slice() {
+            &[ethabi::Token::Uint(index)] => Ok(index),
+            _ => Err(anyhow!(
+                "Unexpected tokens {:?}. Expected a single uint",
+                current_index
+            )),
+        }?;
+
+        let get_guardian_set_method = serde_json::from_str::<Function>(
+            r#"{"inputs":[{"internalType":"uint32","name":"index","type":"uint32"}],"name":"getGuardianSet",
```
Reisen, line 99: We should probably use the `serde_json` `json!` macro here, It's far clearer.

hermes/src/network/pythnet.rs
```diff
@@ -0,0 +1,173 @@
...
+
+                        let (candidate, _) = Pubkey::find_program_address(
+                            &[
+                                b"AccumulatorState",
+                                &accumulator_messages.ring_index().to_be_bytes(),
+                            ],
+                            &system_program::id(),
+                        );
+
+                        if candidate.to_string() == update.value.pubkey {
```
jayantk, line 125: is there a reason this check is necessary, or is it just a sanity check? (maybe leave a comment)

jayantk: I guess there's a possibility that someone creates accounts that look like accumulator updates to confuse hermes?

hermes/src/store/storage/local_storage.rs
```diff
@@ -0,0 +1,163 @@
...
+    cache_size:        u64,
+}
+
+impl LocalStorage {
+    pub fn new_instance(cache_size: u64) -> StorageInstance {
+        Box::new(Self {
+            message_cache: Arc::new(DashMap::new()),
+            accumulator_cache: Cache::builder()
+                .max_capacity(cache_size)
+                .time_to_live(std::time::Duration::from_secs(60 * 60 * 24))
```
jayantk, line 41: extract a constant? (ideally this should be configurable also?)

jayantk: also doesn't 1 day seem like a long time to keep things around in the cache? It could potentially be a lot of data?

hermes/src/store/storage/local_storage.rs
```diff
@@ -0,0 +1,163 @@
...
+                                return None;
+                            }
+                        }
+
+                        // Binary search returns Ok(idx) if the element is found at index idx or Err(idx) if it's not
+                        // found which idx is the index where the element should be inserted to keep the vector sorted.
+                        // Getting idx within any of the match arms will give us the index of the element that is
+                        // closest after or equal to the requested time.
+                        let idx = match key_cache
+                            .binary_search_by_key(&time, |record| record.time().publish_time)
```
jayantk, line 70: note that we need to secondarily sort by slot here. If the records are stored in ascending order of (publish_time, slot), i think we need to search for (publish_time, 0)

hermes/src/store/storage/local_storage.rs
```diff
@@ -0,0 +1,163 @@
...
+};
+
+#[derive(Clone)]
+pub struct LocalStorage {
+    message_cache:     Arc<DashMap<MessageStateKey, VecDeque<MessageState>>>,
+    accumulator_cache: Cache<Slot, AccumulatorState>,
+    cache_size:        u64,
+}
+
+impl LocalStorage {
```
jayantk, line 35: this struct could use some unit tests.

hermes/src/store/storage/local_storage.rs
```diff
@@ -0,0 +1,163 @@
...
+        request_time: RequestTime,
+        filter: MessageStateFilter,
+    ) -> Result<Vec<MessageState>> {
+        ids.into_iter()
+            .flat_map(|id| {
+                let request_time = request_time.clone();
+                let message_types: Vec<MessageType> = match filter {
+                    MessageStateFilter::All => MessageType::iter().collect(),
+                    MessageStateFilter::Only(t) => vec![t],
+                };
```
jayantk, line 133: MessageStateFilter seems a bit overkill. You could just pass in `message_types_to_find` and the caller could provide `MessageType::iter().collect()` if they wanted all of them.

hermes/src/store/types.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub wormhole_merkle_proof: WormholeMerkleMessageProof,
+}
+
+pub type Slot = u64;
+pub type UnixTimestamp = i64;
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum RequestTime {
+    Latest,
+    FirstAfter(UnixTimestamp),
```
jayantk, line 27: (Note for later: some people want FirstBefore as well)

hermes/src/store/types.rs
```diff
@@ -0,0 +1,62 @@
...
+        (self.slot % self.ring_size as u64) as u32
+    }
+}
+
+pub enum Update {
+    Vaa(Vec<u8>),
+    AccumulatorMessages(AccumulatorMessages),
+}
+
+pub struct PriceFeedUpdate {
```
jayantk, line 49: this type and the one below seem like they should go in the API layer (because they're an implementation detail of the current REST API, and will go away in the future). I think you probably want to move `Store.get_price_feeds_with_update_data` to the API layer as well.

jayantk (approved):
This looks very solid to me overall. I think you should merge this and then send any changes as follow-up PRs, because there's nothing major to fix. 

Two asks for follow ups are:
1. Let's figure out the eth guardian set issue. Ideally we would not have the ethereum network dependency, and I'm pretty sure we can make that happen
2. Tests. (I know you've been pushing to get this out so totally understandable that it's not tested right now)

jayantk (approved):
This looks very solid to me overall. I think you should merge this and then send any changes as follow-up PRs, because there's nothing major to fix. 

Two asks for follow ups are:
1. Let's figure out the eth guardian set issue. Ideally we would not have the ethereum network dependency, and I'm pretty sure we can make that happen
2. Tests. (I know you've been pushing to get this out so totally understandable that it's not tested right now)

jayantk (approved):
This looks very solid to me overall. I think you should merge this and then send any changes as follow-up PRs, because there's nothing major to fix. 

Two asks for follow ups are:
1. Let's figure out the eth guardian set issue. Ideally we would not have the ethereum network dependency, and I'm pretty sure we can make that happen
2. Tests. (I know you've been pushing to get this out so totally understandable that it's not tested right now)

hermes/src/store/types.rs
ali-bahjati: Daniel, Guillermo, and I had an interesting conversation about this and we believe first after is enough and actually represents the closing price!

hermes/src/store/storage/local_storage.rs
ali-bahjati: Maybe, I don't have strong feeling about it. I personally feel this way has a more explicit boundary and `MessageType::iter()` is not something obvious.

hermes/src/store/storage/local_storage.rs
ali-bahjati: good catch. It is sorted by slot as well but apparently binary search doesn't guarantee that it gives you the first. I'll fix it.

PR #822, opened 2023-05-18
[cosmwasm] chain ids for pyth mainnet, pyth testnet stable, and pyth testnet edge feeds

0xfirefist:


target_chains/cosmwasm/tools/src/configs.ts
```diff
@@ -0,0 +1,233 @@
...
+export function getChainConfig(
+  chainId: ChainIdTestnet | ChainIdMainnet,
+  deploymentType: DeploymentType
+) {
+  if (deploymentType === "mainnet")
+    return ChainsConfigMainnet[chainId as ChainIdMainnet];
+  return ChainsConfigTestnet[chainId as ChainIdTestnet];
+}
+
+function getPythSources(mainnet: boolean) {
```
jayantk, line 117: isn't `mainnet` now the distinction between stable vs. edge? We should give this variable a different name to clarify that

jayantk: maybe use DeploymentType?

target_chains/cosmwasm/tools/src/helper.ts
```diff
@@ -4,6 +4,8 @@ import AdmZip from "adm-zip";
 import path from "path";
 import { DownloaderHelper } from "node-downloader-helper";
 
+export type DeploymentType = "mainnet" | "testnet-stable" | "testnet-edge";
```
jayantk, line 7: i think this should just be stable vs edge

target_chains/cosmwasm/tools/src/test.ts
```diff
@@ -0,0 +1,105 @@
...
+    choices: ["mainnet", "testnet-stable", "testnet-edge"],
+    demandOption: "Please provide the deployment type",
+  })
+  .help()
+  .alias("help", "h")
+  .wrap(yargs.terminalWidth())
+  .parseSync();
+
+async function run() {
+  const STORAGE_DIR = `./${argv.deploy}/test-contracts`;
```
jayantk, line 32: you've hardcoded a bunch of directories like this in multiple places in this PR. You should extract all of these paths to a single file. that could be a class like FilesystemStore with methods like .getTestContractDir, or it could just be a file with some functions or string constants.

jayantk (changes requested):
I don't think this is quite right. The chainIds should map on to blockchains, and then the deployment types should either be stable or edge. That way you can uniquely identify a contract by (chainId, deployment type).

target_chains/cosmwasm/tools/src/test.ts
```diff
@@ -0,0 +1,112 @@
...
+  DeploymentType,
+  getChainIdsForEdgeDeployment,
+  getChainIdsForStableDeployment,
+  getPythContractAddress,
+  getTestPythContractFileName,
+} from "./helper";
+import { PriceServiceConnection } from "@pythnetwork/price-service-client";
+import { CosmwasmQuerier } from "./chains-manager/chain-querier";
+const argv = yargs(hideBin(process.argv))
+  .usage("USAGE: npm run wormhole-stub -- <command>")
```
jayantk, line 19: command name and the arguments don't seem right for the test

target_chains/cosmwasm/tools/src/test.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+    });
+
+    // 2 fetch the updated price feed
+    pipeline.addStage({
+      id: "fetch-price-feed-update",
+      executor: () => {
+        return pythQuerier.getPriceFeed(pythContractAddr, priceFeedId);
+      },
+    });
```
jayantk, line 106: why do this as a pipeline (as opposed to simply awaiting the results here)?

target_chains/cosmwasm/tools/store/edge/test-contracts/injective_testnet-1.2.0.json
```diff
@@ -0,0 +1,34 @@
+{
+  "push-price-update": {
+    "status": "rejected",
```
jayantk, line 3: this seems bad?

jayantk (approved):
nice!

jayantk (approved):
nice!

target_chains/cosmwasm/tools/store/edge/test-contracts/injective_testnet-1.2.0.json
0xfirefist: The error is expected. We have to get the injective team update the pyth contract address which they are using for the oracle.

target_chains/cosmwasm/tools/src/helper.ts
```diff
@@ -3,6 +3,31 @@ import { rimrafSync } from "rimraf";
...
+
+export function getChainIdsForStableDeployment(): ChainId[] {
+  return [
+    ChainId.INJECTIVE_TESTNET,
+    ChainId.OSMOSIS_TESTNET_4,
+    ChainId.OSMOSIS_TESTNET_5,
+    ChainId.SEI_TESTNET_ATLANTIC_2,
+    ChainId.NEUTRON_TESTNET_PION_1,
+    ChainId.JUNO_TESTNET,
+  ];
```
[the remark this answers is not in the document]
0xfirefist: We can't deploy to each mainnet. Right now we are on Injective and Osmosis both of which have some restrictions.

target_chains/cosmwasm/tools/src/test.ts
0xfirefist: to store the results of the last test.

PR #829, opened 2023-05-22
[arbitrum] fix: max fee per gas less than block base fee

0xfirefist:


price_pusher/src/evm/evm.ts
```diff
@@ -240,6 +240,14 @@ export class EvmPricePusher implements IPricePusher {
           return;
         }
 
+        if (err.message.includes("max fee per gas less than block base fee")) {
+          // We just have to handle this error and return.
+          // LastPushAttempt was stored with the class
+          // Next time the update will be executing, it will check the last attempt
+          // and increase the gas price accordingly.
```
ali-bahjati, line 247: Maybe just log something for visibility?

ali-bahjati (approved):
Thank you 🙇

PR #832, opened 2023-05-22
[Xc admin] frontend to use server

guibescos:
Make xc-admin-frontend send the proposal instructions to the proposer server, that will submit them to the multisig.
You can test it locally by setting `NEXT_PUBLIC_PROPOSER_SERVER_URL` to the right value.

I tried to keep this one small, might be able to cleanup further once it gets approved and we make sure it works.


governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
```diff
@@ -77,39 +78,46 @@ const PermissionDepermissionKey = ({
       const fundingAccount = isRemote
         ? mapKey(multisigAuthority)
         : multisigAuthority
-      priceAccounts.map((priceAccount) => {
-        isPermission
-          ? pythProgramClient.methods
+
+      for (const priceAccount of priceAccounts) {
```
guibescos, line 82: Had to rewrite this logic to `await` for the instructions.

cctdaniel (approved):

PR #837, opened 2023-05-23
[remote-executor] Move overflow check

guibescos:
Addresses PYTH-TOB 12


.github/workflows/remote-executor.yml
```diff
@@ -15,8 +15,9 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: 1.66.1
```
guibescos, line 18: Pin rust for CI.

cctdaniel (approved):

PR #840, opened 2023-05-24
Make proposer server url a serverside prop

guibescos:


governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
```diff
@@ -31,10 +31,12 @@ const PermissionDepermissionKey = ({
   isPermission,
   pythProgramClient,
   squads,
+  PROPOSER_SERVER_URL,
```
jayantk, line 34: why is this all caps? it's a normal variable not a constant in this context, right?

jayantk: same pretty much everywhere below actually

jayantk (approved):

governance/xc_admin/packages/xc_admin_frontend/components/PermissionDepermissionKey.tsx
guibescos: I'll fix this

PR #843, opened 2023-05-24
[remote-executor] Isolate wormhole-solana crate

guibescos:
The goal of this PR is :
- Make remote-executor use the new version of `wormhole-sdk`
- Move the `wormhole-solana` (that wormhole doesn't have) to its own repo and update it to use `wormhole-sdk`

Wormhole team needs to :
- Fix the crate patch issue with serde_wormhole
- Decide what happens to `wormhole-solana`

### DIFF
--- a/.github/workflows/remote-executor.yml
+++ b/.github/workflows/remote-executor.yml
@@ -20,7 +20,7 @@ jobs:
           override: true
       - name: Install Solana
         run: |
-          sh -c "$(curl -sSfL https://release.solana.com/stable/install)"
+          sh -c "$(curl -sSfL https://release.solana.com/v1.14.18/install)"
           echo "/home/runner/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
       - name: Run executor tests
         run: cargo test-bpf --manifest-path ./governance/remote_executor/Cargo.toml
--- a/governance/remote_executor/Cargo.toml
+++ b/governance/remote_executor/Cargo.toml
@@ -6,3 +6,6 @@ members = [
 
 [profile.release]
 overflow-checks = true
+
+[patch.crates-io]
+serde_wormhole = { git = "https://github.com/wormhole-foundation/wormhole", tag = "v2.17.1" }
--- a/governance/remote_executor/cli/src/main.rs
+++ b/governance/remote_executor/cli/src/main.rs
@@ -137,7 +137,6 @@ fn main() -> Result<()> {
                 &vec![&payer],
             )?;
 
-            println!("{}", posted_vaa_key);
             // Now execute
             process_transaction(
                 &rpc_client,
--- a/governance/remote_executor/cli/src/main.rs
+++ b/governance/remote_executor/cli/src/main.rs
@@ -1,4 +1,12 @@
 #![deny(warnings)]
+
+use {
+    serde_wormhole::RawMessage,
+    wormhole_sdk::{
+        vaa::digest,
+        Vaa,
+    },
+};
 pub mod cli;
 use {
     anchor_client::{
@@ -47,7 +55,6 @@ use {
         transaction::Transaction,
     },
     std::str::FromStr,
-    wormhole::VAA,
     wormhole_solana::{
         instructions::{
             post_message,
@@ -86,7 +93,7 @@ fn main() -> Result<()> {
 
             let signature_set_keypair = Keypair::new();
 
-            let vaa = VAA::from_bytes(vaa_bytes.clone())?;
+            let vaa: Vaa<&RawMessage> = serde_wormhole::from_slice(&vaa_bytes)?;
 
             let posted_vaa_key = PostedVAA::key(&wormhole, vaa.digest().unwrap().hash);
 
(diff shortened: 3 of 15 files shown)


governance/remote_executor/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.66.1"
```
guibescos, line 2: Drive-by pin blockchain

governance/remote_executor/programs/remote-executor/src/tests/executor_simulator.rs
```diff
@@ -81,9 +83,12 @@ pub enum VaaAttack {
...
     /// Deploys the executor program as upgradable
     pub fn new() -> ExecutorBench {
-        let mut bpf_data = read_file(find_file("remote_executor.so").unwrap_or_else(|| {
-            panic!("Unable to locate {}", "remote_executor.so");
-        }));
+        let mut bpf_data = read_file(
+            std::env::current_dir()
+                .unwrap()
+                .join(Path::new("../../target/deploy/remote_executor.so")),
+        );
```
guibescos, line 90: Newer versions of solana can't find the binary with the previous logic

.github/workflows/remote-executor.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           override: true
       - name: Install Solana
         run: |
-          sh -c "$(curl -sSfL https://release.solana.com/stable/install)"
+          sh -c "$(curl -sSfL https://release.solana.com/v1.14.18/install)"
```
guibescos, line 23: I had to bump this

governance/remote_executor/Cargo.toml
```diff
@@ -6,3 +6,6 @@ members = [
 
 [profile.release]
 overflow-checks = true
+
+[patch.crates-io]
+serde_wormhole = { git = "https://github.com/wormhole-foundation/wormhole", tag = "v2.17.1" }
```
guibescos, line 11: This patch.crate-io wormhole thing is getting out of control, wormhole needs to fix

wormhole_attester/Cargo.lock
guibescos, line 1: Attester cargo lock updated since attester governance cli uses remote-executor

governance/remote_executor/programs/remote-executor/src/tests/executor_simulator.rs
Reisen: Can you explain why?

governance/remote_executor/programs/remote-executor/Cargo.toml
```diff
@@ -18,11 +18,13 @@ pythtest = []
 
 [dependencies]
 anchor-lang = {version = "0.25.0", features = ["init-if-needed"]}
-wormhole-solana = { git = "https://github.com/guibescos/wormhole", branch = "reisen/sdk-solana"}
-wormhole-core = { git = "https://github.com/guibescos/wormhole", branch = "reisen/sdk-solana"}
+wormhole-solana = { git = "https://github.com/guibescos/wormhole-solana", rev="f14b3b54c1e37e1aaf8c2ac2a5e236832ffdb3c2"}
```
Reisen, line 21: We should really get away from this, maybe we can chat about what needs to be upstreamed. We're running into this in pythnet too .

governance/remote_executor/cli/src/main.rs
```diff
@@ -86,9 +94,10 @@ fn main() -> Result<()> {
 
             let signature_set_keypair = Keypair::new();
 
-            let vaa = VAA::from_bytes(vaa_bytes.clone())?;
+            let vaa: Vaa<&RawMessage> = serde_wormhole::from_slice(&vaa_bytes)?;
+            let (header, body): (Header, Body<&RawMessage>) = vaa.into();
```
Reisen, line 98: Nice nice nice.

governance/remote_executor/Cargo.toml
Reisen: Agreed, reasonable fix for now though. Let's ask for publishing.

governance/remote_executor/programs/remote-executor/Cargo.toml
guibescos: This is a temporary home.

governance/remote_executor/programs/remote-executor/src/tests/executor_simulator.rs
guibescos: Previous versions "1.13.x" would set an environment variable BPF_OUT_DIR and that's how the code would find the binary.
New versions "1.14.x" don't seems to set it, so `find_file` fails.

Reisen: Yeah I recall the `BPF_OUT_DIR` issues as well. Can you mention what might be the reason it's not doing it before we merge this? Don't worry if it takes too much time you can merge over this commit, it would just be nice to know what changed here for other solana related work.

Reisen (approved):

PR #844, opened 2023-05-25
Add serializers for Pyth formats

Reisen:
This PR introduces serde based serialization for Pyth. There are various places we currently hand-roll serialization/de-serialization but have naturally converged on a simple wire format in all places. These serializers capture that format and allow use to use the standard serde derives to do so. This comes with several benefits:

- It prevents us accidentally introducing inconsistencies into our wire formats.
- It gives us a single piece of code to audit.
- It gives us the ability to also get JSON, CBOR, YAML, TOML, or whatever other format we want for our types.
- It provides better error handling.


ali-bahjati (approved):
It's great!

pythnet/pythnet_sdk/src/wire.rs
```diff
@@ -0,0 +1,306 @@
...
+                nested_u8:    50,
+                nested_tuple: (51, 52),
+            },
+            t_prefixed:         vec![0u8; 512].into(),
+        };
+
+        golden_struct.serialize(&mut serializer).unwrap();
+
+        // The serialized output should be as expected.
+        assert_eq!(
```
Reisen, line 249: This isn't the prettiest way to test, and ideally we'd test variations of each value too (I.E checking true and false for bool and empty non-empty strings) but I will come back to do this after hermes is complete because it's a lot more tests to generate.

jayantk (approved):
very nice

pythnet/pythnet_sdk/src/wire.rs
```diff
@@ -0,0 +1,306 @@
...
+    minor_version: u8,
+    trailing:      &'a [u8],
+    proof:         v1::Proof<'a>,
+}
+
+// Proof Format (V1)
+// --------------------------------------------------------------------------------
+// The definitions within each module can be updated with append-only data without
+// requiring a new module to be defined. So for example, new accounts can be added
+// to the end of `AccumulatorAccount` without moving to a `v1`.
```
Reisen, line 52: Note that this is not correct, but Ali is modifying these to fit Hermes which is the right place for those changes anyway, so ignore these.

PR #852, opened 2023-05-29
Update readme with hosting recommendations

jayantk:


ali-bahjati (approved):
Thanks a lot! Can you also add the 4gb ram + 4 core requirement?

PR #853, opened 2023-05-30
Choose unique VAA for query time if it exists

jayantk:
The current get_vaa endpoint returns any vaa whose publishTime equals that of the query. However, there can be more than one such VAA, in which case the server picks arbitrarily. This PR changes the behavior to choose the unique VAA for that time based on lastAttestedPublishTime. 

Note that this PR is imperfect in that it only fixes the problem if the desired VAA is in the cache. If the VAA is not in the cache, we can return a different VAA (either from the cache or the database). We unfortunately can't solve that problem without changing the database schema to include lastAttestedPublishTime, which doesn't seem worthwhile right now. Let's just remember to add that column when indexing the accumulator data.

### DIFF
--- a/wormhole_attester/sdk/js/src/index.ts
+++ b/wormhole_attester/sdk/js/src/index.ts
@@ -34,6 +34,7 @@ export type PriceAttestation = {
   prevPublishTime: UnixTimestamp;
   prevPrice: string;
   prevConf: string;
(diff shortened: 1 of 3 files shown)


wormhole_attester/sdk/js/src/index.ts
ali-bahjati, line 1: Can you bump the version of this package?

PR #855, opened 2023-05-31
Ethereum/parse price feed updates accumulators

swimricky:
### Summary
Add support accumulator update data in `parsePriceFeeds`

### Changes
- Refactor `PythAccumulator.sol` and Pyth.sol` and add support for purely parsing
- Add test for parsePriceFeedUpdates with Accumulator update data

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
@@ -478,7 +478,7 @@ abstract contract Pyth is
                         ) {
                             priceFeeds[k].id = 0;
                         } else {
-                            setPriceFeedsInfo(
+                            fillPriceFeedsInfo(
                                 priceFeeds,
                                 k,
                                 accumulatorPriceId,
@@ -582,11 +582,12 @@ abstract contract Pyth is
     }
 
     function getPriceFeedsIndex(
-        bytes32[] memory priceIds,
+        bytes32[] calldata priceIds,
         bytes32 targetPriceId
     ) private pure returns (uint index) {
         uint k = 0;
-        for (; k < priceIds.length; k++) {
+        uint len = priceIds.length;
+        for (; k < len; k++) {
             if (priceIds[k] == targetPriceId) {
                 break;
             }
@@ -594,7 +595,7 @@ abstract contract Pyth is
         return k;
     }
 
-    function setPriceFeedsInfo(
+    function fillPriceFeedsInfo(
         PythStructs.PriceFeed[] memory priceFeeds,
         uint k,
         bytes32 priceId,
--- a/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
@@ -11,6 +11,7 @@ import "./PythAccumulator.sol";
 import "./PythGetters.sol";
 import "./PythSetters.sol";
 import "./PythInternalStructs.sol";
+import "forge-std/console.sol";
 
 abstract contract Pyth is
     PythGetters,
@@ -575,41 +576,21 @@ abstract contract Pyth is
             }
 
             priceFeeds = new PythStructs.PriceFeed[](priceIds.length);
-
             for (uint i = 0; i < updateData.length; i++) {
                 bytes memory encoded;
-                bool isAccumulator = false;
 
-                {
-                    IWormhole.VM memory vm;
-
-                    if (
-                        updateData[i].length > 4 &&
-                        UnsafeBytesLib.toUint32(updateData[i], 0) ==
-                        ACCUMULATOR_MAGIC
-                    ) {
-                        // [Ricky] - does simple wormhole VM parsing/validation
-                        vm = parseAndVerifyPythVM(updateData[i]);
-                        isAccumulator = true;
-                    } else {
-                        vm = parseAndVerifyBatchAttestationVM(updateData[i]);
-                    }
-                    encoded = vm.payload;
-                }
-
-                /** accumulator update logic */
-                /*
-                    offset = parseAccumulatorHeader;
-                */
-
-                if (isAccumulator) {
+                if (
+                    updateData[i].length > 4 &&
+                    UnsafeBytesLib.toUint32(updateData[i], 0) ==
+                    ACCUMULATOR_MAGIC
+                ) {
                     (
                         PythInternalStructs.PriceInfo[]
                             memory accumulatorPriceInfos,
                         bytes32[] memory accumulatorPriceIds
-                    ) = parsePricesUsingAccumulator(encoded);
+                    ) = parsePricesUsingAccumulator(updateData[i]);
 
-                    for (uint j = 0; j < accumulatorPriceIds.length; i++) {
+                    for (uint j = 0; j < accumulatorPriceIds.length; j++) {
                         bytes32 accumulatorPriceId = accumulatorPriceIds[j];
                         // check whether caller requested for this data
                         uint k = 0;
@@ -654,6 +635,16 @@ abstract contract Pyth is
                         }
                     }
                 } else {
+                    console.log("[parsePriceFeedUpdates] batch");
+
+                    {
+                        IWormhole.VM
+                            memory vm = parseAndVerifyBatchAttestationVM(
+                                updateData[i]
+                            );
+                        encoded = vm.payload;
+                    }
+
                     /** Batch price logic */
                     (
                         uint index,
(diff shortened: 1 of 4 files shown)


target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -521,6 +581,37 @@ abstract contract Pyth is
...
+        uint k = 0;
+        for (; k < priceIds.length; k++) {
+            if (priceIds[k] == targetPriceId) {
+                break;
+            }
+        }
+        return k;
+    }
+
+    function setPriceFeedsInfo(
```
ali-bahjati, line 597: can we rename it to `fillPriceFeedsInfo`? `set` might be a bit confusing (as it's used for state setters)

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -521,6 +581,38 @@ abstract contract Pyth is
         }
     }
 
+    function getPriceFeedsIndex(
```
jayantk, line 584: how about `findIndexOfPriceId`?

also this may have been inlined in the other implementation for gas efficiency reasons... @ali-bahjati will know more

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -432,84 +432,144 @@ abstract contract Pyth is
...
+                        uint publishTime = uint(info.publishTime);
+                        // Check the publish time of the price is within the given range
+                        // if it is not, then set the id to 0 to indicate that this price id
+                        // still does not have a valid price feed. This will allow other updates
+                        // for this price id to be processed.
+                        if (
+                            publishTime < minPublishTime ||
+                            publishTime > maxPublishTime
+                        ) {
+                            priceFeeds[k].id = 0;
```
jayantk, line 479: isn't this guaranteed to be 0 at this point (because line 463 implies this is zero, and you haven't populated it after that)?

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -105,29 +125,56 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
...
-            if (updateType == UpdateType.WormholeMerkle) {
-                updatePricesUsingWormholeMerkle(
-                    UnsafeBytesLib.slice(
-                        accumulatorUpdate,
-                        offset,
-                        accumulatorUpdate.length - offset
-                    )
-                );
-            } else {
+            if (updateType != UpdateType.WormholeMerkle) {
```
jayantk, line 128: This check seems like it's in the wrong place. If the updateType is a different type of payload, it's still a valid header. I think this type needs to be returned so you can do the check on line 56 above

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -42,11 +42,31 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
             revert PythErrors.InvalidUpdateDataSource();
     }
 
-    function updatePricesUsingAccumulator(
-        bytes calldata accumulatorUpdate
-    ) internal {
+    function parsePricesUsingAccumulator(
```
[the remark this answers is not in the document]
jayantk: maybe "extract" instead of "parse"? (I don't love that either, definitely open to other options.)

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -177,35 +224,49 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
...
+    ) private pure returns (uint endOffset, bytes memory encodedMessage) {
         unchecked {
             uint16 messageSize = UnsafeBytesLib.toUint16(encoded, offset);
             offset += 2;
 
-            bytes memory encodedMessage = UnsafeBytesLib.slice(
-                encoded,
-                offset,
-                messageSize
-            );
```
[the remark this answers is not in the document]
jayantk: wait, is `digest` the merkle tree root?  If so, it would be better to call it `merkleRoot`

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -286,37 +370,72 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
...
+
+            return offset;
+        }
+    }
+
+    function processMessage(
+        PythInternalStructs.PriceInfo memory info,
+        bytes32 priceId
+    ) private {
+        unchecked {
```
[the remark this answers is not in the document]
jayantk: actually a bunch of these functions don't seem to need unchecked?

jayantk (commented):
I left you some suggestions for improvements here. I think @ali-bahjati should definitely look at this too because iirc there are weird gas efficiency hacks that he's the most knowledgable about.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
ali-bahjati: I agree, in the old code we did the deserialization in place (for gas saving) and that's why we needed a reset here but here we don't need it.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -432,84 +432,144 @@ abstract contract Pyth is
...
+                if (
+                    updateData[i].length > 4 &&
+                    UnsafeBytesLib.toUint32(updateData[i], 0) ==
+                    ACCUMULATOR_MAGIC
+                ) {
+                    (
+                        PythInternalStructs.PriceInfo[]
+                            memory accumulatorPriceInfos,
+                        bytes32[] memory accumulatorPriceIds
+                    ) = parsePricesUsingAccumulator(updateData[i]);
```
ali-bahjati, line 445: Let's keep in mind that using big memories has a lot of gas usage and in the future PRs we should fix it.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -22,6 +22,9 @@ contract PythWormholeMerkleAccumulatorTest is
 {
     IPyth public pyth;
 
+    // -1 is equal to 0x111111 which is the biggest uint if converted back
```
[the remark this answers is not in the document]
ali-bahjati: my mistake on the another file :sweat:

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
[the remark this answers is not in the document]
ali-bahjati: I agree. Also some methods do more than parsing, they do verification too.

ali-bahjati (commented):
I think the code is great. It needs some improvements that Jayant has pointed out. Many of them are there because of my pervious code bad choices and I'm sorry about it. Some of them were gas optimisation considerations that we will do in a separate PR.

I think we can merge this once comments are addressed and you have added more tests.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
swimricky: yes that's correct thanks

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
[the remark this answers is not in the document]
swimricky: `MerkleTree.isProofValid(encoded, offset, digest, encodedMessage)` constructs the proof from `encoded` and where `encodedMessage` is the hash/leafData of the `PriceFeedMessage`


 yes `digest` is the merkle tree root i'll rename it for clarity along with some of the other parameters

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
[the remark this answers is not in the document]
swimricky: ill fix both thanks for catching this

Reisen:
Comment.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -46,6 +50,42 @@ contract PythWormholeMerkleAccumulatorTest is
         assertEq(emaPrice.publishTime, priceFeedMessage.publishTime);
     }
 
+    function assertParsedPriceFeed(
```
jayantk, line 53: assertPriceFeedEqualsMessage ?

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -205,6 +271,14 @@ contract PythWormholeMerkleAccumulatorTest is
...
 
         assertPriceFeedMessageStored(priceFeedMessages1[0]);
+
+        bytes32[] memory priceIds2 = new bytes32[](1);
+        priceIds2[0] = priceFeedMessages1[0].priceId;
+
+        PythStructs.PriceFeed[] memory priceFeeds2 = pyth.parsePriceFeedUpdates{
+            value: updateFee
+        }(updateData, priceIds2, 0, MAX_UINT64);
+        assertParsedPriceFeedStored(priceFeeds2[0]);
```
jayantk, line 281: doesn't this need to test that priceFeeds2 has length 1 also?

also if you swap the order of the updates in updateData, the code should return the other update? I think it's worth testing that behavior too.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -229,12 +303,22 @@ contract PythWormholeMerkleAccumulatorTest is
         pyth.updatePriceFeeds{value: updateFee}(updateData);
         assertPriceFeedMessageStored(priceFeedMessages1[0]);
 
+        bytes[] memory updateData1 = updateData;
```
jayantk, line 306: rename the updateData variable above instead of doing this assignmnet.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -229,12 +303,22 @@ contract PythWormholeMerkleAccumulatorTest is
...
         // Make sure that the old value is still stored
         assertPriceFeedMessageStored(priceFeedMessages1[0]);
+
+        bytes32[] memory priceIds = new bytes32[](1);
+        priceIds[0] = priceFeedMessages1[0].priceId;
+
+        PythStructs.PriceFeed[] memory priceFeeds = pyth.parsePriceFeedUpdates{
+            value: updateFee
+        }(updateData1, priceIds, 0, MAX_UINT64);
+        assertParsedPriceFeedStored(priceFeeds[0]);
```
jayantk, line 321: this test doesn't seem like it's doing anything useful -- it's testing that this function parses a single update, which is covered by many different tests already. i would delete the change here.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -421,7 +590,82 @@ contract PythWormholeMerkleAccumulatorTest is
...
+        uint numPriceFeeds = (getRandUint() % 10) + 1;
+        PriceFeedMessage[]
+            memory priceFeedMessages = generateRandomPriceFeedMessage(
+                numPriceFeeds
+            );
+        (
+            bytes[] memory updateData,
+            uint updateFee
+        ) = createWormholeMerkleUpdateData(priceFeedMessages);
+        bytes32[] memory priceIds2 = new bytes32[](numPriceFeeds);
```
jayantk, line 621: priceIds ? why priceIds2 ?

same with priceFeeds below

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -421,7 +590,82 @@ contract PythWormholeMerkleAccumulatorTest is
...
+
+        for (uint i = 0; i < priceFeeds2.length; i++) {
+            assertParsedPriceFeed(
+                priceFeeds2[i],
+                priceFeedMessages[i],
+                priceIds2[i]
+            );
+        }
+
+        // update priceFeedMessages
```
jayantk, line 637: I guess this is supposed to test that parsing the feeds doesn't depend on what's on chain? maybe leave a comment?

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -421,7 +590,82 @@ contract PythWormholeMerkleAccumulatorTest is
...
+        for (uint i = 0; i < numPriceFeeds; i++) {
+            priceFeedMessages[i].price = getRandInt64();
+            priceFeedMessages[i].conf = getRandUint64();
+            priceFeedMessages[i].expo = getRandInt32();
+
+            // Increase the publish time if it is not causing an overflow
+            if (priceFeedMessages[i].publishTime != type(uint64).max) {
+                priceFeedMessages[i].publishTime += 1;
```
jayantk, line 645: I think you should set the publishTime to be random -- it literally doesn't matter whether or not this is more recent.

jayantk (commented):
thanks this code is much clearer now. I think there's still some work to be done on the tests though -- see inline comments.

jayantk (commented):
thanks this code is much clearer now. I think there's still some work to be done on the tests though -- see inline comments.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -469,15 +469,13 @@ abstract contract Pyth is
...
-                        // still does not have a valid price feed. This will allow other updates
-                        // for this price id to be processed.
+                        // and only fill the priceFeedsInfo if it is.
+                        // If is not, default id value of 0 will still be set and
+                        // this will allow other updates for this price id to be processed.
                         if (
-                            publishTime < minPublishTime ||
-                            publishTime > maxPublishTime
+                            publishTime >= minPublishTime &&
+                            publishTime <= maxPublishTime
```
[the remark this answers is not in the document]
ali-bahjati: "you can't specify a zero-length interval" is wrong. We don't have zero-length on inclusive end too. 

I decided to have inclusive end interval because it's more intuitive with the minPublishTime and maxPublishTime.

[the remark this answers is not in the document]
ali-bahjati: "you can't specify a zero-length interval" is wrong. We don't have zero-length on inclusive end too. 

I decided to have inclusive end interval because it's more intuitive with the minPublishTime and maxPublishTime.

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -138,94 +184,150 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
                 // emit AccumulatorUpdate(vm.chainId, vm.sequence);
 
                 bytes memory encodedPayload = vm.payload;
-                uint payloadoffset = 0;
+                uint payloadOffset = 0;
```
ali-bahjati, line 187: thanks!

ali-bahjati (commented):
I have two comments that are mostly my opinions and would be great to discuss them:
- It was hard for me to follow the code flow and I think we don't need this many methods.
- I think although it's easier to add code to updatePriceFeeds tests to avoid duplication but it makes tests more complex for a new test reader and makes debugging more difficult. I believe each test should be as simple as possible and test only a single scenario. I would break the tests in `testParsePriceFeedWithWormholeMerkleWorks` in multiple tests and also will probably create a separate file for all the parse price feeds tests.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -167,6 +207,32 @@ contract PythWormholeMerkleAccumulatorTest is
         assertPriceFeedMessageStored(priceFeedMessages1[1]);
         assertPriceFeedMessageStored(priceFeedMessages2[0]);
         assertPriceFeedMessageStored(priceFeedMessages2[1]);
+
```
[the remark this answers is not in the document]
swimricky: There should be no difference between what is stored through `updatePriceFeeds` and what is returned by `parsePriceFeeds` if the updateData is the same aside from the following situation:

The reason that i'm only checking `priceFeedMessage2[1]` is b/c since both `priceFeedMessage1` and `priceFeedMessage2` are passed in the same `updateData` to `updatePriceFeeds` which should update the `priceFeed` information stored based on the latest `publishTime` per `priceFeedMessage.id`.
`parsePriceFeedUpdates` only parses the first occurrence per `priceFeedMessage.id` per `updateData` which is why i split the `updateData` & `priceIds` that i'm passing into `parsePriceFeedUpdates` and checking against.


This test originally was checking the implementation of `updatePriceFeeds` so I do still want to checks from `assertParsedPriceFeedStored`
I'll add comments to clarify these as well as moving the `parsePriceFeedsUpdates` prior to `updatePriceFeeds`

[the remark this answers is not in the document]
swimricky: just for background context, this behavior difference between parsePriceFeedUpdates & updatePriceFeeds was what I was referring to on slack yesterday

swimricky:
> I have two comments that are mostly my opinions and would be great to discuss them:
> 
> * It was hard for me to follow the code flow and I think we don't need this many methods.
> * I think although it's easier to add code to updatePriceFeeds tests to avoid duplication but it makes tests more complex for a new test reader and makes debugging more difficult. I believe each test should be as simple as possible and test only a single scenario. I would break the tests in `testParsePriceFeedWithWormholeMerkleWorks` in multiple tests and also will probably create a separate file for all the parse price feeds tests.

i'll clean up and reduce the number of methods in a separate PR for gas optimization

for your second point about breaking up the tests, i don't mind separating out parsing portion into their own tests. Since the setups between the `parse` and `update` versions  will be largely the same i'm not too sure if the gain in simplicity will really be that significant.  I think it would make sense to bundle them together in some of the tests just to highlight the differences between `updatePriceFeeds` and parsePriceFeedUpdates`

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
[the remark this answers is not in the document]
jayantk: I think the typical convention is >= and <, and following that will be more intuitive to most people.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -653,19 +675,154 @@ contract PythWormholeMerkleAccumulatorTest is
...
+        );
+
+        // Request for parse after the time range should revert.
+        vm.expectRevert(PythErrors.PriceFeedNotFoundWithinRange.selector);
+        pyth.parsePriceFeedUpdates{value: updateFee}(
+            updateData,
+            priceIds,
+            300,
+            MAX_UINT64
+        );
```
jayantk, line 823: test the other direction too (where the interval is before [100, 200])

jayantk (approved):
thanks this looks good to me. I agree with @ali-bahjati that it can be a little hard to understand the tests, but let's look at ways to refactor them to simplify in a subsequent PR. This one is already quite big.

target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -202,36 +203,59 @@ contract PythWormholeMerkleAccumulatorTest is
...
+        assertPriceFeedMessageStored(priceFeedMessages1[1]);
+        assertPriceFeedMessageStored(priceFeedMessages2[0]);
+        assertPriceFeedMessageStored(priceFeedMessages2[1]);
 
-        PythStructs.PriceFeed[] memory priceFeeds2 = pyth.parsePriceFeedUpdates{
-            value: updateFee
-        }(parseUpdateDataInput2, priceIds2, 0, MAX_UINT64);
-        for (uint i = 0; i < priceFeeds2.length; i++) {
-            assertParsedPriceFeedStored(priceFeeds2[i]);
+        PythStructs.PriceFeed[]
```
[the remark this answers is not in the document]
swimricky: the above block was testing the stored price information against the raw priceMessages, the below block is testing the stored price against the parsed price information. i'll add comments for clarity

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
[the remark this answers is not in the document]
ali-bahjati: idk :D maybe you are right and it's very personal. I agree that with array ranges (slices) `[start, end)` is the convention. but when there is min & max, I think `[min, max]` is more intuitive.

PR #858, opened 2023-06-06
[pythnet-sdk] Add merkleRoot

guibescos:
Adds an extra struct MerkleRoot that is just a tuple struct that wraps the actual root as a hash.
This struct is exported and is useful to check proofs when you don't have access to the element set of the tree.


jayantk (approved):

Reisen (approved):

Reisen:
It would be nice if we could `pub` the `MerkleAccumulator` trait methods as well actually but not a big deal, I can do that later, looks good to merge.

PR #867, opened 2023-06-08
Move message structures from pyth-client repo to pythnet-sdk

m30m:
All rust based smart contracts could leverage these structures for deserializing pyth messages. But having them in pyth-client makes all the smart contracts packages depend on irrelevant packages such as solana libraries which caused dependency conflicts

With these structs moved here it would be easier to reuse them in other places

This PR moves structures from [price.rs](https://github.com/pyth-network/pyth-client/blob/b09ee14721f1f20735c66db564326d654e1e552d/program/rust/src/accounts/price.rs#L36) without the `from_price_account` function since it is solana specific.


pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -0,0 +1,330 @@
...
+///
+/// The oracle is not using the Message enum due to the contract size limit and
+/// some of the methods for PriceFeedMessage and TwapMessage are not used by the oracle
+/// for the same reason. Rust compiler doesn't include the unused methods in the contract.
+/// Once we start using the unused structs and methods, the contract size will increase.
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "strum",
```
ali-bahjati, line 30: can you add strum as an optional dependency?

pythnet/pythnet_sdk/src/tests/test_message.rs
```diff
@@ -0,0 +1,96 @@
+use {
```
ali-bahjati, line 1: I know it was here on the client as well. But I think generally unit tests go within the same file as the definition and integration tests go to `tests` directory.

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -0,0 +1,330 @@
...
+    },
+    std::io::{
+        Cursor,
+        Read,
+    },
+};
+
+
+/// Message format for sending data to other chains via the accumulator program
+/// When serialized, each message starts with a unique 1-byte discriminator, followed by the
```
Reisen, line 15: Minor nitpick, maybe clarify this is with the PythNet serialization format, as this won't be true for any other serde based serializer like JSON.

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -0,0 +1,330 @@
...
+    // The size of the serialized message. Note that this is not the same as the size of the struct
+    // (because of the discriminator & struct padding/alignment).
+    pub const MESSAGE_SIZE: usize = 1 + 32 + 8 + 8 + 4 + 8 + 8 + 8 + 8;
+    pub const DISCRIMINATOR: u8 = 0;
+
+    /// Serialize this message as an array of bytes (including the discriminator)
+    /// Note that it would be more idiomatic to return a `Vec`, but that approach adds
+    /// to the size of the compiled binary (which is already close to the size limit).
+    #[allow(unused_assignments)]
+    pub fn to_bytes(self) -> [u8; Self::MESSAGE_SIZE] {
```
Reisen, line 128: I'd say It's fine to keep it this way, but just a thought -- the types being here do make sense, but the solana specific serialization of it might make sense to keep in the contract. At least right now this is here because of solana constraints but on other chains we might be happier using borsh or serde. I'm not particularly die hard about it but wanted to make the point.

guibescos: I think I agree with this point, we hand rolled the serialization in the contract because we needed to optimize for binary size, but that should not be the case in other uses of this library

Reisen: In fact, having just skimmed through all the serialization functions, they all match what we defined in our serde implementation now, so these could I believe definitely stay in Solana and `pythnet_sdk::wire::Serializer` is capable of doing this everywhere else now.

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -0,0 +1,330 @@
...
+        feature = "serde",
+        strum_discriminants(derive(serde::Serialize, serde::Deserialize))
+    )
+)]
+pub enum Message {
+    PriceFeedMessage(PriceFeedMessage),
+    TwapMessage(TwapMessage),
+}
+
+#[allow(dead_code)]
```
guibescos, line 51: Wonder if you need this here

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -0,0 +1,155 @@
...
+    strum_discriminants(derive(
+        Hash,
+        strum::EnumIter,
+        strum::EnumString,
+        strum::IntoStaticStr,
+        strum::ToString,
+    )),
+    cfg_attr(
+        feature = "serde",
+        strum_discriminants(derive(serde::Serialize, serde::Deserialize))
```
ali-bahjati, line 36: We can add move these out too (because serde is a dependency)

ali-bahjati (approved):
Looks great!

PR #870, opened 2023-06-09
Encapsulate parameters to proposeX methods

jayantk:
This is a quick refactor to make this code a little easier to work with before I go about adding more functionality. The code right now instantiates a bunch of parameters in several different places (e.g., the SquadsMesh), which are needed to call `proposeInstructions`. This change makes a class to hold all of these parameters so we can instantiate them all one time instead of repeating the code. 

I suggest looking at `propose.ts` first. The other changes are pretty mechanical.


governance/xc_admin/packages/xc_admin_common/src/propose.ts
```diff
@@ -30,198 +36,282 @@ type SquadInstruction = {
...
+  public vault: PublicKey;
+
+  constructor(
+    wallet: Wallet,
+    cluster: PythCluster,
+    squad: SquadsMesh,
+    vault: PublicKey
+  ) {
+    this.wallet = wallet;
+    this.cluster = cluster;
```
guibescos, line 62: I think you might need `this.cluster = getMultisigCluster(cluster)`

guibescos (approved):
LGTM, I tested on devnet and pythtest-crosschain

governance/xc_admin/packages/xc_admin_common/src/propose.ts
```diff
@@ -30,198 +36,282 @@ type SquadInstruction = {
...
+      return mapKey(localAuthorityPDA);
+    }
+  }
+
+  public wormholeAddress(): PublicKey | undefined {
+    // TODO: we should configure the wormhole address as a vault parameter.
+    return WORMHOLE_ADDRESS[this.cluster];
+  }
+
+  // TODO: does this need a cluster argument?
```
guibescos, line 95: Why would it need a cluster argument

PR #872, opened 2023-06-09
[message-buffer] improve logging for setup message buffer script

swimricky:


ali-bahjati (approved):
One small ask. Can you add replace yarn package.json script to `npx ts-node scripts/setup_message_buffer.ts`? 😅 i missed it in the PR switching to npm.

PR #874, opened 2023-06-10
[price_pusher] Injective: cache sequence number

0xfirefist:


price_pusher/src/injective/injective.ts
```diff
@@ -116,12 +118,14 @@ export class InjectivePricePusher implements IPricePusher {
 
   private async signAndBroadcastMsg(msg: Msgs): Promise<TxResponse> {
     const chainGrpcAuthApi = new ChainGrpcAuthApi(this.grpcEndpoint);
-    const account = await chainGrpcAuthApi.fetchAccount(
+    // Fetch the latest account details only if it's not stored.
+    this.account ??= await chainGrpcAuthApi.fetchAccount(
```
ali-bahjati, line 122: wouldn't be better to periodically sync with the rpc on this? (might be helpful if multiple instances of one key are running)

ali-bahjati (approved):
Nice! I don't have the right context on how it is being used. I left one inline comment that is related to the usage.

jayantk (approved):

price_pusher/src/injective/injective.ts
0xfirefist: we are refetching it if the sequence number is a wrong one. else we won't need to refetch it. 
in that sense it is syncing with the rpc in some way.

PR #878, opened 2023-06-12
[ethereum] - charge updateFee per number of updates

swimricky:
### Summary
- change `getUpdateFee` to calculate the total fee based on the numUpdates for accumulator updates. Still uses same logic for batch price
- lowers optimizer_runs to 2000 in order to get the bytesize down
    - optimizer_runs = 5000 had deployment size of 25046
    image
diff between optimizer_runs = 2000 + this branch and main
image
- `getUpdateFeeWhMerkleX` is high b/c the original implementation returned singleUpdateFee * updateData.length
### Note
- Will make separate PR for optimizing bytesize further and increasing optimizer_runs to get as close to original 5000 runs

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
@@ -624,12 +624,6 @@ abstract contract Pyth is
         return totalNumUpdates * singleUpdateFeeInWei();
     }
 
-    function getFeeBatch(
-        uint updateDataLength
-    ) private view returns (uint requiredFee) {
-        return updateDataLength * singleUpdateFeeInWei();
-    }
-
     function findIndexOfPriceId(
         bytes32[] calldata priceIds,
         bytes32 targetPriceId
--- a/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
@@ -71,16 +71,15 @@ abstract contract Pyth is
     function updatePriceFeeds(
         bytes[] calldata updateData
     ) public payable override {
-        // TODO: Is this fee model still good for accumulator?
-        uint requiredFee = getUpdateFee(updateData);
-        if (msg.value < requiredFee) revert PythErrors.InsufficientFee();
-
+        uint8 totalNumUpdates = 0;
         for (uint i = 0; i < updateData.length; ) {
             if (
                 updateData[i].length > 4 &&
                 UnsafeBytesLib.toUint32(updateData[i], 0) == ACCUMULATOR_MAGIC
             ) {
-                updatePriceInfosFromAccumulatorUpdate(updateData[i]);
+                totalNumUpdates += updatePriceInfosFromAccumulatorUpdate(
+                    updateData[i]
+                );
             } else {
                 updatePriceBatchFromVm(updateData[i]);
             }
@@ -89,6 +88,8 @@ abstract contract Pyth is
                 i++;
             }
         }
+        uint requiredFee = getRequiredFee(totalNumUpdates, updateData.length);
+        if (msg.value < requiredFee) revert PythErrors.InsufficientFee();
     }
 
     /// This method is deprecated, please use the `getUpdateFee(bytes[])` instead.
@@ -101,7 +102,30 @@ abstract contract Pyth is
     function getUpdateFee(
         bytes[] calldata updateData
     ) public view override returns (uint feeAmount) {
-        return singleUpdateFeeInWei() * updateData.length;
+        uint8 totalNumUpdates = 0;
+        for (uint i = 0; i < updateData.length; i++) {
+            if (
+                updateData[i].length > 4 &&
+                UnsafeBytesLib.toUint32(updateData[i], 0) == ACCUMULATOR_MAGIC
+            ) {
+                (
+                    uint offset,
+                    UpdateType updateType
+                ) = extractUpdateTypeFromAccumulatorHeader(updateData[i]);
+                if (updateType != UpdateType.WormholeMerkle) {
+                    revert PythErrors.InvalidUpdateData();
+                }
+                totalNumUpdates += parseWormholeMerkleHeaderNumUpdates(
+                    updateData[i],
+                    offset
+                );
+            }
+        }
+        if (totalNumUpdates > 0) {
+            return totalNumUpdates * singleUpdateFeeInWei();
+        } else {
+            return updateData.length * singleUpdateFeeInWei();
+        }
     }
 
     function verifyPythVM(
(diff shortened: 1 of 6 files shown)


target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -71,16 +71,15 @@ abstract contract Pyth is
     function updatePriceFeeds(
         bytes[] calldata updateData
     ) public payable override {
-        // TODO: Is this fee model still good for accumulator?
-        uint requiredFee = getUpdateFee(updateData);
-        if (msg.value < requiredFee) revert PythErrors.InsufficientFee();
-
+        uint8 totalNumUpdates = 0;
```
swimricky, line 74: @ali-bahjati is `uint8` enough for holding the total number of updates we could get in one call?

ali-bahjati: yeah i'm not sure we can fit more gas-wise (the block gas limit). That's why on wire format it's u8. Do you think people will need more?

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -101,7 +102,30 @@ abstract contract Pyth is
...
+                    updateData[i],
+                    offset
+                );
+            }
+        }
+        if (totalNumUpdates > 0) {
+            return totalNumUpdates * singleUpdateFeeInWei();
+        } else {
+            return updateData.length * singleUpdateFeeInWei();
+        }
```
ali-bahjati, line 128: You can make an else statement above and add totalNumUpdates by 1 if it's not accumulator. It will be cleaner.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -425,52 +449,45 @@ abstract contract Pyth is
...
-                    );
-
+                    bytes memory encoded;
                     (
                         offset,
                         digest,
-                        numUpdates
-                    ) = extractWormholeMerkleHeaderDigestAndNumUpdates(encoded);
+                        numUpdates,
+                        encoded
```
jayantk, line 482: I don't like that this method returns `encoded` now -- it makes the method signature for this function extremely confusing.

BTW I find the overall handling of byte arrays in this code to be confusing. You're using two different techniques to slice out bytes (creating new arrays pointing to slices + offset indexes), so it's very hard for the reader to understand what an index into an array represents.

Imo the best way to handle this would be to have 1 byte array (the whole updateData) and 1 offset that increments as bytes are consumed.

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -170,6 +185,19 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
         }
     }
 
+    function parseWormholeMerkleHeaderNumUpdates(
```
jayantk, line 188: IMO it's better to reuse the method above instead of adding this. I don't think getUpdateFee needs to be efficient now, since it's only used by off-chain callers.

jayantk (commented):
great! I have a couple suggestions to improve this, but should be quick to do.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
jayantk: (please don't address the comment about overall handling of byte arrays in this PR. You can do that separately if you'd like)

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
swimricky: this would be for summing up all the individual `numUpdates` so depends on how big `updateData` could potentially be. I think for now 255 would be enough but we could possibly need more in the future?

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
swimricky: changing this to return `encoded` was a workaround for hitting the stack too deep error. 

I spoke to @ali-bahjati regarding slicing out bytes vs just reading offsets and he said  the implementation being used should be dependent on whether or not we're "throwing away" the data or need to possibly re-read it. I agree that there's areas where the implementation could be more consistent but there's also a few bytesize & gas cost considerations I'm also taking into account. I'll revisit this in a separate PR

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
ali-bahjati: oh ok ok. I might just change it to uint. Since word-size in evm is 32 bytes it shouldn't change much gas-wise.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -89,6 +88,8 @@ abstract contract Pyth is
                 i++;
             }
         }
+        uint requiredFee = getRequiredFee(totalNumUpdates, updateData.length);
```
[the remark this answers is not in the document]
swimricky: how about this instead?
```
 if (totalNumUpdates == 0){
          totalNumUpdates = updateData.length;
 }
 requiredFee = getTotalFee(totalNumUpdates);
```
the signatures between `getFeeAccumulator` & `getFeeBatch` are the same so for saving on bytesize maybe it would be best to not have 2 separate `getFeeXXX` methods?

swimricky:
@ali-bahjati @jayantk 
would not changing the batch price attestation `updateFee` to also be based on `numAttestations` possibly cause users to continue using batch prices  due to cheaper fee?

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
jayantk: ah yes good point about that on-chain usage.

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -591,9 +609,27 @@ abstract contract Pyth is
...
         }
     }
 
+    function getTotalFee(
+        uint totalNumUpdates
+    ) private view returns (uint requiredFee) {
+        return totalNumUpdates * singleUpdateFeeInWei();
+    }
+
+    function getFeeBatch(
```
jayantk, line 627: this is unused now

target_chains/ethereum/contracts/forge-test/GasBenchmark.t.sol
```diff
@@ -391,7 +394,27 @@ contract GasBenchmark is Test, WormholeTestUtils, PythTestUtils {
         pyth.getEmaPrice(priceIds[0]);
     }
 
-    function testBenchmarkGetUpdateFee() public view {
+    function testBenchmarkGetUpdateFeeWhBatch() public view {
         pyth.getUpdateFee(freshPricesWhBatchUpdateData);
     }
+
+    function testBenchrmarkGetUpdateFeeWhMerkle1() public view {
```
jayantk, line 401: typo: benchrmark. also repeated below

jayantk (approved):

PR #883, opened 2023-06-13
[hermes] Add get_price_feed endpoint

ali-bahjati:


jayantk (approved):
lgtm. one thing is i also added the `target_chain=` query param to this method and a couple others (can be done in a separate pr though)

PR #884, opened 2023-06-13
[eth]: add forward compatibility test for accumulator updateData

swimricky:
### Summary
- add tests for checking forward compatibility in `parsePriceFeedUpdates` & `updatePriceFeeds`


target_chains/ethereum/contracts/forge-test/Pyth.WormholeMerkleAccumulator.t.sol
```diff
@@ -85,6 +85,24 @@ contract PythWormholeMerkleAccumulatorTest is
         assertEq(emaPrice.publishTime, priceFeed.emaPrice.publishTime);
     }
 
+    function assertParsedPriceFeed(
+        PythStructs.PriceFeed memory priceFeeds1,
+        PythStructs.PriceFeed memory priceFeed2
+    ) internal {
```
ali-bahjati, line 91: There is a typo here (priceFeeds1). Can we also make it something like assertPriceFeedEqual? it is comparing two price feeds

PR #885, opened 2023-06-13
[xc-admin] Contract management tool

jayantk:
I've been experimenting with different ways to read / write the state of contracts, and I have an approach that seems reasonable. The idea is to make a collection of Contract objects, each of which represents a single deployed on-chain contract. There can be separate types of Contracts to represent different types of programs (target chain vs. oracle program vs. ...). Each contract exposes its state as a json object which can be both read & written. You can then update contract state by manipulating the JSON object and calling the write method. The JSON interface makes this approach very flexible, as we can change its schema to fit whatever kind of contract we have.

Writing the state back to the chain is a little complicated because we need to run operations that pause and wait for external input (e.g., multisig approvals). I've implemented this using a filesystem dependency, which is not perfect, but is how we've solved this problem in all of our previous implementations. I think we can circle back and fix that dependency later by exposing sufficient APIs on all of our various components to let this code directly query the real world state.

There's also a CLI that lets you query and update the state. The upshot of all of this is that you'll be able to make a change to a set of contracts by running a command like:

`npm run cli -- set --network avalanche --type evmTargetChain validTimePeriod=30` 

then approving the multisig tx and rerunning the above command from the same computer. You can rerun the command whenever and it'll do something sane (e.g., if you approve some proposals but not others, it'll submit the approved changes and wait for the remaining pending changes).

There's still a lot of work to do to finish this up, but I wanted to share to get feedback on the general idea. I'll fix things up in smaller changes as we go.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_cli/package.json
+++ b/governance/xc_admin/packages/xc_admin_cli/package.json
@@ -17,7 +17,7 @@
   "scripts": {
     "build": "tsc",
     "format": "prettier --write \"src/**/*.ts\"",
-    "cli": "npm run build && node lib/contracts/cli.js"
+    "cli": "npm run build && node lib/cli.js"
   },
   "dependencies": {
     "@coral-xyz/anchor": "^0.26.0",
--- a/governance/xc_admin/packages/xc_admin_cli/package.json
+++ b/governance/xc_admin/packages/xc_admin_cli/package.json
@@ -16,7 +16,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "format": "prettier --write \"src/**/*.ts\""
+    "format": "prettier --write \"src/**/*.ts\"",
+    "cli": "npm run build && node lib/contracts/cli.js"
   },
   "dependencies": {
     "@coral-xyz/anchor": "^0.26.0",
--- a/governance/xc_admin/packages/xc_admin_cli/src/cli.ts
+++ b/governance/xc_admin/packages/xc_admin_cli/src/cli.ts
@@ -1,5 +1,6 @@
 import { program } from "commander";
-import { loadContractConfig, ContractType } from "xc_admin_common";
+import { loadContractConfig, ContractType, SyncOp } from "xc_admin_common";
+import * as fs from "fs";
 
 const contractsConfig = [
   {
@@ -41,6 +42,7 @@ const networksConfig = {
   },
 };
 
+// TODO: we will need configuration of this stuff to decide which multisig to run.
 const multisigs = [
   {
     name: "",
@@ -88,9 +90,38 @@ program
     }
   });
 
+class Cache {
+  private path: string;
+
+  constructor(path: string) {
+    this.path = path;
+  }
+
+  private opFilePath(op: SyncOp): string {
+    return `${this.path}/${op.id()}.json`;
+  }
+
+  public readOpCache(op: SyncOp): Record<string, any> {
+    const path = this.opFilePath(op);
+    if (fs.existsSync(path)) {
+      return JSON.parse(fs.readFileSync(path).toString("utf-8"));
+    } else {
+      return {};
+    }
+  }
+
+  public writeOpCache(op: SyncOp, cache: Record<string, any>) {
+    fs.writeFileSync(this.opFilePath(op), JSON.stringify(cache));
+  }
+
+  public deleteCache(op: SyncOp) {
+    fs.rmSync(this.opFilePath(op));
+  }
+}
+
 program
   .command("set")
-  .description("Set a configuration parameter for one or more Pyth contract")
+  .description("Set a configuration parameter for one or more Pyth contracts")
   .option("-n, --network <network-id>", "Find contracts on the given network")
   .option("-a, --address <address>", "Find contracts with the given address")
   .option("-t, --type <type-id>", "Find contracts of the given type")
@@ -127,7 +158,20 @@ program
       ops.push(...(await contract.sync(state)));
     }
 
-    console.log(ops);
+    // TODO: extract constant
+    const cacheDir = "cache";
+    fs.mkdirSync(cacheDir, { recursive: true });
+    const cache = new Cache(cacheDir);
+
+    for (const op of ops) {
+      const opCache = cache.readOpCache(op);
+      const isDone = await op.run(opCache);
+      if (isDone) {
+        cache.deleteCache(op);
+      } else {
+        cache.writeOpCache(op, opCache);
+      }
+    }
   });
 
 program.parse();
--- a/governance/xc_admin/packages/xc_admin_cli/src/cli.ts
+++ b/governance/xc_admin/packages/xc_admin_cli/src/cli.ts
@@ -1,7 +1,5 @@
-import { ContractType } from "xc_admin_common/lib/contracts/Contract";
-import { loadContractConfig } from "xc_admin_common/lib/contracts/config";
+import { loadContractConfig, ContractType } from "xc_admin_common";
 
-// TODO: load these from files
 const contractsConfig = [
   {
     type: ContractType.EvmPythUpgradable,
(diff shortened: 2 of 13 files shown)


governance/xc_admin/packages/xc_admin_frontend/next.config.js
```diff
@@ -5,6 +5,7 @@ const nextConfig = {
     externalDir: true,
   },
   webpack(config) {
+    config.experiments = { asyncWebAssembly: true }
```
jayantk, line 8: here and below is fixing some weird build issues from pulling the xc-governance-sdk into xc-admin & the frontend components.

governance/xc_admin/packages/xc_admin_common/src/contracts/Contract.ts
```diff
@@ -0,0 +1,139 @@
+import { ChainId, Instruction } from "@pythnetwork/xc-governance-sdk";
```
guibescos, line 1: Do you see any problem in using `PythGovernanceAction` instead? `Instruction` from `@pythnetwork/xc-governance-sdk`. Goal is enumerating all the available governance actions in one place and using the same class for deserializing them in the proposal page.

guibescos (commented):
Overall it makes sense to me. It seems an amazing improvement for setting up the config of many contracts on many chains at once.
What are your thoughts about having an evm relayer that relays any governance message, much like in Pythnet.

governance/xc_admin/packages/xc_admin_cli/src/cli.ts
```diff
@@ -0,0 +1,178 @@
...
+      url: "https://api.avax.network/ext/bc/C/rpc",
+    },
+    canto: {
+      url: "https://canto.gravitychain.io",
+    },
+  },
+};
+
+// TODO: we will need configuration of this stuff to decide which multisig to run.
+const multisigs = [
```
guibescos, line 47: I'd reuse https://github.com/pyth-network/pyth-crosschain/blob/main/governance/xc_admin/packages/xc_admin_common/src/multisig.ts

governance/xc_admin/packages/xc_admin_common/src/contracts/Contract.ts
```diff
@@ -0,0 +1,139 @@
...
+  Oracle,
+  EvmPythUpgradable,
+  EvmWormholeReceiver,
+}
+
+/**
+ * A unique identifier for a blockchain. Note that we cannot use ChainId for this, as ChainId currently reuses
+ * some ids across mainnet / testnet chains (e.g., ethereum goerli has the same id as ethereum mainnet).
+ */
+export type NetworkId = string;
```
guibescos, line 14: Cartesian product of wormhole network + chainid?

jayantk: we could use that tuple, but it's going to get confusing, especially as we start rolling out stable/edge configurations to each network. In that case ethereum goerli stable will be `[ethereum_goerli, mainnet]` and ethereum goerli edge will be `[ethereum, testnet]`. I think it's better to make a clean break here and define the right concept.

governance/xc_admin/packages/xc_admin_cli/package.json
```diff
@@ -16,7 +16,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "format": "prettier --write \"src/**/*.ts\""
+    "format": "prettier --write \"src/**/*.ts\"",
+    "cli": "npm run build && node lib/cli.js"
```
ali-bahjati, line 20: This command is a bit brittle as the build is not stanalone in the monorepo. Can you either run build using `npx lerna run build --scope="xc_admin_cli" --include-dependencies && ..`

governance/xc_admin/packages/xc_admin_common/src/contracts/Contract.ts
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * Run this operation from a previous state (recorded in cache). The operation can modify cache
+   * to record progress, then returns true if the operation has completed. If this function returns false,
+   * it is waiting on an external operation to complete (e.g., a multisig transaction to be approved).
+   * Re-run this function again once that operation is completed to continue making progress.
+   *
+   * The caller of this function is responsible for preserving the contents of `cache` between calls to
+   * this function.
+   */
+  run(cache: Record<string, any>): Promise<boolean>;
```
ali-bahjati, line 69: You would need to somehow expose the cache changes back to the caller

governance/xc_admin/packages/xc_admin_common/src/contracts/EvmPythUpgradable.ts
```diff
@@ -0,0 +1,98 @@
...
+        new SendGovernanceInstruction(
+          value,
+          authority,
+          this.submitGovernanceVaa
+        )
+    );
+  }
+
+  public async submitGovernanceVaa(vaa: string): Promise<boolean> {
+    // FIXME: also needs the full PythUpgradable ABI
```
ali-bahjati, line 88: You will just need to single abi to executeGovernanceInstruction. You can manually write that down and it's fine. I think full abi is not needed at all.

governance/xc_admin/packages/xc_admin_common/src/contracts/config.ts
```diff
@@ -0,0 +1,58 @@
...
+        contractConfig.address,
+        PythAbi,
+        getEvmProvider(contractConfig.networkId, networksConfig)
+      );
+
+      return new EvmPythUpgradable(
+        contractConfig.networkId,
+        contractConfig.address,
+        ethersContract
+      );
```
ali-bahjati, line 39: It's easier to understand (and less coupled) if this be a factory method in the contract file. And in general we automatically call the factory methods without having switch case at all.

governance/xc_admin/packages/xc_admin_cli/src/cli.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  public writeOpCache(op: SyncOp, cache: Record<string, any>) {
+    fs.writeFileSync(this.opFilePath(op), JSON.stringify(cache));
+  }
+
+  public deleteCache(op: SyncOp) {
+    fs.rmSync(this.opFilePath(op));
+  }
+}
+
```
ali-bahjati, line 122: I think it's better to move the cache to the common package. Ideally the CLI doesn't have any business logic and storage is kind of a business logic. Moving it there allows us to use it in other places if needed too.

governance/xc_admin/packages/xc_admin_cli/src/cli.ts
```diff
@@ -0,0 +1,178 @@
...
+program
+  .command("set")
+  .description("Set a configuration parameter for one or more Pyth contracts")
+  .option("-n, --network <network-id>", "Find contracts on the given network")
+  .option("-a, --address <address>", "Find contracts with the given address")
+  .option("-t, --type <type-id>", "Find contracts of the given type")
+  .argument("<fields...>", "Fields to set on the given contracts")
```
ali-bahjati, line 129: I think having the contract state written in a file is a better idea and easier to understand. Instead of set and get we can have a `sync` method that checks the written file is a match. The current way is prone to mistakes and requires looking at the source.

ali-bahjati (approved):
I like the design and approve it to get you unblocked.

It seems that this code is no utilizing what xc-admin is offering. I think as Guillermo said we can get use of xc-admin structures in the code and it will help with parsing the proposals and displaying them. 

My general thinking is that the logics here should be moved to a separate package and only the wire format live in the xc-admin-common to have a better boundry.

I left some inline comments which are mostly my opinions about the structure of the code. I might be wrong but would be good if you address them in later PRs. 

Lastly, another cool feature that is good if the CL offer is checking the balance of the our deployer key in different networks.

PR #886, opened 2023-06-13
[eth] more update/parsePriceFeed optimizations

swimricky:
### Summary
- changed parameter type from `bytes calldata` to `bytes memory` for some functions
- changed `parsePriceFeedMessage` to use offset instead of making a slice/copy

### Diff
Before
image
After
image

After @ali-bahjati change in #888 
image

### DIFF
--- a/target_chains/ethereum/contracts/contracts/libraries/MerkleTree.sol
+++ b/target_chains/ethereum/contracts/contracts/libraries/MerkleTree.sol
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.0;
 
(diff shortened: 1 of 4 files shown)


swimricky:
A bit paranoid about the changes of `bytes calldata` to `bytes memory` since it seems from every article/suggestion online that `bytes calldata` should  be cheaper ([see memory vs calldata](https://coinsbench.com/advanced-gas-optimizations-tips-for-solidity-85c47f413dc5)

ali-bahjati (approved):
Nice!

PR #888, opened 2023-06-14
[eth] Add unsafe calldata bytes lib

ali-bahjati:


target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -609,12 +611,6 @@ abstract contract Pyth is
                     revert PythErrors.PriceFeedNotFoundWithinRange();
                 }
             }
-
-            {
-                uint requiredFee = getTotalFee(totalNumUpdates);
-                if (msg.value < requiredFee)
-                    revert PythErrors.InsufficientFee();
-            }
```
ali-bahjati, line 617: I ended up removing it because the change to use calldata instead of memory uses 2 stack variables and we reached the limit. So I needed to remove totalNumUpdates variable.

swimricky (approved):
Amazing thanks for this!

PR #891, opened 2023-06-15
Avoid EnumAccess triggering `visit_u64` in Serde Visitor.

Reisen:
This PR works around a WASM bug caused by accidental floating point behaviour triggered by Serde when deserializing an enum variant. The relevant comment can be found in the `de.rs` file but the comment is copied here for PR visibility:

```
        // When serializing/deserializing, serde passes a variant_index into the handlers. We
        // currently write these as u8's and have already parsed' them during deserialize_enum
        // before we reach this point.
        //
        // Normally, when deserializing enum tags from a wire format that does not match the
        // expected size, we would use a u*.into_deserializer() to feed the already parsed
        // result into the visit_u64 visitor method during `__Field` deserialize.
        //
        // The problem with this however is during `visit_u64`, there is a possibility the
        // enum variant is not valid, which triggers Serde to return an `Unexpected` error.
        // These errors have the unfortunate side effect of triggering Rust including float
        // operations in the resulting binary, which breaks WASM environments.
        //
        // To work around this, we rely on the following facts:
        //
        // - variant_index in Serde is always 0 indexed and contiguous
        // - transmute_copy into a 0 sized type is safe
        // - transmute_copy is safe to cast into __Field as long as u8 >= size_of::<__Field>()
        //
        // This behaviour relies on serde not changing its enum deserializer generation, but
        // this would be a major backwards compatibility break for them so we should be safe.
```

### DIFF
--- a/pythnet/pythnet_sdk/src/wire.rs
+++ b/pythnet/pythnet_sdk/src/wire.rs
@@ -364,6 +364,126 @@ mod tests {
         );
     }
 
+    #[test]
+    #[rustfmt::skip]
+    /// This method tests that our EnumAccess workaround does not violate any memory safety rules.
+    /// In particular we want to make sure we avoid transmuting to any type that is not a u8 (we do
+    /// not support > 255 variants anyway).
+    fn test_serde_enum_access_behaviour() {
+        use serde::Deserialize;
+        use serde::Serialize;
+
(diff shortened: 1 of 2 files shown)


ali-bahjati (approved):

ali-bahjati (approved):
Very nice! Thanks for testing it extensively!

PR #893, opened 2023-06-15
[price-service] Update readiness probe

ali-bahjati:
I have made some changes to the liveness that should also be in readiness too. This also refactors and explains the rationale behind those. Liveness is stricter than readiness but some people might use readiness as health check in other infrastructures so it's good to update it.

### DIFF
--- a/price_service/server/src/listen.ts
+++ b/price_service/server/src/listen.ts
@@ -442,7 +442,7 @@ export class Listener implements PriceStore {
 
     // Check whether we have stale feeds and report false if we have too many.
     // This is not related to price service and is probably related to
-    // upstream dependencies.(spy, wormhole, attester, ...). However, as this
+    // upstream dependencies (like spy, wormhole, attester, ...). However, as this
     // has been mostly related to the spy and spy doesn't get recover from this
     // automatically, we are adding it to the price service readiness to stop
     // serving traffic from this unrecoverable state..
--- a/price_service/server/src/listen.ts
+++ b/price_service/server/src/listen.ts
@@ -440,6 +440,30 @@ export class Listener implements PriceStore {
       return false;
     }
 
+    // Check whether we have stale feeds and report false if we have too many.
+    // This is not related to price service and is probably related to
+    // upstream dependencies.(spy, wormhole, attester, ...). However, as this
+    // has been mostly related to the spy and spy doesn't get recover from this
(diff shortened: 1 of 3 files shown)


jayantk (approved):
code change seems fine to me. I don't understand the difference between liveness / readiness as used here though, so would be good to make sure @thmzlt agrees with this change.

price_service/server/src/rest.ts
```diff
@@ -514,7 +514,7 @@ export class RestAPI {
     endpoints.push("/api/stale_feeds?threshold=<staleness_threshold_seconds>");
 
     app.get("/ready", (_, res: Response) => {
-      if (this.isReady!()) {
+      if (this.isReady === undefined || this.isReady!()) {
```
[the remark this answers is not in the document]
ali-bahjati: It is not passed in the testing. it's not a good idea that it be optional but i thought I make it correct.

price_service/server/src/listen.ts
```diff
@@ -440,6 +440,30 @@ export class Listener implements PriceStore {
       return false;
     }
 
+    // Check whether we have stale feeds and report false if we have too many.
```
thmzlt, line 443: I think we can reword this comment to say something like "if too many price feeds are stale it probably means that the price service is not receiving messages from Wormhole at all and is essentially dead".

price_service/server/src/rest.ts
```diff
@@ -523,31 +523,17 @@ export class RestAPI {
...
-        }
-      }
-
-      if (
-        priceIds.length < minimumNumPrices ||
-        stalePriceCnt > stalePriceTreshold
-      ) {
-        res.sendStatus(StatusCodes.SERVICE_UNAVAILABLE);
-      } else {
+      // Reuse the same readiness check for liveness.
```
thmzlt, line 526: Just an observation, but I think there are some assumptions about how the price service is deployed in this comment.

thmzlt (approved):

price_service/server/src/listen.ts
ali-bahjati: 👍

price_service/server/src/rest.ts
ali-bahjati: Right. I reverted the liveness back to be always Ok. instead in our infra we will use readiness check for liveness. This seems better.

PR #894, opened 2023-06-15
[hermes] Move GuardianSet sourcing from Ethereum to PythNet

Reisen:
The code currently relies on reading GuardianSet state updates from Ethereum, this PR switches that to relying on reading GuardianSet accounts from Wormhole deployed on PythNet. We should be safe to do this as Wormhole upgrade all other target chain contracts before Ethereum, so we will not run into a situation where we receive VAA's that are signed by keys we have not yet seen.


hermes/src/store.rs
```diff
@@ -71,7 +74,7 @@ pub mod wormhole;
 pub struct Store {
     pub storage:           StorageInstance,
     pub observed_vaa_seqs: Cache<u64, bool>,
-    pub guardian_set:      RwLock<Option<Vec<GuardianAddress>>>,
+    pub guardian_set:      RwLock<BTreeMap<u32, Vec<[u8; 20]>>>,
```
ali-bahjati, line 77: maybe use `GuardianSet` here too?

ali-bahjati (approved):

jayantk (approved):
nice

PR #895, opened 2023-06-16
Cosmwasm refactor and preparation for parse price feeds

m30m:
This PR separates the parsing logic for applying updates logic so that the former can be used independently for parse_price_feed messages.
Parsing functions do not need `DepsMut` and they have been changed to use a `Deps` dependency instead.


ali-bahjati (approved):
Looks great!

PR #897, opened 2023-06-16
Use pythnet types

ali-bahjati:


pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -29,7 +29,7 @@ use serde::{
         strum::EnumIter,
         strum::EnumString,
         strum::IntoStaticStr,
-        strum::ToString,
+        strum::Display,
```
ali-bahjati, line 32: `strum::ToString` is deprecated in favor of `strum::Display`

Reisen (approved):

PR #901, opened 2023-06-20
[eth] - optimize ReceiverMessages parseAndVerifyVM

swimricky:
### Summary
- optimize`ReceiverMessages.parseAndVerifyVM()`
    - use `UnsafeCalldataBytesLib`
    -  do parse & verification for each section at the same time instead of parsing everything first then verifying
    - return an empty array for `vm.signatures`
    - parse/verify each signature one-by-one
    - use unchecked

### snapshot diffs
before (optimizer_runs = 2000)
image

after (optimizer_runs = 2000 - commit `6042534`)
image

### DIFF
--- a/target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
+++ b/target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
@@ -16,101 +16,107 @@ contract ReceiverMessages is ReceiverGetters {
     using BytesLib for bytes;
 
     /// @dev parseAndVerifyVM serves to parse an encodedVM and wholy validate it for consumption
+    ///     - it intentionally sets vm.signatures to an empty array since it is not needed after it is validated in this function
     function parseAndVerifyVM(
         bytes calldata encodedVM
     )
         public
         view
-        returns (
-            bool valid,
-            string memory reason,
-            uint16 emitterChainId,
-            bytes32 emitterAddress,
-            uint64 sequence,
-            bytes calldata vmPayload
-        )
+        returns (ReceiverStructs.VM memory vm, bool valid, string memory reason)
     {
-        vmPayload = encodedVM;
         uint index = 0;
-
-        {
-            uint8 version = UnsafeCalldataBytesLib.toUint8(encodedVM, index);
-            index += 1;
-            if (version != 1) {
-                revert VmVersionIncompatible();
+        unchecked {
+            {
+                vm.version = UnsafeCalldataBytesLib.toUint8(encodedVM, index);
+                index += 1;
+                if (vm.version != 1) {
+                    revert VmVersionIncompatible();
+                }
             }
-        }
 
-        ReceiverStructs.GuardianSet memory guardianSet;
-        {
-            uint32 guardianSetIndex = UnsafeCalldataBytesLib.toUint32(
+            ReceiverStructs.GuardianSet memory guardianSet;
+            {
+                vm.guardianSetIndex = UnsafeCalldataBytesLib.toUint32(
+                    encodedVM,
+                    index
+                );
+                index += 4;
+                guardianSet = getGuardianSet(vm.guardianSetIndex);
+
+                /**
+                 * @dev Checks whether the guardianSet has zero keys
+                 * WARNING: This keys check is critical to ensure the guardianSet has keys present AND to ensure
+                 * that guardianSet key size doesn't fall to zero and negatively impact quorum assessment.  If guardianSet
+                 * key length is 0 and vm.signatures length is 0, this could compromise the integrity of both vm and
+                 * signature verification.
+                 */
+                if (guardianSet.keys.length == 0) {
+                    return (vm, false, "invalid guardian set");
+                }
+
+                /// @dev Checks if VM guardian set index matches the current index (unless the current set is expired).
+                if (
+                    vm.guardianSetIndex != getCurrentGuardianSetIndex() &&
+                    guardianSet.expirationTime < block.timestamp
+                ) {
+                    return (vm, false, "guardian set has expired");
+                }
+            }
+
+            // Parse Signatures
+            uint256 signersLen = UnsafeCalldataBytesLib.toUint8(
                 encodedVM,
                 index
             );
-            index += 4;
-            guardianSet = getGuardianSet(guardianSetIndex);
-
-            /**
-             * @dev Checks whether the guardianSet has zero keys
-             * WARNING: This keys check is critical to ensure the guardianSet has keys present AND to ensure
-             * that guardianSet key size doesn't fall to zero and negatively impact quorum assessment.  If guardianSet
-             * key length is 0 and vm.signatures length is 0, this could compromise the integrity of both vm and
-             * signature verification.
-             */
-            if (guardianSet.keys.length == 0) {
-                return (
-                    false,
-                    "invalid guardian set",
-                    emitterChainId,
-                    emitterAddress,
-                    sequence,
-                    vmPayload
+            index += 1;
+            {
+                uint hashIndex = index + (signersLen * 66);
+                // Hash the body
+                vm.hash = keccak256(
+                    abi.encodePacked(
+                        keccak256(
+                            UnsafeCalldataBytesLib.sliceFrom(
+                                encodedVM,
+                                hashIndex
+                            )
+                        )
+                    )
                 );
             }
 
-            /// @dev Checks if VM guardian set index matches the current index (unless the current set is expired).
-            if (
-                guardianSetIndex != getCurrentGuardianSetIndex() &&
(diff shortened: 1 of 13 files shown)


target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
```diff
@@ -17,10 +21,167 @@ contract ReceiverMessages is ReceiverGetters {
...
+            if (!signaturesValid) {
+                return (
+                    false,
+                    invalidReason,
+                    emitterChainId,
+                    emitterAddress,
+                    sequence,
+                    vmPayload
+                );
+            }
```
ali-bahjati, line 160: Can you try verifying signatures one by one without creating a new array?

swimricky:
@jayantk @ali-bahjati we've actually cut down the bytesize enough to the point where we can revert the `optimizer_runs` back to the original value of 5000 but you can see in the screenshot above that this actually didn't give much gas improvement relative to what we've already gained so i think having a healthier buffer for bytesize would provide more value

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
```diff
@@ -7,20 +7,173 @@ pragma experimental ABIEncoderV2;
...
+
+            if (
+                (((guardianSet.keys.length * 10) / 3) * 2) / 10 + 1 > signersLen
+            ) {
+                return (vm, false, "no quorum");
+            }
+
+            // purposely setting vm.signatures to empty array since we don't need it anymore
+            // and we've already verified it above
+            vm.signatures = new ReceiverStructs.Signature[](0);
```
jayantk, line 140: this scares me a little because the overall function is no longer a drop-in replacement for the old one. If anyone was depending on the old behavior, then this could break them.

How much of a gas savings was this change? If it wasn't big, may be worthwhile to undo this.

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
```diff
@@ -7,20 +7,173 @@ pragma experimental ABIEncoderV2;
...
+            }
+
+            // Parse Signatures
+            uint256 signersLen = UnsafeCalldataBytesLib.toUint8(
+                encodedVM,
+                index
+            );
+            index += 1;
+            {
+                uint hashIndex = index + (signersLen * 66);
```
jayantk, line 73: suggest commenting where the 66 comes from (I presume the length of each signature?)

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
```diff
@@ -7,20 +7,173 @@ pragma experimental ABIEncoderV2;
...
+            vm.consistencyLevel = UnsafeCalldataBytesLib.toUint8(
+                encodedVM,
+                index
+            );
+            index += 1;
+
+            vm.payload = UnsafeCalldataBytesLib.sliceFrom(encodedVM, index);
+            if (index > encodedVM.length) {
+                return (vm, false, "invalid payload length");
+            }
```
jayantk, line 173: should this if statement occur before the call to sliceFrom ? I think it's checking that there is data left to slice, right?

target_chains/ethereum/contracts/forge-test/utils/WormholeTestUtils.t.sol
```diff
@@ -81,7 +115,7 @@ abstract contract WormholeTestUtils is Test {
 
 contract WormholeTestUtilsTest is Test, WormholeTestUtils {
     function testGenerateVaaWorks() public {
-        IWormhole wormhole = IWormhole(setUpWormhole(5));
+        IWormhole wormhole = IWormhole(setUpWormholeReceiver(5));
```
jayantk, line 118: Can we also test some things that *shouldn't* work?

jayantk (commented):
generally looks good. My main ask here is for tests of invalid cases, so we're sure that all of the needed security checks are in place. (Commenting out various checks in the code should cause test failures, and I don't think that will happen right now.)

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
ali-bahjati: The gas saving is very big. Also, this is Pyth's version of Wormhole and we know that we don't rely on it, so your concern doesn't make much sense.

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
ali-bahjati: It is usually fine to check them later but it's better to move it before.

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
```diff
@@ -7,20 +7,173 @@ pragma experimental ABIEncoderV2;
...
+
+                    bool signatureValid;
+                    string memory invalidReason;
+                    (signatureValid, invalidReason) = verifySignature(
+                        i,
+                        lastIndex,
+                        vm.hash,
+                        sig,
+                        guardianSet
+                    );
```
ali-bahjati, line 117: I think we might get some gas benefit if we don't use the Signature struct here.

ali-bahjati (commented):
Nice, the changes look great and it saves a lot of gas!

I agree that some tests will be helpful here. I'll approve when you add them.

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
swimricky: here are the numbers if we actually fill the `vm.signatures`
overall we lose about half of the gas savings

image

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
swimricky: moved to before. thanks

swimricky:
gas savings for commit `618e6f7`
image

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
swimricky: i updated the code comment on `parseAndVerifyVM` to include a warning about the `vm.signatures` and to use `parseVM` and `verifyVM` if the caller wants to get a populated `vm.signatures`

target_chains/ethereum/contracts/contracts/wormhole-receiver/ReceiverMessages.sol
swimricky: done

target_chains/ethereum/contracts/forge-test/utils/WormholeTestUtils.t.sol
```diff
@@ -71,17 +126,113 @@ abstract contract WormholeTestUtils is Test {
...
             uint8(1), // Version
-            uint32(0), // Guardian set index. it is initialized by 0
+            IWormhole(wormholeReceiverAddr).getCurrentGuardianSetIndex(), // Guardian set index. it is initialized by 0
             numSigners,
             signatures,
             body
         );
     }
+
+    function forgeVaa(
```
ali-bahjati, line 136: Can we have a test with forgeItem="" to make sure it is correct without forging? like [this](https://github.com/pyth-network/pyth-crosschain/pull/908).

target_chains/ethereum/contracts/forge-test/utils/WormholeTestUtils.t.sol
```diff
@@ -102,6 +253,252 @@ contract WormholeTestUtilsTest is Test, WormholeTestUtils {
...
+            payable(whAddr)
+        );
+        // generate the vaa and sign with the initial wormhole guardian set
+        bytes memory vaa = forgeVaa(
+            112,
+            7,
+            0x0000000000000000000000000000000000000000000000000000000000000bad,
+            10,
+            hex"deadbeaf",
+            4,
```
ali-bahjati, line 332: It seems that these values are always the same arguments here. Maybe you can just hardcode them?

ali-bahjati (approved):
Very nice!

target_chains/ethereum/contracts/forge-test/utils/WormholeTestUtils.t.sol
swimricky: done ty

PR #903, opened 2023-06-20
[sui 10/x] - merkle tree

optke3:
## Summary
 - implement and test a Merkle tree data structure in Move using [Sui Move stdlib](https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/move-stdlib/sources/vector.move) and [Sui Wormhole lib](https://github.com/wormhole-foundation/wormhole/tree/main/sui/wormhole/sources)
 - `constructProofs` for constructing a Merkle tree
 - `isProofValid` for checking for inclusion in the Merkle tree

## Testing
```
[pyth-crosschain/target_chains/sui/contracts]$ sui move test
```
Sui version: `sui 0.33.0-f65b02ccc`

### DIFF
--- a/target_chains/sui/contracts/sources/merkle_tree.move
+++ b/target_chains/sui/contracts/sources/merkle_tree.move
@@ -206,17 +206,6 @@ module pyth::merkle {
     }
 
     #[test]
-    fun testXOR(){
-        let x = 1;
-        let y = x^1;
-        assert!(y==0, 0);
-
-        let x = 64;
-        let y = x^1;
-        assert!(y==65, 0);
-    }
-
-    #[test]
     fun testMerkleTreeDepth1(){
         let messages = vector::empty<vector<u8>>();
         vector::push_back(&mut messages, x"1234");
--- a/target_chains/sui/contracts/sources/merkle_tree.move
+++ b/target_chains/sui/contracts/sources/merkle_tree.move
@@ -30,7 +30,9 @@ module pyth::merkle {
     }
 
     fun leafHash(data: &vector<u8>): Bytes20 {
-        bytes::push_u8(&mut *data, MERKLE_LEAF_PREFIX);
+        let v = vector::empty<u8>();
+        vector::push_back(&mut v, MERKLE_LEAF_PREFIX);
+        vector::append(&mut v, *data);
         hash(data)
     }
 
@@ -41,11 +43,16 @@ module pyth::merkle {
         if (greaterThan(childA, childB)) {
             (childA, childB) = (childB, childA);
         };
+        // append data_B to data_A
         let data_A = bytes20::data(&childA);
         let data_B = bytes20::data(&childB);
         vector::append(&mut data_A, data_B);
-        bytes::push_u8(&mut data_A, MERKLE_NODE_PREFIX);
-        hash(&data_A)
+
+        // create a vector containing MERKLE_NODE_PREFIX and append data_A to back
+        let v = vector::empty<u8>();
+        vector::push_back(&mut v, MERKLE_NODE_PREFIX);
+        vector::append(&mut v, data_A);
+        hash(&v)
     }
 
     // greaterThan returns whether a is strictly greater than b
--- a/target_chains/sui/contracts/sources/merkle_tree.move
+++ b/target_chains/sui/contracts/sources/merkle_tree.move
@@ -74,14 +74,14 @@ module pyth::merkle_tree {
     }
 
     // The Sui Move stdlb insert function shifts v[i] and subsequent elements to the right.
-    // We don't want this behavior, so we define our own vector insert function using swapping.
+    // We don't want this behavior, so we define our own setElement function that instead replaces the ith element.
     // Reference: https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/move-stdlib/sources/vector.move
     fun setElement<T: drop>(a: &mut vector<T>, value: T, index: u64){
         vector::push_back<T>(a, value); // push value to end
         vector::swap_remove(a, index); // swap value to correct position and pop last value
     }
 
-    // isProofValid checks whether a merkle proof is valid and returns (valid, offset)
+    // isProofValid returns whether a merkle proof is valid
     fun isProofValid(
         encodedProof: &mut Cursor<u8>,
         root: Bytes20,
--- a/target_chains/sui/contracts/sources/merkle_tree.move
+++ b/target_chains/sui/contracts/sources/merkle_tree.move
@@ -1,6 +1,6 @@
 // Implementation of a Merkle tree in Move. Supports constructing a new tree
 // with a given depth, as well as proving that a leaf node belongs to the tree.
-module pyth::merkle {
+module pyth::merkle_tree {
     use std::vector::{Self};
     use sui::hash::{keccak256};
     use wormhole::bytes::{Self};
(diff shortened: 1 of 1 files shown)


target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+    use pyth::deserialize::{Self};
+
+    const MERKLE_LEAF_PREFIX: u8 = 0;
+    const MERKLE_NODE_PREFIX: u8 = 1;
+    const MERKLE_EMPTY_LEAF_PREFIX: u8 = 2;
+
+    const E_DEPTH_NOT_LARGE_ENOUGH_FOR_MESSAGES: u64 = 1212121;
+
+    // take keccak256 of input data, then return 20 rightmost bytes of result
+    fun hash(bytes: &vector<u8>): Bytes20 {
```
optke3, line 18: not 100% sure if this is what ETH does. Pretty sure it takes the rightmost 20 bytes, but will have to confirm in the testing phase.

ali-bahjati: I think it is the first 20 bytes. [This](https://github.com/pyth-network/pyth-crosschain/blob/main/pythnet/pythnet_sdk/src/hashers/keccak256_160.rs) is the hash we use on pythnet validator.

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+    fun vectorInsert<T: drop>(a: &mut vector<T>, value: T, index: u64){
+        vector::push_back<T>(a, value); // push value to end
+        vector::swap_remove(a, index); // swap value to correct position and pop last value
+    }
+
+    // isProofValid checks whether a merkle proof is valid and returns (valid, offset)
+    fun isProofValid(
+        encodedProof: &mut Cursor<u8>,
+        root: Bytes20,
+        leafData: vector<u8>,
```
m30m, line 82: Not sure if it is ok to mutate this argument. We may need it later on for additional parsing

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    fun emptyLeafHash(): Bytes20 {
+        let v = vector::empty();
+        bytes::push_u8(&mut v, MERKLE_EMPTY_LEAF_PREFIX);
+        hash(&v)
+    }
+
+    fun leafHash(data: &vector<u8>): Bytes20 {
+        bytes::push_u8(&mut *data, MERKLE_LEAF_PREFIX);
```
ali-bahjati, line 34: shouldn't we push_front here? :? it seems push_u8 pushes to the back

target_chains/sui/contracts/sources/merkle_tree.move
ali-bahjati: which argument?

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+        assert!(isProofValid(&mut proofsCursor, root, x"4321")==true, 0);
+        assert!(isProofValid(&mut proofsCursor, root, x"11")==true, 0);
+        assert!(isProofValid(&mut proofsCursor, root, x"22")==true, 0);
+
+        // destroy cursor
+        let _ = cursor::take_rest<u8>(proofsCursor);
+    }
+
+
+    #[test]
```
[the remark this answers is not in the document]
ali-bahjati: maybe use a proof of other node for another one and make sure it's false.

target_chains/sui/contracts/sources/merkle_tree.move
m30m: leafData

m30m: never mind, it's being copied, my bad.

target_chains/sui/contracts/sources/merkle_tree.move
jayantk: it seems dangerous for this function to mutate data, as if someone wants to use it later, then they may be surprised.

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+        childA: Bytes20,
+        childB: Bytes20
+    ): Bytes20 {
+        if (greaterThan(childA, childB)) {
+            (childA, childB) = (childB, childA);
+        };
+        let data_A = bytes20::data(&childA);
+        let data_B = bytes20::data(&childB);
+        vector::append(&mut data_A, data_B);
+        bytes::push_u8(&mut data_A, MERKLE_NODE_PREFIX);
```
jayantk, line 48: this also needs to push to the front

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+            };
+            i = i + 1;
+        };
+        false
+    }
+
+    // The Sui Move stdlb insert function shifts v[i] and subsequent elements to the right.
+    // We don't want this behavior, so we define our own vector insert function using swapping.
+    // Reference: https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/move-stdlib/sources/vector.move
+    fun vectorInsert<T: drop>(a: &mut vector<T>, value: T, index: u64){
```
jayantk, line 73: this is really setting the `i`th element to value. I would call this like `setElement`, as "insert" usually means something that does the shifting behavior above.

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+                siblingDigest
+            );
+            i = i + 1;
+        };
+        bytes20::data(&currentDigest) == bytes20::data(&root)
+    }
+
+    // constructProofs constructs a merkle tree and returns the root of the tree as
+    // a Bytes20 as well as the vector of encoded proofs
+    fun constructProofs(
```
jayantk, line 104: per my comment in slack, i think it would be smart to generate a bunch of valid / invalid messages from the official rust implementation for these tests. That would ensure that some of the details around the hash truncation etc. are correct.

target_chains/sui/contracts/sources/merkle_tree.move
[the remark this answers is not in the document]
jayantk: definitely agree with this comment -- super important to test that the code rejects invalid proofs.

target_chains/sui/contracts/sources/merkle_tree.move
```diff
@@ -0,0 +1,276 @@
...
+        // Fill in bottom row with leaf hashes
+        let j: u64 = 0;
+        while (j < vector::length(messages)){
+            vectorInsert<Bytes20>(&mut tree, leafHash(vector::borrow(messages, j)), (1 << depth) + j);
+            j = j + 1;
+        };
+
+        // Filling the node hashes from bottom to top
+        let k: u8 = depth;
+        while (k>0){
```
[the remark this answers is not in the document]
optke3: unfortunately not for Move...

target_chains/sui/contracts/sources/merkle_tree.move
optke3: data is passed in as a read-only reference, so the original copy is not mutated

optke3: you guys are right about how push_u8 pushes back

optke3: I believe it would have to be `data: &mut vector<u8>` for there to be a danger of mutating the original data, otherwise it makes a copy inside this function

target_chains/sui/contracts/sources/merkle_tree.move
optke3: I can do this in a later PR after those messages are generated

target_chains/sui/contracts/sources/merkle_tree.move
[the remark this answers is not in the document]
m30m: I tried this out and it was pretty good:
https://plugins.jetbrains.com/plugin/14721-move-language
You may need some setup though.

ali-bahjati (approved):

PR #905, opened 2023-06-21
Update price feeds early if part of a batch

jayantk:
I was looking at the sui pusher's txs and I noticed that it currently only updates a subset of the price feeds in the pushed VAA. This behavior is an interaction between the structure of the sui contract (with the hot potato etc) and the price pusher.

I'm fixing this problem by adding a feature we'll want anyway, which is to piggyback additional price updates on to a push. With accumulators, it's very cheap to add an additional update to an existing price push, so we'll want to take advantage of this by pushing more feeds than just the ones that met the triggering conditions.

This PR adds an extra set of triggering conditions for an early update. If at least one feed meets the normal triggering conditions, we'll push a batch of updates that includes the original update + all feeds meeting the early update conditions. Using this, we can also set early_update.time_difference = 0 in the sui config to solve the specific problem as well.

I haven't been able to run this since testnet is down, but i will make sure it works before merging.

### DIFF
--- a/price_pusher/src/evm/command.ts
+++ b/price_pusher/src/evm/command.ts
@@ -91,6 +91,11 @@ export default {
       mnemonic,
       pythContractAddress
     );
+    console.log(
+      `Pushing updates from wallet address: ${pythContractFactory
+        .createWeb3PayerProvider()
+        .getAddress()}`
+    );
 
     const evmListener = new EvmPriceListener(pythContractFactory, priceItems, {
       pollingFrequency,
--- a/price_pusher/src/price-config.ts
+++ b/price_pusher/src/price-config.ts
@@ -15,6 +15,11 @@ const PriceConfigFileSchema: Joi.Schema = Joi.array()
       time_difference: Joi.number().required(),
       price_deviation: Joi.number().required(),
       confidence_ratio: Joi.number().required(),
+      early_update: Joi.object({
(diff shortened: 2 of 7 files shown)


price_pusher/src/evm/command.ts
```diff
@@ -91,6 +91,11 @@ export default {
       mnemonic,
       pythContractAddress
     );
+    console.log(
+      `Pushing updates from wallet address: ${pythContractFactory
+        .createWeb3PayerProvider()
+        .getAddress()}`
+    );
```
jayantk, line 98: drive-by for ease of use

ali-bahjati (approved):
lgtm! :shipit:

price_pusher/src/price-config.ts
```diff
@@ -59,25 +83,25 @@ export function shouldUpdate(
...
   sourceLatestPrice: PriceInfo | undefined,
   targetLatestPrice: PriceInfo | undefined
-): boolean {
+): UpdateCondition {
   const priceId = priceConfig.id;
 
   // There is no price to update the target with.
   if (sourceLatestPrice === undefined) {
-    return false;
+    return UpdateCondition.YES;
```
jason-hy-li-crypto, line 91: is this a bug ?

ali-behjati: yes unfortunately. fix is in #2730

PR #907, opened 2023-06-21
[hermes] Add storage tests + refactor

ali-bahjati:
This PR adds some tests to the local storage to make sure it works fine. During the testing I realized that moka's Cache is not working as we expect it to work (it doesn't evict items as we want to) so we removed it and instead adopted a simple queue (or BTreeMap). We might replace it with a more efficient implementation in the future but will do in separate PRs (if needed). 

Also, I added a new type alias of `FeedId` in the Message struct and renamed `id` to `feed_id` in pythnet_sdk to make the id more clear.  I bumped version to 2.0.0 to get away from solana versioning and have proper version control on this crate. Lastly, updated the cosmwasm contract with this breaking change.

### DIFF
--- a/hermes/src/store.rs
+++ b/hermes/src/store.rs
@@ -48,6 +48,7 @@ use {
     std::{
         collections::{
             BTreeMap,
+            BTreeSet,
             HashSet,
         },
         sync::Arc,
@@ -78,9 +79,11 @@ pub mod storage;
 pub mod types;
 pub mod wormhole;
 
+const OBSERVED_CACHE_SIZE: usize = 1000;
+
 pub struct Store {
     pub storage:                  StorageInstance,
-    pub observed_vaa_seqs:        Cache<u64, bool>,
+    pub observed_vaa_seqs:        RwLock<BTreeSet<u64>>,
     pub guardian_set:             RwLock<BTreeMap<u32, GuardianSet>>,
     pub update_tx:                Sender<()>,
     pub last_completed_update_at: RwLock<Option<Instant>>,
@@ -90,10 +93,7 @@ impl Store {
     pub fn new_with_local_cache(update_tx: Sender<()>, cache_size: u64) -> Arc<Self> {
         Arc::new(Self {
             storage: storage::local_storage::LocalStorage::new_instance(cache_size),
-            observed_vaa_seqs: Cache::builder()
-                .max_capacity(cache_size)
-                .time_to_live(Duration::from_secs(60 * 5))
-                .build(),
+            observed_vaa_seqs: RwLock::new(Default::default()),
             guardian_set: RwLock::new(Default::default()),
             update_tx,
             last_completed_update_at: RwLock::new(None),
@@ -113,7 +113,7 @@ impl Store {
                     return Ok(()); // Ignore VAA from other emitters
                 }
 
-                if self.observed_vaa_seqs.get(&vaa.sequence).is_some() {
+                if self.observed_vaa_seqs.read().await.contains(&vaa.sequence) {
                     return Ok(()); // Ignore VAA if we have already seen it
                 }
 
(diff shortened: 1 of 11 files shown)


hermes/src/store.rs
```diff
@@ -127,7 +127,11 @@ impl Store {
                     }
                 };
 
-                self.observed_vaa_seqs.insert(vaa.sequence, true).await;
+                let mut observed_vaa_seqs = self.observed_vaa_seqs.write().await;
+                observed_vaa_seqs.insert(vaa.sequence);
+                if observed_vaa_seqs.len() > OBSERVED_CACHE_SIZE {
+                    observed_vaa_seqs.pop_first();
+                }
```
jayantk, line 134: i presume that we drop the lock as soon as the mutable ref goes out of scope? if so, should this bit go in a { } block so that we drop the lock as soon as we can?

hermes/src/store/storage/local_storage.rs
```diff
@@ -155,12 +157,507 @@ impl Storage for LocalStorage {
...
+        let mut accumulator_cache = self.accumulator_cache.write().await;
+        accumulator_cache.push_back(state);
+
+        let mut i = accumulator_cache.len().saturating_sub(1);
+        while i > 0 && accumulator_cache[i - 1].slot > accumulator_cache[i].slot {
+            accumulator_cache.swap(i - 1, i);
+            i -= 1;
+        }
+
+        if accumulator_cache.len() > self.cache_size as usize {
```
jayantk, line 169: maybe make this a `while` ? I think `if` is also correct because the size of the deque can only grow by 1 here, but `while` would be more robust to code changes around it (like inserting multiple values at the same time)

hermes/src/store/storage/local_storage.rs
```diff
@@ -155,12 +157,507 @@ impl Storage for LocalStorage {
...
+            accumulators::merkle::MerklePath,
+            hashers::keccak256_160::Keccak160,
+            messages::{
+                Message,
+                PriceFeedMessage,
+            },
+        },
+    };
+
+    pub fn create_dummy_price_feed_message_state(
```
jayantk, line 207: (if you made this take an array of `times_and_slots: [(i64, Slot)]` you could shorten the tests below a bit)

hermes/src/store/storage/local_storage.rs
```diff
@@ -155,12 +157,507 @@ impl Storage for LocalStorage {
...
+        );
+    }
+
+    #[tokio::test]
+    pub async fn test_store_and_receive_multiple_accumulator_state_works() {
+        // Initialize a storage with a cache size of 2 per key and the accumulator state.
+        let storage = LocalStorage::new_instance(2);
+
+        let accumulator_state_at_slot_10 = create_empty_accumulator_state_at_slot(10);
+        let accumulator_state_at_slot_20 = create_empty_accumulator_state_at_slot(20);
```
jayantk, line 600: could also do create_and_store here to reduce code duplication

jayantk (approved):
nice!

hermes/src/store.rs
ali-bahjati: nice catch!

hermes/src/store/storage/local_storage.rs
ali-bahjati: Makes sense

hermes/src/store/storage/local_storage.rs
ali-bahjati: I tried this now but the test readability was not very good. I decided to keep it as it is.

hermes/src/store/storage/local_storage.rs
ali-bahjati: 👍

Reisen (approved):

PR #909, opened 2023-06-22
Make AccumulatorUpdateData deserialization accept minor version bumps

m30m:


pythnet/pythnet_sdk/src/wire.rs
```diff
@@ -385,4 +389,28 @@ mod tests {
         // The deserialized value should be the same as the original.
         assert_eq!(deserialized_update, empty_update);
     }
+
+    // Test if the AccumulatorUpdateData major and minor version increases work as expected
+    #[test]
+    fn test_accumulator_forward_compatibility() {
```
ali-bahjati, line 395: Can you also add a new byte to the message and see how it performs with that?

m30m: In which field do you mean?

ali-bahjati: i mean probably this:
1. create a message
2. serialize it to `data`
3. append some bytes to `data`
4. make sure deserializing the `data` to message struct works

ali-bahjati (approved):
Tyvm

PR #910, opened 2023-06-22
[sui 11/x] - pyth merkle accumulator

optke3:
## Summary
- implement Sui pyth accumulator price feed creation and update logic, including
     - accumulator message parsing logic in `pyth::accumulator`
     - the new price feed update entry point `create_authenticated_price_infos_using_accumulator` in `pyth::pyth` with the following signature:
     - ```
        public fun create_authenticated_price_infos_using_accumulator(
             pyth_state: &PythState,
             accumulator_message: vector<u8>,
             verified_vaa: VAA,
             clock: &Clock,
       ): AuthenticatedPriceInfos<PriceInfo>
       ```
    - `create_price_feeds_using_accumulator` for creating new price feeds using accumulator (similar signature as above)
- rename `HotPotatoVector` to `AuthenticatedVector`

## Notes
- We have two different entry points for the batch price updates and accumulator updates, because they have difference function signatures. The latter requires two inputs (accumulator message, and verified VAA of the vaa bytes encoded in the accumulator), while the former only requires a verified VAA argument
- For `create_authenticated_price_infos_using_accumulator`, the `verified_vaa` argument is the parsed and verified version of the vaa bytes in the accumulator message, so in that sense those vaa bytes in the accumulator are redundant. Verifying the vaa _inside of_ `create_authenticated_price_infos_using_accumulator` with a hardcoded call to `wormhole::vaa::parse_and_verify` is not an option for Sui, because we need to separate business logic and Wormhole logic to prevent Wormhole upgrades to always necessitate Pyth upgrades. Since WH can upgrade at any point, we do not want the Pyth contract to become obsolete due to a hardcoded `wormhole::vaa::parse_and_verify` in the price feed update code path (note that the `wormhole` here refers to a fixed deployment of WH and cannot be changed without upgrading Pyth).

## Tests
- Added `test_create_and_update_single_price_feed_with_accumulator_success`
- Added `test_create_and_update_single_price_feed_with_accumulator_failure`
- Added `test_create_and_update_multiple_price_feeds_with_accumulator_success`
- Added `test_parse_and_verify_accumulator_updates`
- Revived `test_update_cache`
- Revived `test_update_cache_old_update`
- Revived `test_create_price_feeds_invalid_data_source`
- Revived `test_get_update_fee`
- Revived `test_create_and_update_price_feeds_insufficient_fee`


target_chains/sui/contracts/sources/authenticated_price_infos.move
ali-bahjati, line 1: I like the rename to `Authenticated...` but this is a generic authenticated vector, maybe `AuthenticatedVector` be a better name?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -430,10 +474,10 @@ module pyth::pyth_tests{
...
         let governance_contract =
             x"deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef";
-        let initial_guardians =
-            vector[
-                x"beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe"
-            ];
+        // let initial_guardians =
+        //     vector[
+        //         x"beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe"
+        //     ];
```
ali-bahjati, line 480: Since it's an argument now, can you remove it?

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,157 @@
...
+        let trailing_size = deserialize::deserialize_u8(cursor);
+        deserialize::deserialize_vector(cursor, (trailing_size as u64));
+
+        let proof_type = deserialize::deserialize_u8(cursor);
+        assert!(proof_type == 0, E_INVALID_ACCUMULATOR_PAYLOAD);
+
+        // Ignore the vaa in the accumulator message because presumably it has already been verified
+        // and passed to this function as input.
+        let vaa_size = deserialize::deserialize_u16(cursor);
+        let _ = deserialize::deserialize_vector(cursor, (vaa_size as u64));
```
ali-bahjati, line 87: nitpik: I don't have much Move knowledge, but maybe we can remove `let _ =`

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,157 @@
...
+        let price_info = price_info::new_price_info(
+            clock::timestamp_ms(clock) / 1000, // not used anywhere kept for backward compatibility
+            clock::timestamp_ms(clock) / 1000,
+            price_feed::new(
+                price_identifier,
+                pyth::price::new(price, conf, expo, publish_time),
+                pyth::price::new(ema_price, ema_conf, expo, publish_time),
+            )
+        );
+        //cursor::rest(message_cur);
```
ali-bahjati, line 132: Why have you commented this out? We want to consume the rest for future compatibility.

ali-bahjati (commented):
It's a shame that with Sui design our new update requires a new method and the protocols need to update their codes.

It looks good to me. I like to look again when you add more tests to the code.

We also should think about the update fee here. The way it has handled before is not correct. The fee per batch on Eth was because users had no control over batches but in sui people choose which prices to update (after verifying the message). We should charge the base_fee per each price update and this aligns with the accumulator as well (because users choose what do update).

p.s: during the review I realized that some of the tests are commented out from the past, i recommend removing them if they are not going to be used.

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,157 @@
...
+        // TODO - verify VAA targets this chain and contract
+        //verify_data_source(&msg_vaa);
+        let merkle_root_hash = parse_accumulator_merkle_root_from_vaa_payload(vaa::payload(&msg_vaa));
+        vaa::destroy(msg_vaa);
+        parse_and_verify_accumulator_updates(cursor, merkle_root_hash, clock)
+    }
+
+    // parse_and_verify_accumulator_message verifies that the merkle root contained a pre-verified VAA
+    // matches the rest of the content in the accumulator message (encapsulated by cursor).
+    public(friend) fun parse_and_verify_accumulator_message(cursor: &mut Cursor<u8>, vaa_payload: vector<u8>, clock: &Clock): vector<PriceInfo> {
```
m30m, line 73: We don't need both `parse_and_verify_accumulator_message` and `parse_and_verify_accumulator_message_with_worm_state` in the end right?

optke3: yeah, only the former is needed, the latter is marked `#[test_only]`

target_chains/sui/contracts/sources/pyth_accumulator.move
optke3: It's because `&mut message_cur` is passed into this function, but I can also pass in the actual message_cur, instead of destroying it on the outside.
```
            let message_cur = cursor::new(message);
            let price_info = parse_price_feed_message(&mut message_cur, clock);
            cursor::destroy_empty(message_cur);
```

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -85,8 +92,50 @@ module pyth::pyth {
         sources
     }
 
-    /// Create and share new price feed objects if they don't already exist.
-    public fun create_price_feeds(
+    /// Create and share new price feed objects if they don't already exist using accumulator message.
+    public fun create_price_feeds_using_accumulator(
+        pyth_state: &mut PythState,
+        accumulator_message: vector<u8>,
+        vaa: VAA, // the verified version of the vaa bytes encoded within the accumulator_message
```
jayantk, line 99: I suggest making a struct to encapsulate these two arguments, like `VerifiedAccumulatorMessage` . I wouldn't worry about this for this PR though, since this change is already pretty big.

jayantk: you could potentially check some of the header bytes and data sources and stuff when that struct is created as well. That would consolidate the security checks into one place.

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -113,38 +162,85 @@ module pyth::pyth {
...
-                    transfer::public_share_object(new_price_info_object);
-                }
+                transfer::public_share_object(new_price_info_object);
             }
+        }
+    }
+
+
+    // verified_vaa is the verified version of the VAA encoded within the accumulator_message
+    public fun create_authenticated_price_infos_using_accumulator(
```
jayantk, line 204: now that users can customize which price feeds are in the accumulator message, we could change the contract so that it always updates the on-chain state for *all* of the feeds in the message. This function could potentially do that work itself. I think that would simplify the interface for our users, as they wouldn't have to worry about `AuthenticatedVector<PriceInfo>`.

No need to do this change in this PR though.

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -478,108 +599,112 @@ module pyth::pyth_tests{
...
+        let multiple_vaas = vector[
+            x"4ee17a1a4524118de513fddcf82b77454e51be5d6fc9e29fc72dd6c204c0e4fa",
+            x"c72fdf81cfc939d4286c93fbaaae2eec7bae28a5926fa68646b43a279846ccc1",
+            x"d9a8123a793529c31200339820a3210059ecace6c044f81ecad62936e47ca049",
+            x"84e4f21b3e65cef47fda25d15b4eddda1edf720a1d062ccbf441d6396465fbe6",
+            x"9e73f9041476a93701a0b9c7501422cc2aa55d16100bec628cf53e0281b6f72f"
+        ];
+
+        // Pass in multiple VAAs
+        assert!(pyth::get_total_update_fee(&pyth_state, vector::length<vector<u8>>(&multiple_vaas)) == 250, 1);
```
jayantk, line 695: nitpick: `5 * single_update_fee` would be an easier test condition for the reader to understand

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -651,12 +776,329 @@ module pyth::pyth_tests{
...
+
+    #[test_only]
+    const TEST_ACCUMULATOR_SINGLE_FEED: vector<u8> = x"504e41550100000000a0010000000001005d461ac1dfffa8451edda17e4b28a46c8ae912422b2dc0cb7732828c497778ea27147fb95b4d250651931845e7f3e22c46326716bcf82be2874a9c9ab94b6e42000000000000000000000171f8dcb863d176e2c420ad6610cf687359612b6fb392e0642b0ca6b1f186aa3b0000000000000000004155575600000000000000000000000000da936d73429246d131873a0bab90ad7b416510be01005500b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf65f958f4883f9d2a8b5b1008d1fa01db95cf4a8c7000000006491cc757be59f3f377c0d3f423a695e81ad1eb504f8554c3620c3fd02f2ee15ea639b73fa3db9b34a245bdfa015c260c5a8a1180177cf30b2c0bebbb1adfe8f7985d051d2";
+    #[test]
+    fun test_create_and_update_single_price_feed_with_accumulator_success() {
+        use pyth::data_source::Self;
+
+        let base_update_fee = 50;
+        let coins_to_mint = 5000;
+        let emitter_address = x"71f8dcb863d176e2c420ad6610cf687359612b6fb392e0642b0ca6b1f186aa3b";
```
jayantk, line 865: suggest extracting some constants for these byte strings so they can be reused across tests instead of defined repeatedly.

jayantk: i also suggest leaving a comment explaining what's in the accumulator payload above, because I can't really tell what the test does otherwise.

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -651,12 +776,329 @@ module pyth::pyth_tests{
...
+        let initial_guardians = vector[x"7E5F4552091A69125d5DfCb7b8C2659029395Bdf"];
+        let data_sources = vector[data_source::new(1, external_address::new(bytes32::from_bytes(x"71f8dcb863d176e2c420ad6610cf687359612b6fb392e0642b0ca6b1f186aa3b")))];
+        //new(emitter_chain: u64, emitter_address: ExternalAddress)
```
jayantk, line 868: delete this line

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,156 @@
...
+    const E_INVALID_WORMHOLE_MESSAGE: u64 = 454;
+    const E_INVALID_ACCUMULATOR_PAYLOAD: u64 = 554;
+
+    const ACCUMULATOR_UPDATE_WORMHOLE_VERIFICATION_MAGIC: u32 = 1096111958;
+
+    friend pyth::pyth;
+
+    #[test_only]
+    // get_price_feed_updates_from_single_vaa is a test_only function that gets the vector of price info updates
+    // from a Wormhole VAA message.
```
jayantk, line 25: comment seems to be about a different function

target_chains/sui/contracts/sources/pyth_accumulator.move
jayantk: I'm confused about why we need the `_with_worm_state` version of this method for the tests -- seems like we should be testing the code paths in this function here, and not the ones above (which could be different). I think a better test helper would parse the vaa_payload out of the accumulator message so that the tests could call this function.

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,156 @@
...
+        let _merkle_root_slot = deserialize::deserialize_u64(&mut msg_payload_cursor);
+        let _merkle_root_ring_size = deserialize::deserialize_u32(&mut msg_payload_cursor);
+        let merkle_root_hash = deserialize::deserialize_vector(&mut msg_payload_cursor, 20);
+        cursor::take_rest<u8>(msg_payload_cursor);
+        bytes20::new(merkle_root_hash)
+    }
+
+    // Note: this parsing function is adapted from the Aptos Pyth parse_price_feed_message function
+    fun parse_price_feed_message(message_cur: &mut Cursor<u8>, clock: &Clock): PriceInfo {
+        //let message_cur = cursor::new(message);
```
jayantk, line 111: delete

target_chains/sui/contracts/sources/pyth.move
jayantk: i notice that test cases generally have quite a bit of repeated code. Is there any way to reduce that? It's a lot easier for the reader to understand what's important (or different) about each test that way.

Maybe you could make helpers like `test_accumulator_update_is_successful(message, expected_price_infos)` and `test_accumulator_update_fails(message, error code)` or something like that ?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -348,20 +445,23 @@ module pyth::pyth_tests{
...
+    use pyth::price_feed::{Self};
+    use pyth::accumulator::{Self};
+    use pyth::deserialize::{Self};
 
     use wormhole::setup::{Self as wormhole_setup, DeployerCap};
     use wormhole::external_address::{Self};
     use wormhole::bytes32::{Self};
     use wormhole::state::{State as WormState};
     use wormhole::vaa::{Self, VAA};
+    use wormhole::cursor::{Self};
```
jayantk, line 464: does the VAA below need to be declared test_only when the whole module is already test_only ?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -651,12 +776,329 @@ module pyth::pyth_tests{
...
+            &mut pyth_state,
+            auth_price_infos,
+            &mut price_info_object_1,
+            coins,
+            &clock
+        );
+
+        test_scenario::next_tx(&mut scenario, DEPLOYER);
+        authenticated_vector::destroy<PriceInfo>(auth_price_infos);
+
```
jayantk, line 918: shouldn't there be some assertion here or something that the updates happened correctly?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -651,12 +776,329 @@ module pyth::pyth_tests{
...
```
jayantk, line 988: same comment as above

jayantk (commented):
I think looks generally reasonable. I would like to see some refactoring of the test logic though, so that there's less duplicated code. At the moment it's quite difficult to understand what each test does and how it differs from other tests. See inline comments for some suggestions on how to do this.

target_chains/sui/contracts/sources/pyth.move
ali-bahjati: to the best of my knowledge the reason that it's not happening in the first place is a limitation of Sui to pass vector of resources in one call. Sui was supposed to fix it but not sure whether they did that.

target_chains/sui/contracts/sources/authenticated_price_infos.move
ali-bahjati, line 1: Rename this file?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -85,8 +92,50 @@ module pyth::pyth {
...
+                    vaa::emitter_address(&vaa))
+                ),
+            E_INVALID_DATA_SOURCE
+        );
+
+        // decode the price info updates from the VAA payload (first check if it is an accumulator or batch price update)
+        let accumulator_message_cursor = cursor::new(accumulator_message);
+        let header = deserialize::deserialize_u32(&mut accumulator_message_cursor);
+
+        let _price_infos = vector::empty<PriceInfo>();
```
ali-bahjati, line 121: why starting with underscore? it's an indicator of unused variable.

target_chains/sui/contracts/sources/pyth.move
optke3: don't think it's possible atm

target_chains/sui/contracts/sources/pyth.move
optke3: because there is a compile warning:

```
warning[W09003]: unused assignment
    ┌─ ./sources/pyth.move:121:13
    │
121 │         let price_infos = vector::empty<PriceInfo>();
    │             ^^^^^^^^^^^ Unused assignment for variable 'price_infos'. Consider removing, replacing with '_', or prefixing with '_' (e.g., '_price_infos')
```

if we do

```
        let price_infos = vector::empty<PriceInfo>();
        if ((header as u64) == PYTHNET_ACCUMULATOR_UPDATE_MAGIC) {
            price_infos = accumulator::parse_and_verify_accumulator_message(&mut accumulator_message_cursor, vaa::take_payload(vaa), clock);
        }
        else {
            abort E_INVALID_ACCUMULATOR_HEADER
        };
```

optke3: I rewrote the code like this to avoid this situation where the compiler gets tripped up by the if statement:

```
if ((header as u64) != PYTHNET_ACCUMULATOR_UPDATE_MAGIC) {
            abort E_INVALID_ACCUMULATOR_HEADER
        };
        let price_infos = accumulator::parse_and_verify_accumulator_message(&mut accumulator_message_cursor, vaa::take_payload(vaa), clock);
```

target_chains/sui/contracts/sources/pyth.move
optke3: yeah you're right, the `#[test_only]`s is an artifact of refactoring the tests (moving some from from `pyth_accumulator.move` to `pyth.move`). In the former file, the tests were not located in a `test_only` module, but in the latter they are.

target_chains/sui/contracts/sources/pyth.move
optke3: this specific thing is kind of hard to do in Move, since for the failure case, the entire #[test] function has to be annotated with the error code at the top, e.g.

```
#[test]
#[expected_failure(abort_code = pyth::accumulator::E_INVALID_PROOF)]
    fun test_create_and_update_single_price_feed_with_accumulator_failure() {
```
so passing in the error code doesn't help.

I'll try to reduce the amount of shared code between functions as much as possible by factoring out constants and hard-coding them, writing helper functions etc.

target_chains/sui/contracts/sources/pyth.move
optke3: yeah I think that could achieve better conceptual separation, in a later PR I could create a separate public entry-point function

`create_verified_accumulator_message`

and the user can create a programmable txn to call that and obtain a `VerifiedAccumulatorMessage` to be fed into this function

target_chains/sui/contracts/sources/pyth_accumulator.move
optke3: the `_with_worm_state` version of the function is a shortcut used for testing `pyth_accumulator.move` and not the full code path in `pyth.move`. I had to move those tests to pyth.move, because of a dependency cyckle.

As noted in pyth.move, 
`// pyth accumulator tests (included in this file instead of pyth_accumulator.move to avoid dependency cycle - as we need pyth_tests::setup_test)`

optke3: Specifically, `test_get_price_feed_updates_from_accumulator` is used in `test_parse_and_verify_accumulator_updates` in `pyth.move`.

optke3: that test function isn't for testing the full code path, it's merely a shortcut for testing accumulator-specific functions, such as `parse_accumulator_merkle_root_from_vaa_payload`.

optke3: good idea, I removed the `parse_and_verify_accumulator_message_with_worm_state` entirely, and instead added the helper `parse_vaa_bytes_from_accumulator_message` as you suggested.

.github/workflows/sui-contract.yml
ali-bahjati, line 1: This CI seems a bit odd. We don't need macos. If you see similar CI files with github actions you can easily set up rust and probably cmake is there as well. 

This one is also failing.

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -392,16 +497,41 @@ module pyth::pyth_tests{
...
+
+        let proof_type = deserialize::deserialize_u8(&mut cursor);
+        assert!(proof_type == 0, 0);
+
+        let vaa_size = deserialize::deserialize_u16(&mut cursor);
+        let vaa = deserialize::deserialize_vector(&mut cursor, (vaa_size as u64));
+        cursor::take_rest(cursor);
+        vaa::parse_and_verify(worm_state, vaa, clock)
+    }
+
```
ali-bahjati, line 523: How the users are going to do this from the client side? how should we separate it? I don't know whether it's possible or not, but maybe it's good if we have a public function that takes out the vaa bytes and then users can pass that object to wormhole. If that be possible we don't need to put this on the test

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -167,11 +260,11 @@ module pyth::pyth {
...
-        return hot_potato_vector::new(price_updates)
+        return authenticated_vector::new(price_updates)
     }
 
     /// Update a singular Pyth PriceInfoObject (containing a price feed) with the
-    /// price data in the given hot potato vector (a vector of PriceInfo objects).
+    /// price data in the authenticated price infos vector (a vector of PriceInfo objects).
     ///
     /// The javascript https://github.com/pyth-network/pyth-js/tree/main/pyth-sui-js package
     /// should be used to fetch these VAAs from the Price Service. More information about this
```
ali-bahjati, line 270: pyth-sui-js doesn't exist right? Maybe point them to the readme?

target_chains/sui/contracts/sources/pyth.move
ali-bahjati: We also need to add it to the readme for the users.

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,158 @@
...
+        deserialize::deserialize_vector(cursor, (trailing_size as u64));
+
+        let proof_type = deserialize::deserialize_u8(cursor);
+        assert!(proof_type == 0, E_INVALID_ACCUMULATOR_PAYLOAD);
+
+        let vaa_size = deserialize::deserialize_u16(cursor);
+        let vaa_bytes = deserialize::deserialize_vector(cursor, (vaa_size as u64));
+        cursor::take_rest(c);
+        vaa_bytes
+    }
```
ali-bahjati, line 157: There exists a similar method on the `pyth.move` named `get_verified_vaa_from_accumulator_message`

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,158 @@
...
+        let verified_vaa = vaa::parse_and_verify(wormhole_state, vaa_bytes, clock);
+
+        // main entry-point
+        let price_infos = parse_and_verify_accumulator_message(&mut cur, vaa::take_payload(verified_vaa), clock);
+
+        //TODO -  update_cache(updates);
+        // cursor should be empty after parsing the header and accumulator message
+        cursor::destroy_empty(cur);
+        price_infos
+    }
```
ali-bahjati, line 130: It is only used for one test? What is the purpose of it compared with using `get_verified_vaa_from_accumulator_message` and calling the contract methods?

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -646,17 +770,373 @@ module pyth::pyth_tests{
...
+        return_shared(price_info_object_2);
+        return_shared(price_info_object_3);
+
+        clock::destroy_for_testing(clock);
+        test_scenario::end(scenario);
+    }
+
+    #[test]
+    #[expected_failure(abort_code = pyth::pyth::E_INSUFFICIENT_FEE)]
+    fun test_create_and_update_price_feeds_insufficient_fee() {
```
ali-bahjati, line 1134: nitpik: It's better to use accumulator message here (we might remove batch attestation soon)

ali-bahjati (commented):
It looks great! I left some inline comments. Would be great to address the non-nitpik ones. Also I like to see other failure tests added here too.

target_chains/sui/contracts/sources/pyth_accumulator.move
optke3: removed this one

target_chains/sui/contracts/sources/pyth_accumulator.move
optke3: this was written earlier and used solely for testing `pyth_accumulator.move`, but I can remove it now

target_chains/sui/contracts/sources/pyth.move
optke3: We'll just need a typescript function that extracts the vaa bytes from an accumulator message. That's the only extra step.

ali-bahjati (approved):
Nice!

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -85,8 +92,47 @@ module pyth::pyth {
...
+            E_INVALID_DATA_SOURCE
+        );
+
+        // decode the price info updates from the VAA payload (first check if it is an accumulator or batch price update)
+        let accumulator_message_cursor = cursor::new(accumulator_message);
+        let header = deserialize::deserialize_u32(&mut accumulator_message_cursor);
+
+        if ((header as u64) != PYTHNET_ACCUMULATOR_UPDATE_MAGIC) {
+            abort E_INVALID_ACCUMULATOR_HEADER
+        };
```
jayantk, line 123: should this check go in parse_and_verify_accumulator_message? that seems safer and would reduce code duplication -- I see a very similar if block below.

target_chains/sui/contracts/sources/pyth.move
```diff
@@ -646,17 +768,373 @@ module pyth::pyth_tests{
...
+    const TEST_ACCUMULATOR_SINGLE_FEED: vector<u8> = x"504e41550100000000a0010000000001005d461ac1dfffa8451edda17e4b28a46c8ae912422b2dc0cb7732828c497778ea27147fb95b4d250651931845e7f3e22c46326716bcf82be2874a9c9ab94b6e42000000000000000000000171f8dcb863d176e2c420ad6610cf687359612b6fb392e0642b0ca6b1f186aa3b0000000000000000004155575600000000000000000000000000da936d73429246d131873a0bab90ad7b416510be01005500b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf65f958f4883f9d2a8b5b1008d1fa01db95cf4a8c7000000006491cc757be59f3f377c0d3f423a695e81ad1eb504f8554c3620c3fd02f2ee15ea639b73fa3db9b34a245bdfa015c260c5a8a1180177cf30b2c0bebbb1adfe8f7985d051d2";
+    // Info about the accumulator message:
+    //      Price Identifier: 0xb10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6
+    //      Price: 6887568746747646632
+    //      Conf: 13092246197863718329
+    //      Exponent: 1559537863
+    //      EMA Price: 4772242609775910581
+    //      EMA Conf: 358129956189946877
+    //      EMA Expo: 1559537863
+    //      Published Time: 1687276661
```
jayantk, line 864: nitpick: usually a comment like this would go above the variable declaration, not after.

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,116 @@
...
+        // and passed to this function as input.
+        let vaa_size = deserialize::deserialize_u16(cursor);
+        deserialize::deserialize_vector(cursor, (vaa_size as u64));
+
+        let merkle_root_hash = parse_accumulator_merkle_root_from_vaa_payload(vaa_payload);
+        parse_and_verify_accumulator_updates(cursor, merkle_root_hash, clock)
+    }
+
+    // parse_accumulator_merkle_root_from_vaa_payload takes in some VAA bytes, verifies that the vaa
+    // corresponds to a merkle update, and finally returns the keccak hash of the merkle root.
```
jayantk, line 54: nitpick: comment here is slightly confusing. "keccak hash of the merkle root" sounds like it does an extra hash of the root for some reason. I think "merkle root" would be clearer.

target_chains/sui/contracts/sources/pyth_accumulator.move
```diff
@@ -0,0 +1,116 @@
...
+            let message = deserialize::deserialize_vector(cursor, (message_size as u64)); //should be safe to go from u16 to u16
+            let message_cur = cursor::new(message);
+            let price_info = parse_price_feed_message(&mut message_cur, clock);
+            cursor::take_rest(message_cur);
+
+            vector::push_back(&mut price_info_updates, price_info);
+
+            // isProofValid pops the next merkle proof from the front of cursor and checks if it proves that message is part of the
+            // merkle tree defined by merkle_root
+            assert!(merkle_tree::isProofValid(cursor, merkle_root, message), E_INVALID_PROOF);
```
jayantk, line 111: another nitpick: in a separate PR, can you rename the merkle_tree functions so they're also snake_case instead of camelCase?

jayantk (approved):
nice looks solid to me. All my comments are minor, so feel free to merge this and address them later.

PR #911, opened 2023-06-22
Price pusher: fix PTB bug

jayantk:
we forgot an await, and it seems like we're sending multiple concurrent transactions because of that.

I think this await does potentially mess up the loop timing because it stacks with the sleep below, but let's see if this fixes the error and we can adjust the timing logic after.

### DIFF
--- a/price_pusher/package.json
+++ b/price_pusher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-pusher",
-  "version": "5.3.2",
+  "version": "5.4.0",
   "description": "Pyth Price Pusher",
   "homepage": "https://pyth.network",
   "main": "lib/index.js",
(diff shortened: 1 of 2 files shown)


ali-bahjati (approved):
I think it was on purpose though to make sure timing is right. But i can imagine it's not great with the Sui frequency. I think for that high frequency we don't really need price pusher like this we can subscribe to the prices and just keep pushing without wait in a loop.

0xfirefist:
I also think this was on purpose. Say the pushing frequency is 5 seconds and we wait for 2 seconds, the next push will be in 7 seconds. We wanted to push every 5 seconds and not wait on if it was added to the blockchain. 

Also, it defeat the purpose of gasPrice increment in evm based pusher. We could probably keep waiting a long time. 
On injective we are pushing many transaction without waiting for the last one.

jayantk:
i see. let's try this a different way then.

PR #914, opened 2023-06-23
Sui pusher updates

jayantk:
Two small updates:
1. Flag to avoid sending duplicate transactions. This is a different approach to avoiding concurrent transactions that's sup-specific, as opposed to the previous `await` PR (which i closed).
2. CLI flag to set the maximum number of VAAs that can be included in a single sui PTB. At the moment, the SUI cost of our PTBs grows superlinearly with the number VAAs, so we want to split price updates into multiple PTBs to keep the costs down.

### DIFF
--- a/price_pusher/src/sui/sui.ts
+++ b/price_pusher/src/sui/sui.ts
@@ -83,6 +83,8 @@ export class SuiPriceListener extends ChainPriceListener {
 
 export class SuiPricePusher implements IPricePusher {
   private readonly signer: RawSigner;
+  private isAwaitingTx: boolean;
+
   constructor(
     private priceServiceConnection: PriceServiceConnection,
     private pythPackageId: string,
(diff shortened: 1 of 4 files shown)


price_pusher/src/sui/sui.ts
```diff
@@ -110,10 +116,64 @@ export class SuiPricePusher implements IPricePusher {
...
+    const priceFeeds = await this.priceServiceConnection.getLatestPriceFeeds(
+      priceIds
+    );
+    if (priceFeeds === undefined) {
+      console.log("Failed to fetch price updates. Skipping push.");
+      return;
+    }
 
-    const vaas = await this.priceServiceConnection.getLatestVaas(priceIds);
+    const vaaToPriceFeedIds: Record<string, string[]> = {};
```
ali-bahjati, line 134: In general I prefer Map over Record as it's less JS black magic.

ali-bahjati (approved):
Looks good to me. I understand the rationale behind it. Breaking down updates into multiple updates is only possible because we don't override a transaction and we want (and ignore). Maybe it's good to right down the assumptions somewhere in the file.

for me and @0xfirefist  It's good if we do it for other networks in another PR too, there are some unintuitive important tricks per network that can easily go out of hand.

PR #915, opened 2023-06-23
Clean up workflows & release process

jayantk:
We have two different ways of publishing / releasing code right now: you either push a tag to github, or you create a release in the github UI. This PR moves everything to the first way (which is also much simpler).

I also renamed all of the workflows so that it's clear what they do. Every workflow has a prefix like `ci-` for tests, or `push-` for docker image pushes.


.github/workflows/push-xc-admin-frontend-image.yml
```diff
@@ -1,6 +1,5 @@
 name: Build and Push Cross Chain Admin Frontend
 on:
-  pull_request:
```
jayantk, line 3: this is the only docker build / push that happens on PRs right now. It's not necessary to run on PRs though, because the xc_admin_frontend code is built and tested in the typescript ci workflow already.

ali-bahjati: Typescript ci test is not enough for this. This workflow was initially on main and got broken couple of times. It is because one should update the dockerfiles when a dependency added to the project.

guibescos: seems like it was added here : https://github.com/pyth-network/pyth-crosschain/commit/92aa157e856bfa500130f33d5ac1b429b9c4b2f2

ali-bahjati: I see. I think it's ok if it fails from time to time. We don't change the xc_* very often anymore.

ali-bahjati (approved):

PR #917, opened 2023-06-23
[eth] Add setWormholeAddress governance message

ali-bahjati:
This PR adds a new governance message to update wormhole address that is used in Pyth. This will allow us to change the Wormhole that Pyth uses to a gas optimized version of it developed in the PR #901. Per @jayantk suggestion, I added some sanity checks to make sure that the new Wormhole contract works properly and reject otherwise.  

## How to test
I added the tests to the js test suite because the rest of governance messages are tested there. To test you can either use tilt to do that automatically or manually run ganache-cli as a blockchain instance and test it. To do the latter, do the following:
1. Run `npx ganache-cli -e 10000 --deterministic --time="1970-01-02T00:00:00+00:00" --host=0.0.0.0` on a separate terminal to spawn a new network.
2. Run `cp .env.test .env && npx truffle compile --all && npx truffle migrate` to deploy the contracts
3. Run `npm run test-contract` to test the contract.

## Gas difference
The optimizer runs has not changed and hence no gas difference in benchmark tests. I updated truffle and hardhat to correspond with foundry that is used is gas benchmark.

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
@@ -34,6 +34,10 @@ abstract contract PythGovernance is
     );
     event FeeSet(uint oldFee, uint newFee);
     event ValidPeriodSet(uint oldValidPeriod, uint newValidPeriod);
+    event WormholeAddressSet(
+        address oldWormholeAddress,
+        address newWormholeAddress
+    );
 
     function verifyGovernanceVM(
         bytes memory encodedVM
@@ -86,6 +90,13 @@ abstract contract PythGovernance is
         ) {
             // RequestGovernanceDataSourceTransfer can be only part of AuthorizeGovernanceDataSourceTransfer message
             revert PythErrors.InvalidGovernanceMessage();
+        } else if (gi.action == GovernanceAction.SetWormholeAddress) {
+            if (gi.targetChainId == 0)
+                revert PythErrors.InvalidGovernanceTarget();
+            setWormholeAddress(
+                parseSetWormholeAddressPayload(gi.payload),
+                encodedVM
+            );
         } else {
             revert PythErrors.InvalidGovernanceMessage();
         }
@@ -189,5 +200,45 @@ abstract contract PythGovernance is
         setValidTimePeriodSeconds(payload.newValidPeriod);
 
         emit ValidPeriodSet(oldValidPeriod, validTimePeriodSeconds());
+    }
+
+    function setWormholeAddress(
+        SetWormholeAddressPayload memory payload,
+        bytes memory encodedVM
+    ) internal {
+        address oldWormholeAddress = address(wormhole());
+        setWormhole(payload.newWormholeAddress);
+
+        // We want to verify that the new wormhole address is valid, so we make sure that it can
+        // parse and verify the same governance VAA that is used to set it.
+        (IWormhole.VM memory vm, bool valid, ) = wormhole().parseAndVerifyVM(
+            encodedVM
+        );
+
+        if (!valid) revert PythErrors.InvalidGovernanceMessage();
+
+        if (!isValidGovernanceDataSource(vm.emitterChainId, vm.emitterAddress))
+            revert PythErrors.InvalidGovernanceMessage();
+
+        if (vm.sequence != lastExecutedGovernanceSequence())
+            revert PythErrors.InvalidGovernanceMessage();
+
+        GovernanceInstruction memory gi = parseGovernanceInstruction(
+            vm.payload
+        );
+
+        if (gi.action != GovernanceAction.SetWormholeAddress)
+            revert PythErrors.InvalidGovernanceMessage();
+
+        // Purposefully, we don't check whether the chainId is the same as the current chainId because
+        // we might want to change the chain id of the wormhole contract.
+
+        SetWormholeAddressPayload
+            memory newPayload = parseSetWormholeAddressPayload(gi.payload);
+
+        if (newPayload.newWormholeAddress != payload.newWormholeAddress)
+            revert PythErrors.InvalidGovernanceMessage();
+
+        emit WormholeAddressSet(oldWormholeAddress, address(wormhole()));
     }
 }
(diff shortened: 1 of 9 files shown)


target_chains/ethereum/contracts/test/pyth.js
```diff
@@ -1069,6 +1075,88 @@ contract("Pyth", function () {
...
+      .setup(
+        newImpl.address,
+        [testSigner1.address],
+        governance.CHAINS.polygon, // changing the chain id to polygon
+        wormholeGovernanceChainId,
+        wormholeGovernanceContract
+      )
+      .encodeABI();
+
+    const newWormhole = await Wormhole.new(newSetup.address, initData);
```
swimricky, line 1094: should we test this using `WormholeReceiver`? (or add a separate test case for it)

target_chains/ethereum/contracts/hardhat.config.ts
```diff
@@ -80,7 +80,7 @@ module.exports = {
     settings: {
       optimizer: {
         enabled: true,
-        runs: 5000,
+        runs: 2000,
```
swimricky, line 83: thanks for updating this!

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
```diff
@@ -190,4 +201,44 @@ abstract contract PythGovernance is
...
+
+        // We want to verify that the new wormhole address is valid, so we make sure that it can
+        // parse and verify the same governance VAA that is used to set it.
+        (IWormhole.VM memory vm, bool valid, ) = wormhole().parseAndVerifyVM(
+            encodedVM
+        );
+
+        if (!valid) revert PythErrors.InvalidGovernanceMessage();
+
+        if (!isValidGovernanceDataSource(vm.emitterChainId, vm.emitterAddress))
```
swimricky, line 220: we already do some of these checks in `verifyGovernanceVM` in `executeGovernanceInstruction`

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
```diff
@@ -190,4 +201,44 @@ abstract contract PythGovernance is
 
         emit ValidPeriodSet(oldValidPeriod, validTimePeriodSeconds());
     }
+
+    function setWormholeAddress(
+        SetWormholeAddressPayload memory payload,
+        bytes memory encodedVM
+    ) internal {
+        address oldWormholeAddress = address(wormhole());
+        setWormhole(payload.newWormholeAddress);
```
swimricky, line 210: Would it be safer to do 
```
(IWormhole.VM memory vm, bool valid, ) = IWormhole(payload.newWormholeAddress).parseAndVerifyVM(
            encodedVM
        );
// validation checks
setWormhole(payload.newWormholeAddress);
emit WormholeAddressSet(oldWormholeAddress, address(wormhole()));
```

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
```diff
@@ -190,4 +201,44 @@ abstract contract PythGovernance is
...
+            vm.payload
+        );
+
+        if (gi.action != GovernanceAction.SetWormholeAddress)
+            revert PythErrors.InvalidGovernanceMessage();
+
+        // Purposefully, we don't check whether the chainId is the same as the current chainId because
+        // we might want to change the chain id of the wormhole contract.
+
+        SetWormholeAddressPayload
```
swimricky, line 236: is this necessary? We already called `parseSetWormholeAddressPayload(gi.payload)` in `executeGovernanceInstruction`

swimricky (commented):
few questions but overall looks good. if my questions are non-issues lmk and  i'll approve this

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
```diff
@@ -190,4 +201,44 @@ abstract contract PythGovernance is
 
         emit ValidPeriodSet(oldValidPeriod, validTimePeriodSeconds());
     }
+
+    function setWormholeAddress(
+        SetWormholeAddressPayload memory payload,
+        bytes memory encodedVM
+    ) internal {
```
swimricky, line 208: should we restrict this to `private`

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
ali-bahjati: This is called via the new wormhole, it's just an additional sanity check that makes sure the new wormhole is parsing messages corerctly.

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
ali-bahjati: Same as above.

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
ali-bahjati: I think these are equally safe. Checking using the `wormhole()` after setting it just makes sure we use the same command we use in other places for the sanity check.

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
ali-bahjati: good point, i followed the pattern in this file but i don't see any reason why not

target_chains/ethereum/contracts/test/pyth.js
ali-bahjati: good point, yes! i'll do.

target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
jayantk: also suggest that the reverts here should return a more specific error, like "InvalidWormholeContract".

jayantk (approved):
LGTM. Couple minor comments but easy to address.

PR #918, opened 2023-06-24
[sui][price_pusher] Implement gas pool

666lcz:
This PR implements a gas object queue such that we can submit transactions concurrently by using different gas objects.

## Testing

```
npm run start -- sui \
  --endpoint https://wallet-rpc.testnet.sui.io \
  --pyth-package-id 0x975e063f398f720af4f33ec06a927f14ea76ca24f7f8dd544aa62ab9d5d15f44 \
  --pyth-state-id 0xd8afde3a48b4ff7212bd6829a150f43f59043221200d63504d981f62bff2e27a \
  --wormhole-package-id 0xcc029e2810f17f9f43f52262f40026a71fbdca40ed3803ad2884994361910b7e \
  --wormhole-state-id 0xebba4cc4d614f7a7cdbe883acc76d1cc767922bc96778e7b68be0d15fce27c02 \
  --price-feed-to-price-info-object-table-id 0xf8929174008c662266a1adde78e1e8e33016eb7ad37d379481e860b911e40ed5 \
  --price-service-endpoint https://xc-testnet.pyth.network \
  --mnemonic-file ./mnemonic \
  --price-config-file ./price-config.testnet.sample.yaml \
  --pushing-frequency 0.1 \
  --polling-frequency 0.1 \
  --num-gas-objects 50
```

observed 38 `Successfully updated price with transaction digest` within 30 seconds

PR #923, opened 2023-06-26
[price_pusher] Create gas pool for Sui

jayantk:
This PR is a version of https://github.com/pyth-network/pyth-crosschain/pull/918 that fixes a couple minor issues and CI. The  jist of the change is to split the SUI coins owned by the wallet into multiple roughly-equal-size coins, then maintain a queue of those coins in the pusher. Each transaction uses the gas coin from the head of the queue, returning it once the tx is done. This approach allows the blockchain to properly parallelize the execution of multiple concurrent transactions.

### DIFF
--- a/price_pusher/src/sui/sui.ts
+++ b/price_pusher/src/sui/sui.ts
@@ -315,8 +315,8 @@ export class SuiPricePusher implements IPricePusher {
       });
 
       nextGasObject = getTransactionEffects(result)
-        ?.mutated!.map((obj) => obj.reference)
-        .find((ref) => ref.objectId === gasObject.objectId)!;
+        ?.mutated?.map((obj) => obj.reference)
+        .find((ref) => ref.objectId === gasObject.objectId);
 
       console.log(
         "Successfully updated price with transaction digest ",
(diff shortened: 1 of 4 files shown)


price_pusher/src/sui/sui.ts
```diff
@@ -244,34 +288,199 @@ export class SuiPricePusher implements IPricePusher {
...
+      tx.setGasPayment([gasObject]);
+      tx.setGasBudget(this.gasBudget);
+      const result = await this.signer.signAndExecuteTransactionBlock({
+        transactionBlock: tx,
+        options: {
+          showEffects: true,
+        },
+      });
+
+      nextGasObject = getTransactionEffects(result)
```
ali-bahjati, line 317: I'm not understanding it correctly, how nextGasObject couldn't be the old gasObject here?

price_pusher/src/sui/sui.ts
```diff
@@ -244,34 +288,199 @@ export class SuiPricePusher implements IPricePusher {
...
+        )
+      );
+      hasNextPage = paginatedCoins.hasNextPage;
+      cursor = paginatedCoins.nextCursor;
+    }
+
+    if (numCoins !== coins.size) {
+      throw new Error("Unexpected getCoins result: duplicate coins found");
+    }
+    return [...coins].map((item) => JSON.parse(item));
```
ali-bahjati, line 401: why stringify and then again parse?

PR #924, opened 2023-06-27
[eth] - Aave FallbackOracle Integration

swimricky:
### Summary
- Add support for Pyth to be aave's fallback oracle by implementing `IPriceOracleGetter` and a mapping between assets and pyth's priceIds

### Notes
- This contract will be deployed as an immutable contract so no access/authority controls will be needed for `setAssetSources` since it can only be called once during initialization


target_chains/ethereum/contracts/contracts/pyth/PythPriceOracleGetter.sol
```diff
@@ -0,0 +1,40 @@
...
+        address asset
+    ) external view override returns (uint256) {
+        bytes32 priceId = _registryState.assetsPriceIds[asset];
+        if (asset == BASE_CURRENCY) {
+            return BASE_CURRENCY_UNIT;
+        }
+        if (priceId == 0) {
+            revert PythErrors.PriceFeedNotFound();
+        }
+        return uint256(uint64(pyth().getPriceUnsafe(priceId).price));
```
jayantk, line 38: I think we likely need to do some price conversion here using the exponent (?). I'm not sure what exactly Aave expects the uint256 to be, but we need to find out how they're representing decimals and convert to that.

target_chains/ethereum/contracts/contracts/pyth/PythAssetRegistry.sol
```diff
@@ -0,0 +1,67 @@
...
+        address[] memory assets,
+        bytes32[] memory priceIds
+    ) internal {
+        if (assets.length != priceIds.length) {
+            revert InconsistentParamsLength();
+        }
+        for (uint256 i = 0; i < assets.length; i++) {
+            _registryState.assetsPriceIds[assets[i]] = priceIds[i];
+            emit AssetSourceUpdated(assets[i], priceIds[i]);
+        }
```
jayantk, line 56: this function name sounds like it replaces all existing keys in the map, but it doesn't. Maybe `insertAssetSources` ?

target_chains/ethereum/contracts/contracts/pyth/PythAssetRegistry.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: Apache 2
```
jayantk, line 1: maybe put these in a separate folder like `aave` instead of `pyth`

target_chains/ethereum/contracts/contracts/pyth/PythPriceOracleGetter.sol
jayantk: and let's do a staleness check here (with a configurable threshold)

jayantk (commented):
generally looks good to me. couple small comments.

target_chains/ethereum/contracts/contracts/pyth/PythAssetRegistry.sol
swimricky: typically would agree, but since we're planning on deploying this as an immutable contract & only calling `setAssetSources` once when initializing the contract I think it would be better to leave the name as-is. It also follows the Aave convention. Can still update the naming if you feel strongly a bout this

target_chains/ethereum/contracts/contracts/pyth/PythPriceOracleGetter.sol
swimricky: it doesn't look like `AaveOracle` explicitly checks anything related to the exponent/decimals they just always assume that it matches and the values will be normalized correctly. I'll add an explicit conversion on ours to be safe.

also added a `validTimePeriodSeconds` in the constructor and use it in `getPriceNoOlderThan()` for `getAssetPrice`

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
```diff
@@ -0,0 +1,70 @@
...
+        if (priceId == 0) {
+            revert PythErrors.PriceFeedNotFound();
+        }
+        PythStructs.Price memory price = pyth().getPriceNoOlderThan(
+            priceId,
+            PythAssetRegistry.validTimePeriodSeconds()
+        );
+
+        // price.price is always positive and therefore this cast is safe.
+        uint256 normalizedPrice = uint64(price.price);
```
jayantk, line 51: I would avoid this assumption -- technically, we can offer feeds that go negative, e.g. if we start offering futures or something. Now I don't think AAVE would ever use such a feed, but I still think it's worthwhile to check if price is negative and fail if so.

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
```diff
@@ -0,0 +1,70 @@
...
+        int32 normalizerExpo = price.expo + int8(BASE_EXPONENTS);
+
+        if (normalizerExpo < 0) {
+            return normalizedPrice / (10 ** uint32(-normalizerExpo));
+        } else {
+            return normalizedPrice * (10 ** uint32(normalizerExpo));
+        }
+    }
+
+    function baseExponents(uint number) private pure returns (uint8) {
```
jayantk, line 62: seems like it would be more natural to pass the base exponent to the constructor and then do 10^exponent to get the BASE_CURRENCY_UNIT. That prevents misspecification (like right now BASE_CURRENCY_UNIT doesn't have to be a power of 10, and then something weird happens)

jayantk: Also, I think this returns -1 * the exponent ? i.e., if  `BASE_CURRENCY_UNIT * 10^e = 1`, then I think the `e = -18` . You could call this `base_num_decimals` which is typically the term for `-e`

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -0,0 +1,137 @@
...
+            address(pyth),
+            assets,
+            priceIds,
+            address(0x0),
+            BASE_CURRENCY_UNIT,
+            60
+        );
+    }
+
+    function generateRandomPriceFeedMessage(
```
jayantk, line 57: maybe put this in test utils? seems generally useful

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -0,0 +1,137 @@
...
+        while ((1 << depth) < priceFeedMessages.length) {
+            depth++;
+        }
+
+        depth += getRandUint8() % 3;
+
+        updateData[0] = generateWhMerkleUpdate(priceFeedMessages, depth, 1);
+
+        updateFee = pyth.getUpdateFee(updateData);
+    }
```
jayantk, line 90: same with this?

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -0,0 +1,137 @@
...
+            int64 pythPrice = price.price;
+            uint pythNormalizer;
+            uint price2;
+            if (price.expo < 0) {
+                pythNormalizer = 10 ** uint32(-price.expo);
+                price2 = uint64(pythPrice) / pythNormalizer;
+            } else {
+                pythNormalizer = 10 ** uint32(price.expo);
+                price2 = uint64(pythPrice) * pythNormalizer;
+            }
```
jayantk, line 111: I don't like this test because it replicates the same logic that's in the contract. would be better to create a price feed with specific test values in it, then explicitly write down what you think the answer should be.

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    function testGetAssetPriceWorksIfGivenUsdAddress() public {
+        address usdAddress = address(0x0);
+        uint256 assetPrice = pythOracleGetter.getAssetPrice(usdAddress);
+        assertEq(assetPrice, BASE_CURRENCY_UNIT);
+    }
+
+    function testGetAssetRevertsIfPriceNotRecentEnough() public {
+        vm.warp(block.timestamp + 5 days);
```
jayantk, line 123: suggest testing both sides of the boundary here. add 59 seconds, test for success, then add 1 more second and test for failure.

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
```diff
@@ -0,0 +1,70 @@
...
+        }
+        PythStructs.Price memory price = pyth().getPriceNoOlderThan(
+            priceId,
+            PythAssetRegistry.validTimePeriodSeconds()
+        );
+
+        // price.price is always positive and therefore this cast is safe.
+        uint256 normalizedPrice = uint64(price.price);
+        // price.expo is bound between [-12, 12]
+        int32 normalizerExpo = price.expo + int8(BASE_EXPONENTS);
```
jayantk, line 53: maybe we should also explicitly validate that `normalizerExpo` is not too large, such that we can't overflow in the multiplication case below. I think we're good as long 10^normalizerExpo can be written in (256 - 64) bits

jayantk (commented):
I think the math here is right. I have a couple suggestions to be extra cautious here, but should be easy to incorporate.

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
swimricky: the constructor in `AaveOracle` passes in the `BASE_CURRENCY_UNIT` so i followed their structure but if that's not important then i can modify it to what you suggested.

For Aave, if BASE_CURRENCY is USD (0x0) then BASE_CURRENCY_UNIT is 1e8 and `baseExponent()` will return 8 but i will update the naming to be `baseNumDecimals`

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
swimricky: added a separate test for the scenario you described. Keeping this test since there's still value in testing against a bunch of different values to make sure there's no over/underflow

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
```diff
@@ -47,19 +57,34 @@ contract PythPriceOracleGetter is PythAssetRegistry, IPriceOracleGetter {
...
 
-        // price.price is always positive and therefore this cast is safe.
+        // Aave is not using any price feeds < 0 for now.
+        if (price.price <= 0) {
+            revert InvalidNonPositivePrice();
+        }
         uint256 normalizedPrice = uint64(price.price);
         // price.expo is bound between [-12, 12]
-        int32 normalizerExpo = price.expo + int8(BASE_EXPONENTS);
+        int32 normalizerExpo = price.expo + int8(BASE_NUM_DECIMALS);
```
jayantk, line 66: the comment ^ isn't necessary because you have the check below right? I think a better comment would be on the check to say this prevents overflow in normalizedPrice.

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
```diff
@@ -47,19 +57,34 @@ contract PythPriceOracleGetter is PythAssetRegistry, IPriceOracleGetter {
...
-            return normalizedPrice * (10 ** uint32(normalizerExpo));
+        if (normalizedPrice <= 0) {
+            revert InvalidNonPositivePrice();
         }
+
+        return normalizedPrice;
     }
 
-    function baseExponents(uint number) private pure returns (uint8) {
+    function baseNumExponents(uint number) private pure returns (uint8) {
```
jayantk, line 87: baseNumDecimals

I guess it would be nice to validate that the input is a power of 10 as well, though not required.

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -134,4 +146,127 @@ contract PythAaveTest is Test, WormholeTestUtils, PythTestUtils, RandTestUtils {
...
         pythOracleGetter.getAssetPrice(addr);
     }
+
+    function testGetAssetPriceRevertsIfPriceIsNegative() public {
+        PriceFeedMessage[] memory priceFeedMessages = new PriceFeedMessage[](1);
+        PriceFeedMessage memory priceFeedMessage = PriceFeedMessage({
+            priceId: getRandBytes32(),
+            price: int64(-5), // assuming price should always be positive
+            conf: getRandUint64(),
+            expo: int32(getRandInt8() % 13), // pyth contract guarantees that expo between [-12, 12]
```
jayantk, line 156: you don't need this assumption for this contract, so i don't think you should make it. Use the assumptions mandated by the implementation instead.

target_chains/ethereum/contracts/forge-test/Pyth.Aave.t.sol
```diff
@@ -50,77 +51,88 @@ contract PythAaveTest is Test, WormholeTestUtils, PythTestUtils, RandTestUtils {
...
-                emaConf: getRandUint64()
-            });
-        }
-    }
-
-    function createWormholeMerkleUpdateData(
-        PriceFeedMessage[] memory priceFeedMessages
-    ) internal returns (bytes[] memory updateData, uint updateFee) {
-        updateData = new bytes[](1);
+    function testGetAssetPriceWorks() public {
```
jayantk, line 58: suggest having a few more tests like this (e.g., one with a positive exponent, one with zero, etc). You can make a test helper function like `testConversion(pythPrice, pythExpo, aavePrice)` or something and then throw a bunch of inputs at it.

jayantk (approved):
Nice! I have a couple very minor comments but feel free to address and merge, or address separately.

target_chains/ethereum/contracts/contracts/aave/PythPriceOracleGetter.sol
swimricky: added a check in the constructor to validate BASE_NUM_DECIMALS is a power of 10

PR #925, opened 2023-06-27
[hermes] Fix concurrency issue

ali-behjati:
Because of the storage design there was a race condition happening that both wh and pythnet try to store their state in the accumulator state and they ignore the other one because fetch and store are not done together consistently. This PR fixes that by adding `update_accumulator_state` to the Storage interface which takes a callback to apply.

### DIFF
--- a/hermes/src/store/storage.rs
+++ b/hermes/src/store/storage.rs
@@ -134,8 +134,20 @@ pub trait Storage: Send + Sync {
         filter: MessageStateFilter,
     ) -> Result<Vec<MessageState>>;
 
+    /// Store the accumulator state. Please note that this call will replace the
+    /// existing accumulator state for the given state's slot. If you wish to
+    /// update the accumulator state, use `update_accumulator_state` instead.
     async fn store_accumulator_state(&self, state: AccumulatorState) -> Result<()>;
     async fn fetch_accumulator_state(&self, slot: Slot) -> Result<Option<AccumulatorState>>;
+
+    /// Update the accumulator state inplace using the provided callback. The callback
+    /// takes the current state and returns the new state. If there is no accumulator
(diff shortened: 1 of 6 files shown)


hermes/src/store/storage.rs
```diff
@@ -134,8 +134,20 @@ pub trait Storage: Send + Sync {
         filter: MessageStateFilter,
     ) -> Result<Vec<MessageState>>;
 
+    /// Store the accumulator state. Please note that this call will replace the
+    /// existing accumulator state for the given state's slot. If you wish to
+    /// update the accumulator state, use `update_accumulator_state` instead.
     async fn store_accumulator_state(&self, state: AccumulatorState) -> Result<()>;
```
jayantk, line 140: Couple thoughts on this change:
1. is there ever a situation where it's safe to use this `store` function? Maybe you'd be better off deleting it?
2. Does the same change need to be done to storing the message_states above? I think the answer is "no", because you never mutate the content of `MessageState`. But then it's a little weird that the two types of stored values require different handling. You could make this consistent by splitting out the two components of AccumulatorState and fetching/storing them separately. That would be easy to do if you created a cache abstraction that handled the insertion / eviction logic (which would be good anyway, because i see that code is repeated for messages and accumulator states below).

jayantk (approved):
I think this is a fine solution to the problem, though it's a little awkward that the storage doesn't work the same for the two different types of values it stores (see inline comment). Not a big deal though, so up to you whether you want to change it.

hermes/src/store/storage.rs
ali-bahjati: I agree with you and David suggested the same. I will do it in another PR. having a proper abstraction for the simple ordered cache makes sense too.

PR #927, opened 2023-06-29
[sui 13/x] - store fee coins in `PriceInfoObject`

optke3:
## Summary
- Instead of sending fee coins to a fixed address or storing fee coins in `PythState`, a better approach is to store the coins in `PriceInfoObject`s. This preserves parallel price feed updates because there exists a `PriceInfoObject` for each price feed, and also makes it such that no fixed user account is the recipient of the fees.
- `AdminCap` (capability that enables holder to withdraw coins), `withdraw_fee_coins`, `deposit_fee_coins`.

## Notes
- Only one AdminCap is ever created. Do we want to allow cloning of AdminCaps?

## Testing
 - We tested that deposit fee coins works, and that they can be withdrawn using the AdminCap.


target_chains/sui/contracts/sources/pyth.move
```diff
@@ -411,6 +412,13 @@ module pyth::pyth {
     public fun get_total_update_fee(pyth_state: &PythState, n: u64): u64 {
         state::get_base_update_fee(pyth_state) * n
     }
+
+    // withdraw_fee_coins allows the holder of AdminCap to withdraw fee coins accumulated in any PriceInfoObject.
+    // pyth_state is passed in only for the purpose of version-checking via assert_latest_only, or making this function upgradeable
+    public fun withdraw_fee_coins(_admin_cap: &AdminCap, pyth_state: &PythState, price_info_object: &mut PriceInfoObject, ctx: &mut TxContext): coin::Coin<SUI>  {
+        let _ = state::assert_latest_only(pyth_state);
```
m30m, line 419: No need to assign to a variable

optke3:
Closing this PR, since dynamic fee storage was added in another commit that is backwards compatible with the Pyth contracts on mainnet.

https://github.com/pyth-network/pyth-crosschain/commit/9d2d673d0154917ed6bc467a2daf9fe6f6f4659a

The changes introduced in this PR are not backwards compatible because they involved adding new fields to the `PriceInfoObject` struct.

PR #934, opened 2023-06-30
[price-service/client] Crash by default on WsError

ali-bahjati:
This PR also makes the default logger console on warn and error. Also updates the compose files to point to new versions and adds more comments on building the price pusher image.


jayantk (approved):
🙇

PR #937, opened 2023-07-04
Update aptos pusher to support high TPS

jayantk:
The aptos pusher can't submit more than 1 transaction per 3 seconds. The problem seems to be that the submitted transactions don't land on chain, so the pusher tries to reuse an old sequence number, or simulation fails.

This change updates the pusher to fix this problem. It does three main things:
1. add local sequence number tracking and send each transaction with an incremented sequence number
2. remove the simulation / transaction replacement logic. The simulation RPC call fails often when you're submitting many transaction sequentially, presumably because the transaction is using the wrong blockchain state.
3. Stop using the update_if_necessary contract method, because we don't have a good way to track the "current" on-chain state as of when the tx is submitted.

### DIFF
--- a/price_pusher/package.json
+++ b/price_pusher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-pusher",
-  "version": "5.4.2",
+  "version": "5.4.3",
   "description": "Pyth Price Pusher",
   "homepage": "https://pyth.network",
   "main": "lib/index.js",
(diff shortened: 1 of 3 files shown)


ali-bahjati:
I understand that aptos and sui and injective have their own dedicated pushers but I still believe that high frequency centralized updates were not the intention of the entire price pusher design and is just its opposite. Two major reasons for using sequences and tx simulation everywhere was cost saving while increasing the reliablity (using the same sequence overrides usually with a higher gas). Now with this new design running two pushers i am not sure the same features are preserved (probably not). Btw we are running a single instance of pusher which can be very bad. 
I still think a high frequency pusher is a much simpler separate thing and i ask that at least we start documenting these assumptions per different network to avoid accumulating this amount of undocumented context. I asked last time as well but didn’t hear back 🚶‍♂️

PR #941, opened 2023-07-06
fix: Use generic_err constructor instead of literal

apollo-sturdy:
The cosmwasm-std `StdError` struct has a feature-gated field `backtraces` that you forgot to fill. This breaks compilation for anyone who depends on the Pyth SDK and tries to enable the backtraces feature. This PR simply uses the `StdError::generic_err` instead of creating the struct literal so that you don't have create the backtrace yourself.

See this issue for more info: https://github.com/CosmWasm/cosmwasm/issues/1760

### DIFF
--- a/target_chains/cosmwasm/Cargo.lock
+++ b/target_chains/cosmwasm/Cargo.lock
@@ -1471,7 +1471,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-sdk-cw"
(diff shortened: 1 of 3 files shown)


jayantk (approved):
thanks!

jayantk:
I checked that this passes the pre-commit checks locally. I think pre-commit tends to fail on external contributions for some reason.

PR #942, opened 2023-07-06
[sui 16/x] - update imports

optke3:
## Summary
- use official Mysten [sui.js](https://www.npmjs.com/package/@mysten/sui.js) lib in various Pyth TS scripts, including ones for deployment and contract initialization (previously there was a bug in the official package, so we had to fix it and use our own package)


m30m (approved):
LGTM, can you remove this package from dependencies before merging?

PR #946, opened 2023-07-10
[Sui 17/x] - accumulator price update from client-side 

optke3:
## Summary
- do price update from TS client using accumulator msg
- create price info objects for new Pyth testnet deployment

TS scripts for
- getting latest accumulator message from Hermes endpoint for target price feed
- updating price using accumulator msg
- creating and signing a generic Pyth governance message
- updating on-chain Pyth data sources by executing a governance action
- getting the VAA bytes out of an accumulator message (`parse_vaa_bytes_from_accumulator_message`)

Move unit test for 
- doing accumulator update with WH testnet guardian + Pyth testnet data sources

## Testing
- [accumulator price update txn result](https://suiexplorer.com/txblock/6rPdgxpHwFQPtZD2gMN78cnebWAL5myipYwhpyLNcg3y?network=testnet)


target_chains/sui/contracts/sources/pyth.move
```diff
@@ -919,6 +926,91 @@ module pyth::pyth_tests{
         test_scenario::end(scenario);
     }
 
+    #[test]
+    // for testnet debugging
+    fun test_create_and_update_single_price_feed_with_accumulator_message_for_testnet_success() {
```
m30m, line 931: What is the difference between this test and the previous test?If it's just the testnet, can we reuse some of the assertion logic?

optke3: the difference is in the initial guardians, data sources, and accumulator msg. The  `expected_price_info` is different, so had to redefine that. What assertion logic do you suggest be reused?

m30m:
Please fix pre-commits before merging

PR #949, opened 2023-07-11
[Sui 18/x] - governance action tests

optke3:
## Summary
- generate testnet governance VAAs for setting data sources, base update fee, stale price threshold
- confirm that these VAAs can be verified and executed via the `execute_governance_instruction` code path, and furthermore that they correctly update `PythState`


m30m (approved):
LGTM, It would be great if we can compress the testing logic even further. Like with a utility test function that gets a vaa and executes it. Maybe also another one liner for cleanup...

PR #953, opened 2023-07-13
[Sui 19/x] - sui contract manager additions

optke3:


contract_manager/src/chains.ts
```diff
@@ -231,16 +211,25 @@ export class SuiChain extends Chain {
       BigInt(fee),
       BigInt(exponent)
     ).encode();
-    return this.wrapWithWormholeGovernancePayload(3, setFee);
```
m30m, line 234: Since now the implementation is the same as the parent class, you don't need to override them anymore and can just delete the function.

contract_manager/src/chains.ts
```diff
@@ -231,16 +211,25 @@ export class SuiChain extends Chain {
...
+    return setFee
   }
 
   generateGovernanceSetDataSources(datasources: DataSource[]): Buffer {
     const setDataSource = new SetDataSources(
       this.wormholeChainName,
       datasources
     ).encode();
-    return this.wrapWithWormholeGovernancePayload(2, setDataSource);
```
m30m, line 242: Same as above

contract_manager/src/chains.ts
```diff
@@ -231,16 +211,25 @@ export class SuiChain extends Chain {
...
+    return setDataSource
   }
+
+  generateGovernanceSetStalePriceThreshold(newValidStalePriceThreshold: bigint): Buffer {
```
m30m, line 225: You can also move this to the parent Chain class, so that it can be used on other chains too

m30m (approved):

m30m:
Some minor changes needed + please fix pre-commit before merging

PR #955, opened 2023-07-14
Clean up create price feed scripts and use contract manager

m30m:


contract_manager/src/sui.ts
```diff
@@ -209,6 +209,33 @@ export class SuiContract extends Contract {
     return this.executeTransaction(tx, keypair);
   }
 
+  async executeCreatePriceFeeds(keypair: Ed25519Keypair, vaa: Buffer) {
```
optke3, line 212: we should also add a similar function `executeCreatePriceFeedsUsingAccumulator`, which calls [create_price_feeds_using_accumulator](https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/sui/contracts/sources/pyth.move#L100C16-L100C52), but this can be done in a later PR.

target_chains/sui/scripts/pyth/create_all_price_feeds.ts
```diff
@@ -1,118 +1,63 @@
...
+  console.log("wallet address: ", keypair.getPublicKey().toSuiAddress());
 
   // Fetch all price IDs
-  let { data } = await axios.get(price_feed_id_url);
-  const price_feed_ids = data;
+  const price_feed_ids = await connection.getPriceFeedIds();
   console.log("num price feed ids: ", price_feed_ids.length);
 
   // Create price feeds 20 at a time
   for (let chunk of _.chunk(price_feed_ids, 20)) {
```
optke3, line 49: we need to add some sleep logic here, perhaps 10-20 seconds between each iteration of the for-loop, otherwise this script will hit rate-limits and fail to create all price feeds

target_chains/sui/scripts/pyth/create_all_price_feeds.ts
```diff
@@ -1,118 +1,63 @@
...
-const walletPrivateKey = process.env.SUI_MAINNET; // <= NOTE: Update this when changing network
-const price_feed_id_url = "https://xc-mainnet.pyth.network/api/price_feed_ids"; // <= NOTE: Update this when changing network
-const connection = new PriceServiceConnection(
-  "https://xc-mainnet.pyth.network", // <= NOTE: Update this when changing network
-  {
-    priceFeedRequestConfig: {
-      binary: true,
+const parser = yargs(hideBin(process.argv))
+  .scriptName("create_all_price_feeds.ts")
+  .usage("Usage: $0 --contract <contract_id> --private-key <private_key>")
```
optke3, line 12: perhaps rename this to `package_id`? In the Sui docs they call it package instead of contract. The terminology seems standardized in their ecosystem

target_chains/sui/scripts/pyth/create_all_price_feeds.ts
```diff
@@ -1,118 +1,63 @@
...
+  .options({
+    contract: {
```
optke3, line 14: ditto

target_chains/sui/scripts/pyth/create_price_feed.ts
```diff
@@ -1,99 +1,45 @@
...
-);
+import { Ed25519Keypair } from "@mysten/sui.js";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+import { DefaultStore, SuiContract } from "@pythnetwork/pyth-contract-manager";
+
+const parser = yargs(hideBin(process.argv))
+  .scriptName("create_price_feed.ts")
+  .usage(
+    "Usage: $0 --contract <contract_id> --private-key <private_key> --vaa <vaa>"
```
optke3, line 9: ditto

target_chains/sui/scripts/pyth/create_price_feed.ts
```diff
@@ -1,99 +1,45 @@
...
+  )
+  .options({
+    contract: {
```
optke3, line 12: ditto

target_chains/sui/scripts/pyth/get_price_info_object_ids.ts
```diff
@@ -1,16 +1,30 @@
...
-const provider = new JsonRpcProvider(
-  new Connection({ fullnode: "https://fullnode.mainnet.sui.io:443" }) // <= NOTE: Update this when changing network
-);
-const objectId =
-  "0x14b4697477d24c30c8eecc31dd1bd49a3115a9fe0db6bd4fd570cf14640b79a0"; // <= NOTE: Update this when changing network AND with new deployment
-// ================================================================
-
+const parser = yargs(hideBin(process.argv))
+  .scriptName("get_price_info_object_ids.ts")
+  .usage("Usage: $0 --contract <contract_id>")
```
optke3, line 11: ditto

target_chains/sui/scripts/pyth/create_all_price_feeds.ts
m30m: Well, this is actually not the package_id it's a combination of the chain and state_id as defined in the contract manager. I agree that we should document it better.

PR #957, opened 2023-07-14
Abolish xc governance sdk

m30m:


contract_manager/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+node_modules/
```
ali-bahjati, line 2: this is in the global gitignore

ali-bahjati (approved):
Looks good to me. 

In general I think xc-admin-common is getting more complex and we can extract crosschain governance out of it (and also some multisig works) but maybe I'm wrong.

PR #964, opened 2023-07-19
Xc admin frontend refactor

m30m:
Small refactors and move the proposal instructions fetching inside the proposal component to avoid excessive rpc usage when the page is loaded


cctdaniel (approved):
thanks!

PR #966, opened 2023-07-19
[price_pusher] Refresh sui object versions

jayantk:
It seems like sui pusher's reliability problem is caused by the object versions of the gas coins in this code getting mismatched with the ones on-chain. This PR refreshes the version of the gas objects every time a transaction fails, which hopefully will fix the issue.

It's not easy to test this since it takes a long time to hit the error in the first place, but it should be safe to deploy this and see what happens. It's definitely not going to make things worse.

### DIFF
--- a/price_pusher/package.json
+++ b/price_pusher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-pusher",
-  "version": "5.4.5",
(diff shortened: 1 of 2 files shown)


ali-bahjati (approved):
I don't understand it well enough but technically lgtm.

PR #969, opened 2023-07-21
[hermes] Bump dependencies in Rust and Go

Reisen:
After digging deep to try and resolve a memory leak error in hermes, I tracked it down to libp2p. It seems this was already fixed previously [in go-libp2p-pubsub](https://github.com/libp2p/go-libp2p-pubsub/pull/528) and so wasn't a problem in our own codebase after all.

This PR therefore bumps libp2p to the latest versions to get ahead of problems like this, it also fixes the `Cargo.lock` pinned versions for our existing Rust dependencies so it continues to build against latest nightly.


ali-bahjati (approved):
Can you bump hermes version before merging it? (So we can deploy and test it)

PR #971, opened 2023-07-21
[contract-manager] Evm upgrades

m30m:


contract_manager/src/base.ts
```diff
@@ -20,6 +20,18 @@ export abstract class Storable {
...
+  conf: string;
+  publishTime: string;
+  expo: string;
+}
+
+export interface PriceFeed {
+  price: Price;
+  emaPrice: Price;
+}
+
```
ali-bahjati, line 34: Hmmm, we have a PriceFeed definition in price-service-sdk. I think it's good to have it as it is now and don't using it. Just saying it for awareness.

contract_manager/src/chains.ts
```diff
@@ -139,44 +203,78 @@ export class SuiChain extends Chain {
...
+      parsed.id,
+      parsed.mainnet,
+      parsed.wormholeChainName,
+      parsed.rpcUrl,
+      parsed.networkId
+    );
+  }
+
+  getRpcUrl(): string {
+    if (this.rpcUrl.includes("$INFURA_KEY") && !process.env.INFURA_KEY)
```
ali-bahjati, line 256: nitpik: maybe it's better to do this check on the constructor

contract_manager/src/entities.ts
```diff
@@ -206,26 +236,30 @@ export class WormholeMultiSigTransaction {
 
   async getState() {
     const proposal = await this.squad.getTransaction(this.address);
-    return proposal.status; //TODO: make it typed and parse the status to a more readable format
+    return Object.keys(proposal.status)[0];
```
ali-bahjati, line 239: JS 🪄 🧙

contract_manager/src/evm.ts
```diff
@@ -107,18 +146,87 @@ export class EVMContract extends Contract {
...
+    privateKey: string,
+    abi: any,
+    bytecode: string
+  ): Promise<EVMContract> {
+    const web3 = new Web3(chain.getRpcUrl());
+    const signer = web3.eth.accounts.privateKeyToAccount(privateKey);
+    web3.eth.accounts.wallet.add(signer);
+    const contract = new web3.eth.Contract(abi);
+    contract.options.data = bytecode;
+    const deployTx = contract.deploy({ data: "" });
```
ali-bahjati, line 166: why we couldn't pass bytecode as data 🤔 can you comment that?

contract_manager/src/evm.ts
```diff
@@ -107,18 +146,87 @@ export class EVMContract extends Contract {
...
+    const gas = await deployTx.estimateGas();
+    let gasPrice = await web3.eth.getGasPrice();
+
+    if (!chain.isMainnet()) {
```
ali-bahjati, line 170: can you comment why we do this?

contract_manager/src/evm.ts
```diff
@@ -107,18 +146,87 @@ export class EVMContract extends Contract {
...
+  async getLastExecutedGovernanceSequence() {
+    const pythContract = await this.getContract();
+    return Number(
+      await pythContract.methods.lastExecutedGovernanceSequence().call()
+    );
+  }
+
   async getPriceFeed(feedId: string) {
     const pythContract = this.getContract();
+    const feed = "0x" + feedId;
```
ali-bahjati, line 210: there is a method named `queryPriceFeed` which you can use to get both of them.

contract_manager/src/evm.ts
```diff
@@ -163,9 +271,36 @@ export class EVMContract extends Contract {
...
+      value: updateFee,
+    });
+    let gasPrice = await web3.eth.getGasPrice();
+    if (!this.chain.isMainnet()) {
+      gasPrice = (BigInt(gasPrice) * 2n).toString();
+    }
+    return transactionObject.send({
+      from: address,
+      value: updateFee,
+      gas: gasEstiamte * 2,
```
ali-bahjati, line 296: same comment on documenting why we do these.

target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
```diff
@@ -16,48 +21,61 @@ import {
...
+   * Returns a signer from a mnemonic and prefix to use for the executor
+   * @param mnemonic
+   * @param prefix chain address prefix (example - osmo)
+   */
+  static async getSignerFromMnemonic(mnemonic: string, prefix: string) {
+    const directSecp256k1HdWallet = await DirectSecp256k1HdWallet.fromMnemonic(
+      mnemonic,
+      { prefix }
+    );
+    await directSecp256k1HdWallet.getAccounts();
```
ali-bahjati, line 51: why this call is needed? Can you document it? 🤔

contract_manager/src/entities.ts
ali-bahjati, line 1: What is the responsibility of this module? Is it the controller? If so is this name the best name? 🤔

ali-bahjati (approved):
Fantastic! I left some minor inline comments.

I need to look at the codebase entirely to understand all the design decisions and grasp how different components relate to each other (and can be used) but this PR is very solid.

I also recommend moving chains definitions to a sub-module like this:
```
|- chains
  |- index.ts (or chains.ts on the folder outside)
  |- aptos.ts
  |- evm.ts
  |- ...
|- store.ts
|- entitities.ts
```

contract_manager/src/base.ts
```diff
@@ -42,12 +54,30 @@ export abstract class Contract extends Storable {
...
+  /**
+   * Returns the price feed for the given feed id
+   * @param feedId hex encoded feed id without 0x prefix
+   */
+  abstract getPriceFeed(feedId: string): Promise<PriceFeed | undefined>;
+
   /**
    * Executes the governance instruction contained in the VAA using the sender credentials
-   * @param sender based on the contract type, this can be a private key, a mnemonic, a wallet, etc.
+   * @param senderPrivateKey private key of the sender in hex format without 0x prefix
```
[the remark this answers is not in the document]
m30m: So far, I'm seeing hex format as the more common input format. I think if the implementation needs the uint8 array or Buffer it can easily convert the privateKey via `Buffer.from`

contract_manager/src/entities.ts
m30m: This was the first (and only file in the beginning) on this package. I'll organize it soon

contract_manager/src/evm.ts
m30m: I will try that 👍  This was a direct copy-paste from some example contract deployment code on the web

contract_manager/src/evm.ts
m30m: This was a dirty attempt to eliminate any need for manually specifying gas and gasPrice values per chain. For the `gasPrice` I was not sure how costly would it be to double the value on mainnet chains, so I only do it for testnet chains for now. If this is not an issue, I can do it everywhere and we don't need to worry a lot about manual entry of these values.
For the `gas` value, the `* 2` is just a very loose upper bound of how much gas is expected to be used.

target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
m30m: good catch

contract_manager/src/entities.ts
m30m: Added some comments

contract_manager/src/contracts/evm.ts
```diff
@@ -107,18 +148,82 @@ export class EVMContract extends Contract {
...
+  async getLastExecutedGovernanceSequence() {
+    const pythContract = await this.getContract();
+    return Number(
+      await pythContract.methods.lastExecutedGovernanceSequence().call()
+    );
+  }
+
   async getPriceFeed(feedId: string) {
     const pythContract = this.getContract();
+    const feed = "0x" + feedId;
```
jayantk, line 207: i think a class or type alias would be useful for price feed ids as well

contract_manager/src/contracts/evm.ts
```diff
@@ -163,9 +268,32 @@ export class EVMContract extends Contract {
     };
   }
 
-  async executeGovernanceInstruction(privateKey: string, vaa: Buffer) {
-    const web3 = new Web3(this.chain.rpcUrl);
-    const { address } = web3.eth.accounts.wallet.add(privateKey);
+  async executeUpdatePriceFeed(senderPrivateKey: string, vaas: Buffer[]) {
```
jayantk, line 271: fwiw i like this passing in the price updates version better than the one above that queries the price service internally

contract_manager/src/contracts/evm.ts
```diff
@@ -163,9 +268,32 @@ export class EVMContract extends Contract {
...
+    const web3 = new Web3(this.chain.getRpcUrl());
+    const { address } = web3.eth.accounts.wallet.add(senderPrivateKey);
+    const pythContract = new web3.eth.Contract(EXTENDED_PYTH_ABI, this.address);
+    const priceFeedUpdateData = vaas.map((vaa) => "0x" + vaa.toString("hex"));
+    const updateFee = await pythContract.methods
+      .getUpdateFee(priceFeedUpdateData)
+      .call();
+    const transactionObject =
+      pythContract.methods.updatePriceFeeds(priceFeedUpdateData);
+    const gasEstiamte = await transactionObject.estimateGas({
```
jayantk, line 281: gasEstimate

contract_manager/src/contracts/evm.ts
```diff
@@ -163,9 +268,32 @@ export class EVMContract extends Contract {
...
+      from: address,
+      gas: 100000000,
+      value: updateFee,
+    });
+    return transactionObject.send({
+      from: address,
+      value: updateFee,
+      gas: gasEstiamte * GAS_ESTIMATE_MULTIPLIER,
+      gasPrice: await this.chain.getGasPrice(),
+    });
```
jayantk, line 291: if you do this a bunch, consider making a helper function like sendWithGasEstimate

contract_manager/src/governance.ts
```diff
@@ -125,6 +128,33 @@ export class SubmittedWormholeMessage {
   }
 }
 
+/**
+ * A general executor that tries to find any contract that can execute a given VAA and executes it
+ * @param senderPrivateKey the private key to execute the governance instruction with
+ * @param vaa the VAA to execute
+ */
+export async function executeVaa(senderPrivateKey: string, vaa: Buffer) {
```
jayantk, line 136: this is going to be great

contract_manager/src/governance.ts
```diff
@@ -206,26 +236,32 @@ export class WormholeMultiSigTransaction {
...
+          await SubmittedWormholeMessage.fromTransactionSignature(
+            signature,
+            this.cluster
+          )
         );
       } catch (e: any) {
         if (!(e instanceof InvalidTransactionError)) throw e;
       }
     }
+    if (msgs.length > 0) return msgs;
```
jayantk, line 264: you sure about this check? seems like if none of the txs sent wormhole messages, then the proposal executes successfully and returns 0 messages

governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
```diff
@@ -113,6 +113,8 @@ export class PythGovernanceHeader {
       module = MODULE_TARGET;
       action = TargetAction[this.action as keyof typeof TargetAction];
     }
+    if (toChainId(this.targetChainId) === undefined)
+      throw new Error(`Invalid chain id ${this.targetChainId}`);
```
jayantk, line 117: fwiw we should make validity-checked types for all these things so that we don't need to put checks like this everywhere. ChainName shouldn't be able to represent an invalid chain.

contract_manager/store/chains/EVMChain/aurora_testnet.yaml
```diff
@@ -0,0 +1,6 @@
+id: aurora_testnet
+wormholeChainName: aurora
+mainnet: false
+rpcUrl: https://testnet.aurora.dev
+networkId: 1313161555
+type: EVMChain
```
jayantk, line 6: the ship may have sailed on this but i think this should be capitalized like `EvmChain`. I generally prefer to treat acronyms as a "word" for the purpose of camelcasing -- there's a whole discussion of this in the google coding conventions which i thought was persuasive. 

definitely not important though

contract_manager/src/chains.ts
jayantk: I agree with ali on this, though also understand your point. Ideally we would have a way to only instantiate the chains the user requests, such that we can perform validity checks on instantiation. Not sure if that's easy to do though.

at the very least, this seems worth a code comment explaining why it's like this and what it would take to improve the situation.

jayantk (approved):
very cool. I left a bunch of comments but nothing major.

contract_manager/src/chains.ts
```diff
@@ -7,56 +7,110 @@ import {
...
    * Returns the payload for a governance contract upgrade instruction for contracts deployed on this chain
    * @param upgradeInfo based on the contract type, this can be a contract address, codeId, package digest, etc.
    */
   abstract generateGovernanceUpgradePayload(upgradeInfo: any): Buffer;
 }
 
+export class GlobalChain extends Chain {
+  static type: string = "GlobalChain";
+  constructor() {
+    super("global", true, "unset");
```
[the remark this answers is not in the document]
m30m: Technically, it is one global chain x 2 multisig emitters. testnet contracts are listening to one emitter while mainnet contracts are listening to the other one.

contract_manager/src/governance.ts
m30m: So, this class was mainly designed for Wormhole messages, so I thought it should throw an exception if there is actually no wormhole instructions inside. I can add a check in constructor to verify this fact, but if someone is manually creating this object, hopefully they know what they are doing. I will add some comments. In the end, I think the main usage of this class will be in manual/dev testing as a cranker would do this automatically on our pre-configured vaults.

contract_manager/store/chains/EVMChain/aurora_testnet.yaml
m30m: Yeah, you mentioned this before and I thought I fixed it everywhere. I missed this one, will make the changes 👍

contract_manager/src/contracts/aptos.ts
```diff
@@ -60,6 +63,41 @@ export class AptosContract extends Contract {
...
+    if (priceInfo.price.negative) price = "-" + price;
+    return {
+      conf: priceInfo.conf,
+      publishTime: priceInfo.timestamp,
+      expo,
+      price,
+    };
+  }
+
+  async getPriceFeed(feedId: string) {
```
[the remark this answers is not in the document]
m30m: In general, I didn't annotate the implementations or write any docs if the abstract parent function was annotated. I thought most IDEs are smart enough to show relevant information like this:
image
If that's not the case I can copy paste it in the implementations too.

contract_manager/src/chains.ts
m30m: I decided to replace the current RPC endpoints with the public ones as the simplest solution and remove this logic😆

m30m:
Thanks for all the comments, some more cleanups and changes are definitely still remaining to do and will be done in followup PRs.

PR #980, opened 2023-07-25
[hermes] Refactor Storage

ali-bahjati:
This PR:
- Removes Storage trait and LocalStorage and implement everything on Storage struct. Storage is an unnecessary abstraction.
- Removes update_accumulator_state and breaks it down to accumulator_messages and
wormhole_merkle_state. This will reduce the complexity of modifying the data in-place.
- Use BTreeMap instead of Vector for storing ordered messages due to simplicity. Our write rate is not much and the increased time complexity is worth the simplicity. 

The reason for the changes above are reducing unnecessary complexity around the Storage and reducing over-engineerings. When we do this we don't need to have a dynamic reference and this will allow us to refine the interfaces (to take an iterator as input) much easier. Initially we had multiple Storage backends in mind but I believe we won't get close to it any time soon and we can make the change when the time comes.

You might wonder why we don't abstract different storage items as a generic ordered key/value store. I thought about this (and implemented it a bit) but the main reason that I stopped is slight differences in various items. MessageState is different in terms of storage and is going to support data retrieval from benchmarks api. AccumulatorMessages and WormholeMerkleState are the same but their backing BTreeMap is so simple that doesn't need another abstraction.


jayantk (approved):
seems reasonable to me. The diff on Storage is a bit hard to follow but I think this is mostly moving stuff over from LocalStorage + changing to the BTree. I didn't read this too closely but lmk if there's anything in there that you would like a closer look at.

Reisen (approved):

PR #981, opened 2023-07-25
[contract-manager] Evm contract manager verification

m30m:


contract_manager/src/contracts/evm.ts
```diff
@@ -197,10 +278,24 @@ export class EvmContract extends Contract {
...
    */
   async getCode(): Promise<string> {
+    // TODO: handle proxy contracts
     const web3 = new Web3(this.chain.getRpcUrl());
     return web3.eth.getCode(this.address);
   }
 
+  /**
+   * Returns the keccak256 digest of the contract bytecode after replacing any occurences of the contract addr in the bytecode with 0
+   * This is used to verify that the contract code is the same on all chains
```
ali-bahjati, line 288: Can you explain more why the address is in the contract bytecode?

PR #985, opened 2023-07-26
test(hermes): add tests for store

ali-bahjati:


hermes/src/store.rs
```diff
@@ -288,14 +295,365 @@ impl Store {
...
+        let (store, mut update_rx) = setup_store(10).await;
+
+        let price_feed_message = create_dummy_price_feed_message(100, 10);
+        // Populate the store with some inputs via calling them all at once and joining them.
+        let res = join_all(
+            generate_update(vec![Message::PriceFeedMessage(price_feed_message)], 10, 20)
+                .into_iter()
+                .map(|u| store.store_update(u)),
+        )
+        .await;
```
jayantk, line 436: maybe make a store_updates helper so you can do something like `store_updates(generate_updates(...)).await`

you could actually put the assert below into the helper, as that seems like it's a sanity check not part of the test logic (?)

hermes/src/store.rs
```diff
@@ -288,14 +295,365 @@ impl Store {
...
+    pub fn create_dummy_price_feed_message(seed: u8, publish_time: i64) -> PriceFeedMessage {
+        PriceFeedMessage {
+            feed_id: [seed; 32],
+            price: seed as _,
+            conf: seed as _,
+            exponent: 0,
+            ema_conf: seed as _,
+            ema_price: seed as _,
+            publish_time,
+            prev_publish_time: publish_time,
```
jayantk, line 404: fwiw you will probably need to set this field to a different value from publish_time in the future, because I think the "retrieve VAA for time" function needs to use this to disambiguate between the multiple updates at the same publish_time

hermes/src/store/storage.rs
```diff
@@ -45,6 +45,11 @@ pub struct MessageStateTime {
 pub struct MessageState {
     pub slot:        Slot,
     pub message:     Message,
+    /// The raw updated message.
+    ///
+    /// We need to store the raw message binary because Message
```
jayantk, line 50: nit: the Message struct

hermes/src/store/wormhole.rs
```diff
@@ -120,7 +119,15 @@ pub async fn verify_vaa<'a>(
...
 
-    let quorum = (guardian_set.keys.len() * 2 + 2) / 3;
+    // TODO: This check bypass checking the signatures on tests.
+    // Ideally we need to test the signatures but currently Wormhole
+    // doesn't give us any easy way for it.
+    let quorum = if cfg!(test) {
+        0
+    } else {
+        (guardian_set.keys.len() * 2) / 3 + 1
+    };
```
jayantk, line 129: i'm confused by this. in the old logic, if guardian_set.keys.len() == 0, then the quorum was 0 right? seems like the change to the arithmetic here is what made the cfg!(test) necessary?

jayantk (approved):

hermes/src/store.rs
ali-bahjati: Let's make this change in the future that we add the benchmarks endpoint. I'll add a comment explaining it.

ali-bahjati: 👍

hermes/src/store.rs
ali-bahjati: 👍

PR #988, opened 2023-07-27
chore(ethereum): deploy to base and syndr_testnet

ali-bahjati:


contract_manager/store/chains/EvmChain/base.yaml
```diff
@@ -0,0 +1,6 @@
+id: base
+wormholeChainName: base
+mainnet: true
+rpcUrl: https://developer-access-mainnet.base.org/
```
ali-bahjati, line 4: This is publicly available on chainlist, that's why i didn't hide it.

m30m (approved):

PR #990, opened 2023-07-27
[hermes] add utoipa for API docs

jayantk:
This PR adds utoipa for API docs to the /docs route. I just added docs for one endpoint, and they're not 100% correct. It's actually a little finicky to get the specification right, so we'll have to mess around with it a bit to understand how it works better.

I also drive-by fixed build.rs to detect the current build target and compile the go binary for that architecture.

### DIFF
--- a/hermes/Cargo.toml
+++ b/hermes/Cargo.toml
@@ -58,6 +58,8 @@ structopt              = { version = "0.3.26" }
 strum                  = { version = "0.24.1", features = ["derive"] }
 tokio                  = { version = "1.26.0", features = ["full"] }
 tower-http             = { version = "0.4.0", features = ["cors"] }
+utoipa = { version = "3.4.0", features = ["axum_extras"] }
+utoipa-swagger-ui = { version = "3.1.4", features = ["axum"] }
 wormhole-sdk           = { git = "https://github.com/wormhole-foundation/wormhole", tag = "v2.17.1" }
 
 [patch.crates-io]
--- a/hermes/Cargo.toml
+++ b/hermes/Cargo.toml
@@ -58,8 +58,8 @@ structopt              = { version = "0.3.26" }
 strum                  = { version = "0.24.1", features = ["derive"] }
 tokio                  = { version = "1.26.0", features = ["full"] }
 tower-http             = { version = "0.4.0", features = ["cors"] }
-utoipa = { version = "3.4.0", features = ["axum_extras"] }
-utoipa-swagger-ui = { version = "3.1.4", features = ["axum"] }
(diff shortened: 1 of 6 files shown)


hermes/src/api/rest.rs
```diff
@@ -95,15 +96,32 @@ pub async fn latest_vaas(
...
     ids:     Vec<PriceIdInput>,
     #[serde(default)]
+    #[param(value_type = Option<bool>, required = false, nullable = true)]
     verbose: bool,
     #[serde(default)]
+    #[param(value_type = Option<bool>, required = false, nullable = true)]
     binary:  bool,
 }
 
+/// Get the latest prices by price feed ids.
```
ali-bahjati, line 111: the formatting is broken.

hermes/build.rs
```diff
@@ -84,11 +86,30 @@ fn main() {
...
+            // CGO_ENABLED required for building amd64 on mac os
+            cmd.env("GOARCH", "amd64").env("CGO_ENABLED", "1");
+        }
+        "aarch64" => {
+            cmd.env("GOARCH", "arm64");
+        }
+        // Add other target architectures as needed
+        _ => {
+            panic!("Unsupported target architecture: {}", rust_target_arch);
+        }
```
ali-bahjati, line 101: Can you remove this to allow building in those targets as well? As long as rust and go use the same architecture this code should compile.

hermes/Cargo.toml
```diff
@@ -58,6 +58,8 @@ structopt              = { version = "0.3.26" }
 strum                  = { version = "0.24.1", features = ["derive"] }
 tokio                  = { version = "1.26.0", features = ["full"] }
 tower-http             = { version = "0.4.0", features = ["cors"] }
+utoipa = { version = "3.4.0", features = ["axum_extras"] }
```
ali-bahjati, line 61: Can you add some whitespases here? to keep them match with the rest of the file. (it's not enforces but good to have)

hermes/build.rs
```diff
@@ -84,11 +86,30 @@ fn main() {
...
     println!("cargo:rustc-link-lib=static=pythnet");
     println!("cargo:rustc-link-lib=resolv");
 
-    let status = cmd.status().unwrap();
-    assert!(status.success());
+    let go_build_output = cmd.output().expect("Failed to execute Go build command");
+    if !go_build_output.status.success() {
+        let error_message = String::from_utf8_lossy(&go_build_output.stderr);
+        panic!("Go build failed:\n{}", error_message);
+    }
```
ali-bahjati, line 114: 👍

ali-bahjati (approved):

PR #991, opened 2023-07-27
[Sui 20/x]- refactor Pyth governance code path

optke3:
## Summary
- Make it so that Pyth governance messages no longer adhere to the WH governance message format and instead comply with the Pyth governance message standard.
- Specifically, we eliminate some payload headers in the Pyth governance VAA which contain redundant info
- Use sequence number only for replay protection

## Notes
- Previously we used the `wormhole::governance_message` module helpers to verify governance messages for Pyth. This module assumes a specific format for the governance messages, which Pyth governance messages do not adhere to. Despite this, we were still able to use this module by modifying the Pyth governance messages a bit, resulting in some redundancy (specifically redundant header info).
- Remove all dependencies on `wormhole::governance_message`
- Define custom structs to separate the VAA verification process from VAA usage and do replay protection ourselves 
- Use `WormholeVAAVerificationReceipt` struct instead of `DecreeReceipt` to store the result of a Pyth-verified governance VAA

## Testing
We generate new VAAs and write unit tests for:
-  `SetFee`
- `SetDataSources`
- `SetStalePriceThreshold`


target_chains/sui/contracts/sources/governance/governance.move
```diff
@@ -7,43 +7,106 @@ module pyth::governance {
...
+    }
+
+    public fun destroy(receipt: WormholeVAAVerificationReceipt) {
+        let WormholeVAAVerificationReceipt{payload: _, digest: _, sequence: _} = receipt;
+    }
+
+   // We define a custom verify_vaa function instead of using wormhole::governance_message::verify_vaa
+   // because that function makes extra assumptions about the VAA payload headers. Pyth uses a
+   // different header format compared to Wormhole, so
+    public fun verify_vaa(
```
m30m, line 45: I believe we don't need additional verification on top of what wormhole is doing in the general `parse_and_verify_vaa` function. The exact match of the guardian_set_index is only necessary for wormhole upgrades, not ours. We just need to accept a `VAA` and check the data source with `is_valid_governance_data_source`

optke3: I think you're right about not having to check the guardian set index here, but I think we still need to check that the emitter address and emitter chain correspond to Pyth governance address and chain.

m30m: I believe that's done in `is_valid_governance_data_source` right?

target_chains/sui/contracts/sources/governance/governance.move
```diff
@@ -7,43 +7,88 @@ module pyth::governance {
...
         // Get the sequence number of the governance VAA that was used to
         // generate the receipt.
-        let sequence = governance_message::sequence(&receipt);
+        let sequence = receipt.sequence;
 
         // Require that new sequence number is greater than last executed sequence number.
         assert!(sequence > state::get_last_executed_governance_sequence(pyth_state),
             E_CANNOT_EXECUTE_GOVERNANCE_ACTION_WITH_OBSOLETE_SEQUENCE_NUMBER);
 
+        // consume VAA digest for replay protection
```
m30m, line 83: Is this also necessary? Do we have this logic on other contracts too? I think by just increasing the last executed sequence number, we should be safe

optke3: May remove the `ConsumedVAAs` struct from `PythState` in a future PR, since it is no longer being used.

m30m (approved):

PR #993, opened 2023-07-28
[xc-admin] Governance continue proposal

m30m:


contract_manager/store/chains/CosmWasmChain/injective.yaml
```diff
@@ -1,3 +1,4 @@
+querierEndpoint: https://injective-rpc.quickapi.com:443
```
ali-bahjati, line 1: For my understanding, how querierEndpoint can be different than executorEndpoint?

ali-bahjati (approved):

contract_manager/store/chains/CosmWasmChain/injective.yaml
m30m: In the old tooling of cosmwasm, this was only the case for injective (do not know exactly why). Right now, for injective, the executor endpoint is set automatically by their own sdk. So, I can remove the executor endpoint config in general. I will do this in another PR.

PR #995, opened 2023-07-28
[hermes] Add docs for 1 method

jayantk:
This PR adds reasonable docs for the 1 method that is documented. Everything seems to work *except* the schemas for PriceId and Price, which are defined in a different crate so will need some special handling.

This was relatively painful to get working (I had to trial and error a bunch of stuff), but I think it's pretty repeatable at this point. We just have to copy paste the exact same macros over to the remaining API endpoints.

### DIFF
--- a/hermes/src/api.rs
+++ b/hermes/src/api.rs
@@ -3,24 +3,18 @@ use {
     crate::store::Store,
     anyhow::Result,
     axum::{
+        extract::Extension,
         routing::get,
         Router,
     },
+    serde_qs::axum::QsQueryConfig,
     std::sync::Arc,
     tokio::{
         signal,
         sync::mpsc::Receiver,
     },
     tower_http::cors::CorsLayer,
-    utoipa::{
-        openapi::security::{
-            ApiKey,
-            ApiKeyValue,
-            SecurityScheme,
-        },
-        Modify,
-        OpenApi,
-    },
+    utoipa::OpenApi,
     utoipa_swagger_ui::SwaggerUi,
 };
 
@@ -54,7 +48,7 @@ pub async fn run(store: Arc<Store>, mut update_rx: Receiver<()>, rpc_addr: Strin
       rest::latest_price_feeds,
     ),
     components(
-      schemas(types::RpcPriceFeedMetadata, types::RpcPriceFeed)
+      schemas(types::RpcPriceFeedMetadata, types::RpcPriceFeed, types::PriceIdInput)
     ),
     tags(
       (name = "hermes", description = "Pyth Real-Time Pricing API")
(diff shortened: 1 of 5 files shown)


hermes/src/api.rs
```diff
@@ -80,7 +74,8 @@ pub async fn run(store: Arc<Store>, mut update_rx: Receiver<()>, rpc_addr: Strin
         .route("/api/get_vaa_ccip", get(rest::get_vaa_ccip))
         .route("/api/price_feed_ids", get(rest::price_feed_ids))
         .with_state(state.clone())
-        .layer(CorsLayer::permissive()); // Permissive CORS layer to allow all origins
+        .layer(CorsLayer::permissive()) // Permissive CORS layer to allow all origins
+        .layer(Extension(QsQueryConfig::new(5, false))); // non-strict mode permits escaped [] in URL parameters
```
jayantk, line 78: i added this because the `ids[]` parameter gets escaped by OpenAPI. It seems like this is a common enough problem in HTTP clients that there's an option to permit it here, so i think it's worth enabling that.

hermes/src/api/rest.rs
```diff
@@ -97,26 +97,34 @@ pub async fn latest_vaas(
 }
 
 #[derive(Debug, serde::Deserialize, IntoParams)]
+#[into_params(parameter_in=Query)]
 pub struct LatestPriceFeedsQueryParams {
-    #[param(value_type = String)]
+    /// Get the most recent price update for these price feed ids.
+    /// Provide this parameter multiple times to retrieve multiple price updates,
+    /// ids[]=a12...&ids[]=b4c...
```
jayantk, line 104: these code comments become the API documentation for each parameter

hermes/src/api/types.rs
```diff
@@ -19,23 +19,21 @@ use {
...
+/// A price id is a 32-byte hex string, optionally prefixed with "0x".
+/// Price ids are case insensitive.
+///
+/// Examples:
+/// * 0x63f341689d98a12ef60a5cff1d7f85c70a9e17bf1575f0e7c0b2512d48b1c8b3
+/// * 63f341689d98a12ef60a5cff1d7f85c70a9e17bf1575f0e7c0b2512d48b1c8b3
+///
+/// See https://pyth.network/developers/price-feed-ids for a list of all price feed ids.
+#[derive(Debug, Clone, Deref, DerefMut, ToSchema)]
+#[schema(value_type=String)]
```
jayantk, line 36: I tried for a while, but couldn't figure out how to get utoipa to enforce that this matches a regex for hexadecimal strings 🤷

ali-bahjati: I think it's fine

hermes/src/api/rest.rs
```diff
@@ -97,26 +97,34 @@ pub async fn latest_vaas(
...
+    #[param(
+        rename = "ids[]",
```
ali-bahjati, line 106: Will swagger ui work well with multi parameter input?

hermes/src/api.rs
ali-bahjati: Can you document what is 5 and why it is chosen? It's the allowed depth i assume :?

ali-bahjati (approved):
I couldn't check it manually but the code looks good

PR #997, opened 2023-07-31
[hermes] Document most of the endpoints

jayantk:
Add docs for the rest of the endpoints and fix the swagger ui errors.

I think the result here still needs a careful readthrough -- i'm pretty sure some examples and such are still missing -- but this should be a good step in the right direction.


hermes/src/api.rs
```diff
@@ -78,7 +83,7 @@ pub async fn run(store: Arc<Store>, mut update_rx: Receiver<()>, rpc_addr: Strin
         .layer(CorsLayer::permissive())
         // non-strict mode permits escaped [] in URL parameters.
         // 5 is the allowed depth (also the default value for this parameter).
-        .layer(Extension(QsQueryConfig::new(false)));
+        .layer(Extension(QsQueryConfig::new(5, false)));
```
Reisen, line 86: Might be nice to lift this to a const to keep in sync as I believe you have another 5 from another PR as well. Can place in the root `lib.rs` as a project-wide const.

hermes/src/api/types.rs
```diff
@@ -105,3 +106,72 @@ impl RpcPriceFeed {
         }
     }
 }
+
+/// A price with a degree of uncertainty at a certain time, represented as a price +- a confidence
```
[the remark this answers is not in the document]
Reisen: I'd actually argue this is the right thing, distinct RPC types is good practice and this forcing us to do it is beneficial. Having distinct Rpc and internal types provides a [lot of benefits when splitting data domains](https://www.parsonsmatt.org/2018/03/22/three_layer_haskell_cake.html) and protects against accidental API breakage.

Reisen (approved):

hermes/src/api.rs
jayantk: will get this const the next one so I don't have to wait for CI again.

PR #1002, opened 2023-08-03
[hermes] Finish up docs

jayantk:
I went through everything and updated the final values that were missing. I also found a way to extract out replicated example values to a single file so that they aren't hardcoded in a bunch of places.

### DIFF
--- a/hermes/src/doc_examples.rs
+++ b/hermes/src/doc_examples.rs
@@ -1,5 +1,11 @@
 use crate::store::types::UnixTimestamp;
 
+// Example values for the utoipa API docs.
+// Note that each of these expressions is only evaluated once when the documentation is created,
+// so the examples don't auto-update over time. We may want to adjust these examples for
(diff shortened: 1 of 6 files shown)


hermes/src/doc_examples.rs
```diff
@@ -0,0 +1,32 @@
+use crate::store::types::UnixTimestamp;
+
+// Example values for the utoipa API docs.
+// Note that each of these expressions is only evaluated once when the documentation is created,
+// so the examples don't auto-update over time. We may want to adjust these examples for
+// stable/beta in the future so that the example values in the docs work for both hermes and hermes-beta.
+// (Currently all example values are for stable.)
```
jayantk, line 7: there's probably a smart way to populate these values on initialization. I'm not sure what the best way to do it is though.

ali-bahjati: [lazh_static](https://docs.rs/lazy_static/latest/lazy_static/)?

ali-bahjati (approved):

PR #1004, opened 2023-08-04
[Sui 21/x] - pyth-sui-js SDK

optke3:
## Summary
- Pyth Sui JS SDK
- `SuiPriceServiceConnection` class
- `SuiRelay` example for fetching update data and updating a price feed
- `helpers` for updating price feed using accumulator message / batch price attestation, getting latest package ids given Wormhole or Pyth state, parsing VAA from accumulator msg, etc.
- add updated examples to `README`

## TODOs
- publish the `pyth-sui-js` npm package so that it is available to users via `npm install --save @pythnetwork/pyth-sui-js`


target_chains/sui/sdk/js/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { SuiPriceServiceConnection } from "./SuiPriceServiceConnection";
+export { SuiPythClient } from "./client";
```
ali-bahjati, line 2: Good to export other useful types as well like [this](https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/ethereum/sdk/js/src/index.ts#L3C1-L10)

target_chains/sui/sdk/js/src/SuiPriceServiceConnection.ts
```diff
@@ -0,0 +1,21 @@
...
+export class SuiPriceServiceConnection extends PriceServiceConnection {
+  /**
+   * Gets price update data (either batch price attestation VAAs or accumulator messages, depending on the chosen endpoint), which then
+   * can be submitted to the Pyth contract to update the prices. This will throw an axios error if there is a network problem or
+   * the price service returns a non-ok response (e.g: Invalid price ids)
+   *
+   * @param priceIds Array of hex-encoded price ids.
+   * @returns Array of price update data.
+   */
+  async getPriceFeedsUpdateData(priceIds: HexString[]): Promise<number[][]> {
```
ali-bahjati, line 16: `number[]` is a weird type. I prefer passing arrays of buffers instead and do the conversion on the edges as Buffer is more known as the data blob type.

target_chains/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,298 @@
...
+    return priceInfoObjects;
+  }
+  async createPriceFeed(tx: TransactionBlock, updates: number[][]) {
+    const wormholePackageId = await this.getWormholePackageId();
+    const packageId = await this.getPythPackageId();
+    if (updates.every((update) => this.isAccumulatorMsg(update))) {
+      if (updates.length > 1) {
+        throw new Error(
+          "SDK does not support sending multiple accumulator messages in a single transaction"
+        );
```
ali-bahjati, line 158: Why?

target_chains/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,298 @@
...
+      throw new Error("Can't mix accumulator and non-accumulator messages");
+    }
+
+    const priceInfoObjects: ObjectId[] = [];
+    for (const feedId of feedIds) {
+      const priceInfoObjectId = await this.getPriceFeedObjectId(feedId);
+      if (!priceInfoObjectId) {
+        throw new Error(
+          `Price feed ${feedId} not found, please create it first`
+        );
```
ali-bahjati, line 129: Ideally we don't want to create this ourself each time. Isn't it better to add the object creation part to the tx if possible?

target_chains/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,298 @@
...
+      }
+      priceInfoObjects.push(priceInfoObjectId);
+      const coin = tx.splitCoins(tx.gas, [tx.pure(1)]);
```
ali-bahjati, line 132: Is it assuming the update cost is always 1 per price feed? I think we should query it from on-chain.

ali-bahjati (commented):
It's the most complex sdk that we have had so far! I left some inline comments. 

Also on sui/scripts/pyth:
  - Can we remove Yarn in favor of our package?
  - It's good to make it a local package which relies local dependencies instead of their released version

target_chains/sui/sdk/js/src/client.ts
m30m: Just for the sake of simplicity of the sdk and the fact that it's rare.

ali-bahjati (approved):
Thanks! I'm still a bit unclear on the createPrice part and the single sized update data but i don't have much context to judge.

Please add this package to the global package.json file before merging.
https://github.com/pyth-network/pyth-crosschain/blob/main/package.json

target_chains/sui/sdk/js/src/client.ts
m30m: I will try but last time I checked it was not that easy.

m30m: PriceInfoObjects are shared objects and should be specified by user as inputs to the transaction. Therefore, it's not possible to create and use them in the same transaction.

PR #1006, opened 2023-08-08
fix(hermes): send close message on WS closure

ali-bahjati:
Some WS clients were receiving "Abnormal Connection Closure" errors. Usually clients handle this without any problem but not all of them. This PR fixes the issue by sending specific close message upon receiving close message from client. I tested it using Postman (to produce the error and check the fix).

This PR also refactors the ws.rs code by making it simpler and using tokio::time::Interval instead of a manually implemented interval. Lastly, it updates the axum package to include newer patches.

### DIFF
--- a/hermes/src/api/ws.rs
+++ b/hermes/src/api/ws.rs
@@ -39,7 +39,6 @@ use {
     },
     std::{
         collections::HashMap,
-        pin::Pin,
         sync::{
             atomic::{
                 AtomicUsize,
@@ -88,7 +87,7 @@ pub struct Subscriber {
     receiver:                SplitStream<WebSocket>,
     sender:                  SplitSink<WebSocket, Message>,
     price_feeds_with_config: HashMap<PriceIdentifier, PriceFeedClientConfig>,
-    ping_interval_future:    Pin<Box<tokio::time::Sleep>>,
+    ping_interval:           tokio::time::Interval,
     responded_to_ping:       bool,
 }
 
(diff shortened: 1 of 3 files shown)


hermes/src/api/ws.rs
```diff
@@ -170,6 +160,8 @@ impl Subscriber {
                     "Config missing, price feed list was poisoned during iteration."
                 ))?;
 
+            // `sender.feed` sends a message to the client but does not flush it, so we can send
```
Reisen, line 163: Probably "buffers a message to the client" to be explicit.

hermes/src/api/ws.rs
```diff
@@ -108,15 +107,15 @@ impl Subscriber {
...
+            ping_interval: tokio::time::interval(PING_INTERVAL_DURATION),
             responded_to_ping: true, // We start with true so we don't close the connection immediately
         }
     }
 
     pub async fn run(&mut self) {
         while !self.closed {
             if let Err(e) = self.handle_next().await {
-                log::warn!("Subscriber {}: Error handling next message: {}", self.id, e);
+                log::debug!("Subscriber {}: Error handling next message: {}", self.id, e);
```
ali-bahjati, line 118: None of the client errors are important and none are actually a problem that we should care. That's why it's changed to debug.

hermes/src/api/ws.rs
```diff
@@ -261,8 +259,10 @@ pub async fn notify_updates(ws_state: Arc<WsState>) {
             .map(|subscriber| async move {
                 match subscriber.send(()).await {
                     Ok(_) => None,
-                    Err(e) => {
-                        log::debug!("Error sending update to subscriber: {}", e);
+                    Err(_) => {
+                        // The channel is closed. When a client closes the connection
```
Reisen, line 263: I found this wording a little off, alternate wording:

```
// An error here indicates the channel is closed (which may happen either when the
// client has send Message::Close or some other abrupt disconnection). We remove
// subscribers only when send fails so we can handle closure only once when we are
// able to see send() fail.
```

Reisen (approved):
The tokio::interval is nice! Minor doc comments.

hermes/src/api/ws.rs
ali-bahjati: Thanks!

PR #1007, opened 2023-08-08
[price-pusher] Fix fee calculation on injective price pusher

m30m:


price_pusher/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-pusher",
-  "version": "5.4.6",
+  "version": "5.4.7",
```
ali-bahjati, line 3: The version is bumped manually and the lock is not updated. 
Please update the versions using 
`npm version [x.y.z|patch|minor|major] --no-git-tag-add`

ali-bahjati (approved):
The change seems good to me. Can you add more context on the PR description on why this is needed now because it seems that we are removing an optimization that i assume got there because of rate-limit or sth.

PR #1009, opened 2023-08-09
[evm] Wormhole receiver deploy all

m30m:
Scripts for deploying the wormhole-receiver to all chains and some small improvements on contract manager side


ali-bahjati (approved):
Nice!

PR #1011, opened 2023-08-11
[sui] Sui dx improvement

m30m:


target_chains/sui/scripts/pyth/pyth_deploy.ts
```diff
@@ -1,99 +1,204 @@
...
 import { execSync } from "child_process";
 
-import dotenv from "dotenv";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+import {
+  DefaultStore,
+  getDefaultDeploymentConfig,
+  SuiChain,
+} from "contract_manager";
```
optke3, line 20: did you mean to do something like this?

`} from "../../../../contract_manager/src/index";`

the code isn't compiling for me as is

optke3: hmm, it seems to work after I run

`npx lerna run build in the root directory
`
do you happen to know why that is?

target_chains/sui/scripts/pyth/create_price_feed.ts
```diff
@@ -1,99 +1,53 @@
...
+    "Usage: $0 --contract <contract_id> --feed-id <feed-id> --private-key <private-key>"
+  )
+  .options({
+    contract: {
+      type: "string",
+      demandOption: true,
+      desc: "Contract to update price feeds for (e.g sui_testnet_0xe8c2ddcd5b10e8ed98e53b12fcf8f0f6fd9315f810ae61fa4001858851f21c88)",
+    },
+    "feed-id": {
+      type: "array",
```
optke3, line 17:
```suggestion
 type: "string",
```

target_chains/sui/scripts/pyth/create_price_feed.ts
```diff
@@ -1,99 +1,53 @@
...
+      demandOption: true,
+      desc: "Price feed ids to create without the leading 0x (e.g f9c0172ba10dfa4d19088d94f5bf61d3b54d5bd7483a322a982e1373ee8ea31b). Can be provided multiple times for multiple feed updates",
```
optke3, line 19: confirmed that this script works

wonder why it doesn't work when you include the leading `0x` when specifying price feed ids (in which case it assumes it's an array of numbers and won't convert it to the type `[]string` properly)

m30m: It's because Javascript accepts `0x` prefixed strings as number literals

contract_manager/src/contracts/sui.ts
```diff
@@ -185,18 +192,114 @@ export class SuiContract extends Contract {
...
+        arguments: [
+          tx.object(this.wormholeStateId),
+          tx.pure([...vaa]),
+          tx.object(SUI_CLOCK_OBJECT_ID),
+        ],
+      });
+      verifiedVaas.push(verifiedVaa);
+    }
+    return verifiedVaas;
+  }
```
ali-bahjati, line 228: Can we reuse the login on the sui sdk?

m30m: Yeah, that was the plan after merging the other PR

contract_manager/src/contracts/sui.ts
```diff
@@ -185,18 +173,56 @@ export class SuiContract extends Contract {
...
+    const keypair = Ed25519Keypair.fromSecretKey(
+      Buffer.from(senderPrivateKey, "hex")
+    );
+    let result = await this.executeTransaction(tx, keypair);
+    return result.digest;
+  }
+  async executeCreatePriceFeed(senderPrivateKey: string, vaas: Buffer[]) {
+    const tx = new TransactionBlock();
+    const client = this.getSdkClient();
+    await client.createPriceFeed(tx, vaas);
```
ali-bahjati, line 219: I think I missed this on the previous review but `createPriceFeeds` seem to be a better name.

contract_manager/src/contracts/sui.ts
```diff
@@ -185,18 +173,56 @@ export class SuiContract extends Contract {
...
-      arguments: [tx.object(this.stateId), decreeReceipt],
+      arguments: [tx.object(this.stateId), verificationReceipt],
     });
 
     return this.executeTransaction(tx, keypair);
   }
 
   async executeUpdatePriceFeed(senderPrivateKey: string, vaas: Buffer[]) {
-    throw new Error("Not implemented");
+    throw new Error("Use executeUpdatePriceFeedWithFeeds instead");
```
ali-bahjati, line 191: I understand the complication here but it's not great to break the abstraction. Can you add some comments on why it's difficult to do and how we should make it work if we want to in the future?

target_chains/sui/cli/src/cli.ts
```diff
@@ -0,0 +1,302 @@
...
+        })
+        .usage(
+          "$0 create --contract <contract-id> --feed-id <feed-id> --private-key <private-key>"
+        );
+    },
+    async (argv) => {
+      const contract = getContract(argv.contract);
+      const defaultEndpoint = contract.getChain().isMainnet()
+        ? "https://xc-mainnet.pyth.network"
+        : "https://xc-testnet.pyth.network";
```
ali-bahjati, line 82: Can we move this to hermes?

target_chains/sui/cli/src/cli.ts
```diff
@@ -0,0 +1,302 @@
...
+  path: {
+    type: "string",
+    default: "../../contracts",
+    desc: "Path to the sui contracts, will use ../../contracts by default",
+  },
+  endpoint: {
+    type: "string",
+    desc: "Price service endpoint to use, defaults to https://xc-mainnet.pyth.network for mainnet and https://xc-testnet.pyth.network for testnet",
+  },
+  "feed-id": {
```
ali-bahjati, line 47: maye feed-ids? :think:

ali-bahjati (approved):
Nicely done! I left some minor inline comments.

contract_manager/src/contracts/sui.ts
m30m: oh, I had this implicit convention that functions starting with `execute` actually do something on-chain. That's why I named it like this

target_chains/sui/cli/src/cli.ts
m30m: I thought about using it like this:
`cli create-feed --feed-id f1 --feed-id f2`
Which makes more sense to use the singular word. Something like `docker run -p 80:80 -p 3000:3000`

PR #1012, opened 2023-08-11
fix(price_pusher): wait on tx to succeed

ali-bahjati:


m30m (approved):
LGTM, I'm in favor of simpler codes, so if the localCache for sequence number is not necessary and we are not hitting any RPC limits + always fetching the sequenceNumber doesn't lead to other issues (nodes lagging behind) maybe removing the cache all together works better.

ali-bahjati:
We are broadcasting updates every second and it takes some time for a tx to land on-chain. If we use the rpc sequence number many of our tx es will fail due to this.

PR #1024, opened 2023-08-22
[contract-manager] General improvements

m30m:


contract_manager/src/base.ts
```diff
@@ -1,6 +1,22 @@
 import { DataSource } from "xc_admin_common";
 import { Chain } from "./chains";
 
+export interface TxResult {
+  id: string;
+  info: any; // chain specific info
+}
+
+export type DeploymentType = "stable" | "edge";
```
ali-bahjati, line 9: let's rename "edge" to "beta".

contract_manager/src/contracts/aptos.ts
```diff
@@ -44,23 +44,28 @@ export class AptosContract extends Contract {
     return this.sendTransaction(senderPrivateKey, txPayload);
   }
 
-  private sendTransaction(
-    senderPrivateKey: string,
+  private async sendTransaction(
+    senderPrivateKey: PrivateKey,
     txPayload: TxnBuilderTypes.TransactionPayloadEntryFunction
   ) {
```
ali-bahjati, line 50: I know it's inferable but i think specifying the type is more clear.

ali-bahjati (approved):

contract_manager/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  rules: {
+    "@typescript-eslint/no-explicit-any": "off",
```
ali-bahjati, line 7: Can't we instead turn it off for the only place that it's defined? In general it's good to keep this enabled by default?

m30m: We had ~40 warnings, fixed most and ignored 4 of them

ali-bahjati (approved):

PR #1025, opened 2023-08-23
[cosmwasm] CosmWasm integration with contract manager

m30m:


target_chains/cosmwasm/deploy-scripts/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+   ```
+   npm run instantiate-pyth --contract-version <X.Y.Z> --private-key <YOUR_PRIVATE_KEY_HEX> --deploy <stable or edge>
+   ```
+
+4. You can test the new contract via contract manager scripts like this:
+
+   ```
+   cd ../../../contract_manager
+   npx ts-node src/update_pricefeed.ts --private-key <YOUR_PRIVATE_KEY_HEX> --contract <CONTRACT_ID> --feed-id <FEED_ID>
```
ali-bahjati, line 23:
```suggestion
   npx ts-node scripts/update_pricefeed.ts --private-key <YOUR_PRIVATE_KEY_HEX> --contract <CONTRACT_ID> --feed-id <FEED_ID>
```

target_chains/cosmwasm/deploy-scripts/README.md
```diff
@@ -0,0 +1,27 @@
...
+   ```
+
+5. Commit the new entries added to contract manager yaml files
```
ali-bahjati, line 26:
```suggestion
5. Make a PR  the new entries added to contract manager yaml files
```

target_chains/cosmwasm/deploy-scripts/src/instantiate-pyth.ts
ali-bahjati, line 1: I think if we rely on releases we should make sure that the releases cannot be modified easily. Otherwise a single hacked account is enough to cause an incorrect upgrade which can be dangerous. I think relying on builds for our deployment is better.

ali-bahjati (approved):
Looks good to me. I think that relying on releases is dangerous but it was there before (although i think it had hard-coded checksum) so I approve this PR. It's good to discuss it more and if needed fix it in a subsequent PR.

PR #1029, opened 2023-08-25
refactor(hermes): clean exit

Reisen:
This PR refactors hermes to have a cleaner shut down. It removes all sudden process exits and indicates to the current tasks they should shut down. This introduces about 1 second of shut down latency but allows for a clean exit. We currently have no tasks that corrupt from an abrupt shut down but this might not be true if we introduce stateful storage etc so adding it now will make it easier in future.

### DIFF
--- a/hermes/src/store.rs
+++ b/hermes/src/store.rs
@@ -341,7 +341,7 @@ mod test {
     /// Generate list of updates for the given list of messages at a given slot with given sequence
     ///
     /// Sequence in Vaas is used to filter duplicate messages (as by wormhole design there is only
-    /// one message per sequence)
+    /// one message per sequence).
     pub fn generate_update(messages: Vec<Message>, slot: Slot, sequence: u64) -> Vec<Update> {
         let mut updates = Vec::new();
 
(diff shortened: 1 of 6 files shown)


hermes/src/store.rs
```diff
@@ -341,7 +341,8 @@ mod test {
     /// Generate list of updates for the given list of messages at a given slot with given sequence
     ///
     /// Sequence in Vaas is used to filter duplicate messages (as by wormhole design there is only
-    /// one message per sequence)
+    /// one message per sequence).
+    #[cfg(test)]
```
ali-bahjati, line 345: isn't the entire test module under this cfg?

hermes/src/store/storage.rs
```diff
@@ -296,6 +296,7 @@ mod test {
         }
     }
 
+    #[cfg(test)]
```
ali-bahjati, line 299: isn't the entire test module under this cfg?

hermes/src/store.rs
Reisen: Ah I missed that, will remove.

ali-bahjati (approved):

PR #1031, opened 2023-08-25
feat(hermes): add benchmarks for histroical data

ali-bahjati:
This PR adds supports for historical price feeds through our benchmarks API. It is written in a way to compress the separate update data from multiple prices together. We currently don't offer an endpoint for fetching multiple historical prices at once but will offer it in the future with a newer API design or will add it to the current design. Some of the interfaces between the modules are changed to allow addition of benchmarks without complicating the design.


hermes/src/doc_examples.rs
ali-bahjati, line 1: changes via clippy

hermes/src/api/types.rs
```diff
@@ -78,30 +78,29 @@ impl RpcPriceFeed {
         verbose: bool,
         binary: bool,
     ) -> Self {
-        let price_feed_message = price_feed_update.price_feed;
+        let price_feed = price_feed_update.price_feed;
```
Reisen, line 81:
```suggestion
        let price_feed = price_feed_update.price_feed;
        let price = price_feed.get_price_unchecked();
```

Nitpick, probably optimises away anyway as you have it.

hermes/src/store/benchmarks.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+
+pub async fn get_price_feeds_with_update_data_from_benchmarks(
+    endpoint: Url,
+    price_ids: Vec<PriceIdentifier>,
+    publish_time: UnixTimestamp,
+) -> Result<PriceFeedsWithUpdateData> {
+    let endpoint = endpoint
+        .join(format!("/v1/updates/price/{}", publish_time).as_str())
```
Reisen, line 68:
```suggestion
        .join(&format!("/v1/updates/price/{}", publish_time))
```

hermes/src/store/benchmarks.rs
```diff
@@ -0,0 +1,109 @@
...
+impl TryFrom<BenchmarkPriceFeed> for PriceFeedUpdate {
+    type Error = anyhow::Error;
+
+    fn try_from(benchmark_price_feed: BenchmarkPriceFeed) -> Result<Self> {
+        let update_data = match benchmark_price_feed.binary.encoding {
+            BlobEncoding::Base64 => {
+                base64_standard_engine.decode(benchmark_price_feed.binary.data)?
+            }
+            BlobEncoding::Hex => hex::decode(benchmark_price_feed.binary.data)?,
+        };
```
Reisen, line 50: Readability nitpick.

```suggestion
        let data = &benchmark.binary.data;
        let update_data = match benchmark.binary.encoding {
            BlobEncoding::Base64 => base64_standard_engine.decode(data)?,
            BlobEncoding::Hex => hex::decode(data)?,
        };
```

hermes/src/store.rs
```diff
@@ -277,14 +304,41 @@ impl Store {
...
+                            price_ids,
+                            publish_time,
+                        )
+                        .await;
+                    }
+                }
+
+                Err(e)
+            }
+        }
```
Reisen, line 339: Nitpick: Can use `or_else` to simplify the error handling and remove a level of indentation.

```suggestion
    self.get_price_feeds_with_update_data_from_storage(price_ids.clone(), request_time.clone())
        .await
        .or_else(|e| {
            if let RequestTime::FirstAfter(publish_time) = request_time {
                if let Some(endpoint) = &self.benchmarks_endpoint {
                    return benchmarks::get_price_feeds_with_update_data_from_benchmarks(
                        endpoint.clone(),
                        price_ids,
                        publish_time,
                    )
                    .await;
                }
            }
            Err(e)
        })
```

hermes/src/store.rs
```diff
@@ -265,10 +278,24 @@ impl Store {
             .iter()
             .map(|message_state| match message_state.message {
                 Message::PriceFeedMessage(price_feed) => Ok(PriceFeedUpdate {
```
Reisen, line 280: `impl From<PriceFeedMessage> for PriceFeed` ?

hermes/src/store.rs
```diff
@@ -277,14 +304,41 @@ impl Store {
...
+    ) -> Result<PriceFeedsWithUpdateData> {
+        match self
+            .get_price_feeds_with_update_data_from_storage(price_ids.clone(), request_time.clone())
+            .await
+        {
+            Ok(price_feeds_with_update_data) => Ok(price_feeds_with_update_data),
+            Err(e) => {
+                if let RequestTime::FirstAfter(publish_time) = request_time {
+                    if let Some(benchmarks_endpoint) = &self.benchmarks_endpoint {
+                        return benchmarks::get_price_feeds_with_update_data_from_benchmarks(
```
Reisen, line 328: I'm very wary of the fact our storage abstraction falls back to a remote RPC. I feel like we should be separating out the abstraction but I'm not sure the best way just yet so I think this is fine for this PR, but indicates we might not have the right abstraction as we extend hermes in the future.

Reisen (approved):

hermes/src/api/types.rs
ali-bahjati: Thanks

hermes/src/store.rs
ali-bahjati: I will address it a newer PR

hermes/src/store.rs
ali-bahjati: It's not on this package. Do you think it's good to add pyth-sdk dependency in pythnet sdk? 🤔

Reisen (approved):

hermes/src/store.rs
Reisen: Could also do `impl Into<PriceFeed> for PriceFeedMessage`, but It's no big deal.

PR #1035, opened 2023-08-30
feat(hermes): add datadog profiling

Reisen:
This adds DataDog profiling to the Go portion of Hermes. Datadog provides two API's, APM tracing and full profiling. APM tracing allows defining meaningful spans over the code to measure, wheras profiling exposes all runtime metrics for the profiling mode specified.

We only have one small hot path in the message handling so the application is relatively simple, initially I opted for APM but tracing seems better.

There is a second option, which is to add `ddprof` to the Rust portion but I'm opting against this for now because while it will provide tracing for both the Rust and Go portion, the tracing will not be Go aware and be much more brittle. Ideally it would be nice to enable both, and prevent `ddprof` from analyzing past the Go boundary, but this seems difficult to achieve and I'm not sure the Rust portion will really benefit from tracing yet.

This also renames `GO_LOG_ALL` to `GO_LOG` which now takes a log level, so can be configured similar to `RUST_LOG` with `GO_LOG=info`


hermes/src/network/p2p.go
```diff
@@ -74,12 +77,34 @@ func RegisterObservationCallback(f C.callback_t, network_id, bootstrap_addrs, li
...
 		defer recoverRerun()
 
-		// Record the current time
+		// Record the current time for profiling and backoff purposes.
 		startTime = time.Now().UnixNano()
 
-		ctx := context.Background()
+		// Start the datadog profiler. Only starts if an environment is
+		// specified and a datadog API key is provided.
+		if os.Getenv("DD_PROFILING_ENV") != "" && os.Getenv("DD_API_KEY") != "" {
```
ali-bahjati, line 89: wouldn't it be better if we pass them by cli arguments (which supports env variable as well)?

Reisen: Yes I'll change it, I added them with changes to the `GO_LOG` which I intended to behave similar to `RUST_LOG` (no arg), this would be better.

ali-bahjati (approved):

PR #1038, opened 2023-09-02
feat(hermes): add tracing

Reisen:
Adding tracing/metrics to Hermes.


hermes/src/network/p2p.go
```diff
@@ -51,7 +51,7 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	libp2ptls "github.com/libp2p/go-libp2p/p2p/security/tls"
 	libp2pquic "github.com/libp2p/go-libp2p/p2p/transport/quic"
-	libp2pquicreuse "github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
+	// libp2pquicreuse "github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
```
ali-bahjati, line 54: i think you pushed this file

hermes/src/network/p2p.rs
```diff
@@ -72,28 +72,39 @@ lazy_static::lazy_static! {
...
     // Create a fixed slice from the pointer and length.
     let vaa = unsafe { std::slice::from_raw_parts(o.vaa, o.vaa_len) }.to_owned();
-    // The chances of the mutex getting poisioned is very low and if it happens
-    // there is no way for us to recover from it.
+
+    let hash = &{
+        let mut bytes = <sha3::Sha3_256 as sha3::Digest>::new();
+        sha3::Digest::update(&mut bytes, &vaa);
+        hex::encode(sha3::Digest::finalize(bytes))
+    };
```
ali-bahjati, line 84: what is this used for?

hermes/src/main.rs
```diff
@@ -70,9 +72,20 @@ async fn init() -> Result<()> {
 }
 
 #[tokio::main]
+#[tracing::instrument]
 async fn main() -> Result<()> {
     env_logger::init();
 
+    // Initialize a Tracing Subscriber
+    tracing::subscriber::set_global_default(
```
Reisen, line 80: Should be able to compose this with [tracing-opentelemetry](https://crates.io/crates/tracing-opentelemetry) if we want this in DataDog and so on.

hermes/src/network/p2p.rs
```diff
@@ -53,16 +53,41 @@ pub struct ObservationC {
...
 }
 
-pub type Observation = Vec<u8>;
+/// A wrapper around a VAA observed from Wormhole.
+///
+/// This wrapper tracks a Span that allows tracking the VAA through the system. This Span is
+/// expected to be tied to the `proxy` Span and so logging will always be traced back to the
+/// associated `proxy` Span regardless of where in the system it is being used.
+#[derive(Clone, Debug)]
+pub struct Vaa {
```
Reisen, line 62: This is temporary, I think we probably want `struct Traceable<T> { span: Span, data: T }` to support the idea of traceable data more generically, this was just a quick way to test.

hermes/src/network/p2p.rs
Reisen: It's associated with the Span that's stored with the VAA, so all logs related to this VAA are tagged with its hash, makes it easy to filter logs by VAA to get a focused trace on events related to this VAA as it moves through the program. Could probably use sequence number instead I just wanted to see for sure that duplicate VAAs were handled right.

hermes/src/store.rs
```diff
@@ -157,16 +162,35 @@ impl Store {
...
+
                         proof.slot
                     }
                 }
             }
+
             Update::AccumulatorMessages(accumulator_messages) => {
                 let slot = accumulator_messages.slot;
-                log::info!("Storing accumulator messages for slot {:?}.", slot,);
+                if let Some(state) = self.storage.fetch_wormhole_merkle_state(slot).await? {
```
Reisen, line 183: Temporary hack.

hermes/src/store/proof/wormhole_merkle.rs
```diff
@@ -108,9 +108,13 @@ pub fn construct_update_data(mut message_states: Vec<&MessageState>) -> Result<V
...
                         .clone();
 
+                    vaa.span.in_scope(|| {
+                        tracing::info!("Constructing update data for {} Messages.", messages.len())
+                    });
+
                     Ok(to_vec::<_, byteorder::BE>(&AccumulatorUpdateData::new(
                         Proof::WormholeMerkle {
-                            vaa:     vaa.into(),
+                            vaa:     (*vaa).to_owned().into(),
```
Reisen, line 117: Temporary hack, would not be able to store Span's with database storage for example.

ali-bahjati (approved):

PR #1039, opened 2023-09-04
fix: underflow issue

Sahilgat:
Fixes a minor underflow comparison issue in `OracleSwap.sol` example


ali-bahjati (approved):
Thanks for your contribution.

PR #1040, opened 2023-09-04
Price account v2 for remote clusters

guibescos:
Update xc-admin to create new Pythnet accounts with the new size


drozdziak1 (approved):
LGTM. I assume size choice logic is there to prevent uncontrolled v2 allocations on the native deployments with users accessing accounts directly (devnet, mainnet).

PR #1041, opened 2023-09-04
Update dockerfile

guibescos:
xc-admin cranks now depends on the solidity sdk library, so we need to do this


m30m (approved):

.github/workflows/lerna.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
guibescos, line 10: Drive-by to pass CI

PR #1042, opened 2023-09-05
fix(hermes): match p2p config with wh spy

ali-bahjati:


ali-bahjati:
Closing because #1038 includes this changeset.

PR #1049, opened 2023-09-13
chore: improve precommit ci

ali-bahjati:
By default, pre-commit ci runs for all files which takes so long. This is because `extra_args` is `--all-files` by default. This pr changes the `extra_args` on PRs to make sure only the changes in the PR are checked.


m30m (approved):
Nice!

PR #1062, opened 2023-09-22
feat(hermes): drop additional signatures from VAAs

ali-bahjati:


hermes/src/wormhole.rs
```diff
@@ -102,11 +102,31 @@ pub async fn verify_vaa<'a>(
...
+
+        if let Some(last_signed_id) = last_signed_id {
+            if last_signed_id >= signer_id {
+                return Err(anyhow!(
+                    "Signatures are not sorted by signer ID. Last signer ID: {}, current signer ID: {}",
+                    last_signed_id,
+                    signer_id
+                ));
+            }
+        }
```
Reisen, line 118: Matter of opinion but could:

```suggestion
        if let Some(true) = last_signed_id.map(|v| v >= signer_id) {
            return Err(anyhow!(
                "Signatures are not sorted by signer ID. Last signer ID: {}, current signer ID: {}",
                last_signed_id,
                signer_id
            ));
        }
```

hermes/src/wormhole.rs
```diff
@@ -102,11 +102,31 @@ pub async fn verify_vaa<'a>(
...
+        last_signed_id = Some(signer_id);
```
Reisen, line 119: Nitpick: could be at the bottom of the function, feels like the correct place to go as It's after everything.

hermes/src/wormhole.rs
```diff
@@ -141,15 +161,20 @@ pub async fn verify_vaa<'a>(
...
-        return Err(anyhow!(
+    // Check if we have enough correct signatures and update the header to
+    // contain only the minimal required signatures (quorum) to save gas
+    // upon VAA verification.
+    match valid_signatures.get(0..quorum) {
+        Some(valid_signatures) => {
+            header.signatures = valid_signatures.to_vec();
+            Ok((header, body).into())
+        }
+        None => Err(anyhow!(
```
Reisen, line 172: There's a few changes you can make to this function to make this a little clearer, remove the mutability, and flatten the flow a little bit to make it easier to follow (always nice to help the compiler with immutability optimizations), feel free to ignore as It's no big deal but up to you if you [like anything about this flow](https://pape.rs/2uMGXrdd) or to keep it the way it currently is, either is fine.

Reisen (approved):

PR #1069, opened 2023-09-28
Optimize merkle tree computations gas usage by minimizing vector data movements

m30m:
Analysis was done with [sui replay tool](https://github.com/MystenLabs/sui/pull/12680):
```
env MOVE_VM_PROFILE=1  ./sui-tool replay --rpc https://fullnode.mainnet.sui.io:443 tx -t DAKY6wVqkM1jybpCwa8HPL9XrNpEEVJF6cvJHeA7JbCo
env MOVE_VM_PROFILE=1  ./sui-tool replay --rpc https://fullnode.mainnet.sui.io:443 tx -t 5Kw1SvhAAaST5CVHByC5BtZwnCbj7fP6EFgqsahJ5ZQf
```
These are both transactions updating an identical set of feeds (3 feeds) on mainnet contracts:
Before optimization:
image
After optimization:
image

There is a 6x improvement visible although some of it might be related to sui gas bucketing. After this optimization, the cost of updating a single batch should (0.0062) is similar to updating 3 individual feeds (0.0069).

### DIFF
--- a/target_chains/sui/contracts/sources/merkle_tree.move
+++ b/target_chains/sui/contracts/sources/merkle_tree.move
@@ -3,7 +3,6 @@
 module pyth::merkle_tree {
     use std::vector::{Self};
     use sui::hash::{keccak256};
-    use wormhole::bytes::{Self};
     use wormhole::bytes20::{Self, Bytes20, data};
     use wormhole::cursor::{Self, Cursor};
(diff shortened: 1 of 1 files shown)


ali-bahjati (approved):
Nice! Interesting that the bytes operations are expensive and doing them one by one makes it cheaper.

PR #1073, opened 2023-09-29
chore: run ts/js ci only on relevant files

ali-bahjati:


.github/workflows/ci-typescript.yml
```diff
@@ -1,8 +1,22 @@
 name: JS/TS checks
 on:
   pull_request:
+    paths:
+      - "**/*.ts"
+      - "**/*.js"
+      - "**/*.sol"
```
ali-bahjati, line 7: we build some solidity contracts using npm

PR #1077, opened 2023-10-04
patch(hermes): improve ws reliability

ali-bahjati:
- Add max message size for incoming messages
- Add sent message rate limit and ip whitelisting


Reisen (approved):
Looks nice! One question on how the rate limiter works, could I get around it by repeatedly connecting and disconnecting?

PR #1078, opened 2023-10-05
feat(hermes): move to wormhole gRPC from P2P

Reisen:
This PR removes the current p2p code used to listen to Wormhole's P2P network and replaces it with the Wormhole gRPC definitions, using a Wormhole node gRPC endpoint to listen for VAAs instead. This let's us use standard wormhole nodes as our source of data and also allows us to move to more robust/load balanced solutions.

- Removes P2P options from `Config` and adds `gRPC` address.
- Removes old `wormhole.rs` module in favor of `network/wormhole.rs`
- Adds gRPC definitions from upstream Wormhole for using gRPC endpoints.
- Replaces p2p service with wormhole service for fetching and processing VAAs.


hermes/src/config/wormhole.rs
```diff
@@ -1,48 +1,27 @@
...
-    pub listen_addrs: Vec<Multiaddr>,
-
-    /// Network ID for Wormhole
-    #[arg(long = "wormhole-network-id")]
-    #[arg(default_value = DEFAULT_NETWORK_ID)]
-    #[arg(env = "WORMHOLE_NETWORK_ID")]
-    pub network_id: String,
+    /// gRPC endpoint for a Wormhole node.
+    ///
+    /// This can either be a standard Wormhole node gRPC endpoint or a beacon endpoint if
```
ali-bahjati, line 21: does node has the same api as `spy`? maybe call it spy_rpc_addr :?

thmzlt: I agree.

hermes/src/network/wormhole.rs
```diff
@@ -0,0 +1,340 @@
...
+            tracing::error!(error = ?e, "Wormhole gRPC service failed.");
+        }
+
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
+    }
+    Ok(())
+}
+
+async fn run(state: Arc<State>) -> Result<()> {
+    let mut client = SpyRpcServiceClient::connect("grpc://127.0.0.1:7073").await?;
```
thmzlt, line 159: Maybe pass the address as a value from the outside?

hermes/src/config/wormhole.rs
Reisen: Hmm, I thought the goal was to have them have the same API?

hermes/src/network/wormhole.rs
Reisen: That was the intent I missed this, will fix.

hermes/src/network/wormhole.rs
```diff
@@ -0,0 +1,340 @@
...
+        vaa.emitter_chain == Chain::Pythnet,
+        "VAA from non-Pythnet Chain."
+    );
+    ensure!(
+        vaa.emitter_address == Address(ACCUMULATOR_EMITTER_ADDRESS),
+        "VAA from non-Pythnet Emitter: {}",
+        vaa.emitter_address
+    );
+    Ok(())
+}
```
ali-bahjati, line 231: we might be able to move this to filters for spy

ali-bahjati (approved):

PR #1091, opened 2023-10-11
Fix the race condition by putting the setters at the end of the function

guibescos:
There was a race condition where a thread (with the wrong parameters) would set some of these fields to the wrong value before it could be cancelled. Hopefully this fixes it.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
+++ b/governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
@@ -101,39 +101,31 @@ export const useMultisig = (): MultisigHookData => {
           wallet: new NodeWallet(new Keypair()),
         })
         if (cancelled) return
-        setUpgradeMultisigAccount(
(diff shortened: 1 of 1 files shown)


governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
```diff
@@ -101,39 +101,31 @@ export const useMultisig = (): MultisigHookData => {
...
-          await readOnlySquads.getMultisig(UPGRADE_MULTISIG[multisigCluster])
+        const upgradeMultisigAccount = await readOnlySquads.getMultisig(
+          UPGRADE_MULTISIG[multisigCluster]
+        )
+
+        if (cancelled) return
+        const priceFeedMultisigAccount = await readOnlySquads.getMultisig(
+          PRICE_FEED_MULTISIG[multisigCluster]
         )
-        try {
```
guibescos, line 107: I thing this try blocks were useful when price feed multisigs didn't exist everywhere

cctdaniel (approved):

PR #1093, opened 2023-10-11
Initial version of random numbers contract

jayantk:
This PR adds the core protocol logic for random number generation. It needs a bunch of tests and little features, but the main ideas are all there.

Note that there's a private key in the deploy script that got deleted in the last commit. That private key is from foundry's default local VM, so it's not a problem that it's revealed in git history.


target_chains/ethereum/contracts/contracts/random/PythRandom.sol
```diff
@@ -0,0 +1,324 @@
...
+//    refuses to reveal the ith random number.
+// 5. The provider reveals x_i to the user.
+// 6. The user submits both the provider's revealed number x_i and their own x_U to the contract.
+// 7. The contract verifies hash(x_i) == x_{i-1} to prove that x_i is the i'th random number. The contract also checks that hash(x_U) == h_U.
+//    The contract stores x_i as the i'th random number to reuse for future verifications.
+// 8. If both of the above conditions are satisfied, the random number r = hash(x_i, x_U).
+//    (Optional) as an added security mechanism, this step can further incorporate the blockhash of the request transaction,
+//    r = hash(x_i, x_U, blockhash).
+//
+// This protocol has the same security properties as the 2-party randomness protocol above: as long as either
```
ali-bahjati, line 47: I think it's not entirely true. I discussed it previously with @jayantk.

target_chains/ethereum/contracts/contracts/random/PythRandomState.sol
```diff
@@ -0,0 +1,51 @@
...
+        // Metadata for the current commitment. Providers may optionally use this field to to help
+        // manage rotations (i.e., to pick the sequence number from the correct hash chain).
+        bytes32 commitmentMetadata;
+        // The first sequence number that is *not* included in the current commitment (i.e., an exclusive end index).
+        // The contract maintains the invariant that sequenceNumber <= endSequenceNumber.
+        // If sequenceNumber == endSequenceNumber, the provider must rotate their commitment to add additional random values.
+        uint64 endSequenceNumber;
+    }
+
+    // TODO: add block number?
```
ali-bahjati, line 36: it is added right?

target_chains/ethereum/contracts/forge-test/PythRandom.t.sol
```diff
@@ -0,0 +1,483 @@
...
+            provider1,
+            newHashChainOffset,
+            10
+        );
+        vm.prank(provider1);
+        random.register(
+            provider1FeeInWei,
+            newHashChain[0],
+            bytes32(keccak256(abi.encodePacked(uint256(0x0100)))),
+            newHashChainOffset + 10
```
ali-bahjati, line 328: Shouldn't this be 10?

target_chains/ethereum/contracts/forge-test/PythRandom.t.sol
```diff
@@ -0,0 +1,483 @@
...
+        );
+        assertInvariants();
+
+        uint64 sequenceNumber3 = request(user2, provider1, 42, false);
+        // Rotating the provider key uses a sequence number
+        assertEq(sequenceNumber3, sequenceNumber2 + 2);
+
+        // Requests that were in-flight at the time of rotation use the commitment from the time of request
```
ali-bahjati, line 336: nice test!

ali-bahjati (approved):
Very nice!

target_chains/ethereum/contracts/contracts/random/PythRandom.sol
jayantk: yeah, i think you're right that it's somewhat easier to attack the hash function itself. I looked at a few papers on the topic and it seems like they do the same thing we're doing here though (no added random value per element). I figure let's go with this for the moment, and we can ask TOB at the audit stage what they think.

target_chains/ethereum/contracts/forge-test/PythRandom.t.sol
```diff
@@ -0,0 +1,483 @@
...
+import "@pythnetwork/pyth-sdk-solidity/PythErrors.sol";
+import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
+import "./utils/WormholeTestUtils.t.sol";
+import "./utils/PythTestUtils.t.sol";
+import "./utils/RandTestUtils.t.sol";
+import "../contracts/random/PythRandom.sol";
+
+// TODO
+// - what's the impact of # of in-flight requests on gas usage?
+// - fuzz test?
```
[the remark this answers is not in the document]
jayantk: oh, sorry this comment is confusing. The issue is that verifying a provider's value requires hashing multiple times, where the number of times is approximately the number of in-flight requests. I think that should be fine since hashing is cheap, but it's probably worth having some explicit metrics.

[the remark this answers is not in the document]
jayantk: (i clarified the code comment)

PR #1096, opened 2023-10-12
feat(hermes): use ip from request headers for ratelimiting

ali-bahjati:
Previously we were using ip address from connecting socket for rate limiting but it is often an internal IP in production environments (as the service is behind a load balancer or proxy). This PR uses ip inside headers (by default `X-Forwarded-For`) instead to make sure we get the actual requester ip.

### DIFF
--- a/hermes/Cargo.lock
+++ b/hermes/Cargo.lock
@@ -481,7 +481,7 @@ dependencies = [
  "async-trait",
  "axum-core",
  "axum-macros",
- "base64 0.21.2",
+ "base64 0.21.4",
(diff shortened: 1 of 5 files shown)


Reisen:
I don't think we should rely on Cloudflare headers, in theory `X-Forwarded-For` should be being set by Cloudflare and should be identical, this is the standard header set (for example if you used `proxy_pass` in nginx this is what would be set). Can you check if it's available and switch to that, and use the cloudflare one as last resort?

Reisen (approved):

PR #1102, opened 2023-10-16
chore(evm): deploy to new networks

ali-bahjati:
chiliz_spicy and scroll


m30m (approved):
It seems that `target_chains/ethereum/contracts/networks/88888.json` is missing but LGTM

ali-bahjati:
I haven't deployed to that chain (chiliz mainnet) yet. Will the add the related files once I do.

PR #1103, opened 2023-10-16
Initial version of random number server

jayantk:
I'm working off of @Reisen 's initial version. This server implements all of the required protocol functionality for a single blockchain. It exposes an API endpoint that lets you retrieve a random number, and that endpoint only returns if the number has been requested on-chain.

The code also has some useful command line utilities for interacting with the contract. One of these commands runs the protocol end-to-end to produce a random number (which you can run to test that things are working).


pyth-rng/rustfmt.toml
ali-bahjati, line 1: we have this on top level (to be consistent everywhere). it's not needed here.

pyth-rng/src/config/generate.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub ethereum: EthereumOptions,
+
+    /// Submit a randomness request to this provider
+    #[arg(long = "provider")]
+    #[arg(default_value = "0x368397bDc956b4F23847bE244f350Bde4615F25E")]
+    pub provider: String,
+
+    #[arg(long = "url")]
+    #[arg(default_value = super::DEFAULT_HTTP_ADDR)]
+    pub url: String,
```
ali-bahjati, line 23: We can do type enforcement here (as long as they implement TryFrom<String>)

pyth-rng/src/config/get_request.rs
```diff
@@ -0,0 +1,24 @@
...
+#[group(id = "GetRequest")]
+pub struct GetRequestOptions {
+    #[command(flatten)]
+    pub ethereum: EthereumOptions,
+
+    /// Retrieve a randomness request to this provider
+    #[arg(long = "provider")]
+    #[arg(env = "PYTH_PROVIDER")]
+    #[arg(default_value = "0x368397bDc956b4F23847bE244f350Bde4615F25E")]
+    pub provider: String,
```
ali-bahjati, line 17: We can do stronger type enforcement here too.

pyth-rng/src/config/register_provider.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[command(flatten)]
+    pub ethereum: EthereumOptions,
+
+    #[command(flatten)]
+    pub randomness: RandomnessOptions,
+
+    /// The fee to charge (in wei) for each requested random number
+    #[arg(long = "pyth-contract-fee")]
+    #[arg(default_value = "100")]
+    pub fee: u64,
```
ali-bahjati, line 22: `u64` can be low in some chains (where the native token doesn't have much value)

pyth-rng/src/ethereum.rs
```diff
@@ -0,0 +1,121 @@
...
+        &self,
+        provider: &Address,
+        user_randomness: &[u8; 32],
+        use_blockhash: bool,
+    ) -> Result<u64, Box<dyn Error>> {
+        let hashed_randomness: [u8; 32] = Keccak256::digest(user_randomness).into();
+
+        if let Some(r) = self
+            .request(*provider, hashed_randomness, use_blockhash)
+            .value(200)
```
ali-bahjati, line 71: what is this value?

pyth-rng/src/command/generate.rs
```diff
@@ -0,0 +1,60 @@
...
+        .get(opts.url.join("/v1/revelation")?)
+        .query(&[("sequence", sequence_number)])
+        .build()?;
+    let resp = client
+        .execute(request_url)
+        .await?
+        .json::<GetRandomValueResponse>()
+        .await?;
+
+    println!(
```
Reisen, line 41: We should swap these out for `log` or `tracing` at some point.

pyth-rng/src/command/generate.rs
```diff
@@ -0,0 +1,60 @@
...
+    let client = reqwest::Client::new();
+    let request_url = client
+        .get(opts.url.join("/v1/revelation")?)
+        .query(&[("sequence", sequence_number)])
+        .build()?;
+    let resp = client
+        .execute(request_url)
+        .await?
+        .json::<GetRandomValueResponse>()
+        .await?;
```
Reisen, line 39:
```suggestion
let resp: GetRandomValueResponse = reqwest::get(opts.url.join("/v1/revelation")?)
    .query(&[("sequence", sequence_number)])
    .send()
    .await?
    .json()
    .await?;
```

pyth-rng/src/command/run.rs
```diff
@@ -0,0 +1,95 @@
...
+        error::Error,
+        sync::Arc,
+    },
+    tower_http::cors::CorsLayer,
+    utoipa::OpenApi,
+    utoipa_swagger_ui::SwaggerUi,
+};
+
+pub async fn run(opts: &RunOptions) -> Result<(), Box<dyn Error>> {
+    #[derive(OpenApi)]
```
Reisen, line 29: I'm not sure why rustfmt doesn't catch these but they're formatted so roughly here, I can catch these later but It's a shame they don't auto-format, I guess It's to do with being embedded within the derive.

pyth-rng/src/config.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Register a new provider with the Pyth Random oracle.
+    RegisterProvider(RegisterProviderOptions),
+
+    /// Request a random number from the contract.
+    RequestRandomness(RequestRandomnessOptions),
+
+    /// Generate a random number by running the entire protocol end-to-end
+    Generate(GenerateOptions),
+
+    GetRequest(GetRequestOptions),
```
Reisen, line 50: Docstring would be nice to make `--help` complete.

pyth-rng/src/config/register_provider.rs
```diff
@@ -0,0 +1,24 @@
...
+    },
+    clap::Args,
+    ethers::types::U256,
+};
+
+#[derive(Args, Clone, Debug)]
+#[command(next_help_heading = "Register Provider Options")]
+#[group(id = "RegisterProvider")]
+pub struct RegisterProviderOptions {
+    #[command(flatten)]
```
Reisen, line 14: I'm surprised this works, I'm pretty sure they'll get argument name conflicts due to flatten, if this compiles though ignore this comment, but I have a feeling it might trip up the help command if you try and `--help`, you'll likely get a runtime error. There isn't a great way to work around this but one common pattern would be to lift the common arguments up to before the command when they're common to multiple methods.

pyth-rng/src/ethereum.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// This method is a version of the autogenned `reveal` method that parses the emitted logs
+    /// to return the generated random number.
+    pub async fn reveal_wrapper(
+        &self,
+        provider: &Address,
+        sequence_number: u64,
+        user_randomness: &[u8; 32],
+        provider_randomness: &[u8; 32],
+    ) -> Result<[u8; 32], Box<dyn Error>> {
```
Reisen, line 100: We should really swap all these `Result` types (which I just added to get us going) with `anyhow::Result`, as we're using anyhow here anyway, and anyhow's result type offers all kinds of traceback/debugging nicety.

pyth-rng/src/config/register_provider.rs
jayantk: yeah i dunno. --help seems to work just fine. (I copied this from hermes)

Reisen (approved):

PR #1110, opened 2023-10-18
[pyth-rng] Add support for multiple blockchains

jayantk:
This PR adds support for a single service to return random numbers for multiple blockchains. Main changes here are:
* Extracted ethereum network parameters to a config file
* Make private-key optional for ethereum network instantiation (so you don't need it for every CLI command)
* Make the state store a map from chain ids -> random number chains

### DIFF
--- a/pyth-rng/config.yaml
+++ b/pyth-rng/config.yaml
@@ -1,4 +1,4 @@
 chains:
-  - chain_id: "optimism-goerli"
+  optimism-goerli:
     geth_rpc_addr: https://goerli.optimism.io
     contract_addr: 0x28F16Af4D87523910b843a801454AEde5F9B0459
--- a/pyth-rng/config.yaml
+++ b/pyth-rng/config.yaml
@@ -1,0 +1,4 @@
+chains:
+  - chain_id: "optimism-goerli"
+    geth_rpc_addr: https://goerli.optimism.io
+    contract_addr: 0x28F16Af4D87523910b843a801454AEde5F9B0459
--- a/pyth-rng/src/command/run.rs
+++ b/pyth-rng/src/command/run.rs
@@ -17,6 +17,7 @@ use {
         Router,
     },
     std::{
+        collections::HashMap,
         error::Error,
         sync::Arc,
     },
@@ -30,10 +31,11 @@ pub async fn run(opts: &RunOptions) -> Result<(), Box<dyn Error>> {
     #[openapi(
     paths(
     crate::api::revelation,
+    crate::api::chain_ids,
     ),
     components(
     schemas(
-    crate::api::GetRandomValueResponse
+    crate::api::GetRandomValueResponse,
     )
     ),
     tags(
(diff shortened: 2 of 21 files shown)


pyth-rng/src/config.rs
```diff
@@ -47,31 +52,31 @@ pub enum Options {
...
+    pub config: String,
+}
 
-    /// URL of a Geth RPC endpoint to use for interacting with the blockchain.
-    #[arg(long = "geth-rpc-addr")]
-    #[arg(env = "GETH_RPC_ADDR")]
-    #[arg(default_value = "https://goerli.optimism.io")]
-    pub geth_rpc_addr: String,
+impl ConfigOptions {
+    pub fn load(&self) -> Result<Config, Box<dyn Error>> {
```
Reisen, line 71: It would be better to move this method to `Config`, as this is a constructor for `Config` in disguise:

```rust
impl Config {
    fn from_file(path: &PathBuf) -> anyhow::Result<Self> {
        let yaml_content = fs::read_to_string(&self.config)?;
        let config: Config = serde_yaml::from_str(&yaml_content)?;
        ...
    }
}
```

Which you can call later with `Config::from_file(opts.config)?;` where you currently do this.

pyth-rng/src/config.rs
```diff
@@ -90,3 +95,60 @@ pub struct RandomnessOptions {
...
+        if duplicates.len() > 0 {
+            Err(anyhow!(format!(
+                "Config has duplicated chain ids: {}",
+                duplicates.join(",")
+            ))
+            .into())
+        } else {
+            Ok(())
+        }
+    }
```
Reisen, line 126: It would be nice to incorporate this into the parsing to avoid the decoupled check code, ala [dont validate, parse](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). Should be able to do this with something that looks like this:


```rust
mod chains {
    pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<HashMap<ChainId, String>, D::Error> {
        let vec: Vec<String> = Vec::deserialize(deserializer)?;
        let mut map = HashMap::new();
        ensure!(vec.duplicates_by(|v| v.chain_id).count() == 0, "duplicate chains in config");
        vec.into_iter()
            .try_fold(
                HashMap::new(),
                |mut acc, item| acc.insert(item.chain_id, item)
            )
    }
}

#[derive(Clone, Debug, serde::Deserialize)]
pub struct Config {
    #[serde(with = "chains")]
    pub chains: Vec<EthereumConfig>,
}
```

Note that along with the other suggestion of moving `load` to `from_file` cleans this up a lot.

pyth-rng/src/config.rs
```diff
@@ -90,3 +95,60 @@ pub struct RandomnessOptions {
...
+            .find(|x| x.chain_id == *chain_id)
+            .map(|c| c.clone())
+            .ok_or(
+                anyhow!(format!(
+                    "Could not find chain id {} in the configuration file",
+                    &chain_id
+                ))
+                .into(),
+            )
+    }
```
Reisen, line 140: This method can go with the above suggestion, as you can just do `config.chains.get(chain_id)?` thanks to having a hashmap.

pyth-rng/src/command/run.rs
```diff
@@ -42,34 +44,52 @@ pub async fn run(opts: &RunOptions) -> Result<(), Box<dyn Error>> {
...
+            offsets:     vec![provider_info
+                .original_commitment_sequence_number
+                .try_into()?],
+            hash_chains: vec![hash_chain],
+        };
+
+        if chain_state.reveal(provider_info.original_commitment_sequence_number)?
+            != provider_info.original_commitment
+        {
+            return Err(anyhow!(format!("The root of the generated hash chain for chain id {} does not match the commitment. Are the secret and chain length configured correctly?", &chain_config.chain_id)).into());
```
Reisen, line 74: `anyhow!` already formats, so you can do `anyhow!("Woah {}", 1);` over `anyhow!(format!("Woah {}", 1));`

pyth-rng/config.yaml
```diff
@@ -0,0 +1,4 @@
+chains:
```
Reisen, line 1: At the risk of bikeshedding, I'm not the biggest fan of config files and prefer CLI interfaces like repeated args, I'm open to anything here but did we ever come to a conclusion on what configuration files among Pyth projects should be? Is it to be YAML everywhere? Just want to double check before we end up with multiple configuration styles.

jayantk: Yeah so I went back and forth on this config file. For this particular use case though, we're going to end up with a big list of different ethereum chains with corresponding RPC urls, and it seemed like this would end up being quite awkward to pass on the command line. I've kept everything else on the command line for now though.

We use YAML in most places for config (except typescript which uses .env sometimes, but that really sucks).

pyth-rng/src/config.rs
jayantk: i kept this because the error wrapping is useful (reduces code duplication because this is called in a few places). but implementation is much simpler with the hashmap in the config object

pyth-rng/src/config.rs
jayantk: ah, smart. I ended up using the default serde HashMap support (didn't realize that was in there), but this is much better.

I didn't extend the chain config to an enum, as i think this is easy to do later when we need it.

Reisen (approved):

PR #1124, opened 2023-10-29
Entropy Solidity SDK & usage example 

jayantk:
I wanted to make an example for how to use entropy. In order to do that, I needed to extract out the interface of the entropy contract into an SDK. 

There is still work to do on the example. I've only added the solidity contract for the moment and checked that it builds. I think we also need to generate ABIs in the solidity SDK and then show how to interact with the example contract via typescript. I think that can wait for a future PR though because there's already a decent amount of stuff here.


fortuna/config.yaml
```diff
@@ -2,3 +2,9 @@ chains:
   optimism-goerli:
     geth_rpc_addr: https://goerli.optimism.io
     contract_addr: 0x28F16Af4D87523910b843a801454AEde5F9B0459
+  avalanche-fuji:
+    geth_rpc_addr: https://api.avax-test.network/ext/bc/C/rpc
+    contract_addr: 0xD42c7a708E74AD19401D907a14146F006c851Ee3
+  eos-evm-testnet:
+    geth_rpc_addr: https://api.testnet.evm.eosnetwork.com/
+    contract_addr: 0xD42c7a708E74AD19401D907a14146F006c851Ee3
```
jayantk, line 10: drive-by: deployed on some other networks

target_chains/ethereum/entropy_sdk/solidity/IEntropy.sol
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: Apache 2
```
jayantk, line 1: this is copy/paste of the interface from PythRandom.

I've called this IEntropy to match the new name scheme. I'll circle back and rename the other contracts as well (didn't want to intermingle in this one PR).

fortuna/config.yaml
ali-bahjati: ToDo: contract manager support

Reisen (approved):

PR #1129, opened 2023-11-01
[fortuna] Support multiple random value encodings

jayantk:
Per the API design guidelines, add support for multiple binary encodings to fortuna. This PR makes the endpoint take a `?encoding=` query parameter. Binary values are returned as a json object of the form: `{ encoding: "hex", data: "123ab" }`

(this is a nice-to-have, but I think it's going to save us a lot of questions from users of the form "how do i convert this array to a hex string")

### DIFF
--- a/fortuna/src/api/revelation.rs
+++ b/fortuna/src/api/revelation.rs
@@ -8,18 +8,19 @@ use {
     axum::{
         extract::{
             Path,
+            Query,
             State,
         },
         Json,
     },
     pythnet_sdk::wire::array,
+    serde_with::serde_as,
     utoipa::{
         IntoParams,
         ToSchema,
     },
 };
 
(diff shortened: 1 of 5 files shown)


fortuna/src/api/revelation.rs
```diff
@@ -80,15 +84,67 @@ pub async fn revelation(
...
+impl Blob {
+    pub fn new(encoding: BinaryEncoding, data: [u8; 32]) -> Blob {
+        match encoding {
+            BinaryEncoding::Hex => Blob::Hex { data },
+            BinaryEncoding::Base64 => Blob::Base64 { data },
+            BinaryEncoding::Array => Blob::Array { data },
+        }
+    }
+
+    pub fn data(&self) -> &[u8; 32] {
```
jayantk, line 143: this kind of sucks -- is there a better way to do this?

Reisen (approved):

PR #1133, opened 2023-11-06
[fortuna] Add tests

jayantk:
Add some tests for the `revelations` endpoint. I also factored out a trait that lets us mock the on-chain contract for testing purposes, which I'm using for these tests.

Some of this code is definitely well beyond my level of real rust knowledge, so please let me know if there are better ways to do any of the things I'm doing here.


fortuna/src/api.rs
```diff
@@ -51,13 +55,25 @@ pub struct ApiState {
     pub metrics: Arc<Metrics>,
 }
 
+impl ApiState {
+    pub fn new(chains: &[(ChainId, BlockchainState)]) -> ApiState {
+        let map: HashMap<ChainId, BlockchainState> =
+            chains.into_iter().map(|r| (*r).clone()).collect();
```
Reisen, line 61: `.map(|r| (*r).clone()` can be replaced with just `.cloned()`

fortuna/src/api.rs
```diff
@@ -137,3 +153,229 @@ impl IntoResponse for RestError {
...
+            TestResponse,
+            TestServer,
+        },
+        ethers::prelude::Address,
+        once_cell::sync::Lazy,
+        std::sync::Arc,
+    };
+
+    const PROVIDER: Address = Address::zero();
+    const OTHER_PROVIDER: Lazy<Address> = Lazy::new(|| Address::from_low_u64_be(1));
```
Reisen, line 203: I generally think `lazy_static` is a better library/interface for this, it's sad they provide no `const` initializers from arrays. Though it seems this type is just an alias for H160 and the content of that is public, so you might be able to do:

```
const OTHER_PROVIDER: Address = Address([1u8; 20]);
```

fortuna/src/api.rs
```diff
@@ -137,3 +153,229 @@ impl IntoResponse for RestError {
...
+    const ETH_CHAIN: Lazy<Arc<HashChainState>> = Lazy::new(|| {
```
Reisen, line 204: The problem with having these as Arc's is their state will persist between other tests. In general I think It's a bad idea to do this, especially as Rust runs tests asynchronously leading to non-deterministic and unpredictable behaviour. I think you're better off moving these inside your test_server() initializer and returning them there, or just setting them up at the start of the test.

fortuna/src/chain/reader.rs
```diff
@@ -0,0 +1,89 @@
...
+    anyhow::Result,
+    axum::async_trait,
+    ethers::types::Address,
+};
+
+/// EntropyReader is the read-only interface of the Entropy contract.
+#[async_trait]
+pub trait EntropyReader: Send + Sync {
+    /// Get an in-flight request (if it exists)
+    /// Note that if we support additional blockchains in the future, the type of `provider` may
```
Reisen, line 11: I think you might want to deal with the generic here now, because this trait is a public interface. If we add a new chain this will break peoples working code and if It's not generic it defeats the purpose of the trait in the first place. You cant get here quite fast with a less likely to break trait using associated types:

```
pub trait EntropyReader: Send + Sync {
    type Provider;
    async fn get_request(&self, provider: Self::Provider, sequence_number: u64)
        -> Result<Option<Request>>;
}
```

Then in your ethereum.rs you can do:

```
impl EntropyReader for PythContract {
    type Provider = Address;
```

To associate the Ethereum type for that implementation.

Reisen (changes requested):

Reisen (approved):

PR #1142, opened 2023-11-21
feat(contract-manager): add script to fetch account balances

ali-bahjati:


target_chains/cosmwasm/tools/src/chains-manager/cosmwasm.ts
```diff
@@ -63,7 +64,20 @@ export class CosmwasmExecutor implements ChainExecutor {
...
+    const denom = GasPrice.fromString(this.gasPrice).denom;
+    const balance = await cosmwasmClient.getBalance(address, denom);
+
+    // By default the coins have 6 decimal places in CosmWasm
+    // and the denom is usually `u<chain>`.
+    return Number(balance.amount) / 10 ** 6;
+  }
+
+  async getAddress(): Promise<string> {
     return (await this.signer.getAccounts())[0].address;
```
m30m, line 81: such a shame that this is async 😢 and made the whole function async in the end.

contract_manager/scripts/fetch_account_balance.ts
```diff
@@ -0,0 +1,58 @@
...
+  address: string | undefined;
+  balance: number | undefined;
+};
+
+async function getBalance(
+  chain: string,
+  privateKey: PrivateKey
+): Promise<AccountBalance | undefined> {
+  try {
+    const address = await DefaultStore.chains[chain].getAccountAddress(
```
m30m, line 30: I would put this out of the try/catch and remove the `undefined` possibility from the `address` field in `AccountBalance`. It's better to crash if we can't get the address based on private key 😅 and I don't think the cosmwasm library uses any network to do this, it's just async 🤷

m30m (approved):
looks great! added a minor comment.

contract_manager/scripts/fetch_account_balance.ts
ali-bahjati: Did it.

PR #1144, opened 2023-11-23
[xc-admin] Ship 64 pubs

guibescos:
Enable 64 publishers for pythtest networks

### DIFF
--- a/governance/xc_admin/packages/xc_admin_common/src/cluster.ts
+++ b/governance/xc_admin/packages/xc_admin_common/src/cluster.ts
@@ -27,3 +27,22 @@ export function getMultisigCluster(cluster: PythCluster): Cluster | "localnet" {
       return cluster;
   }
 }
+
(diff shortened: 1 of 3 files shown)


governance/xc_admin/packages/xc_admin_common/src/cluster.ts
```diff
@@ -27,3 +27,22 @@ export function getMultisigCluster(cluster: PythCluster): Cluster | "localnet" {
...
+export function getMaximumNumberOfPublishers(cluster: PythCluster) {
+  switch (cluster) {
+    case "mainnet-beta":
+      return 32;
+    case "devnet":
+      return 32;
+    case "testnet":
+      return 32;
+    case "pythnet":
+      return 32;
```
drozdziak1, line 40: Note: I think this is good, we should prevent a pythnet issue if there's a regression

drozdziak1 (approved):

PR #1145, opened 2023-11-24
Solana receiver decode vaa accumulator

swimricky:


target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    #[allow(unused_variables)]
+    pub fn post_updates(
+        ctx: Context<PostUpdates>,
+        vaa_hash: [u8; 32], // used for pda seeds
+        price_updates: Vec<Vec<u8>>,
+    ) -> Result<()> {
+        let vaa = &ctx.accounts.posted_vaa; // let posted_vaa_data = PostedVaaData::try_deserialize_unchecked(&mut &**vaa.try_borrow_data()?)?;
+        let wh_message = WormholeMessage::try_from_bytes(vaa.payload.as_slice())
+            .map_err(|_| ReceiverError::InvalidWormholeMessage)?;
+        msg!("constructed wh_message {:?}", wh_message);
```
jayantk, line 184: the checks for the message emitter need to be added here (and removed from the instruction constraints)

.pre-commit-config.yaml
```diff
@@ -99,15 +99,15 @@ repos:
...
       # Hooks for solana receiver contract
-      - id: cargo-fmt-solana-receiver
+      - id: cargo-fmt-pyth_solana_receiver
         name: Cargo format for solana target chain contract
         language: "rust"
         entry: cargo +nightly-2023-03-01 fmt --manifest-path ./target_chains/solana/Cargo.toml --all -- --config-path rustfmt.toml
         pass_filenames: false
         files: target_chains/solana
-      - id: cargo-clippy-solana-receiver
+      - id: cargo-clippy-pyth_solana_receiver
```
jayantk, line 108: please follow the naming convention in this file to use hyphens here not underscores

target_chains/solana/programs/pyth_solana_receiver/tests/program_test/mod.rs
```diff
@@ -40,72 +40,24 @@ pub struct ProgramSimulator {
...
-            lamports:   Rent::default().minimum_balance(programdata_vec.len()),
-            data:       programdata_vec,
-            owner:      bpf_loader_upgradeable::ID,
-            executable: false,
-            rent_epoch: Epoch::default(),
-        };
-
-        // Add to both accounts to program test, now the program is deploy as upgradable
-        program_test.add_account(program_key, program_account);
-        program_test.add_account(programdata_key, programdata_account);
```
jayantk, line 98: (FYI, we've been doing it this way ^ because it lets us test the program upgrade flow in the simulator. Not sure if the replacement code enables that or not)

tbh we should really pull this out to a rust library somewhere and import it because we've copy/pasted this code like 10 times now.

jayantk (changes requested):
Thanks. I think there are some issues here though -- see inline comments

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        Ok(())
+    }
+
+    /// Verifies the accumulator update data header then invokes a CPI call to wormhole::postVAA
+    ///
+    /// * `data` - Bytes of the AccumulatorUpdateData response from hermes with the updates omitted
+    ///           (i.e. the `updates` field is an empty array). The updates are removed so that
+    ///           all the data needed for postVaa can fit in one txn.
+    pub fn post_accumulator_update_vaa(
```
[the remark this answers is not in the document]
ali-bahjati: I think adding some extra types and structs would help to separate the boundary between the pythnet wireformat and the solana receiver wireformat.

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        Ok(())
+    }
+
+    #[allow(unused_variables)]
+    pub fn post_updates(
+        ctx: Context<PostUpdates>,
+        vaa_hash: [u8; 32], // used for pda seeds
+        price_updates: Vec<Vec<u8>>,
+    ) -> Result<()> {
+        let vaa = &ctx.accounts.posted_vaa; // let posted_vaa_data = PostedVaaData::try_deserialize_unchecked(&mut &**vaa.try_borrow_data()?)?;
```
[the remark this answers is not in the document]
swimricky: forgot to clean up from testing thanks

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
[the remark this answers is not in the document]
swimricky: the reason this is a wrapper instead of having the client do `postVaa` directly was so that we could do on-chain verification of the header in the `accumulatorUpdateData` using the `AccumulatorUpdateData::try_from_slice`

the client/cli code is currently doing the exact same thing when it first parses the accumulatorUpdateData bytes before it invokes this ix but i thought it would be good to have this verification on-chain as well

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub nonce:             u32,
+    pub emitter_chain:     u16,
+    pub emitter_address:   [u8; 32],
+    pub sequence:          u64,
+    pub consistency_level: u8,
+    pub payload:           Vec<u8>,
+}
+
+#[derive(Default, AnchorSerialize, AnchorDeserialize, Clone, PartialEq, Eq)]
+pub struct GuardianSet {
```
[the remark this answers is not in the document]
swimricky: in the cli, i'm currently using a mix of the wormhole repo, the wormhole repo fork with the solana-sdk and the wormhole-anchor-sdk. 
the solana-sdk has this but it's using an outdated version of borsh/solana which was why i re-implemented this here for now.

 ideally i would like to make a PR to the wormhole-anchor-sdk  to add this and some of the other functionality from the sdk-solana repo and remove the solana-sdk fork dependency completely

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
swimricky: can you elaborate on why it needs to be here vs. as an instruction constraint?

target_chains/solana/programs/pyth_solana_receiver/tests/program_test/mod.rs
swimricky: i see that makes sense. I did a quick look into this and one alternative to being able to test program upgrade flows would be to use a ts test with solana-local-validator and submit commands like this https://github.com/Squads-Protocol/squads-mpl/blob/main/tests/squads-mpl.ts

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
+pub mod error;
```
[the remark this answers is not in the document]
swimricky: sorry forgot about this again will keep this in mind for future changes

target_chains/solana/programs/pyth_solana_receiver/src/state.rs
```diff
@@ -1,25 +1,32 @@
...
 impl AccountDeserialize for AnchorVaa {
     // Manual implementation because this account does not have an anchor discriminator
     fn try_deserialize(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
-        Self::try_deserialize_unchecked(buf)
+        require!(buf.len() >= 3, ErrorCode::AccountDiscriminatorNotFound);
+        let given_disc = &buf[..3];
+        require!(
+            *given_disc == *b"vaa",
+            ReceiverError::PostedVaaHeaderWrongMagicNumber
+        );
```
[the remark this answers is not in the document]
swimricky: anchor will automatically call this method for any context accounts that have this type and the `PostedVaaData` type from the womrhole-anchor-sdk doesn't have this check (which i think is a bug). This is why i wrapped the `PostedVaaData` inside an `AnchorVaa`. I will make a PR to the wormhole-anchor-sdk for this fix then we can get rid of the `AnchorVaa` completely

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
swimricky: i added the message emitter checks here and in the postVaa wrapper. although there's no downside to having it in both, i think it makes more sense being checked in the postAccumulatorUpdateVaa ix than here.

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+#[derive(Accounts)]
+pub struct DecodePostedVaa<'info> {
+    #[account(mut)]
+    pub payer:      Signer<'info>,
+    #[account(
+        constraint = (Chain::from(posted_vaa.emitter_chain()) == Solana || Chain::from(posted_vaa.emitter_chain()) == Pythnet) @ ReceiverError::EmitterChainNotSolanaOrPythnet,
+    )]
```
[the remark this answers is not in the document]
swimricky: I changed this to  passing in the expected emitter_chain as an ix paramter for now but ideally it should be checked using some type of `config` pda that's can only be written to by governance.

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub payer:      Signer<'info>,
+    #[account(
+        constraint = Chain::from(posted_vaa.emitter_chain()) == Pythnet @ ReceiverError::EmitterChainNotSolanaOrPythnet,
+        seeds = [
+            SEED_PREFIX_POSTED_VAA,
+            &vaa_hash
+        ],
+        seeds::program = wormhole_anchor::program::Wormhole::id(),
+        bump
+    )]
```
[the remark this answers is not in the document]
swimricky: similar to above but im passing in the program as an account now. this should also probably be ready/checked against a `config` pda as well

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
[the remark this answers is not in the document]
jayantk: can you move this struct to the CLI where it's actually used then?

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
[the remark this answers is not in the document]
jayantk: I think the header verification check isn't adding any value here. Someone could circumvent this check by calling post vaa themselves, and then calling the subsequent update method.

All of our security checks need to happen in the `update` method themselves. In this particular case, I don't think the magic number / version checks are security critical so they could potentially be skipped for solana (given that there's a sort of weird flow needed to do the vaa posting).

target_chains/solana/programs/pyth_solana_receiver/tests/program_test/mod.rs
jayantk: Please don't use typescript for the tests.  It's way more annoying and difficult to write the tests that way because you can't rely as directly on the rust types and stuff.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        vaa_hash: [u8; 32], // used for pda seeds
+        emitter_chain: u16,
+        price_updates: Vec<Vec<u8>>,
+    ) -> Result<()> {
+        let vaa = &ctx.accounts.posted_vaa;
+        require_eq!(
+            vaa.emitter_chain(),
+            emitter_chain,
+            ReceiverError::InvalidEmitterChain
+        );
```
jayantk, line 131: I think you also need to check that the emitter address is the pyth program on the provided chain.

I suggest referencing the EVM implementation and making sure that every security check there is included here.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// TODO:
+    ///    - use a `config` account that can only be modified by governance for checking emitter_chain
+    ///      and other constraints
+    ///
+
+    #[allow(unused_variables)]
+    pub fn post_updates(
+        ctx: Context<PostUpdates>,
+        vaa_hash: [u8; 32], // used for pda seeds
+        emitter_chain: u16,
```
jayantk, line 123: don't put a security critical parameter as an argument. this is just asking for us to forget about this TODO and leave a security vuln in the codebase. I would prefer you hardcoding a constant with the TODO above than this.

In general, I think we should strive to ensure that *every* PR leaves the code in a state that is OK from a security perspective. It may not be as configurable or flexible as we would like, or the functionality may not be there, but at least it's secure. That approach ensures that our own human fallibility doesn't cause security holes in our code.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
jayantk: (this comment is tied to the theme above -- every PR should leave the code in a state that is OK from a security perspective)

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
jayantk: I believe someone can post a vaa directly (by calling PostVAA themselves), then pass that VAA account to this function. If they do that, any security checks in this function are bypassed. (This is also why i think this function is kind of useless)

These checks should not be an instruction constraint because the data source emitter (chain id and address) is configurable via a governance message on all other chains. Using the instruction constraint is much less flexible than using a configuration parameter.

Also you need to check not only that the chain is correct, but also that the emitter address is correct.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    ///
+    /// TODO:
+    ///    - use a `config` account that can only be modified by governance for checking emitter_chain
+    ///      and other constraints
+    ///
+
+    #[allow(unused_variables)]
+    pub fn post_updates(
+        ctx: Context<PostUpdates>,
+        vaa_hash: [u8; 32], // used for pda seeds
```
guibescos, line 122: I don't think you need vaa_hash as a parameters here. We are trusting the posted VAA based on it being owned by wormhole + the discriminator

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        seeds = [
+            SEED_PREFIX_POSTED_VAA,
+            &vaa_hash
+        ],
+        seeds::program = post_vaa_program.key(),
+        bump
+    )]
+    pub posted_vaa:       Box<Account<'info, AnchorVaa>>,
+    /// CHECK: program that called post_vaa
+    pub post_vaa_program: AccountInfo<'info>,
```
guibescos, line 185: I think you don't need this account

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        for price_update in price_updates {
+            let merkle_price_update =
+                from_slice::<byteorder::BE, MerklePriceUpdate>(price_update.as_slice())
+                    .map_err(|_| ReceiverError::DeserializeUpdateFailed)?;
+            let message_vec = Vec::from(merkle_price_update.message);
+            if !root.check(merkle_price_update.proof, &message_vec) {
+                return err!(ReceiverError::InvalidPriceUpdate);
+            }
+            let msg = from_slice::<byteorder::BE, Message>(&message_vec)
+                .map_err(|_| ReceiverError::InvalidAccumulatorMessage)?;
```
guibescos, line 151: shouldn't this be a `DeserializedMessageFailed`?

guibescos (commented):
For governance I recommend you just create a config account that has the config (emitter_address, emitter_chain) and an authority that can update this fields. 
We can use the remote executor as the authority later.

swimricky:
> For governance I recommend you just create a config account that has the config (emitter_address, emitter_chain) and an authority that can update this fields. We can use the remote executor as the authority later.

thanks for this, that was the eventual plan but i'm going to add the config/governance in a separate PR since this one already got a little bit out of hand size wise

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: i don't mind removing this but technically any vaa that has been posted by the wormhole program would be owned by wormhole and have the right discriminator. The merkle root prove would fail however in that situation if the wrong vaa was used

target_chains/solana/programs/pyth_solana_receiver/src/lib.rs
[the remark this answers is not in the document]
swimricky: changing this again. Now it calls `Pubkey::find_program_address([seeds], wormhole_anchor::program::Wormhole::id())` and checks against the posted_vaa account's key in the ix but leaving a TODO that the program/owner of the posted_vaa will eventually be read from a config instead of hardcoded. I'm assuming this is in case we want to eventually write our own wormhole wrapper or need to update this

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: i removed the `vaa_hash` from the ix parameters but added the signature_set account into the `PostUpdates` context since it also contains the hash & we can verify that it's the correct signature_set since the `vaa` has a field for the signature set pubkey.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: i removed this and changed the key derivation to be in the ix instead. right now im hard-coding that the owner of the posted_vaa program is the wormhole program but based off this feedback from @jayantk https://github.com/pyth-network/pyth-crosschain/pull/1145#discussion_r1404437787
i think we want it to be configurable so i will include it in the PR that adds a config account that only the authority can write to

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: updated this thanks

target_chains/solana/cli/src/main.rs
```diff
@@ -37,90 +59,208 @@ use {
...
+                    println!("update_bytes_len: {}", update_bytes_len);
+
+                    let post_updates_accounts = pyth_solana_receiver::accounts::PostUpdates {
+                        payer:         payer.pubkey(),
+                        posted_vaa:    vaa_pubkey,
+                        signature_set: *posted_vaa_data.vaa.signature_set(),
+                    }
+                    .to_account_metas(None);
+                    let post_updates_ix_data = pyth_solana_receiver::instruction::PostUpdates {
+                        price_updates: update_bytes,
```
jayantk, line 252: we should probably have a struct type for the price_updates argument that works with anchor's built in serialization/deserialization so it's easier for users to construct and work with.

target_chains/solana/programs/pyth-solana-receiver/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+#[error_code]
+pub enum ReceiverError {
+    #[msg("The emitter chain of the VAA is invalid.")]
+    InvalidEmitterChain,
+    #[msg("The emitter address of the VAA is invalid.")]
+    InvalidEmitterAddress,
+    #[msg("The posted VAA has wrong magic number.")]
+    PostedVaaHeaderWrongMagicNumber,
+    #[msg("An error occured when deserializing the VAA.")]
+    DeserializeVAAFailed,
```
jayantk, line 12: inconsistent capitalization of VAA here. It's Vaa in the other errors.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+declare_id!("DvPfMBZJJwKgJsv2WJA8bFwUMn8nFd5Xpioc6foC3rse");
+pub const POST_VAA: u8 = 2;
+
+#[program]
+pub mod pyth_solana_receiver {
+    use super::*;
+
+    /// Verify the updates using the posted_vaa account
+    ///  * `vaa_hash` hash post of the post_vaa data to derive the address of the post_vaa account
```
jayantk, line 39: comment is stale

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        let root: MerkleRoot<Keccak160> = MerkleRoot::new(match wh_message.payload {
+            WormholePayload::Merkle(merkle_root) => merkle_root.root,
+        });
+
+        let mut count_updates = 0;
+
+        let price_updates_len = price_updates.len();
+        for price_update in price_updates {
+            let merkle_price_update =
+                from_slice::<byteorder::BE, MerklePriceUpdate>(price_update.as_slice())
```
jayantk, line 88: wait we already have a struct for these updates. Why not pass `Vec<MerklePriceUpdate>` as the argument to this instruction then and then implement whatever trait anchor needs to get the serialization right?

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        let (expected_vaa_pubkey, _) = Pubkey::find_program_address(
+            &[SEED_PREFIX_POSTED_VAA, &signature_set.hash],
+            //TODO: expected program owner of the posted_vaa account should come from config account that can only be modified by governance
+            &wormhole_anchor_sdk::wormhole::program::id(),
+        );
+        require_keys_eq!(
+            vaa.key(),
+            expected_vaa_pubkey,
+            ReceiverError::InvalidVaaAccountKey
+        );
```
jayantk, line 60: What are the security requirements for verifying that the posted VAA is correctly signed? I don't understand why we need the signature_set here. How do other programs do this?

jayantk: this wormhole example seems like it's using the vaa_hash https://github.com/wormhole-foundation/wormhole-scaffolding/blob/main/solana/programs/01_hello_world/src/context.rs#L243

not sure if that's the current best practice or not. hard to tell from their docs.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+                _ => return err!(ReceiverError::InvalidAccumulatorMessageType),
+            }
+        }
+        msg!("verified {} / {} updates", count_updates, price_updates_len);
+        Ok(())
+    }
+}
+
+#[derive(Accounts)]
+#[instruction(vaa_hash: [u8; 32])]
```
jayantk, line 115: I don't think vaa_hash is used anywhere anymore?

target_chains/solana/programs/pyth-solana-receiver/src/state.rs
```diff
@@ -1,25 +1,30 @@
...
     // Manual implementation because this account does not have an anchor discriminator
     fn try_deserialize(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
-        Self::try_deserialize_unchecked(buf)
+        require!(buf.len() >= 3, ErrorCode::AccountDiscriminatorNotFound);
+        let given_disc = &buf[..3];
+        require!(
+            *given_disc == *b"vaa",
+            ReceiverError::PostedVaaHeaderWrongMagicNumber
+        );
+        Self::try_deserialize_unchecked(&mut &buf[3..])
```
jayantk, line 27: This code doesn't make sense. The unchecked method should get the same buffer as the checked method -- it's just supposed to ignore the discriminator.

target_chains/solana/programs/pyth-solana-receiver/tests/program_test/mod.rs
```diff
@@ -40,72 +40,24 @@ pub struct ProgramSimulator {
...
-        );
-
-
-        let mut program_test = ProgramTest::default();
-        let program_key = Pubkey::try_from(ID).unwrap();
-        // This PDA is the actual address in the real world
-        // https://docs.rs/solana-program/1.6.4/solana_program/bpf_loader_upgradeable/index.html
-        let (programdata_key, _) =
-            Pubkey::find_program_address(&[&program_key.to_bytes()], &bpf_loader_upgradeable::id());
+        let mut program_test = ProgramTest::new("pyth_solana_receiver", ID, None);
```
jayantk, line 50: i really think you shouldn't change this deployment code. We're doing it this way for a reason (to test upgrades) and you're just going to have to rewrite it again.

target_chains/solana/cli/src/main.rs
```diff
@@ -37,90 +59,208 @@ use {
...
         Account,
         Config as WormholeConfig,
-        GuardianSet,
+        GuardianSet as WormholeSolanaGuardianSet,
         VAA as WormholeSolanaVAA,
     },
 };
 
+#[derive(Default, AnchorSerialize, AnchorDeserialize, Clone, PartialEq, Eq)]
+pub struct GuardianSet {
```
jayantk, line 68: can you please document why this struct is necessary? You said it in the github comment, but that should go in the code for posterity.

target_chains/solana/cli/src/main.rs
```diff
@@ -37,90 +59,208 @@ use {
...
+            let wormhole = wormhole_anchor_sdk::wormhole::program::id();
+            let rpc_client = RpcClient::new(url);
+
+            println!("[1/5] Decode the AccumulatorUpdateData");
+            let accumulator_update_data_bytes: Vec<u8> =
+                base64::decode(accumulator_update_data_str)?;
+            let accumulator_update_data =
+                AccumulatorUpdateData::try_from_slice(accumulator_update_data_bytes.as_slice())?;
+
+            match &accumulator_update_data.proof {
```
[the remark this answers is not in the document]
jayantk: (you can do this in a separate PR later on though)

target_chains/solana/cli/src/main.rs
```diff
@@ -37,90 +59,208 @@ use {
...
+
+                    let mut verify_count = 0;
+                    for update in updates {
+                        let message_vec = Vec::from(update.message.clone());
+                        if !root.check(update.proof.clone(), &message_vec) {
+                            println!("[ERR] failed to verify update");
+                        } else {
+                            verify_count += 1;
+                        }
+                    }
```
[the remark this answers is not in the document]
swimricky: sorry this was leftover when i was debugging/testing that the merkle root and updates were working as expected i will clean this up

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: `MerklePriceUpdate` comes from an external crate and the `Keccak160` for the proof doesn't implement `BorshSerialize` or `BorshDeserialize`.

 I will look into making a separate PR and add a todo for this so that we can pass in `Vec<MerklePriceUpdate>` directly like you suggested since this PR is already starting to get a bit large.

target_chains/solana/programs/pyth-solana-receiver/tests/program_test/mod.rs
swimricky: ok will look into reverting this to original state

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+pub struct PostUpdates<'info> {
+    #[account(mut)]
+    pub payer:         Signer<'info>,
+    pub posted_vaa:    Box<Account<'info, AnchorVaa>>,
+    pub signature_set: Box<Account<'info, SignatureSetData>>,
```
swimricky, line 120: @jayantk @guibescos since this is using the `wormhole-anchor-sdk:SignatureSetData` account type, the owner is set to be the wormhole pid. since the `posted_vaa` pid/owner will eventually be checked using a config account, do we also want the signature set account owner to be configurable?

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: I still don't get passing the signature set signature set. 
A posted VAA is a VAA whose signatures have already been checked. 
People can pass whatever valid VAA to this program, if the proof doesn't match the VAA it will just fail.

target_chains/solana/programs/pyth-solana-receiver/tests/program_test/mod.rs
swimricky: i reset it back to the original state but just a note that i tried adding back the original update ix and using that test case and it failed with a strange `InvalidAccountData`. I tried debugging the source of this but wasn't able to figure it out. i'll have to look into it later. The test runs fine when i use `ProgramTest::new("pyth_solana_receiver", ID, None);`

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
swimricky: the signature set is just a different way to pass the `vaa_hash` since it also contains the hash of the data. this is just an extra check b/c like you said, if the proof doesn't match the vaa then it'll fail.

target_chains/solana/cli/src/main.rs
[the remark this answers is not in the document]
swimricky: added todo for this

target_chains/solana/cli/src/main.rs
swimricky: added to do for this

target_chains/solana/programs/pyth-solana-receiver/tests/program_test/mod.rs
swimricky: also is there a repo that tests the program upgrade flow? https://github.com/pyth-network/pyth-client/blob/main/program/rust/src/tests/pyth_simulator.rs also has this setup but it doesn't look like the actual program upgrade is tested there, the upgrade authority is checked for the `upd_permissions` test

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: I don't understand why you need this extra check, let's just keep it simple

target_chains/solana/cli/src/main.rs
```diff
@@ -37,90 +53,190 @@ use {
...
+                            )?;
+
+                            rpc_client.get_account_data(&vaa_pubkey).ok()
+                        }
+                    };
+
+                    println!("[5/5] Post updates from AccumulatorUpdateData and use the PostedVAA on solana using pyth-solana-receiver::PostUpdates");
+                    // TODO need to figure out max number of updates that can be sent in 1 txn
+
+                    let _posted_vaa_data =
```
guibescos, line 204: Unused

target_chains/solana/programs/pyth-solana-receiver/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[msg("Received an invalid wormhole message")]
+    InvalidWormholeMessage,
+    #[msg("Received an invalid price update")]
+    InvalidPriceUpdate,
+    #[msg("Received an invalid accumulator message type")]
+    InvalidAccumulatorMessageType,
+    #[msg("Received an invalid signature set")]
+    InvalidSignatureSet,
+    #[msg("The pubkey of the the posted vaa account is invalid")]
+    InvalidVaaAccountKey,
```
guibescos, line 26: Many errors here are no longer used

jayantk (approved):

target_chains/solana/programs/pyth-solana-receiver/tests/program_test/mod.rs
```diff
@@ -28,9 +27,47 @@ use {
...
+        "cargo-build-bpf did not exit with 0 (code {:?})",
+        status.code()
+        );
+    }
+
+    let target_dir = concat!(env!("CARGO_MANIFEST_DIR"), "/../../target");
+
+    PathBuf::from(target_dir).join("deploy/pyth_solana_receiver.so")
+    };
+}
```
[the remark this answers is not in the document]
jayantk: i guess this can wait for a later pr though. this pr is pretty big already

PR #1148, opened 2023-11-24
[entropy] First cut at EVM executor

jayantk:
Initial version of an EVM contract that can execute arbitrary calls encoded in wormhole messages. The basic idea here is to add a new governance module for this type of contract, then give it a payload that encodes a contract address and evm calldata. 

This contract definitely needs more work, but i thought i'd get something out quickly for feedback. Things that it's missing are:
1. more testing
2. some thought about how to manage upgrades
3. how to handle multiple authorities on the same chain. If we have addresses with different levels of priviledges, do we deploy multiple executor contracts? Or is there a way to have them all in the same contract?


target_chains/ethereum/contracts/contracts/executor/Executor.sol
```diff
@@ -0,0 +1,161 @@
...
+        bool success;
+        (success, response) = address(gi.callAddress).call(gi.callData);
+
+        // Check if the call was successful or not.
+        if (!success) {
+            // If there is return data, the delegate call reverted with a reason or a custom error, which we bubble up.
+            if (response.length > 0) {
+                assembly {
+                    let returndata_size := mload(response)
+                    revert(add(32, response), returndata_size)
```
jayantk, line 87: I copied this logic from another proxy contract https://github.com/PaulRBerg/prb-proxy

ali-bahjati: Hmm in my mind it should be add(20, response). Let me see the other one.

ali-bahjati: Oh no my bad, 32 is a word size.

ali-bahjati: Maybe add a documentation to to explain that the first word of response is the length, so when we call revert we add response pointer by a word (32 bytes) to give the pointer to the beginning of the revert data and pass the size as the second argument.

target_chains/ethereum/contracts/contracts/executor/Executor.sol
```diff
@@ -0,0 +1,161 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.12;
+
+import "../pyth/PythGovernanceInstructions.sol";
```
ali-bahjati, line 4: I recommend removing this dependency so in the future we can use this alone on Pyth itself.

target_chains/ethereum/contracts/forge-test/Executor.t.sol
```diff
@@ -0,0 +1,79 @@
...
+            OWNER_EMITTER,
+            sequence,
+            payload,
+            NUM_SIGNERS
+        );
+
+        executor.execute(vaa);
+    }
+
+    function testBasic() public {
```
ali-bahjati, line 60: I recommend having a test contract that checks the owner and test with that to make sure that the owner is that contract.

ali-bahjati (commented):
It's very nice! I left some inline comments. I think we need to implement another instruction to update the governance data source itself and we are done after.

PR #1149, opened 2023-11-25
[entropy] Change provider interface to simplify supporting many chains

jayantk:
Currently, the Entropy contract has a single `bytes32` field that providers can use to store arbitrary metadata associated with their hash chain. We're using this to store a random seed that the provider can use to regenerate the chain. However, the chain on the server is a function of both the random seed *and* a chain length parameter (and also a secret which of course isn't on the blockchain). Thus, at the moment, every blockchain must use the exact same chain length. This is brittle and will cause operational problems down the line.

This PR changes the field to be `bytes` and the providers can choose to store however much data they want in there.

### DIFF
--- a/fortuna/src/abi.json
+++ b/fortuna/src/abi.json
@@ -1,46 +1,5 @@
 [
   {
-    "inputs": [
-      {
-        "internalType": "uint256",
-        "name": "pythFeeInWei",
-        "type": "uint256"
-      }
-    ],
-    "stateMutability": "nonpayable",
-    "type": "constructor"
-  },
-  {
-    "inputs": [],
-    "name": "AssertionFailure",
-    "type": "error"
-  },
-  {
-    "inputs": [],
-    "name": "IncorrectProviderRevelation",
-    "type": "error"
-  },
-  {
-    "inputs": [],
-    "name": "IncorrectUserRevelation",
-    "type": "error"
-  },
-  {
-    "inputs": [],
-    "name": "InsufficientFee",
-    "type": "error"
-  },
-  {
-    "inputs": [],
-    "name": "NoSuchProvider",
-    "type": "error"
-  },
-  {
-    "inputs": [],
-    "name": "OutOfRandomness",
-    "type": "error"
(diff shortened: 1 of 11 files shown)


target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -186,7 +186,6 @@ contract Entropy is IEntropy, EntropyState {
         req.providerCommitment = providerInfo.currentCommitment;
         req.providerCommitmentSequenceNumber = providerInfo
             .currentCommitmentSequenceNumber;
-        req.providerCommitmentMetadata = providerInfo.commitmentMetadata;
```
jayantk, line 189: deleted this field from req (it's not needed)

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -44,19 +44,14 @@ contract EntropyTest is Test {
         random.register(
             provider1FeeInWei,
             provider1Proofs[0],
-            bytes32(keccak256(abi.encodePacked(uint256(0x0100)))),
+            hex"0100",
```
jayantk, line 47: drive-by: shorter way to write this

fortuna/src/state.rs
```diff
@@ -36,17 +33,18 @@ impl PebbleHashChain {
     }
 
     pub fn from_config(
-        opts: &RandomnessOptions,
+        secret: &str,
         chain_id: &ChainId,
-        random: [u8; 32],
+        random: &[u8; 32],
```
ali-bahjati, line 38: I think seed is a better name here.

fortuna/src/abi.json
```diff
@@ -1,45 +1,4 @@
 [
-  {
```
ali-bahjati, line 2: Why these got deleted? :?

ali-bahjati (approved):
Looks good! I left some minor inline comments.

PR #1150, opened 2023-11-25
[entropy] Add a default provider 

jayantk:
The current interface forces users to pick a provider. This is awkward when there's really only one provider initially. This PR lets governance choose a default and adds some convenience methods for invoking the request/reveal flow with the default. This should simplify things for most users.


target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -64,7 +65,9 @@ contract EntropyTest is Test {
         uint64 startSequenceNumber,
         uint64 size
     ) public pure returns (bytes32[] memory hashChain) {
-        bytes32 initialValue = keccak256(abi.encodePacked(startSequenceNumber));
+        bytes32 initialValue = keccak256(
+            abi.encodePacked(provider, startSequenceNumber)
```
jayantk, line 69: this was a bug in previous versions where the hash chains for both providers were exactly the same. turns out i wasn't using provider2's hash chain in any of the other tests though.

ali-bahjati (commented):
While in solidity it is allowed to have multiple functions with same name and different signatures I have found that the explorers (etherscan) and client libraries (ethers, web3) are not good at handling them. For example they struggle to understand which request does `contract.methods.request` refers to and you need to use something like `contract.methods["request(bytes32,bool)"].call...`. So I recommend using different function names for the default ones.

jayantk:
Update to this PR here. I decided it doesn't make sense to create other versions of request/reveal/getFee if they can't share the same name. Instead, people can just call `request(getDefaultProvider(), ...)` to get the default.

I also added a `uri` parameter for each provider. The idea here is to let the blockchain serve as the registry that maps providers -> where you can retrieve their random numbers. I have not yet specified the expected API that clients should expect at this URI, but we can do that later when we develop client libraries.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -68,6 +80,7 @@ contract EntropyTest is Test {
     }
 
     // Test helper method for requesting a random value as user from provider.
+    // If provider is all-zeros, use the default provider in the contract.
```
ali-bahjati, line 83: Are you actually using it? I'm not seeing this be implemented too :?

ali-bahjati (approved):
Nice!

PR #1153, opened 2023-11-27
[entropy] Entropy gas benchmarks and optimizations

jayantk:
Working on gas optimizations for entropy. This PR got a little out of hand but I'm glad I did it before the audit because the gas optimizations turned out to be nontrivial.

This PR adds a very basic gas benchmark for entropy, then does some code optimizations to bring down the gas usage. There are three main optimizations. Two of these are easy and one is less easy.
* (easy) use uint128 for fees instead of uint256
* (easy) Reduce the size of the Request struct. Store `hash(providerCommitment, userCommitment)` instead of both values, and store the number of hashes instead of the provider's sequence number.
* (less easy) store requests in a 2 level hash table (an array and a mapping) instead of just a mapping. There's an extensive code comment explaining this design, but basically it allows you to reuse nonzero memory for multiple requests, which lets you pay 5k gas to write instead of 21k gas each time. 

The upshot of all of these optimizations is something like this:

Code before any of these changes
request: 157k
reveal: 11k
total: 168k 

After the 2 easy optimizations
request: 65k
reveal: 10k
total: 75k

With all three optimizations
request: 37k
reveal: 11k
total: 48k (72% savings over original, 36% over the two easy versions only)

One issue with the third optimization is that it creates an unlucky case where a new request needs to be stored in the same slot as another concurrent request. This case comes in at 104k + 11k = 115k if it occurs. However, this case should be rare. If there are N active requests, you have an N/32 chance of the next concurrent request (i.e., a new request before any of the existing ones complete) hitting this case. Furthermore, we can always increase 32 to a larger number.

Reviewer guide: start by looking at EntropyStructs and EntropyState, then read the new functions at the end of Entropy.sol. The rest of the changes are pretty straightforward from there.

### DIFF
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -6,6 +6,7 @@ import "@pythnetwork/entropy-sdk-solidity/EntropyStructs.sol";
 import "@pythnetwork/entropy-sdk-solidity/EntropyErrors.sol";
 import "@pythnetwork/entropy-sdk-solidity/EntropyEvents.sol";
 import "@pythnetwork/entropy-sdk-solidity/IEntropy.sol";
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import "./EntropyState.sol";
 
 // Entropy implements a secure 2-party random number generation procedure. The protocol
@@ -186,7 +187,7 @@ contract Entropy is IEntropy, EntropyState {
         uint128 requiredFee = getFee(provider);
         if (msg.value < requiredFee) revert EntropyErrors.InsufficientFee();
         providerInfo.accruedFeesInWei += providerInfo.feeInWei;
-        _state.accruedPythFeesInWei += (uint128(msg.value) -
+        _state.accruedPythFeesInWei += (SafeCast.toUint128(msg.value) -
             providerInfo.feeInWei);
 
         // Store the user's commitment so that we can fulfill the request later.
@@ -198,7 +199,7 @@ contract Entropy is IEntropy, EntropyState {
         );
         req.provider = provider;
         req.sequenceNumber = assignedSequenceNumber;
-        req.numHashes = uint32(
+        req.numHashes = SafeCast.toUint32(
             assignedSequenceNumber -
                 providerInfo.currentCommitmentSequenceNumber
         );
@@ -207,7 +208,7 @@ contract Entropy is IEntropy, EntropyState {
         );
 
         if (useBlockHash) {
-            req.blockNumber = uint96(block.number);
+            req.blockNumber = SafeCast.toUint96(block.number);
         } else {
             req.blockNumber = 0;
         }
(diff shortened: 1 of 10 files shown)


target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -54,19 +53,6 @@ contract EntropyTest is Test {
...
-        uint64 size
-    ) public pure returns (bytes32[] memory hashChain) {
-        bytes32 initialValue = keccak256(abi.encodePacked(startSequenceNumber));
-        hashChain = new bytes32[](size);
-        for (uint64 i = 0; i < size; i++) {
-            hashChain[size - (i + 1)] = initialValue;
-            initialValue = keccak256(bytes.concat(initialValue));
-        }
-    }
-
```
jayantk, line 69: extracted this to EntropyTestUtils

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -305,31 +330,85 @@ contract Entropy is IEntropy, EntropyState {
...
+        (, uint8 shortKey) = requestKey(provider, sequenceNumber);
+
+        req = _state.requests[shortKey];
+        if (isActive(req)) {
+            // There's already a prior active request in the storage slot we want to use.
+            // Overflow the prior request to the requestsOverflow mapping.
+            // This operation is expensive, but should be rare. If overflow happens frequently, increase
+            // the size of the requests array to support more concurrent active requests.
+            (bytes32 reqKey, ) = requestKey(req.provider, req.sequenceNumber);
+            _state.requestsOverflow[reqKey] = req;
```
ali-bahjati, line 402: Why are you moving the previous request and not the new request?

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -305,31 +330,85 @@ contract Entropy is IEntropy, EntropyState {
...
+            req.sequenceNumber = 0;
+        } else {
+            delete _state.requestsOverflow[key];
+        }
+    }
+
+    // Allocate storage space for a new in-flight request. This method returns a pointer to a storage slot
+    // that the caller should overwrite with the new request. Note that the memory at this storage slot may
+    // -- and will -- be filled with arbitrary values, so the caller *must* overwrite every field of the returned
+    // struct.
```
ali-bahjati, line 388: Why you don't zero it out before returning? In the end it shouldn't add much cost. (The storage cost is calculated in the end I believe)

ali-bahjati (approved):
Very nice! I left some inline comments, mostly questions.

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
jayantk: ah this is intentional. If you do it the other way, if there are requests that never get revealed, then those requests will gradually fill up the requests array, and then all new requests will end up in the mapping and pay the higher gas cost. (I left a code comment to this effect also)

ali-bahjati: smart!

PR #1154, opened 2023-11-28
[entropy] governance

0xfirefist:
- [ ] add tests
- [ ] add upgradability

### DIFF
--- a/target_chains/ethereum/entropy_sdk/solidity/EntropyErrors.sol
+++ b/target_chains/ethereum/entropy_sdk/solidity/EntropyErrors.sol
@@ -19,4 +19,6 @@ library EntropyErrors {
     error IncorrectUserRevelation();
     // The provider's revealed random value did not match their commitment.
     error IncorrectProviderRevelation();
+    // Governance message is invalid (e.g., deserialization error).
+    error InvalidGovernanceMessage();
 }
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -74,7 +74,8 @@ import "./EntropyState.sol";
 // cases where the user chooses not to reveal.
 contract Entropy is IEntropy, EntropyState {
     // TODO: Use an upgradeable proxy
-    constructor(uint pythFeeInWei, address defaultProvider) {
+    constructor(address admin, uint pythFeeInWei, address defaultProvider) {
+        _state.admin = admin;
         _state.accruedPythFeesInWei = 0;
         _state.pythFeeInWei = pythFeeInWei;
         _state.defaultProvider = defaultProvider;
--- a/target_chains/ethereum/contracts/contracts/entropy/EntropyGovernance.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/EntropyGovernance.sol
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: Apache 2
+pragma solidity ^0.8.0;
+
+import "@pythnetwork/entropy-sdk-solidity/EntropyErrors.sol";
+
+import "./EntropyState.sol";
+import "./EntropyGovernanceInstructions.sol";
+import "../libraries/external/BytesLib.sol";
+
+/**
+ * @dev `Governance` defines a means to enacting changes to the Pyth contract.
+ */
+abstract contract EntropyGovernance is
+    EntropyState,
+    EntropyGovernanceInstructions
+{
+    using BytesLib for bytes;
+
+    modifier onlyAdmin() {
+        require(msg.sender == _state.admin);
+        _;
+    }
+
+    function executeGovernanceInstruction(
+        bytes memory payload
+    ) external onlyAdmin {
+        GovernanceInstruction memory gi = parseGovernanceInstruction(payload);
+
+        if (gi.action == GovernanceAction.UpgradeContract) {
+            // FIXME: this probably needs a check on the chain id
+            upgradeContract(parseUpgradeContractPayload(gi.payload));
+        } else if (gi.action == GovernanceAction.SetAdmin) {
+            _state.admin = parseSetAdminPayload(gi.payload).newAdmin;
+        } else if (gi.action == GovernanceAction.SetFee) {
+            _state.pythFeeInWei = parseSetFeePayload(gi.payload).newFee;
+        } else {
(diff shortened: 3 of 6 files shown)


target_chains/ethereum/contracts/contracts/entropy/EntropyUpgradable.sol
```diff
@@ -0,0 +1,94 @@
...
+        uint pythFeeInWei,
+        address defaultProvider
+    ) public initializer {
+        __Ownable_init();
+        __UUPSUpgradeable_init();
+
+        Entropy._initialize(admin, pythFeeInWei, defaultProvider);
+
+        // We need to transfer the ownership from deployer to the new owner
+        transferOwnership(owner);
```
jayantk, line 39: note for the future:

we should add a `safeTransferOwnership` function that has two stages:
- proposeOwnershipTransfer(newOwner)
- acceptOwnershipTransfer()

where the second one checks that msg.sender is the newOwner and calls transferOwnership. This approach guarantees that the new owner can send transactions, so we don't accidentally transfer to the wrong address.

This can be done in a separate PR though.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -530,4 +548,76 @@ contract EntropyTest is Test {
         assertEq(providerInfo1.uri, provider1Uri);
         assertEq(providerInfo1.commitmentMetadata, provider1CommitmentMetadata);
     }
+
+    function testSetAdminByAdmin() public {
```
jayantk, line 552: consider making a separate test file for this stuff like `EntropyAdminTest`. These tests don't depend on the same setup as the existing tests in this file, so two files will be easier for future readers to understand.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -530,4 +548,76 @@ contract EntropyTest is Test {
...
+
+    function testSetAdminByOwner() public {
+        vm.prank(owner);
+        random.setAdmin(admin2);
+        assertEq(random.getAdmin(), admin2);
+    }
+
+    function testFailSetAdminByUnauthorised() public {
+        vm.prank(admin2);
+        random.setAdmin(admin);
```
jayantk, line 566: I think `testFail` is deprecated. I think the right way to do this is to use `vm.expectRevert()` before the call that you expect to fail.

target_chains/ethereum/contracts/forge-test/utils/EntropyTestContracts/EntropyDifferentMagic.t.sol
```diff
@@ -0,0 +1,96 @@
...
+
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "@pythnetwork/entropy-sdk-solidity/EntropyErrors.sol";
+
+import "../../../contracts/entropy/EntropyGovernance.sol";
+import "../../../contracts/entropy/Entropy.sol";
+
+contract EntropyDifferentMagic is
```
jayantk, line 12: Do you really need all of this code here to test the magic number thing? I expect you can remove some parts of it (like the upgradeTo stuff). 

it's generally good practice to minimize the amount of code you have hanging around, as code ends up being a maintenence burden over time.

target_chains/ethereum/entropy_sdk/solidity/EntropyErrors.sol
```diff
@@ -19,4 +19,8 @@ library EntropyErrors {
     error IncorrectUserRevelation();
     // The provider's revealed random value did not match their commitment.
     error IncorrectProviderRevelation();
+    // Governance message is invalid (e.g., deserialization error).
+    error InvalidUpgradeMagic();
+    // Invalid authorisation (e.g., invalid admin or owner).
+    error InvalidAuthorisation();
```
jayantk, line 25: how about `Unauthorized`

jayantk (commented):
this looks solid to me. I left a couple minor comments.

I would like @ali-bahjati to take a look as well though -- he's more knowledgable on the upgradable proxy stuff -- so I will let him submit the final approval.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -33,8 +38,21 @@ contract EntropyTest is Test, EntropyTestUtils {
...
+    address public admin = address(9);
+    address public admin2 = address(10);
+
     function setUp() public {
-        random = new Entropy(pythFeeInWei, provider1, false);
+        EntropyUpgradable _random = new EntropyUpgradable();
+        // deploy proxy contract and point it to implementation
+        proxy = new UUPSProxy(address(_random), "");
+        // wrap in ABI to support easier calls
+        random = EntropyUpgradable(address(proxy));
```
ali-bahjati, line 50: Why you are not using ERC1967 directly and created UUPSProxy? :think:

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -528,4 +546,76 @@ contract EntropyTest is Test, EntropyTestUtils {
...
+    function testFailUpgradeByUnauthorised() public {
+        vm.prank(provider1);
+        random.upgradeTo(address(random2));
+    }
+
+    function testFailDifferentMagicContractUpgrade() public {
+        vm.prank(owner);
+        random.upgradeTo(address(randomDifferentMagic));
+    }
 }
```
ali-bahjati, line 621: There can be another case that the magic function doesn't exist but it's fine. (It will revert with no reason)

ali-bahjati (approved):
It is very nice! I have asked some questions in inline comments. Please address the comments and fix the ci before merging.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
0xfirefist: lol, i copied things from a place and missed it somehow. thanks !!!

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
0xfirefist: adding a comment for this

PR #1156, opened 2023-11-29
[entropy] Add authorization for reveal method

jayantk:
Require the reveal method to be called by the same msg.sender that called request. This prevents a denial of service attack (see inline comment)

### DIFF
--- a/target_chains/ethereum/contracts/forge-test/Entropy.t.sol
+++ b/target_chains/ethereum/contracts/forge-test/Entropy.t.sol
@@ -87,12 +87,13 @@ contract EntropyTest is Test, EntropyTestUtils {
         bool useBlockhash
     ) public returns (uint64 sequenceNumber) {
         vm.deal(user, fee);
-        vm.prank(user);
+        vm.startPrank(user);
         sequenceNumber = random.request{value: fee}(
             provider,
             random.constructUserCommitment(bytes32(randomNumber)),
             useBlockhash
         );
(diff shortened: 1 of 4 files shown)


target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -87,12 +87,13 @@ contract EntropyTest is Test, EntropyTestUtils {
         bool useBlockhash
     ) public returns (uint64 sequenceNumber) {
         vm.deal(user, fee);
-        vm.prank(user);
+        vm.startPrank(user);
```
jayantk, line 90: this was a bug. i think the prior prank only applied to the first call, which is actually constructUserCommitment

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -309,21 +380,13 @@ contract EntropyTest is Test, EntropyTestUtils {
...
         );
-
-        // You can only reveal the random number once. This isn't a feature of the contract per se, but it is
-        // the expected behavior.
-        assertRevealReverts(
-            provider1,
-            sequenceNumber,
-            42,
-            provider1Proofs[sequenceNumber]
-        );
```
jayantk, line 326: this code block was copied from another test and isn't needed here

ali-bahjati (approved):

PR #1161, opened 2023-12-02
[fortuna] Add confirmation delay before responding

jayantk:
This PR adds a configurable confirmation delay to fortuna for each blockchain. The purpose of this delay is to ensure that the request has confirmed and will not be rolled back in a chain reorg before revealing the random number.

In order to implement this change, I had to update the entropy contract to store the block number regardless of whether or not the blockhash was used. I did this using 1 bit of the number to save storage slots (sigh).

### DIFF
--- a/fortuna/src/api/revelation.rs
+++ b/fortuna/src/api/revelation.rs
@@ -65,8 +65,10 @@ pub async fn revelation(
         .await
         .map_err(|_| RestError::TemporarilyUnavailable)?;
 
+    let current_block_number = state.contract.get_block_number().await;
+
     match maybe_request {
-        Some(_) => {
+        Some(r) if current_block_number - state.reveal_delay >= r.block_number => {
             let value = &state
                 .state
                 .reveal(sequence)
@@ -77,6 +79,7 @@ pub async fn revelation(
                 value: encoded_value,
             }))
         }
+        Some(_) => Err(RestError::PendingConfirmation),
         None => Err(RestError::NoPendingRequest),
     }
 }
--- a/fortuna/src/api/revelation.rs
+++ b/fortuna/src/api/revelation.rs
@@ -65,10 +65,15 @@ pub async fn revelation(
         .await
         .map_err(|_| RestError::TemporarilyUnavailable)?;
 
-    let current_block_number = state.contract.get_block_number().await;
+    // TODO: paralellize requests
+    let current_block_number = state
+        .contract
+        .get_block_number()
+        .await
+        .map_err(|_| RestError::TemporarilyUnavailable)?;
 
     match maybe_request {
-        Some(r) if current_block_number - state.reveal_delay >= r.block_number => {
+        Some(r) if current_block_number - state.confirmation_blocks >= r.block_number => {
             let value = &state
                 .state
                 .reveal(sequence)
--- a/fortuna/src/chain/ethereum.rs
+++ b/fortuna/src/chain/ethereum.rs
@@ -17,7 +17,10 @@ use {
             abigen,
             EthLogDecode,
         },
-        core::types::Address,
+        core::{
+            k256::U256,
+            types::Address,
+        },
         middleware::{
             transformer::{
                 Transformer,
@@ -36,7 +39,10 @@ use {
             LocalWallet,
             Signer,
         },
-        types::transaction::eip2718::TypedTransaction,
+        types::{
+            transaction::eip2718::TypedTransaction,
+            U64,
+        },
     },
     sha3::{
         Digest,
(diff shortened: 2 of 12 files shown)


fortuna/src/config/register_provider.rs
```diff
@@ -35,7 +34,7 @@ pub struct RegisterProviderOptions {
     /// The fee to charge (in wei) for each requested random number
     #[arg(long = "pyth-contract-fee")]
     #[arg(default_value = "100")]
-    pub fee: U256,
+    pub fee: u128,
```
jayantk, line 37: this is fallout from updating the ABI. this changed on-chain in a previous pr and didn't get updated here

fortuna/src/api/revelation.rs
```diff
@@ -59,14 +60,16 @@ pub async fn revelation(
...
+
+    let current_block_number_fut = state.contract.get_block_number();
+
+    let (maybe_request, current_block_number) =
+        try_join!(maybe_request_fut, current_block_number_fut)
+            .map_err(|_| RestError::TemporarilyUnavailable)?;
 
     match maybe_request {
-        Some(_) => {
+        Some(r) if current_block_number - state.confirmation_blocks >= r.block_number => {
```
ali-bahjati, line 72: it shouldn't happen but let's make it saturating sub to be extra careful

ali-bahjati (commented):
The code looks correct.
However, this optimization to embed something in an existing type freaks me out and increases the likelihood of making mistakes in the future. I am not sure it is entirely worth it but you will know better. The least thing is changing the name of the variable everywhere so it doesn't get mistaken by just a `blockNumber`.

fortuna/src/chain/ethereum.rs
```diff
@@ -194,9 +197,15 @@ impl EntropyReader for PythContract {
             Ok(Some(reader::Request {
                 provider:        r.provider,
                 sequence_number: r.sequence_number,
+                block_number:    r.block_number.abs().try_into()?,
```
ali-bahjati, line 200: To support my comment. This `abs` here is so subtle (and without docs) that can get easily missed if someone edits this code.

fortuna/src/api/revelation.rs
```diff
@@ -59,14 +60,16 @@ pub async fn revelation(
...
-        .get_request(state.provider_address, sequence)
-        .await
-        .map_err(|_| RestError::TemporarilyUnavailable)?;
+    let maybe_request_fut = state.contract.get_request(state.provider_address, sequence);
+
+    let current_block_number_fut = state.contract.get_block_number();
+
+    let (maybe_request, current_block_number) =
+        try_join!(maybe_request_fut, current_block_number_fut)
+            .map_err(|_| RestError::TemporarilyUnavailable)?;
```
Reisen, line 69: Minor point but it might be nice to log which of these failed by logging the error just in case.

fortuna/src/api/revelation.rs
```diff
@@ -59,14 +60,16 @@ pub async fn revelation(
         .get(&chain_id)
         .ok_or_else(|| RestError::InvalidChainId)?;
 
-    let maybe_request = state
-        .contract
-        .get_request(state.provider_address, sequence)
-        .await
-        .map_err(|_| RestError::TemporarilyUnavailable)?;
+    let maybe_request_fut = state.contract.get_request(state.provider_address, sequence);
```
Reisen, line 63: Obviously not for this PR at all but just wanted to note the fact that the endpoint is doing contract interactions and isn't abstracted away for other chains yet, we probably don't want to be directly interacting with state in our endpoints without business logic abstracting the chain implementation.

fortuna/src/config.rs
```diff
@@ -115,6 +118,9 @@ pub struct EthereumConfig {
     /// Address of a Pyth Randomness contract to interact with.
     pub contract_addr: Address,
 
+    /// How many blocks to wait before revealing the random number.
+    pub confirmation_blocks: BlockNumber,
```
Reisen, line 122: Very very minor nitpick and possibly bikeshedding, but I think the name is a bit confusing, `reveal_delay_blocks` might be a bit better.

fortuna/src/chain/ethereum.rs
Reisen: Would suggest `struct BlockHeight()` with a validating constructor to make this never slip through, in general I'd advocate introducing more types and avoiding `type` in general.

fortuna/src/api/revelation.rs
jayantk: I’m confused by this comment

the `contract` referenced here is a `dyn EntropyReader`, which is an abstraction that is (mostly) chain agnostic. (It's chain agnostic except for that it uses ethereum addresses, but we could easily make the address fields into strings to support other chains. It's a trivial refactor, which is also the reason I haven't done it yet)

ali-bahjati (approved):

Reisen (approved):

PR #1166, opened 2023-12-05
fix(hermes): handle price feed removal properly

ali-bahjati:


hermes/src/state/cache.rs
```diff
@@ -206,17 +208,40 @@ impl AggregateCache for crate::state::State {
...
+    /// This method takes the current feed ids and prunes the cache for the keys
+    /// that are not present in the current feed ids.
+    ///
+    /// There is a side-effect of this: if a key gets removed, we will
+    /// lose the cache for that key and cannot retrieve it for historical
+    /// price queries.
+    async fn prune_removed_keys(&self, current_keys: HashSet<MessageStateKey>) {
+        let mut message_cache = self.cache.message_cache.write().await;
+
+        // Sometiems, some keys are removed from the accumulator. We track which keys are not
```
m30m, line 231: typo

hermes/src/api/ws.rs
```diff
@@ -350,22 +350,36 @@ impl Subscriber {
...
         {
+            Ok(updates) => updates,
+            Err(_) => {
+                // The error can only happen when a price feed was available
+                // and is no longer there as we check the price feed ids upon
+                // subscription. In this case we send an error message and
+                // close the connection.
+                self.sender
+                    .send(
+                        serde_json::to_string(&ServerResponseMessage::Err {
```
m30m, line 369: Can't we just ignore the removed feeds? In the current solution, when a price feed is removed, all websocket connections need to be opened up again and we will miss a few slots in between.

hermes/src/state/cache.rs
```diff
@@ -206,17 +208,40 @@ impl AggregateCache for crate::state::State {
...
+        // present in the message states and remove them from the cache.
+        let keys_in_cache = message_cache
+            .iter()
+            .map(|(key, _)| key.clone())
+            .collect::<HashSet<_>>();
+
+        for key in keys_in_cache {
+            if !current_keys.contains(&key) {
+                message_cache.remove(&key);
```
m30m, line 240: Since this is not very often, it would be great if we add some log here to see it in action once live.

ali-bahjati: I added a info log level for it.

hermes/src/api/ws.rs
ali-bahjati: Yeah it's a fair point and I have changed the code to follow what you says. However I think it is an unintuitive business logic and might surprise consumers. I wouldn't do it if our data ingestion wasn't relying on this.

The better approach would be what you had suggested before, a subscribeToAll method (or a `*`) and it is intuitive for a subscription to all to keep the connection while a feed is removed. However I didn't do it because it is a big change and we are going to move away from this API soon.

hermes/src/aggregate.rs
```diff
@@ -296,18 +305,17 @@ pub async fn store_update(state: &State, update: Update) -> Result<()> {
     Ok(())
 }
 
-#[tracing::instrument(skip(state, accumulator_messages, wormhole_merkle_state))]
+#[tracing::instrument(skip(accumulator_messages, wormhole_merkle_state))]
 async fn build_message_states(
```
Reisen, line 309: This is pure now, can remove async.

hermes/src/api/ws.rs
```diff
@@ -350,22 +350,46 @@ impl Subscriber {
...
+            Err(_) => {
+                // The error can only happen when a price feed was available
+                // and is no longer there as we check the price feed ids upon
+                // subscription. In this case we just remove the non-existing
+                // price feed from the list and will keep sending updates for
+                // the rest.
+                let available_price_feed_ids =
+                    crate::aggregate::get_price_feed_ids(&*self.store).await;
+
+                price_feed_ids.iter().for_each(|price_feed_id| {
```
Reisen, line 371: Can probably combine for_each and collect below with a filter of some kind.

PR #1168, opened 2023-12-06
chore(target_chains/eth): revert solidity version

ali-bahjati:
This change reverts the solidity version to 0.8.4 as the current version troubles the Pyth price feeds contract deployment. Some tests are changed to use the methods available in the older version.

Also, the Pyth contract version is updated. This is because we made a change to the Governance structs (to add Stacks) that results in a new bytecode.


m30m (approved):

ali-bahjati:
@jayantk well i needed to move away from `abi.encodeCall` to make it work.

ali-bahjati:
@0xfirefist good point. Actually we have 0.8.0 everywhere so I changed these to 0.8.0 too.

PR #1175, opened 2023-12-11
feat(target-chains/solana): Implement governance for the solana pull oracle

guibescos:
Threw together an implementation of governance.
Here are the main differences with the evm implementation : 
- It is governed directly by a solana key, for remote governance we can use the remote executor
- No valid time period so far as I'm focusing on parsePriceFeeds functionality
- governanceDataSource is replaced by `governance_authority` the Solana key that can govern the contract.


target_chains/solana/programs/pyth-solana-receiver/Cargo.toml
```diff
@@ -13,7 +13,7 @@ no-entrypoint = []
 no-idl = []
 no-log-ix-name = []
 cpi = ["no-entrypoint"]
-default = []
+default = ["mainnet"]
```
guibescos, line 16: This makes it more convenient to test by not having to pass --feature mainnet everytime

target_chains/solana/programs/pyth-solana-receiver/src/error.rs
```diff
@@ -22,4 +22,10 @@ pub enum ReceiverError {
     InvalidPriceUpdate,
     #[msg("Received an invalid accumulator message type")]
     InvalidAccumulatorMessageType,
+    #[msg("The signer is not authorized to perform this governance action")]
+    GovernanceAuthorityMismatch,
+    #[msg("The signer is not authorized to accept the governance authority")]
+    TargetGovernanceAuthorityMismatch,
+    #[msg("The governance authority needs to request a transfer first")]
+    InexistentGovernanceAuthorityTransferRequest,
```
jayantk, line 30: Nonexistent?

target_chains/solana/programs/pyth-solana-receiver/src/state/config.rs
```diff
@@ -0,0 +1,23 @@
...
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct DataSource {
+    pub chain:   u16,
+    pub emitter: Pubkey,
+}
+
+impl Config {
+    pub const LEN: usize = 1000; // PLACEHOLDER LEN, DECIDE HOW BIG WE WANT TO MAKE IT
+}
```
jayantk, line 23: this seems like the kind of thing we can forget to update later. 1000 is definitely big enough right now, but could get tight if we add to the struct.

maybe add a test where you make a Config struct with some reasonable values (like 2 data sources etc) then assert that the LEN > the size of that struct? Or maybe even the size of the struct * 2 or something? 

(I'm not convinced about the idea ^. Main thing is let's not forget to update this later.)

jayantk (approved):
nice

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -103,6 +153,42 @@ pub mod pyth_solana_receiver {
...
+    #[account(seeds = [CONFIG_SEED.as_ref()], bump)]
+    pub config: Account<'info, Config>,
+}
+
+#[derive(Accounts)]
+pub struct AuthorizeGovernanceAuthorityTransfer<'info> {
+    #[account(constraint =
+        payer.key() == config.target_governance_authority.ok_or(error!(ReceiverError::InexistentGovernanceAuthorityTransferRequest))? @
+        ReceiverError::TargetGovernanceAuthorityMismatch
+    )]
```
ali-bahjati, line 185: dark magic of anchor :D

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -27,6 +32,51 @@ declare_id!("DvPfMBZJJwKgJsv2WJA8bFwUMn8nFd5Xpioc6foC3rse");
...
+        let config = &mut ctx.accounts.config;
+        config.target_governance_authority = Some(target_governance_authority);
+        Ok(())
+    }
+
+    pub fn authorize_governance_authority_transfer(
+        ctx: Context<AuthorizeGovernanceAuthorityTransfer>,
+    ) -> Result<()> {
+        let config = &mut ctx.accounts.config;
+        config.governance_authority = config.target_governance_authority.unwrap();
```
ali-bahjati, line 54: i know it's checked to be not None on anchor but still scares me a bit.

ali-bahjati (approved):

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: mm worst case it panics?

ali-bahjati: yeah but in this case I recommend using `expect("this is checked in anchor and should never fail")` to make the code more clear.

PR #1177, opened 2023-12-12
[entropy] audit - 2. lack of contract existence check

0xfirefist:
Fixes:

2. Lack of contract existence check on low-level call


target_chains/ethereum/contracts/contracts/executor/ExecutorErrors.sol
```diff
@@ -14,4 +14,6 @@ library ExecutorErrors {
     error DeserializationError();
     // The message is not intended for this contract.
     error InvalidGovernanceTarget();
+    // The message is not intended for a valid contract address.
```
jayantk, line 17: the comment here is imprecise. The "message" refers to the wormhole message. I think you want to say "the target address for the contract call is not a contract"

jayantk (approved):

PR #1179, opened 2023-12-12
[entropy] audit 4. users can influence the Entropy revealed result

0xfirefist:


target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -253,7 +253,16 @@ abstract contract Entropy is IEntropy, EntropyState {
 
         bytes32 blockHash = bytes32(uint256(0));
         if (req.useBlockhash) {
-            blockHash = blockhash(req.blockNumber);
+            bytes32 _blockHash = blockhash(req.blockNumber);
+
+            // The request can only be stale if the blockhash is zero. The other case when blockhash
+            // can be zero is when the request and reveal is in the same block which can't happen as
+            // the service will reveal only when the request is finalized in a block
+            // Though this will fail for that case too.
```
jayantk, line 261: this comment is confusing. I don't know what a "stale request" is, and this comment assumes I do. Then it proceeds to say "well also it fails when the blockhash is 0 for a different reason".

I think the comment should explain why the blockhash can be zero, and why we need this check in those cases.

Also suggest changing the error to be `BlockhashUnavailable`, which is a more precise description of the problem.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -405,6 +405,7 @@ contract EntropyTest is Test, EntropyTestUtils {
             true
         );
 
+        vm.roll(1236);
```
jayantk, line 408: why is this 1236 and not 1235? I would have thought 1235 would work too.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -415,6 +416,86 @@ contract EntropyTest is Test, EntropyTestUtils {
...
+        sequenceNumber = request(user2, provider1, 42, true);
+
+        vm.roll(1234 + 257);
+        assertRevealReverts(
+            user2,
+            provider1,
+            sequenceNumber,
+            42,
+            provider1Proofs[sequenceNumber]
+        );
```
jayantk, line 496: I think this test would be better if there was exactly 1 `sequenceNumber` generated at the beginning, then you test the two revert conditions, then you test that the reveal succeeds afterward.

That way of writing this test is less error prone, because it ensures the block number is the only variable that is changing between the reveal calls.

jayantk (approved):
please address the comments before merging. They're easy though so no need to get another review from me.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
0xfirefist: no reason as such.

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
0xfirefist: oh yeah!

PR #1180, opened 2023-12-13
[pyth-solana-receiver-cli] Initialize solana wormhole receiver

guibescos:
This is code to initialize a "forked" wormhole contract. There are two reasons this is useful:
- Will allow me to use stable price feeds in devnet
- We might want to experiment with optimizations to shrink transactions further. In particular, I want to experiment with letting the user choose how many wormhole signatures they want to verify


target_chains/solana/cli/src/main.rs
```diff
@@ -55,6 +58,13 @@ use {
         VAA as WormholeSolanaVAA,
     },
 };
+
+const INITIAL_GUARDIAN: &str = "58cc3ae5c097b213ce3c81979e1b9f9570746aa5";
```
jayantk, line 62: comment that this is wormhole mainnet

jayantk (approved):

PR #1183, opened 2023-12-13
[xc-admin] Add activate button for drafts

guibescos:
A button to activate proposals that are in draft mode. This happens when xc-admin-proposer fails to publish the entire proposal.

Also I add waiting 1 second between retries when broadcast proposal creating transactions.


governance/xc_admin/packages/xc_admin_common/src/propose.ts
```diff
@@ -409,6 +409,7 @@ export class MultisigVault {
             needToFetchBlockhash = true;
           }
           console.log(e);
+          await new Promise((r) => setTimeout(r, 1000));
```
m30m, line 412: I'd suggest increasing this to 5 seconds, we are not in a hurry and 1 second may not be enough sometimes.

guibescos: mmm let's try 5 seconds if we still struggle

guibescos: yes

guibescos: Made it 3 seconds for now

PR #1187, opened 2023-12-14
[fortuna] update config and abi

0xfirefist:


fortuna/src/chain/ethereum.rs
```diff
@@ -189,7 +189,8 @@ impl EntropyReader for PythContract {
     ) -> Result<Option<reader::Request>> {
         let r = self
             .get_request(provider_address, sequence_number)
-            .block(ethers::core::types::BlockNumber::Finalized)
+            // TODO: This doesn't work for lighlink right now. Figure out how to do this in lightlink
+            // .block(ethers::core::types::BlockNumber::Finalized)
```
ali-bahjati, line 193: I think it might be the case for many network. In the future you can add a cli argument for this.

ali-bahjati (approved):

fortuna/src/chain/ethereum.rs
0xfirefist: yes, i will figure this one out. for now let's go on lightlink

PR #1193, opened 2023-12-18
[fortuna] sample config

0xfirefist:


fortuna/.gitignore
```diff
@@ -1 +1,2 @@
 /target
+config.yaml
```
jayantk, line 2: i think you may also need to `git rm config.yaml`. I think it's probably still in the repo.

jayantk (approved):

PR #1194, opened 2023-12-18
[executor] make it upgradeable

0xfirefist:


target_chains/ethereum/contracts/contracts/executor/Executor.sol
```diff
@@ -4,64 +4,91 @@ pragma solidity ^0.8.0;
...
     struct GovernanceInstruction {
         PythGovernanceInstructions.GovernanceModule module;
         ExecutorAction action;
         uint16 targetChainId;
+        bytes payload;
+    }
+
+    struct UpgradeContractPayload {
+        address newImplementation;
+    }
```
jayantk, line 33: do we really need other instruction type? I think the contract could be its own owner, and then you could send the normal execute instruction telling the executor to call its own upgrade method.

jayantk (commented):
I think there's a simpler way to do this -- see inline comment

PR #1198, opened 2023-12-19
[executor] implement upgradability

0xfirefist:


target_chains/ethereum/contracts/contracts/executor/ExecutorErrors.sol
```diff
@@ -16,4 +16,6 @@ library ExecutorErrors {
     error InvalidGovernanceTarget();
     // The target address for the contract call is not a contract
     error InvalidContractTarget();
+    // The governance message is not valid
+    error InvalidGovernanceMessage();
```
m30m, line 20: Maybe make the error a bit more specific? Like `InvalidMagicValue`? Since the governance message is valid but the  contract it points to is not

m30m (approved):
Made a renaming comment, lgtm

target_chains/ethereum/contracts/contracts/executor/ExecutorErrors.sol
0xfirefist: yeah ok

PR #1201, opened 2023-12-21
[audit] Secrets appear in environment variables and command line arguments

0xfirefist:
Context: Right now Fortuna, the offchain service we run for Entropy, takes in secrets/private keys as command line arguments. We want to change that and instead use files. 

I am resolving this in the following steps:
- Accept both secret, and secret file path using the same argument
- Update deployment config to pass file path
- Update Fortuna to accept only file path

Similarly a separate PR will be raised for private key.


fortuna/src/command/register_provider.rs
```diff
@@ -31,13 +31,15 @@ pub async fn register_provider(opts: &RegisterProviderOptions) -> Result<()> {
 
     // Create a new random hash chain.
     let random = rand::random::<[u8; 32]>();
+    let secret: String;
+    match opts.randomness.load_secret() {
+        Ok(loaded_secret) => secret = loaded_secret,
+        Err(_err) => secret = opts.randomness.secret_file.clone(),
+    }
```
guibescos, line 38: In rust you can do 
`let secret = match ...`

fortuna/src/config.rs
```diff
@@ -76,10 +76,12 @@ pub struct ConfigOptions {
 #[command(next_help_heading = "Randomness Options")]
 #[group(id = "Randomness")]
 pub struct RandomnessOptions {
-    /// A secret used for generating new hash chains. A 64-char hex string.
+    /// Path to file containing a secret which is a 64-char hex string.
+    /// The secret is used for generating new hash chains
+    /// Or the secret itself. TODO: this will be remove in another PR.
```
guibescos, line 81: this will be remove**d**

fortuna/src/command/run.rs
```diff
@@ -47,6 +47,12 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
     struct ApiDoc;
 
     let config = Config::load(&opts.config.config)?;
+    let secret: String;
+    match opts.randomness.load_secret() {
```
guibescos, line 51: let `secret = match opts.randomness.load_secret()`

fortuna/src/config.rs
```diff
@@ -88,6 +90,13 @@ pub struct RandomnessOptions {
     pub chain_length: u64,
 }
 
+impl RandomnessOptions {
+    pub fn load_secret(&self) -> Result<String> {
+        let secret = fs::read_to_string(&self.secret_file)?;
```
guibescos, line 95: I think you can return `fs::read_to_string(&self.secret_file)`

```
pub fn load_secret(&self) -> Result<String> {
   fs::read_to_string(&self.secret_file)
}
```

guibescos (approved):
Please address comments. Are you going to address "private_key" too?

PR #1210, opened 2024-01-05
[solana] Post update atomic

guibescos:
Adds an instruction to post a price update in 1 transaction.
This function can only be called with 5 guardian signatures or less.

There will be two guys to post a price update.
- post_price_update : 2 transactions but verify 13 signatures and the same VAA can be used to post multiple price updates (for different price feeds)
- post_price_update_atomic : 1 transactions, <= 5 signatures

### DIFF
--- a/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
+++ b/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
@@ -346,19 +346,21 @@ pub struct PostUpdates<'info> {
 pub struct PostUpdatesAtomic<'info> {
     #[account(mut)]
     pub payer:                Signer<'info>,
-    #[account(owner = config.wormhole,
+    #[account(
         seeds = [
             GuardianSet::SEED_PREFIX,
             guardian_set.inner().index.to_be_bytes().as_ref()
         ],
-        bump,)]
+        seeds::program = config.wormhole,
+        bump,
+        owner = config.wormhole)]
     pub guardian_set:         Account<'info, AccountVariant<GuardianSet>>,
     #[account(seeds = [CONFIG_SEED.as_ref()], bump)]
     pub config:               Account<'info, Config>,
     #[account(mut, seeds = [TREASURY_SEED.as_ref()], bump)]
     /// CHECK: This is just a PDA controlled by the program. There is currently no way to withdraw funds from it.
     pub treasury:             AccountInfo<'info>,
-    #[account(init, payer =payer, space = PriceUpdateV1::LEN)]
+    #[account(init, payer = payer, space = PriceUpdateV1::LEN)]
     pub price_update_account: Account<'info, PriceUpdateV1>,
     pub system_program:       Program<'info, System>,
 }
--- a/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
+++ b/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
@@ -51,7 +51,6 @@ use {
 pub mod error;
 pub mod state;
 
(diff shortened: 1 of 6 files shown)


target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -243,3 +350,106 @@ impl crate::accounts::PostUpdates {
...
     }
 }
+
+struct VaaComponents {
+    verified_signatures: u8,
+    emitter_address:     [u8; 32],
+    emitter_chain:       u16,
+}
+
+fn post_price_update_from_vaa<'info>(
```
guibescos, line 360: This is a copy paste refactor to avoid duplicating this code in :
`post_updates` and `post_updates_atomic`

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -90,80 +104,115 @@ pub mod pyth_solana_receiver {
...
+
+        // Make sure the encoded guardian set index agrees with the guardian set account's index.
+        let guardian_set = ctx.accounts.guardian_set.inner();
+        require_eq!(
+            vaa.guardian_set_index(),
+            guardian_set.index,
+            ReceiverError::GuardianSetMismatch
+        );
+
+        // Do we have enough signatures for quorum?
```
jayantk, line 129: comment seems wrong (this doesn't seem to check anything)

jayantk: maybe leave a comment here that we don't need to check the number of signatures, but users of posted price updates are *required* to check the `verified_signatures` field themselves.

jayantk: by the way, asking pyth users to check `verified_signatures` is somewhat scary. It's the kind of check people can easily forget about. we should think about how to make this less error prone.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -217,6 +265,35 @@ pub struct PostUpdates<'info> {
...
+        bump,
+        owner = config.wormhole)]
+    pub guardian_set:         Account<'info, AccountVariant<GuardianSet>>,
+    #[account(seeds = [CONFIG_SEED.as_ref()], bump)]
+    pub config:               Account<'info, Config>,
+    #[account(mut, seeds = [TREASURY_SEED.as_ref()], bump)]
+    /// CHECK: This is just a PDA controlled by the program. There is currently no way to withdraw funds from it.
+    pub treasury:             AccountInfo<'info>,
+    #[account(init, payer = payer, space = PriceUpdateV1::LEN)]
+    pub price_update_account: Account<'info, PriceUpdateV1>,
```
jayantk, line 287: (not an issue right now but: are we going to have a way for people to reclaim the storage associated with these accounts?)

jayantk (approved):
nice!

this code definitely needs some solid testing -- maybe worth leaving some notes around what tests to add in a unified place so that you don't forget about them later.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: Yes

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: I have an idea about this

PR #1213, opened 2024-01-08
Fix pythnet-sdk ci

guibescos:
pythnet-sdk CI has been failing for a bit.
Bumping nightly version in the pre-commits worked.
I think the error was caused by a new version of `ahash` that our `nightly-2023-03-01` didn't like

### DIFF
--- a/pythnet/pythnet_sdk/src/messages.rs
+++ b/pythnet/pythnet_sdk/src/messages.rs
@@ -3,6 +3,7 @@ use quickcheck::Arbitrary;
 use {
     borsh::{
         BorshDeserialize,
+        BorshSchema,
         BorshSerialize,
     },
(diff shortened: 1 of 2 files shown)


pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -75,7 +76,15 @@ pub type FeedId = [u8; 32];
 
 #[repr(C)]
 #[derive(
-    Debug, Copy, Clone, PartialEq, Serialize, Deserialize, BorshDeserialize, BorshSerialize,
+    Debug,
```
guibescos, line 79: Bundling with this change that I want for the solana receiver contract

ali-bahjati (approved):

PR #1216, opened 2024-01-08
[solana] Reuse accounts

guibescos:
This makes it easy to reuse accounts.
You can also use this to build a push oracle by sequentially posting a price feed's update to the same account.

### DIFF
--- a/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
+++ b/target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
@@ -266,6 +266,8 @@ pub struct PostUpdates<'info> {
     /// CHECK: This is just a PDA controlled by the program. There is currently no way to withdraw funds from it.
     #[account(mut)]
     pub treasury:             AccountInfo<'info>,
+    /// The contraint is such that either the price_update_account is uninitialized or the payer is the write_authority.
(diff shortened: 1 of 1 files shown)


target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -289,7 +289,7 @@ pub struct PostUpdatesAtomic<'info> {
     #[account(mut, seeds = [TREASURY_SEED.as_ref()], bump)]
     /// CHECK: This is just a PDA controlled by the program. There is currently no way to withdraw funds from it.
     pub treasury:             AccountInfo<'info>,
-    #[account(init, payer = payer, space = PriceUpdateV1::LEN)]
+    #[account(init_if_needed, constraint = price_update_account.write_authority == Pubkey::default() || price_update_account.write_authority == payer.key(), payer = payer, space = PriceUpdateV1::LEN)]
```
guibescos, line 292: It can only be equal to Pubkey::default() (i.e. all zeros) when it just got initialized.

jayantk: maybe worth leaving a code comment about ^

jayantk (approved):

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: Done

PR #1218, opened 2024-01-09
[fortuna] bind provider and contract address

0xfirefist:
There was one discussion in slack about having multiple contract on a chain. Therefore, I bind contract address too when computing the hash chain.


jayantk (approved):
nice

PR #1219, opened 2024-01-09
[entropy] Add entropy ABIs to solidity sdk

jayantk:
Generate the ABIs for the contracts and put them in the solidity SDK. The JS sdk will depend on the solidity SDK to get access to these ABIs

I've set this up exactly the same as how we've done it for price feeds. I slightly generalized the CI check for ABIs so that it applies to both sdks.


target_chains/ethereum/entropy_sdk/solidity/scripts/generateAbi.js
```diff
@@ -0,0 +1,54 @@
+const fs = require("fs");
```
jayantk, line 1: I copy/pasted this from the other solidity SDK. It's somewhat unfortunate that this code is duplicated, but we don't have a good place for shared javascript and I don't want to solve that problem right now.

0xfirefist (approved):

PR #1220, opened 2024-01-10
[fortuna] support env var in yaml config

0xfirefist:


fortuna/src/config.rs
```diff
@@ -115,6 +120,29 @@ impl Config {
...
+    // It will replace substrings that matches the pattern ${VAR_NAME} with an 
+    // environment variable VAR_NAME. If there is no such env var, it will return 
+    // an error.
+    fn inject_env(yaml_content: &str) -> Result<String> {
+        let re = Regex::new(r"\$\{([a-zA-Z_][0-9a-zA-Z_]*)\}").unwrap();
+        let mut var_error: Option<env::VarError> = None;
+        let new_yaml_content =
+            re.replace_all(&yaml_content, |caps: &Captures| match env::var(&caps[1]) {
+                Ok(val) => val,
+                Err(err) => {
```
ali-bahjati, line 134: why not return immediately with error?

fortuna/src/config.rs
```diff
@@ -105,7 +110,7 @@ impl Config {
         // Open and read the YAML file
         // TODO: the default serde deserialization doesn't enforce unique keys
         let yaml_content = fs::read_to_string(path)?;
-        let config: Config = serde_yaml::from_str(&yaml_content)?;
+        let config: Config = serde_yaml::from_str(&Config::inject_env(&yaml_content)?)?;
```
jayantk, line 113: I think you should explicitly pass a map of substitutions into the `load` function and down into the config replacement. That will make these functions into pure functions (which are easy to test, use in other contexts, etc)

jayantk: this also forces you to make the dependencies of the code explicit instead of implicit

jayantk (changes requested):
I left a couple implementation suggestions here

One other note is that this feels complicated. What exactly are we trying to achieve here, and is there a simpler way to do it? 

I think you're trying to keep the API keys a secret? If so, can we do something simpler like put the whole config.yaml file in 1password?

Or is it such a big deal to put the API keys in the deployments repo? These aren't super sensitive, and I see there are some other API keys in there already.

fortuna/src/main.rs
```diff
@@ -22,6 +23,9 @@ pub mod state;
 #[tokio::main]
 #[tracing::instrument]
 async fn main() -> Result<()> {
+    // Load the environment variable from .env file
+    dotenv().ok();
```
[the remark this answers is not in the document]
0xfirefist: I am thinking of a future state where we will need multiple env vars like 5-6. 
We don't need dotenv for deployments. We are not using .env file there. 

But for local development. I think it's better to have them in env file.

PR #1221, opened 2024-01-10
[wip] Entropy JS sdk (??)

jayantk:
I'm not sure this is worth it.


target_chains/ethereum/entropy_sdk/js/src/examples/EvmRandomDraw.ts
```diff
@@ -0,0 +1,126 @@
...
+    providerOrUrl: network,
+  });
+  console.log(`wallet address: ${walletProvider.getAddress(0)}`);
+
+  // @ts-ignore
+  const web3 = new Web3(walletProvider);
+
+  const userRandom = web3.utils.randomHex(32);
+  console.log(`userRandom: ${userRandom}`);
+  const commitment = web3.utils.keccak256(userRandom);
```
jayantk, line 59: users will need this code snippet to generate a random number and the corresponding commitment. However, this is tied to the web3.js library, which not everyone uses (some use ethers etc) so not sure people will want that dependency.

target_chains/ethereum/entropy_sdk/js/src/examples/EvmRandomDraw.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const connection = new FortunaConnection(fortunaUrl);
+  const providerRandomHex = await connection.retrieveRandomNumber(
+    sequenceNumber
+  );
+  const providerRandom = `0x${providerRandomHex}`;
+  console.log(`   number    : ${providerRandom}`);
+
+  let blockFinalized = false;
+  while (!blockFinalized) {
```
jayantk, line 103: this loop to wait for block finalization may also be useful for users

target_chains/ethereum/entropy_sdk/js/src/FortunaConnection.ts
```diff
@@ -0,0 +1,26 @@
+import axios, { AxiosInstance } from "axios";
+import axiosRetry from "axios-retry";
+
+export class FortunaConnection {
```
jayantk, line 4: this class doesn't seem worth it imo

PR #1222, opened 2024-01-11
[contract-manager] More utility scripts and docs

m30m:


ali-bahjati (approved):
amazing! I think with the upgrade script you can remove my evm deployment script.

m30m:
I still think your script is useful since the new one is only practical for `PythUpgradable` and doesn't accept any deployment args. For something more custom and low level it might come in handy 🤔

PR #1226, opened 2024-01-12
[Solana] tests

guibescos:
It's just a happy path for now, but this is a good time to review the fixtures and setup in general


target_chains/solana/program_simulator/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "program-simulator"
```
guibescos, line 2: I segregated this one in case we wanna use it for another program

target_chains/solana/program_simulator/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    },
+    solana_sdk::{
+        signature::{
+            Keypair,
+            Signer,
+        },
+        transaction::Transaction,
+    },
+};
+
```
guibescos, line 22: I dropped the logic to add a program as upgradable because it doesn't seem to work with 1.16.20, and I don't think we'll need this for the Solana Receiver

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -303,69 +301,6 @@ pub struct PostUpdatesAtomicParams {
     pub merkle_price_update: MerklePriceUpdate,
 }
 
-impl crate::accounts::Initialize {
-    pub fn populate(payer: &Pubkey) -> Self {
-        let config = Pubkey::find_program_address(&[CONFIG_SEED.as_ref()], &crate::ID).0;
```
guibescos, line 308: These get moved to `sdk`

target_chains/solana/program_simulator/Cargo.toml
jayantk: thank you! would be great to avoid copy/pasting it

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            exponent:          11,
+            publish_time:      12,
+            prev_publish_time: 13,
+            ema_price:         14,
+            ema_conf:          15,
+        },
+    ]
+}
+
+pub fn dummy_price_updates() -> (MerkleTree<Keccak160>, Vec<MerklePriceUpdate>) {
```
jayantk, line 104: (minor) maybe this should take a Vec<Message> so it's reusable for more tests

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            message: PrefixedVec::<u16, u8>::from(x.clone()),
+            proof:   merkle_tree_accumulator.prove(x.as_ref()).unwrap(),
+        })
+        .collect();
+
+    (merkle_tree_accumulator, merkle_price_updates)
+}
+
+
+pub async fn build_merkle_root_encoded_vaa(
```
jayantk, line 129: why is this async?

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub encoded_vaa_address:  Pubkey,
+    pub merkle_price_updates: Vec<MerklePriceUpdate>,
+}
+/**
+ * Setup to test the Pyth Receiver. The return values are a tuple composed of :
+ * - The program simulator, which is used to send transactions
+ * - The pubkey of an encoded VAA account, which is pre-populated and can be used to test post_updates
+ * - A vector of MerklePriceUpdate, corresponding to that VAA
+ */
+pub async fn setup_pyth_receiver() -> ProgramTestFixtures {
```
jayantk, line 182: this seems like an odd combination of functionality. It seems like most tests will want to send their own set of price updates (?) Will this ever be reused?

jayantk (approved):
LGTM. I think you will need to refactor the test helper methods to be more flexible -- right now, they're not going to be very helpful for other tests -- but feel free to merge this and refactor as you go.

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
jayantk: (it's a little annoying that you need to set up the encoded vaa accounts before the simulator gets created. Tests probably want the ability to modify the set of encoded VAA accounts themselves. Not sure if there's a good way to facilitate that though.

Even if you can't address that, passing in the set of accounts/merkle trees here would make this code more flexible)

guibescos: I'll think about this in future development

PR #1237, opened 2024-01-23
[cosmwasm]  Use ParsedVaa from wormhole contract

guibescos:
I noticed this while playing around with the cosmos contract. 
We were copy pasting ParsedVAA and WormholeQueryMsg instead of importing them from cw_wormhole.
The reason for this seems to be incompatiblity with osmosis. However I was able to build properly with osmosis, so it think the reason doesn't exist anymore and we can import from the wormhole crate.


target_chains/cosmwasm/contracts/pyth/src/contract.rs
```diff
@@ -1095,28 +1095,28 @@ mod test {
...
     fn check_sufficient_fee(deps: &Deps, data: &[Binary]) {
         let mut info = mock_info("123", coins(100, "foo").as_slice());
-        let result = is_fee_sufficient(&deps, info.clone(), &data);
+        let result = is_fee_sufficient(deps, info.clone(), data);
         assert_eq!(result, Ok(true));
 
         // insufficient fee in base denom -> false
         info.funds = coins(50, "foo");
-        let result = is_fee_sufficient(&deps, info.clone(), &data);
+        let result = is_fee_sufficient(deps, info.clone(), data);
```
guibescos, line 1103: Clippy did this

target_chains/cosmwasm/contracts/pyth/src/wormhole.rs
```diff
@@ -1,46 +0,0 @@
-// These types are copied from the Wormhole contract. See the links with each type to see the original code
-// The reason to do so was dependency conflict. Wormhole contracts were using a very old version of a dependency
-// which is not compatible with the one used by osmosis-sdk. And since we weren't using anything else from
```
guibescos, line 3: Seems like this comment is outdated

jayantk (approved):
I think if it builds, it should be fine?

m30m:
Thanks! It would be great if you can run `Release Pyth Cosmwasm Contract` workflow against this branch just to make sure everything works.

target_chains/cosmwasm/contracts/pyth/Cargo.toml
```diff
@@ -40,6 +40,7 @@ cosmwasm-schema = "1.1.9"
 osmosis-std = "0.15.2"
 pyth-sdk-cw = { path = "../../sdk/rust" }
 pythnet-sdk = { path = "../../../../pythnet/pythnet_sdk" }
+wormhole-cosmwasm =  {git = "https://github.com/wormhole-foundation/wormhole"}
```
ali-bahjati, line 43: can you please lock it to a commit?
I think wormhole folks should really consider releasing their packages.

PR #1238, opened 2024-01-23
[solana/cosmwasm/near] Rust tests refactor

guibescos:
I have the very ambitious project to "move up" a lot of the code that we use to generate accumulator updates in the Rust tests for the rust ecosystems (Cosmwasm, Near and Solana).

This PR is about deciding where this code should live.
I made a feature `gen` in `pythnet-sdk` that exposes the generator functions.
The reason why it's feature gated is that I use `wormhole_sdk` and I think it's good to be able to use `pythnet`-sdk without any wormhole dependencies.


pythnet/pythnet_sdk/src/gen/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    byteorder::BigEndian,
+    serde_wormhole::RawMessage,
+    wormhole_sdk::{
+        Address,
+        Chain,
+        Vaa,
+    },
+};
+
+fn default_emitter_addr() -> [u8; 32] {
```
guibescos, line 31: Some examples of the functions that will live here.

jayantk (commented):
I like the idea of moving the code somewhere where it can be shared. I'm not sure whether it makes sense in this crate or not though, so I will let @ali-bahjati and @Reisen opine on that.

If they don't like this approach, I think another option would be to make a separate crate like `pythnet-sdk-test-utils` (?)

ali-bahjati:
I definitely think this is a good change. Here are my thoughts:

I suggest changing the feature name and module to `test-utils` and keep functions that generate structs here and reuse them in rust ones and have a submodule called `gen` for fixture generation.

Then as a structural refactor I think it's better to rename the pythnet-sdk to accumulators-sdk and we can then move it to `libs/accumulators-sdk/rust` (from the repo root).

ali-bahjati (approved):
It seems to have the near changes included. Perhaps we can remove them?

Apart from that it looks good to me as a good move and we can later do more structural changes (which I mentioned in the comments).

target_chains/near/receiver/tests/workspaces.rs
```diff
@@ -1197,8 +1100,8 @@ async fn test_borsh_field_cmopat() {
     assert_eq!(
         decoded_price,
         PriceTester {
-            price:          i64::MAX.into(),
-            conf:           u64::MAX.into(),
+            price:          i64::MAX,
+            conf:           u64::MAX,
```
guibescos, line 1104: Clippy did this

PR #1244, opened 2024-01-25
Fix contract manager store saving/loading procedure

m30m:
yaml file names were based on contract types but they were using the old format entropy contracts were not serialized


contract_manager/store/contracts/EvmEntropyContracts.yaml
```diff
@@ -0,0 +1,3 @@
+- chain: base_goerli
```
0xfirefist, line 1: Did you deploy entropy to base goerli?

0xfirefist (approved):

PR #1248, opened 2024-01-25
[fortuna] add uri to chain configs

0xfirefist:


fortuna/src/config.rs
```diff
@@ -134,6 +134,10 @@ pub struct EthereumConfig {
     /// How many blocks to wait before revealing the random number.
     pub reveal_delay_blocks: BlockNumber,
 
+    /// The provider uri to register on chain.
+    #[serde(default)]
+    pub provider_uri: String,
```
jayantk, line 139: two comments about this:
1. it's unfortunate that we will have to repeat the base URL of the service multiple times in the config. We could auto-derive these URIs from the base URL of the service, and that seems ok because this service itself defines the convention for how each chain's URLs is derived from the base.
2. It doesn't make sense for this argument to live in the chain config. Blockchains are not equal to providers. The other provider arguments (like the secret key) are specified elsewhere. 

I think a better approach here would be to make a provider config file that was used specifically by your script to keep on-chain provider info in sync. That config could also hold the fee parameters, for example. In that approach, you wouldn't change register provider or setup provider -- those commands could still be called from the command line as they are right now.

fortuna/src/command/register_provider.rs
```diff
@@ -59,7 +59,7 @@ pub async fn register_provider(opts: &RegisterProviderOptions) -> Result<()> {
             commitment,
             bincode::serialize(&commitment_metadata)?.into(),
             commitment_length,
-            bincode::serialize(&opts.uri)?.into(),
+            bincode::serialize(&chain_config.provider_uri)?.into(),
```
jayantk, line 62: This seems wrong to me. The chain config as it stands is a factory that lets you instantiate an Entropy contract on a blockchain. I think you want all of the provider information being registered to be explicitly passed in to this function, not implied from the chain config.

jayantk (commented):
I get why you want to put this in a file, but I don't think it belongs in the same file as the chain config. See inline comments

PR #1252, opened 2024-01-26
add tests for evm execute

0xfirefist:


m30m (approved):
It would be great if you can add some tests for `u64be` and `u256be`

PR #1254, opened 2024-01-29
feat(contract_manager): script for EVM feeds contract deployment

ali-bahjati:
This change adds a script to replace truffle deployment scripts. This gives us more control and flexibility over the deployment process.


contract_manager/src/contracts/evm.ts
```diff
@@ -404,7 +404,7 @@ export class EvmEntropyContract extends Storable {
     return {
       chain: this.chain.getId(),
       address: this.address,
-      type: EvmPriceFeedContract.type,
+      type: EvmEntropyContract.type,
```
ali-bahjati, line 407: it was a bug in json serialization of entropy contracts

contract_manager/src/contracts/evm.ts
```diff
@@ -598,7 +598,7 @@ export class EvmPriceFeedContract extends PriceFeedContract {
       pythContract.methods.updatePriceFeeds(priceFeedUpdateData);
     const gasEstiamte = await transactionObject.estimateGas({
       from: address,
-      gas: 100000000,
+      gas: 15000000,
```
ali-bahjati, line 601: the previous number was too high for Hedera. @m30m can we get away from this parameter entirely?

contract_manager/scripts/upgrade_evm_pricefeed_contracts.ts
```diff
@@ -44,18 +44,18 @@ const parser = yargs(hideBin(process.argv))
     },
   });
 
-async function run_if_not_cached(
-  cache_key: string,
+async function runIfNotCached(
```
ali-bahjati, line 47: format the naming style

contract_manager/scripts/deploy_evm_pricefeed_contracts.ts
```diff
@@ -0,0 +1,298 @@
...
+  );
+
+  const wormholeEvmContract = new WormholeEvmContract(
+    chain,
+    wormholeReceiverAddr
+  );
+
+  if (config.type === "stable") {
+    console.log(`Syncing mainnet guardian sets for ${chain.getId()}...`);
+    // TODO: Add a way to pass gas configs to this
```
ali-bahjati, line 195: Since this is a generic method I can't easily add the gas options to it. Something to refactor later in the future.

contract_manager/src/contracts/evm.ts
```diff
@@ -598,7 +598,7 @@ export class EvmPriceFeedContract extends PriceFeedContract {
       pythContract.methods.updatePriceFeeds(priceFeedUpdateData);
     const gasEstiamte = await transactionObject.estimateGas({
```
0xfirefist, line 599: gasEstimate

contract_manager/scripts/deploy_evm_pricefeed_contracts.ts
```diff
@@ -0,0 +1,298 @@
...
+    },
+  });
+
+async function deployIfNotCached(
+  chain: EvmChain,
+  config: DeploymentConfig,
+  artifactName: string,
+  deployArgs: any[] // eslint-disable-line  @typescript-eslint/no-explicit-any
+): Promise<string> {
+  const cache = existsSync(CACHE_FILE)
```
0xfirefist, line 97: Can we use runIfNotCached here?

contract_manager/scripts/deploy_evm_pricefeed_contracts.ts
```diff
@@ -0,0 +1,298 @@
...
+      type: "string",
+      demandOption: true,
+      desc: "Private key to use for the deployment",
+    },
+    chain: {
+      type: "array",
+      demandOption: true,
+      desc: "Chain to upload the contract on. Can be one of the evm chains available in the store",
+    },
+    type: {
```
0xfirefist, line 52: Can remove this? Not being used below.

contract_manager/src/contracts/evm.ts
m30m: I tried to include paramters only if necessary. Don't clearly remember why we needed this one but there is a high chance that removing it will break something 😓

contract_manager/scripts/deploy_evm_pricefeed_contracts.ts
ali-bahjati: Thanks for catching it. actually i wanted to use it 😅

contract_manager/scripts/deploy_evm_pricefeed_contracts.ts
ali-bahjati: Yeah it's a good idea. Let me do it in a separate PR (because this one is already very big)

m30m (approved):

PR #1255, opened 2024-01-29
[solana] Solana tests 2

guibescos:
The goal of this PR is to expand the testing coverage to `post_updates_atomic` 
For this we add a way to created signed VAAs to `pythnet_sdk/test-utils` . We also simulate the guardian set account in `setup_pyth_receiver`.


.github/workflows/ci-solana-contract.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    paths:
+      - target_chains/cosmwasm/**
+      - pythnet/pythnet_sdk/**
+  push:
+    branches:
+      - main
+    paths:
+      - target_chains/cosmwasm/**
+      - pythnet/pythnet_sdk/**
```
jayantk, line 13: should both of these sets of paths also include target_chains/solana?

pythnet/pythnet_sdk/src/test_utils/mod.rs
```diff
@@ -67,6 +78,18 @@ pub const DEFAULT_VALID_TIME_PERIOD: u64 = 180;
...
+
+pub fn create_dummy_guardians() -> Vec<SecretKey> {
+    let mut result: Vec<SecretKey> = vec![];
+    for i in 0..NUM_GUARDIANS {
+        let mut secret_key_bytes = [0u8; 32];
+        secret_key_bytes[0] = i + 1;
+        result.push(SecretKey::parse(&secret_key_bytes).unwrap());
+    }
+    result
+}
```
jayantk, line 92: the name of this method sounds like it could return a different guardian set each time. Suggest calling it `dummy_guardians` instead

pythnet/pythnet_sdk/src/test_utils/mod.rs
```diff
@@ -155,12 +178,43 @@ pub fn create_vaa_from_payload(
...
+        emitter_address,
         sequence,
         payload: <Box<RawMessage>>::from(payload.to_vec()),
         ..Default::default()
     };
-    vaa
+
+    let digest = libsecp256k1Message::parse_slice(&body.digest().unwrap().secp256k_hash).unwrap();
+
+    let signatures: Vec<(Signature, RecoveryId)> = guardians[0..NUM_SIGNATURES]
```
jayantk, line 193: maybe the number of signatures should be an argument to the function

jayantk: also do you want it to be the case that it's always the first N guardians who signed? are there possible tests where you want a random subset of guardians or something?

I would think about the range of configurability you need downstream and then try to build that in to the test methods here. 

Another similar question would be whether you need the ability to define & work with different guardian sets in the tests. In that case, you may want to make the signing guardian keys an argument to this function.

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
```diff
@@ -169,37 +87,61 @@ pub fn build_merkle_root_encoded_vaa(
...
  * Setup to test the Pyth Receiver. The return values are a tuple composed of :
  * - The program simulator, which is used to send transactions
- * - The pubkey of an encoded VAA account, which is pre-populated and can be used to test post_updates
- * - A vector of MerklePriceUpdate, corresponding to that VAA
+ * - The pubkeys of the encoded VAA accounts corresponding to the VAAs passed as argument, these accounts are prepopulated and can be used to test post_updates
  */
-pub async fn setup_pyth_receiver(
-    price_feed_messages: Vec<PriceFeedMessage>,
-) -> ProgramTestFixtures {
+pub async fn setup_pyth_receiver(vaas: Vec<Vec<u8>>) -> ProgramTestFixtures {
```
jayantk, line 127: I suggest passing the VAA accounts as `RawMessage` (or whatever the right wormhole type is) instead of `Vec<u8>` for clarity. Someone coming to this code is going to have a hard time figuring out how to populate these vectors of bytes.

jayantk (approved):

.github/workflows/ci-solana-contract.yml
guibescos: I can't believe I forgot to replace this

pythnet/pythnet_sdk/src/test_utils/mod.rs
guibescos: Ok so I switches to having random subsets of guardians

target_chains/solana/programs/pyth-solana-receiver/tests/common/mod.rs
guibescos: Done

PR #1257, opened 2024-01-29
[Fortuna] register fortuna uri on chain

0xfirefist:


fortuna/src/api/revelation.rs
```diff
@@ -29,6 +29,10 @@ use {
 ///
 /// Every blockchain supported by this service has a distinct sequence of random numbers and chain_id.
 /// Callers must pass the appropriate chain_id to ensure they fetch the correct random number.
+///
+/// We are registering the provider on chain with the following url:
+/// `base_uri/v1/chains/{chain_id}/revelations/`
```
jayantk, line 34: nit: `{base_uri}`

fortuna/src/api/revelation.rs
```diff
@@ -154,3 +158,8 @@ impl Blob {
         }
     }
 }
+
+/// It expects base_uri to not end with /
+pub fn get_register_uri(base_uri: &str, chain_id: &str) -> String {
+    format!("{0}/v1/chains/{1}/revelations", base_uri, chain_id)
```
jayantk, line 164: I think we should register `{0}/v1/chains/{1}`, which allows us to add more endpoints underneath if needed.

fortuna/src/api/revelation.rs
jayantk: Also, this comment doesn't make sense on this method. Someone looking to find out how the chain registration works isn't going to look at the code comment for a specific endpoint on the API.

fortuna/src/api/revelation.rs
jayantk: Also, I suggest using a crate to handle this path joining problem so you don't have to worry about the trailing slashes. Code snippet here from chat GPT demonstrating how to do it:

```
use url::Url;

fn main() {
    // Base URL
    let base_url = Url::parse("https://example.com").expect("Invalid base URL");

    // Path to be joined
    let path = "/some/path";

    // Join base URL with the path
    let result_url = base_url.join(path).expect("Failed to join URL and path");

    // Print the result URL
    println!("Result URL: {}", result_url);
}
```

fortuna/src/api/revelation.rs
jayantk: I think it would be better to put this code comment and the function get_register_uri in `api.rs`

jayantk (approved):
I have some minor comments on this. If you agree with my comments & address them, then feel free to merge without another review. If you disagree, let's discuss.

PR #1259, opened 2024-01-30
Add entropy error signatures for easier access and debugging

m30m:
Used the same format existing in PythErrors.sol
This website was very handy:
https://abi.hashex.org/


0xfirefist (approved):
Thank you so much ser! Even I was thinking about this

ali-bahjati:
There's also `cast sig` command that gives you the signature.

PR #1262, opened 2024-01-30
[solana] Add wrong vaa tests

guibescos:
I'm trying to exhaustively test all errors that the receiver can throw. This covers 12 errors of the 21.


target_chains/solana/programs/pyth-solana-receiver/src/error.rs
```diff
@@ -2,16 +2,9 @@ use anchor_lang::prelude::*;
 
 #[error_code]
 pub enum ReceiverError {
```
guibescos, line 4: Reordered the errors to match the order of the tests

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -324,6 +324,12 @@ fn deserialize_guardian_set_checked(
         ReceiverError::InvalidGuardianSetPda
     );
 
+    let timestamp = Clock::get().map(Into::into)?;
+    require!(
+        guardian_set.inner().is_active(&timestamp),
+        ReceiverError::GuardianSetExpired
+    );
+
```
guibescos, line 332: I forgot this check 🤦

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -143,7 +143,7 @@ pub mod pyth_solana_receiver {
             // We do not allow for non-increasing guardian signature indices.
             let index = usize::from(sig.guardian_index());
             if let Some(last_index) = last_guardian_index {
-                require!(index > last_index, ReceiverError::InvalidGuardianIndex);
+                require!(index > last_index, ReceiverError::InvalidGuardianOrder);
```
guibescos, line 146: Changing this error because the same error code was used in another location

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
ali-bahjati: On-chain benchmarks users don't use a data way in the past. if we want to be accurate we can pass the price timestamp here but i don't think it's necessary.

target_chains/solana/program_simulator/src/lib.rs
```diff
@@ -49,9 +57,13 @@ impl ProgramSimulator {
         signers: &Vec<&Keypair>,
         payer: Option<&Keypair>,
     ) -> Result<(), BanksClientError> {
+        let compute_units_ixs =
+            compute_budget::ComputeBudgetInstruction::set_compute_unit_limit(2000000);
```
jayantk, line 61: Adding the compute unit limit here may surprise callers of this method. Suggest making the limit an argument to the function so it's explicit

target_chains/solana/program_simulator/src/lib.rs
```diff
@@ -94,3 +106,10 @@ impl ProgramSimulator {
         Ok(T::deserialize(&mut &account.data[8..])?)
     }
 }
+
+pub fn into_transation_error<T: Into<anchor_lang::prelude::Error>>(error: T) -> TransactionError {
```
jayantk, line 110: transaction is spelled wrong

target_chains/solana/program_simulator/src/lib.rs
jayantk: or call the method `process_ix_with_default_compute_limit`

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
jayantk: i think this is ok for now. we haven't solved this problem on any of the other chains either.

jayantk (approved):

target_chains/solana/program_simulator/src/lib.rs
guibescos: Renamed

PR #1264, opened 2024-01-30
[contract_manager] upgrade evm executor contracts

0xfirefist:


m30m (approved):
This script is very similar to the other upgrade script. I think a lot of the logic can be shared via some utility functions. Please try to refactor this before merging.

PR #1267, opened 2024-01-30
[Solana] Load balance treasury

guibescos:
This PR implements a feature where instead of having one treasury account we extend to having `u8::MAX`.
The goal is to avoid the treasury account getting write-locked and being the bottleneck for people using the pull oracle.


jayantk (approved):
nice!

PR #1272, opened 2024-02-01
chore: remove attester and price service

ali-bahjati:
We are still keeping the attester sdk as it is used in some contracts. We can remove it when the contracts are updated.


jayantk (approved):
end of an era

guibescos (changes requested):
CI failing

guibescos (approved):

PR #1277, opened 2024-02-04
[hermes] add /v2/price_feeds endpoint

cctdaniel:
TODO: support `market_hours` when the on-chain component is done


hermes/src/network/pythnet.rs
```diff
@@ -136,11 +136,10 @@ pub async fn run(store: Arc<State>, pythnet_ws_endpoint: String) -> Result<()> {
...
         },
-        filters:        Some(vec![RpcFilterType::Memcmp(Memcmp {
-            offset:   0,
-            bytes:    MemcmpEncodedBytes::Bytes(b"PAS1".to_vec()),
-            encoding: None,
-        })]),
+        filters:        Some(vec![RpcFilterType::Memcmp(Memcmp::new(
+            0,                                           // offset
+            MemcmpEncodedBytes::Bytes(b"PAS1".to_vec()), // bytes
+        ))]),
```
cctdaniel, line 142: getting 
```
warning: use of deprecated field `solana_client::rpc_filter::Memcmp::offset`: Field will be made private in future. Please use a constructor method instead.
   --> src/network/pythnet.rs:140:13
    |
140 |             offset:   0,
    |             ^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated field `solana_client::rpc_filter::Memcmp::bytes`: Field will be made private in future. Please use a constructor method instead.
   --> src/network/pythnet.rs:141:13
    |
141 |             bytes:    MemcmpEncodedBytes::Bytes(b"PAS1".to_vec()),
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: use of deprecated field `solana_client::rpc_filter::Memcmp::encoding`: Field has no server-side effect. Specify encoding with `MemcmpEncodedBytes` variant instead. Field will be made private in future. Please use a constructor method instead.
   --> src/network/pythnet.rs:142:13
    |
142 |             encoding: None,
    |             ^^^^^^^^^^^^^^
    ```
    possibly due to upgrading solana packages hence the change here to fix the warnings

File: hermes/src/api/types.rs [hunk 2]
```diff
@@ -52,7 +53,7 @@ impl From<PriceIdInput> for PriceIdentifier {
 
 type Base64String = String;
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
```
cctdaniel, line 56: drive-by refactors

hermes/src/price_feeds.rs
```diff
@@ -0,0 +1,92 @@
+//! Price Feeds
```
ali-bahjati, line 1: I think it is more suitable to move this inside the `networks/pythnet.rs` as it is kind of the adapter to the Pythnet network.

hermes/src/api/types.rs
```diff
@@ -308,3 +309,20 @@ impl TryFrom<PriceUpdate> for PriceFeedsWithUpdateData {
         })
     }
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct PriceFeedV2 {
```
ali-bahjati, line 314: This name is a bit confusing, isn't it? I recommend `PriceMetadata`

cctdaniel: hmm we agreed on the design docs that price feeds would refer to the metadata itself and price updates would refer to the updates, maybe we can discuss more on this

hermes/src/network/pythnet.rs
```diff
@@ -284,7 +294,7 @@ pub async fn spawn(opts: RunOptions, state: Arc<State>) -> Result<()> {
         })
     };
 
-    let task_guadian_watcher = {
+    let task_guardian_watcher = {
```
cctdaniel, line 297: typo fix

hermes/src/price_feeds_metadata.rs
```diff
@@ -0,0 +1,83 @@
...
+pub trait PriceFeedMetadataProvider {
+    async fn get_price_feeds_metadata(
+        &self,
+        query: Option<String>,
+        asset_type: Option<AssetType>,
+    ) -> Result<Vec<PriceFeedMetadata>>;
+}
+
+#[async_trait::async_trait]
+impl PriceFeedMetadataProvider for crate::state::State {
```
Reisen, line 42: I see why you've added this, but the problem with this design is essentially it treats `State` as a type that implements business logic. What will happen if we follow this pattern is `State` will slowly accumulate unrelated logic into a glob of random logic. It's better to have this logic be in some other modular pattern that state is passed into and used, but not _on_ `State`.

hermes/src/price_feeds_metadata.rs
```diff
@@ -0,0 +1,83 @@
...
+    async fn get_price_feeds_metadata(
+        &self,
+        query: Option<String>,
+        asset_type: Option<AssetType>,
+    ) -> Result<Vec<PriceFeedMetadata>> {
+        let mut price_feeds_metadata = {
+            let feeds = retrieve_price_feeds_metadata(&self).await?;
+            if feeds.is_empty() {
+                // If the result is an empty Vec, fetch and store new price feeds
+                fetch_and_store_price_feeds_metadata(&self).await?
```
Reisen, line 52: This would be classed as `surprising` behaviour to the method caller, as the method implies it just fetches data. We shouldn't mutate within a getter, It's better to return an error here. I think you've done this because on first launch there won't be any state yet, but the better way to do this would be to explicitly fetch this state on launch somewhere before this code is called and rely on making sure `/ready` is aware.

hermes/src/network/pythnet.rs
```diff
@@ -317,6 +327,81 @@ pub async fn spawn(opts: RunOptions, state: Arc<State>) -> Result<()> {
...
+    let map_acct = load_mapping_account(&mapping_data)?;
+    for prod_pkey in &map_acct.products {
+        let prod_data = client.get_account_data(prod_pkey).await?;
+
+        if *prod_pkey == Pubkey::default() {
+            continue;
+        }
+        let prod_acct = load_product_account(&prod_data)?;
+
+        let attributes = prod_acct
```
Reisen, line 381: Ref: A

hermes/src/price_feeds_metadata.rs
```diff
@@ -0,0 +1,83 @@
...
+            } else {
+                feeds
+            }
+        };
+
+
+        // Filter by query if provided
```
Reisen, line 59: So this code implies we don't have good structured data to work with. It's better to implement this logic when you're first parsing into `attributes`. You can read more about this idea [here](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) but what I would suggest is to look at the code where I've commented `Ref: A` and think of this as where we validate.

Instead of reading out the strings, let's strongly type this with:

```
struct Attributes {
    symbol: String,
    asset_type: AssetType,
    ...
}
```

Note that these are not `Option` so we're requiring at read time that our products are compliant. This means later when we get to this code you can replace all the above logic with just:

```
price_feeds_metadata.retain(|feed| feed.symbol.contains(&query_str));
price_feeds_metadata.retain(|feed| feed.asset_type == asset_type);
```

TL;DR we're moving all our validation/parsing logic th the constructor so our logic can be pure logic.

hermes/src/price_feeds_metadata.rs
```diff
@@ -0,0 +1,83 @@
...
+                    symbol.to_lowercase().contains(&query_str.to_lowercase())
+                })
+            });
+        }
+
+        // Filter by asset_type if provided
+        if let Some(asset_type) = &asset_type {
+            price_feeds_metadata.retain(|feed| {
+                feed.attributes.get("asset_type").map_or(false, |type_str| {
+                    type_str.to_lowercase()
```
Reisen, line 72: It took me a minute to understand but I think if I'm reading right you're using serde_json as a trick to get the string representation of the enum?

There are couple better ways to do this:

- Implement `Display` which is the canonical Rusty way to implement stringifying a datatype.
- Use `strum` which I don't think we currently have but should probably add, which will give you `.to_string()` for free on our enums.

Reisen (changes requested):

hermes/src/network/pythnet.rs
```diff
@@ -317,6 +330,115 @@ pub async fn spawn(opts: RunOptions, state: Arc<State>) -> Result<()> {
...
+    let mapping_data = rpc_client.get_account_data(mapping_address).await?;
+    let mapping_acct = load_mapping_account(&mapping_data)?;
+
+    // Split product keys into chunks of 150 to avoid too many open files error (error trying to connect: tcp open error: Too many open files (os error 24))
+    for product_keys_chunk in mapping_acct
+        .products
+        .iter()
+        .filter(|&prod_pkey| *prod_pkey != Pubkey::default())
+        .collect::<Vec<_>>()
+        .chunks(150)
```
cctdaniel, line 392: tried with 200 but still intermittently getting `os error 24` so went with 150 to be safe

hermes/src/price_feeds_metadata.rs
cctdaniel: this makes sense, will follow up on this with a separate PR

hermes/src/state.rs
```diff
@@ -50,13 +53,24 @@ pub struct State {
...
+    pub price_feeds_metadata: RwLock<Vec<PriceFeedMetadata>>,
 }
 
 impl State {
     pub fn new(
         update_tx: Sender<AggregationEvent>,
         cache_size: u64,
         benchmarks_endpoint: Option<Url>,
+        rpc_http_endpoint: String,
+        mapping_address: Pubkey,
```
ali-bahjati, line 73: Do we really need them here?

hermes/src/state.rs
```diff
@@ -67,10 +81,27 @@ impl State {
...
+
+pub async fn store_price_feeds_metadata(
+    state: &State,
+    price_feeds_metadata: &[PriceFeedMetadata],
+) -> Result<()> {
+    let mut price_feeds_metadata_write_guard = state.price_feeds_metadata.write().await;
+    *price_feeds_metadata_write_guard = price_feeds_metadata.to_vec();
+    Ok(())
+}
+
```
ali-bahjati, line 104: I think it would be better to move them to `price_feeds_metadata` module (and potentially remove them)

hermes/src/config.rs
```diff
@@ -49,6 +49,10 @@ pub struct RunOptions {
     /// Metrics Options
     #[command(flatten)]
     pub metrics: metrics::Options,
+
+    /// Update interval for price_feeds_cache in seconds. Default is 600 seconds.
+    #[arg(long = "price-feeds-cache-update-interval", default_value_t = 600)]
+    pub price_feeds_cache_update_interval: u64,
```
ali-bahjati, line 55: I think we don't need to make this configurable but if we do, Pythnet module is a better place.

hermes/src/network/pythnet.rs
```diff
@@ -317,6 +328,115 @@ pub async fn spawn(opts: RunOptions, state: Arc<State>) -> Result<()> {
...
+                Ok(prod_data) => {
+                    let prod_acct = match load_product_account(&prod_data) {
+                        Ok(prod_acct) => prod_acct,
+                        Err(e) => {
+                            println!("Error loading product account: {}", e);
+                            continue;
+                        }
+                    };
+
+                    // Ref A
```
ali-bahjati, line 412: I think I'm missing the context, what do you mean by this comment?

ali-bahjati: Let's add a ToDo for a stricter typing here.

cctdaniel: this was added by mistake, removed

hermes/src/api/types.rs
```diff
@@ -316,3 +324,38 @@ impl TryFrom<PriceUpdate> for PriceFeedsWithUpdateData {
...
+            "{}",
+            match self {
+                // Match each variant to its string representation
+                AssetType::Crypto => "Crypto",
+                AssetType::FX => "FX",
+                AssetType::Equity => "Equity",
+                AssetType::Metals => "Metals",
+                AssetType::Rates => "Rates",
+            }
+        )
```
Reisen, line 359: Minor suggestion but if we're matching the names to the same thing as a string you can get away with:

```
write!(f, "{:?}", self)
```

hermes/src/network/pythnet.rs
```diff
@@ -317,6 +331,117 @@ pub async fn spawn(opts: RunOptions, state: Arc<State>) -> Result<()> {
...
+                // This loop with a sleep interval of 1 second allows the task to check for an exit signal at a
+                // fine-grained interval. Instead of sleeping directly for the entire `price_feeds_update_interval`,
+                // which could delay the response to an exit signal, this approach ensures the task can exit promptly
+                // if `crate::SHOULD_EXIT` is set, enhancing the responsiveness of the service to shutdown requests.
+                for _ in 0..DEFAULT_PRICE_FEEDS_CACHE_UPDATE_INTERVAL {
+                    if crate::SHOULD_EXIT.load(Ordering::Acquire) {
+                        break;
+                    }
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                }
```
Reisen, line 357: We should probably replace the exit variable with a watch channel at some point.

Reisen (approved):

PR #1279, opened 2024-02-05
[solana] Fix idl

guibescos:
The goal of this PR is just to add a useful script that adds external types (types imported from other rust crates) to the Anchor idl for the solana program.


target_chains/solana/programs/pyth-solana-receiver/src/state/price_update.rs
```diff
@@ -14,7 +14,7 @@ use {
  * If partial, at least config.minimum signatures have been verified, but in the case config.minimum_signatures changes in the future we also include the number of signatures that were checked */
 #[derive(AnchorSerialize, AnchorDeserialize, Copy, Clone, PartialEq, BorshSchema, Debug)]
 pub enum VerificationLevel {
-    Partial(u8),
+    Partial { num_signatures: u8 },
```
guibescos, line 17: This becomes a struct enum because Anchor is bad at generating idl for tuple enums

PR #1294, opened 2024-02-12
Sui updates

m30m:
Upgraded to `0.49.1` which was fully backward incompatible 😭 
Some getter functions (like `getExecutionStatusError`) were removed so replaced them with their internal logic
Upgraded typescript version when needed.
Removed batch attestation support
The typesystem got a lot stricter and had to ignore some errors for now. Otherwise, there would be lots of code for type narrowing.

Left to do:
- [ ] test out the new pusher
- [ ] change the sdk to create the new price feed if it doesn't exist instead of throwing error


ali-bahjati (approved):

ali-bahjati (approved):
Nice. I think it would be good if we use proper types instead of ignoring the any error but as that code is rarely touched and we are unsure about the ongoing support it's ok to keep it this way.

PR #1302, opened 2024-02-19
refactor(solana): separate anchor state to another crate

guibescos:
In the smart contract we are pinning anchor to 0.28.0 because of a Wormhole dependency.
I want consumers to be able to import our types in their code, but it's not great if they have to pin anchor to that version.
The compromise is moving the structs to a different crate that doesn't pin anchor and that can be imported both by the on-chain program and by consumer programs.

### DIFF
--- a/target_chains/solana/pyth_solana_receiver_state/Cargo.toml
+++ b/target_chains/solana/pyth_solana_receiver_state/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "pyth-solana-receiver-state"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["lib"]
+name = "pyth_solana_receiver_state"
+
+
+[dependencies]
+anchor-lang = ">=0.28.0"
+pythnet-sdk = { path = "../../../pythnet/pythnet_sdk"}
+solana-program = "*"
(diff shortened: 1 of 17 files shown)


target_chains/solana/pyth_solana_receiver_state/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["lib"]
+name = "pyth_solana_receiver_state"
+
+
+[dependencies]
+anchor-lang = ">=0.28.0"
```
guibescos, line 12: Here I can use a loose anchor so the consumer can pick whichever version they want

target_chains/solana/pyth_solana_receiver_state/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+use anchor_lang::declare_id;
+
+pub mod config;
+pub mod price_update;
+
+declare_id!("rec5EKMGg6MxZYaMdyBfgwp4d5rB9T1VQH5pJv5LtFJ");
```
jayantk, line 6: I forget how this declare_id thing works but I presume this definition is required here as well as in target_chains/solana/programs/pyth-solana-receiver/src/lib.rs ?

I note that the same string is declared in both places, maybe worth extracting it to a constant?

jayantk (approved):

target_chains/solana/pyth_solana_receiver_state/src/lib.rs
guibescos: Done

PR #1303, opened 2024-02-19
feat(solana) : rust sdk

guibescos:
Add some functions that people can use in their smart contract to consumer `PriceUpdateV1` accounts.
I need to add some more unit tests.

### DIFF
--- a/target_chains/solana/pyth_solana_receiver_state/src/error.rs
+++ b/target_chains/solana/pyth_solana_receiver_state/src/error.rs
@@ -1,12 +1,12 @@
-use thiserror::Error;
+use anchor_lang::error_code;
 
-#[derive(Error, Debug)]
+#[error_code]
 pub enum GetPriceError {
-    #[error("Price Too Old")]
+    #[msg("Price Too Old")]
     PriceTooOld,
-    #[error("Wrong Feed Id")]
+    #[msg("Wrong Feed Id")]
     WrongFeedId,
-    #[error("Wrong Verification Level")]
+    #[msg("Wrong Verification Level")]
     WrongVerificationLevel,
 }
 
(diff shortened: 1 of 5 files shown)


target_chains/solana/pyth_solana_receiver_state/src/error.rs
```diff
@@ -0,0 +1,22 @@
+use anchor_lang::error_code;
+
+#[error_code]
+pub enum GetPriceError {
+    #[msg("Price Too Old")]
+    PriceTooOld = 10000, // Big number to avoid conflicts with other error codes
+    #[msg("Wrong Feed Id")]
+    WrongFeedId,
+    #[msg("Wrong Verification Level")]
+    WrongVerificationLevel,
```
jayantk, line 10: nitpick: I don't like using phrases like "wrong" "incorrect" etc in error names because they're not very precise. The error here is caused by the verification level in the account not being high enough for the requester. So is there a more precise way to say that? Maybe "InsufficientVerificationLevel"?

jayantk: same with the error above actually. `MismatchedFeedId` might be more precise

target_chains/solana/pyth_solana_receiver_state/src/price_update.rs
```diff
@@ -31,10 +52,87 @@ impl PriceUpdateV1 {
...
+            maximum_age,
+            feed_id,
+            VerificationLevel::Full,
+        )
+    }
+}
+
+/**
+ * This function takes a hex string and returns a FeedId
+ * The hex string should be 66 characters long, and should start with "0x"
```
jayantk, line 118: suggest supporting both the 0x prefixed version and the non-0x prefixed. This is a common cause of confusion in many places.

jayantk (approved):

PR #1306, opened 2024-02-20
[price_pusher] Add near command

MagicGordon:


price_pusher/src/near/near.ts
```diff
@@ -0,0 +1,194 @@
...
+      args: {
+        data
+      },
+      gas: "300000000000000" as any,
+      attachedDeposit: updateFee
+    });
+  }
+
+  private getConnection(network: string, accountId: string, nodeUrl: string): Connection {
+    const content = fs.readFileSync(path.join(os.homedir(), ".near-credentials", network, accountId + ".json"));
```
ali-bahjati, line 174: This is not very good. Could you get the path from the arguments like other pushers and use that?

ali-bahjati (changes requested):
Thanks a lot for contributing to us. This code looks great. I will approve after my inline comment is addressed.

price_pusher/src/near/near.ts
MagicGordon: Okay, added the optional parameter 'private-key-path'.

MagicGordon: This path is where NEAR by default stores private keys, and users are highly likely to place their private keys here. Additionally, the official near-cli also reads private keys from this location. Therefore, the parameter for specifying the private key file path has been set as optional.

ali-bahjati (approved):

PR #1308, opened 2024-02-20
feat(price-service-sdk) implement accumulator update data deserialization

guibescos:
I needed to separate the vaa part from the merkle price updates part in the js sdk for Solana.
This other ecosystem that did something like this was Sui. 
I moved this code to `@pythnetwork/price-service-sdk`, so it can be shared between Solana and Sui.


ali-bahjati (approved):

ali-bahjati:
perhaps bump the versions too?

PR #1309, opened 2024-02-20
[entropy] Remove warning in SDK docs

jayantk:
We are in mainnet now


m30m (approved):
`package.json` version doesn't match `package-lock.json`. You should use `npm version patch` so that npm automatically changes both.

PR #1318, opened 2024-02-22
chore(contract_manager): add coredao testnet and mode addresses

0xfirefist:


m30m (approved):
fix pre-commit before merging

PR #1319, opened 2024-02-22
Express relay python searcher

anihamde:
- Reorg Js searcher sdk
- Drop python searcher sdk (w/o websocket; to be added)


express_relay/sdk/python/searcher/examples/simple_searcher.py
```diff
@@ -0,0 +1,160 @@
...
+        help="Default amount of bid for liquidation opportunities",
+    )
+    parser.add_argument(
+        "--liquidation-server-url",
+        type=str,
+        required=True,
+        help="Liquidation server endpoint to use for fetching opportunities and submitting bids",
+    )
+    parser.add_argument(
+        "--use-ws",
```
m30m, line 98: we should encourage users to do things the right way. and the right way here it to just use websockets instead of polling. So let's remove this argument (with lines 123-156) and always use websocket.

express_relay/sdk/python/searcher/searcher_utils.py
```diff
@@ -0,0 +1,139 @@
...
+                "id": str(self.ws_msg_counter),
+            }
+            await ws.send(json.dumps(json_subscribe))
+            self.ws_msg_counter += 1
+
+            while True:
+                msg = json.loads(await ws.recv())
+                status = msg.get("status")
+                if status and status != "success":
+                    raise Exception(f"Error in websocket subscription: {msg.get('result')}")
```
m30m, line 72: define a more specific Exception for the sdk and use that instead. Something like:
```py
class ClientException(Exception):
    pass
```

express_relay/sdk/python/searcher/searcher_utils.py
```diff
@@ -0,0 +1,139 @@
...
+                    ),
+                    params={"chain_id": self.chain_id},
+                )
+            ).json()
+
+        opportunities = [OpportunityParamsWithMetadata.from_dict(opp) for opp in opportunities]
+
+        return opportunities
+
+    async def ws_liquidation_opportunities(self, opportunity_handler):
```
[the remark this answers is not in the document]
m30m: I can't understand what this function does based on it's name

express_relay/sdk/python/searcher/searcher_utils.py
```diff
@@ -0,0 +1,139 @@
+from typing import TypedDict
```
m30m, line 1: useless imports

express_relay/sdk/python/searcher/examples/simple_searcher.py
```diff
@@ -0,0 +1,160 @@
...
+        opp: OpportunityParamsWithMetadata,
+    ) -> BidInfo | None:
+        """
+        Assesses whether a liquidation opportunity is worth liquidating; if so, returns a BidInfo object. Otherwise returns None.
+        This function determines whether the given opportunity deals with the specified repay and receipt tokens that the searcher wishes to transact in and whether it is profitable to execute the liquidation.
+        There are many ways to evaluate this, but the most common way is to check that the value of the amount the searcher will receive from the liquidation exceeds the value of the amount repaid.
+        Individual searchers will have their own methods to determine market impact and the profitability of conducting a liquidation. This function can be expanded to include external prices to perform this evaluation.
+        In this simple searcher, the function always (naively) returns a BidInfo object with the default bid and a valid_until timestamp.
+        Args:
+            default_bid: The default amount of bid for liquidation opportunities.
```
m30m, line 30: args do not match

express_relay/sdk/python/searcher/README.md
```diff
@@ -0,0 +1,33 @@
...
+You can use the `openapi-generator` command line tool to auto generate the types you will need for the client. To install, run
+
+```
+$ brew install openapi-generator
+```
+
+To generate the types from the Express Relay server, run
+
+```
+$ openapi-generator generate -i https://per-staging.dourolabs.app/docs/openapi.json -g python -o schema
```
m30m, line 22: limit what's generated here:
```
openapi-generator generate -i http://localhost:9000/docs/openapi.json -g python -o schema --additional-properties=generateSourceCodeOnly=true --global-property models,modelTests=false,modelDocs=false```

File: express_relay/sdk/python/searcher/searcher_utils.py [hunk 7]
```diff
@@ -0,0 +1,139 @@
...
+        if self.liquidation_server_url.startswith("https"):
+            self.ws_endpoint = f"wss{self.liquidation_server_url[5:]}/v1/ws"
+        elif self.liquidation_server_url.startswith("http"):
+            self.ws_endpoint = f"ws{self.liquidation_server_url[4:]}/v1/ws"
+        else:
+            raise ValueError("Invalid liquidation server URL")
+        self.ws_msg_counter = 0
+
+    async def get_liquidation_opportunities(self) -> list[OpportunityParamsWithMetadata]:
+        async with httpx.AsyncClient() as client:
```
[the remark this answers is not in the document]
anihamde: This can lead to some resource leakage, better to instantiate the client in the context so that it shuts down upon the end of this function call.

express_relay/sdk/python/searcher/examples/simple_searcher.py
```diff
@@ -0,0 +1,160 @@
...
+from searcher_utils import BidInfo, SearcherClient
+
+from schema.openapi_client.models.opportunity_bid import OpportunityBid
+from schema.openapi_client.models.opportunity_params_with_metadata import OpportunityParamsWithMetadata
+
+logger = logging.getLogger(__name__)
+
+VALID_UNTIL = 1_000_000_000_000
+
+class SimpleSearcher(SearcherClient):
```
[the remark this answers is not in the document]
anihamde: You could also instantiate a new SearcherClient object within SimpleSearcher. However, SimpleSearcher uses a lot of the functionality of SearcherClient, so it seems like a wrapper around the basic client functionality (interact with the server). Some of the logic is wrapped, e.g. the ws path calls a function in SearcherClient that then routes to the opportunity handler function defined in SimpleSearcher, which includes a call to submit_bid in SearcherClient. This makes it more natural to express SimpleSearcher as an extension of the client

express_relay/sdk/python/searcher/searcher_utils.py
```diff
@@ -0,0 +1,139 @@
...
+                if status and status != "success":
+                    raise Exception(f"Error in websocket subscription: {msg.get('result')}")
+
+                try:
+                    if msg.get("type") != "new_opportunity":
+                        continue
+
+                    opportunity = msg["opportunity"]
+                    opportunity = OpportunityParamsWithMetadata.from_dict(opportunity)
+                    await opportunity_handler(opportunity)
```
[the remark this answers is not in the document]
anihamde: i think this is necessary to submit a bid within the loop. what is the issue with blocking, if there's a timeout

express_relay/sdk/python/searcher/examples/simple_searcher.py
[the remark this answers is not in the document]
m30m: The fact that it uses a lot of the functionalities does not mean we should subclass it here. If we were about to extend its functionality or change some behavior, this would make sense. But now we are putting the logic of searcher (bidding, assessing opportunities) in the same place that we put the logic of connecting to the server while these are very separate concerns.

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        json_subscribe = {
+            "method": "subscribe",
+            "params": {
+                "chain_ids": chain_ids,
+            },
+        }
+        await self.send_ws_message(json_subscribe)
+
+    async def unsubscribe_chains(self, chain_ids: str):
```
m30m, line 124: wrong type

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,265 @@
...
+                continue
+
+            opportunity = msg["opportunity"]
+            opportunity = OpportunityParamsWithMetadata.from_dict(opportunity)
+            try:
+                await opportunity_callback(opportunity)
+            except Exception as e:
+                raise ExpressRelayClientException(f"Error in opportunity handler: {e}")
+
+    async def submit_opportunity(self, opportunity: OpportunityParams, timeout: int = 10) -> httpx.Response:
```
m30m, line 165: this should return the bid id

express_relay/sdk/python/express_relay_utils/express_relay_types.py
```diff
@@ -0,0 +1,400 @@
...
+        permission_key: The permission key to bid on.
+    """
+
+    amount: int
+    calldata: HexString
+    chain_id: str
+    contract: Address
+    permission_key: HexString
+
+    def to_dict(self):
```
m30m, line 91: There should be a better way to do this only by customizing the `amount` field serialization instead of specifying everything.

In general defining `to_dict` and `from_dict` seems wrong as we are defining additional methods for serialization instead of overriding the default behavior that pydantic provides using model_dump/model_validate.

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,404 @@
...
+        async for msg in self.ws:
+            msg_json = json.loads(msg)
+
+            if msg_json.get("type"):
+                if msg_json.get("type") == "new_opportunity":
+                    if opportunity_callback is not None:
+                        opportunity = Opportunity.model_validate(
+                            msg_json.get("opportunity")
+                        )
+                        if opportunity:
```
m30m, line 273: I don't think opportunity can be None here. model_validate throws an exception if invalid

express_relay/sdk/python/express_relay_utils/express_relay_types.py
```diff
@@ -0,0 +1,336 @@
...
+    creation_time: IntString
+    opportunity_id: UUIDString
+
+    @model_validator(mode="after")
+    def check_version(self):
+        if self.version not in ["v1"]:
+            warnings.warn(
+                f"Cannot handle opportunity version: {self.version}. Please upgrade your client."
+            )
+            return None
```
m30m, line 217: according to the docs, we should either return the instance or throw an exception inside these validators. So maybe this is not the best place to handle this case. I don't know what the behavior is if we return None

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,404 @@
...
+        subscribe_to_updates: bool = True,
+        **kwargs,
+    ) -> UUID:
+        """
+        Submits a bid on an opportunity to the server via websocket.
+
+        Args:
+            opportunity_bid: An object representing the bid to submit on an opportunity.
+            subscribe_to_updates: A boolean indicating whether to subscribe to the bid status updates.
+            kwargs: Keyword arguments to pass to the HTTP post request.
```
m30m, line 210: this parameter is now a bit ill defined. If `subscribe_to_updates` is True, this is useless, but we don't communicate that with the users. Anyway, I think we should reuse a single httpx client and allow the user to pass in any extra configuration required in the constructor.

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,404 @@
...
+                    "amount": opportunity_bid.amount,
+                    "executor": opportunity_bid.executor,
+                    "permission_key": opportunity_bid.permission_key,
+                    "signature": opportunity_bid.signature,
+                    "valid_until": opportunity_bid.valid_until,
+                },
+            )
+            bid_id = UUID(result.get("id"))
+        else:
+            async with httpx.AsyncClient() as client:
```
m30m, line 229: I searched a bit and defining a new client per method call doesn't seem like a good idea:
https://www.python-httpx.org/advanced/clients/
I suggest creating a single client in the constructor and use it here. Also we can accept the httpx client configuration in the constructor and set it once and for all the methods.

express_relay/sdk/python/express_relay_utils/express_relay_types.py
```diff
@@ -0,0 +1,336 @@
...
+class OpportunityParams(BaseModel):
+    """
+    Attributes:
+        params: The parameters of the opportunity.
+    """
+
+    params: Union[OpportunityParamsV1] = Field(..., discriminator="version")
+
+    def to_dict(self):
+        return self.params.to_dict()
```
m30m, line 182: I think this needs to be fixed. mypy showed some errors :D

.pre-commit-config.yaml
```diff
@@ -111,3 +111,19 @@ repos:
...
+      - id: black
+        name: black
+        entry: poetry -C express_relay/sdk/python/express_relay_utils run black express_relay/sdk/python/express_relay_utils
+        pass_filenames: false
+        language: "system"
+      - id: pyflakes
+        name: pyflakes
+        entry: poetry -C express_relay/sdk/python/express_relay_utils run pyflakes express_relay/sdk/python/express_relay_utils
+        pass_filenames: false
+        language: "system"
```
m30m, line 124: shouldn't these be `python`?

express_relay/sdk/python/README.md
```diff
@@ -0,0 +1,25 @@
+# Pyth Express Relay Python SDK
```
m30m, line 1: Remove Pyth

express_relay/sdk/python/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[tool.poetry]
+name = "express-relay-utils"
+version = "0.0.1"
+description = "Utilities for searchers and protocols to interact with the express relay protocol."
```
m30m, line 4: Express Relay

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,404 @@
...
+
+    async def get_ws_loop(self) -> asyncio.Task:
+        """
+        Returns the websocket handler loop.
+        """
+        await self.start_ws()
+
+        return self.ws_loop
+
+    async def send_ws_message(self, method: str, params: dict) -> dict:
```
m30m, line 97: I think it makes more sense for this function to accept a ClientMessage and the callers be responsible to craft the correct payloads

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,404 @@
...
+                ._replace(path="/v1/opportunities")
+                .geturl(),
+                params=params,
+                timeout=timeout_secs,
+            )
+
+        resp.raise_for_status()
+
+        opportunities = [
+            Opportunity.model_validate(opportunity) for opportunity in resp.json()
```
m30m, line 320: I think here is a better place to check the version before passing it to model_validate. Or we can catch the incompatible version errors here

.pre-commit-config.yaml
anihamde: this hook directly invokes poetry, not python, see https://sam.hooke.me/note/2023/09/poetry-pre-commit-hooks/

express_relay/sdk/python/express_relay_utils/express_relay_client.py
anihamde: yeah i wanted to avoid multiple duplicate calls to `ClientMessage.model_validate` throughout the code. but i dont have a strong take on this, so can refactor it

express_relay/sdk/python/express_relay_utils/express_relay_types.py
anihamde: yes but i think it's cleaner to handle the verification within this type. i'll add some more customized checking to throw an error but catch this while processing.

express_relay/sdk/python/express_relay_utils/express_relay_client.py
m30m: This is because the link does not assume we are doing this in multiple functions. Each time the user call this functions, a new client is instantiated which is suboptimal:
https://www.python-httpx.org/async/#opening-and-closing-clients

express_relay/sdk/python/express_relay_utils/express_relay_client.py
```diff
@@ -0,0 +1,403 @@
...
+                ._replace(path="/v1/opportunities")
+                .geturl(),
+                params=params,
+                timeout=timeout_secs,
+            )
+
+        resp.raise_for_status()
+
+        opportunities = [
+            Opportunity.process_opportunity_dict(opportunity)
```
m30m, line 318: this list should not include None values

.github/workflows/pypi-express-relay-utils.yml
```diff
@@ -0,0 +1,21 @@
+name: Upload express-relay-utils Python Package to PyPI
+
+on:
+  push:
+    tags:
+      - "v*"
```
m30m, line 6: let's make this a bit more specific. maybe python-v*

express_relay/sdk/python/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[tool.poetry]
+name = "express-relay-utils"
+version = "0.0.1"
```
m30m, line 3: let's make this 0.2.0 to match the server version

express_relay/sdk/python/mypy.ini
```diff
@@ -0,0 +1,17 @@
...
+plugins = pydantic.mypy
+
+follow_imports = silent
+warn_redundant_casts = True
+warn_unused_ignores = True
+disallow_any_generics = True
+check_untyped_defs = True
+no_implicit_reexport = True
+
+; # for strict mypy: (this is the tricky one :-))
```
m30m, line 11: I don't think we will enable this anytime soon let's just remove it

express_relay/sdk/python/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[tool.poetry]
+name = "express-relay-utils"
```
m30m, line 2: let's remove `utils` from here and the folder structure

express_relay/sdk/python/express_relay_utils/express_relay_types.py
```diff
@@ -0,0 +1,318 @@
...
+    version: str
+    creation_time: IntString
+    opportunity_id: UUIDString
+
+    supported_versions: ClassVar[list[str]] = ["v1"]
+    unsupported_version_error_text: ClassVar[str] = (
+        "Cannot handle opportunity version: {%s}. Please upgrade your client."
+    )
+
+    @model_validator(mode="after")
```
m30m, line 213: I think we should do this before the validation. If the new schema doesn't have some of these fields we will receive another exception which we will not handle

express_relay/sdk/python/express_relay_utils/express_relay_types.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Args:
+            opportunity_dict: The opportunity dictionary to convert.
+
+        Returns:
+            The opportunity as a class object.
+        """
+        try:
+            return cls.model_validate(opportunity_dict)
+        except Exception as e:
```
m30m, line 232: instead of catching a general exception we can throw something custom like UnsupportedOpportunity() and just catch that here.

m30m (approved):
Please fix this and the previous comments remaining before merge

PR #1329, opened 2024-02-29
feat(per-js-sdk): Add support for bid state and raw bid submission

m30m:


express_relay/sdk/js/src/index.ts
```diff
@@ -179,6 +219,16 @@ export class Client {
             await this.websocketOpportunityCallback(convertedOpportunity);
           }
         }
+      } else if ("type" in message && message.type === "bid_status_update") {
+        if (this.websocketBidStatusCallback !== undefined) {
+          await this.websocketBidStatusCallback(message);
+        }
+      } else if ("id" in message && message.id) {
```
anihamde, line 226: can you add some comments here for what this path is supposed to be?

express_relay/sdk/js/src/index.ts
```diff
@@ -218,33 +268,66 @@ export class Client {
...
-    return this.sendWebsocketMessage({
+    await this.sendWebsocketMessage({
       method: "subscribe",
       params: {
         chain_ids: chains,
       },
     });
   }
 
+  async submitOpportunityBidViaWebsocket(bid: OpportunityBid): Promise<BidId> {
```
anihamde, line 296: i think there's probably better clearer naming here, but that probably depends on us developing clearer terminology for these different parts of the server. rn the terms OpportunityBid vs Bid are not very clear at first inspection, maybe something like submitLiquidationAdapterBid vs. submitBid?

express_relay/sdk/js/src/index.ts
```diff
@@ -409,27 +492,67 @@ export class Client {
     };
   }
 
+  private toServerOpportunityBid(
+    bid: OpportunityBid
+  ): components["schemas"]["OpportunityBid"] {
+    return {
+      amount: bid.bid.amount.toString(),
```
anihamde, line 499: does the type generation have a way to auto convert to json format? or do like json.stringify or something? then you don't have to rewrite this whenever the type changes in versions

express_relay/sdk/js/src/index.ts
```diff
@@ -99,18 +104,55 @@ export type OpportunityBid = {
   bid: BidInfo;
 };
 
+/**
+ * Represents a raw bid on acquiring a permission key
+ */
+export type Bid = {
```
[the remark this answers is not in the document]
m30m: this is not quite the same type unfortunately. there is a `snake_case` -> `camelCase` conversion and I chose types that made more sense in a js sdk instead of the general openapi types. I'm not sure if there is an easy way to apply these changes on top of the openapi type.

express_relay/sdk/js/src/index.ts
m30m: They just [suggest](https://openapi-ts.pages.dev/advanced#embrace-snake-case-in-js) using snake_case everywhere but it's a bit too late for that. I guess anything magical here would remove the current type safety checks we have. Do you have anything in mind?

express_relay/sdk/js/src/index.ts
m30m: I agree that our terminology is a bit confusing but not sure about your suggestion. the term `adapter` has never been exposed in the api so I think it makes it more confusing here. I guess I have made a decision early on the summarise the LiquidationOpportunity term by dropping the liquidation part. LiquidationOpportunityBid is more verbose but also a bit long. I'm open to suggestions but terminology changes should be consistent across different docs and services.

anihamde: I think this depends on what we call the LiquidationAdapter server to differentiate it from the auction server. maybe we ought to refine that naming distinction, and that will make it clearer: for example, naming it "Liquidation Forum"?

express_relay/sdk/js/src/index.ts
```diff
@@ -218,33 +269,66 @@ export class Client {
...
+      },
+    });
+    if (result === null) {
+      throw new ClientError("Empty response in websocket for bid submission");
+    }
+    return result.id;
+  }
+
+  async submitBidViaWebsocket(bid: Bid): Promise<BidId> {
+    const result = await this.sendWebsocketMessage({
```
ali-bahjati, line 312: nit: i think `sendWebsocketMessage` can be renamed to something that indicated there is a response but I can't think of a good name myself.

express_relay/sdk/js/src/index.ts
ali-bahjati: nit: I think if the type is identical to server type you can consider using https://quicktype.io/ to generate types with conversion based on the JSON Schema of the API.

ali-bahjati (approved):

express_relay/sdk/js/src/examples/SimpleSearcher.ts
```diff
@@ -1,9 +1,67 @@
...
+const DAY_IN_SECONDS = 60 * 60 * 24;
+
+class SimpleSearcher {
+  constructor(
+    public client: Client,
+    public chainId: string,
+    public privateKey: string
+  ) {
+    this.client.setBidStatusHandler(this.bidStatusHandler.bind(this));
+    this.client.setOpportunityHandler(this.opportunityHandler.bind(this));
```
ali-bahjati, line 16: I still think that this is not a non-intutive API and recommend moving it to the constructor of Client.

ali-bahjati (approved):
Nice!
I left some minor inline comments.

Looking at the sdk I also have a mixed feeling about having two ways of doing things. This can also appear confusing to people (and the trade-offs are not very clear). Do we still need to maintain HTTP apis. If so, I recommend documenting the trade-offs.

express_relay/sdk/js/src/index.ts
jayantk: exposing the `viaWebsocket` to users seems like it's exposing an implementation detail of how this works. There's a functional difference between this and the post method below though, which is that you get updates on the bid if you use websocket. 

I think a better way to expose this functionality would be to have one method like `submitOpportunityBid` that has a default parameter like `subscribeToUpdates = true`, and then switches between websocket and post depending on that.

anihamde (approved):

express_relay/sdk/js/src/examples/SimpleSearcher.ts
```diff
@@ -1,9 +1,68 @@
...
+      this.bidStatusHandler.bind(this)
+    );
+  }
+
+  async bidStatusHandler(bidStatus: BidStatusUpdate) {
+    console.log(`Bid status for bid ${bidStatus.id}: ${bidStatus.status}`);
+  }
+
+  async opportunityHandler(opportunity: Opportunity) {
+    const bid = BigInt(argv.bid);
```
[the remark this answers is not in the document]
m30m: this was useful internally as a dummy test to make sure auction server ignores lower bids. I'd run two scripts with different bid params and see who wins. I removed it from the docs, but will leave it here since there is some default value and it will not complicate the users testing flow.

PR #1334, opened 2024-02-29
fix: solana audit 3

guibescos:
- allow post_update_atomic to return VerificationLevel::Full if it meets guardian quorum
- add cancel_governance_authority_transfer instruction

### DIFF
--- a/target_chains/solana/programs/pyth-solana-receiver/tests/test_post_price_update_from_vaa.rs
+++ b/target_chains/solana/programs/pyth-solana-receiver/tests/test_post_price_update_from_vaa.rs
@@ -34,6 +34,7 @@ use {
             create_accumulator_message,
             create_dummy_price_feed_message,
             create_dummy_twap_message,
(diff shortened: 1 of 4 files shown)


target_chains/solana/programs/pyth-solana-receiver/tests/test_post_price_update_from_vaa.rs
```diff
@@ -391,7 +398,7 @@ async fn test_post_price_update_from_vaa() {
     assert_eq!(price_update_account.write_authority, poster.pubkey());
     assert_eq!(
         price_update_account.verification_level,
-        VerificationLevel::Partial { num_signatures: 13 }
+        VerificationLevel::Full
```
jayantk, line 401: this is because the transaction above errored? (maybe worth leaving a comment above this saying "check that account data isn't updated when the update fails"?)

jayantk (approved):

PR #1343, opened 2024-03-07
chore(target_chains/ethereum): add new networks

ali-bahjati:
Add Mantle Sepolia, Merlin (testnet and mainnet), and Sei EVM devnet.


KemarTiti (approved):
lgtm

PR #1330, opened 2024-02-29
docs: code docs for solana sdk

guibescos:

PR #1356, opened 2024-03-13
feat: transaction builder

guibescos:
This PR addresses a comment in https://github.com/pyth-network/pyth-crosschain/pull/1330. It tries to make the sdk more ergonomic by providing a transaction builder.
You can see how to use the transaction builder in the README.


target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -39,12 +39,150 @@ import {
...
+    this.closeInstructions = [];
+    this.priceFeedIdToPriceUpdateAccount = {};
+  }
+
+  /**
+   * Add instructions to post price updates to the builder.
+   *
+   * @param priceUpdateDataArray the output of the `@pythnetwork/price-service-client`'s `PriceServiceConnection.getLatestVaas`. This is an array of verifiable price updates.
+   */
+  async withPostPriceUpdates(priceUpdateDataArray: string[]) {
```
guibescos, line 80: Should we enforce that 
`withPostPriceUpdates`, `withPriceConsumerInstructions` and `withCloseInstructions` need to be called sequentially?

ali-bahjati: I think it would be nice to expose them.

I thought there should be a neat way to do it with types. Look at [here](https://stackoverflow.com/questions/73072541/call-chain-order-dependent-builder-pattern-in-typescript-compiler-limitation) or [here](https://stackoverflow.com/questions/45291644/builder-pattern-using-typescript-interfaces) to see how you can impose order using some type markers. Feel free to ignore it if you think it gets complex.

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -39,12 +39,150 @@ import {
...
+   *    ETH_PRICE_FEED_ID,
+   * ]);
+   *
+   * const transactionBuilder = pythSolanaReceiver.newTransactionBuilder();
+   * await transactionBuilder.withPostPartiallyVerifiedPriceUpdates(priceUpdateData);
+   * await transactionBuilder.withPriceConsumerInstructions(...)
+   * ...
+   * ```
+   */
+  async withPostPartiallyVerifiedPriceUpdates(priceUpdateDataArray: string[]) {
```
ali-bahjati, line 117: very nit: normally `withX` implied that you return it and i chain them together. 
`B.withX().withY().build()`. Maybe `addX` be a name that doesn't imply it.

jayantk: yeah i agree with this. generally with builders there's always a question of "does this method mutate the internal state, or does it return a new builder". The `with` naming suggests that it returns a new builder

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
jayantk: some of constraints here aren't strictly necessary right? Like in theory I could call `withPriceConsumerInstructions` multiple times adding different sets of instructions and the code will work (?)

My inclination would be to avoid adding constraints on usage that aren't necessary for correctness -- you're just making the builder less flexible. see also the comment below about `withCloseInstructions` which I think helps resolve the one correctness constraint that exists.

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -39,12 +39,150 @@ import {
...
+      await getInstructions(this.priceFeedIdToPriceUpdateAccount)
+    );
+  }
+
+  /**
+   * Add instructions to close the encoded vaa accounts and the price update accounts created by `withPostPriceUpdates` and `withPostPartiallyVerifiedPriceUpdates` to reclaim rent.
+   */
+  withCloseInstructions() {
+    this.addInstructions(this.closeInstructions);
+  }
```
jayantk, line 178: The pattern here for closing the price accounts is confusing. I think a simpler approach would be to:

1. Pass a boolean `closeAccount` flag to `withPostPriceUpdates` (etc) above
2. Automatically append `this.closeInstructions` to the instruction list in `getVersionedTransactions` (or anywhere the builder is built)
3. Delete this method

jayantk (approved):
Generally LGTM. I left a couple suggestions on the approach here, but feel free to address whenever and merge this without another review from me.

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
guibescos: Done

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -39,12 +44,194 @@ import {
...
+  /**
+   * This method is used to retrieve the address of the price update account where the price update for a given price feed id will be posted.
+   * If multiple price updates for the same price feed id will be posted with the same builder, the address of the account corresponding to the last update to get posted will be returned.
+   * */
+  getPriceUpdateAccount(priceFeedId: string): PublicKey {
+    const priceUpdateAccount =
+      this.priceFeedIdToPriceUpdateAccount[priceFeedId];
+    if (!priceUpdateAccount) {
+      throw new Error(
+        `A price update account for the price feed ID ${priceFeedId} is being consumed before it was posted. Make sure to call addPostPriceUpdates or addPriceConsumerInstructions before calling addPriceConsumerInstructions.`
```
ali-bahjati, line 223: This comment is ambiguous

PR #1363, opened 2024-03-15
fix: release price service sdk

guibescos:


ali-bahjati (approved):
Maybe we can relax the >= dependency now? back to *?

guibescos:
> Maybe we can relax the >= dependency now? back to *?

Let's leave it like this as it's accurate

PR #1366, opened 2024-03-18
feat(fortuna-v2): implement a keeper service for entropy v2

0xfirefist:


fortuna/src/keeper.rs
```diff
@@ -0,0 +1,420 @@
...
+    tx: mpsc::Sender<BlockRange>,
+    rx_exit: watch::Receiver<bool>,
+) -> Result<()> {
+    tracing::info!(
+        "Watching blocks to handle new events for chain: {}",
+        &chain_id
+    );
+    while !*rx_exit.borrow() {
+        // for a http provider it only supports streaming
+        let provider = Provider::<Http>::try_from(&chain_eth_config.geth_rpc_addr)?;
```
m30m, line 263: why not websocket?

fortuna/src/command/run.rs
```diff
@@ -46,9 +95,162 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
+        // Permissive CORS layer to allow all origins
+        .layer(CorsLayer::permissive());
+
+    tracing::info!("Starting server on: {:?}", &socket_addr);
+    // Binds the axum's server to the configured address and port. This is a blocking call and will
+    // not return until the server is shutdown.
+    axum::Server::try_bind(&socket_addr)?
+        .serve(app.into_make_service())
+        .with_graceful_shutdown(async {
+            while !*rx_exit.borrow() {
```
m30m, line 119: await on changed

fortuna/src/command/run.rs
```diff
@@ -46,9 +95,162 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
+
+            tracing::info!("Shutting down RPC server...");
+        })
+        .await?;
+
+    Ok(())
+}
+
+
+pub async fn run_keeper_with_exit_check(
```
m30m, line 131: after looking at all the logic, it seems that we don't essentially need graceful shutdown on the keeper side. I think we can remove the exit checks in this path.

fortuna/src/config/run.rs
```diff
@@ -27,4 +31,16 @@ pub struct RunOptions {
     #[arg(long = "provider")]
     #[arg(env = "FORTUNA_PROVIDER")]
     pub provider: Address,
+
+    /// Path to a file containing a 20-byte (40 char) hex encoded Ethereum private key.
+    /// This key is required to submit transactions (such as registering with the contract).
+    #[arg(long = "private-key")]
```
m30m, line 37: maybe `keeper-private-key`?
Also mention that this key should not have any specific privilege but just used for submitting transactions.

fortuna/src/keeper.rs
0xfirefist: we don't have wss:://{links}

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,420 @@
...
+        time::{
+            self,
+            sleep,
+            Duration,
+        },
+    },
+    tracing,
+};
+
+async fn is_valid_request(
```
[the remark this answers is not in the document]
0xfirefist: yeah agree. I added this when I didn't thought of simulating a request. Removed

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    match sim_res {
+        Ok(_) => {
+            let res = contract
+                .reveal_with_callback_wrapper(
+                    event.provider_address,
+                    event.sequence_number,
+                    event.user_random_number,
+                    provider_revelation,
+                    nonce_manager.next(),
```
m30m, line 64: we should use nonce manager to send these transactions instead of manually getting the next nonce

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,348 @@
...
+    nonce_manager: Arc<NonceManagerMiddleware<Provider<Http>>>,
+    contract: Arc<SignablePythContract>,
+    gas_limit: U256,
+    chain_config: api::BlockchainState,
+) -> Result<()> {
+    tracing::info!("Starting backlog handler for chain: {}", &chain_id);
+    let backlog_blocks: u64 = 10_000;
+    let blocks_at_a_time = 100;
+
+    let mut from_block = if backlog_blocks > latest_safe_block {
```
m30m, line 112: use `saturating_sub`

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,348 @@
...
+        0
+    } else {
+        latest_safe_block - backlog_blocks
+    };
+    let last_block = latest_safe_block;
+
+    while from_block < last_block {
+        tracing::info!("Waiting for 5 seconds before processing the a lot of blocks");
+        time::sleep(Duration::from_secs(5)).await;
```
m30m, line 121: why do we need this?

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,348 @@
...
+    let blocks_at_a_time = 100;
+
+    let mut from_block = if backlog_blocks > latest_safe_block {
+        0
+    } else {
+        latest_safe_block - backlog_blocks
+    };
+    let last_block = latest_safe_block;
+
+    while from_block < last_block {
```
m30m, line 119: shouldn't this be `from_block <= last_block`?

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,348 @@
...
+                    {
+                        // TODO: some retry mechanisms here
+                        tracing::error!("Error processing event: {:?}", e);
+                    }
+                }
+
+                from_block = to_block + 1;
+
+                tracing::info!(
+                    "Backlog processed for chain: {} from block: {}",
```
m30m, line 158: add `to block` here as well

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,361 @@
...
+                                "Error while getting latest safe block for chain: {}",
+                                &chain_id
+                            );
+                            continue;
+                        }
+                    }
+                }
+            },
+            Err(e) => {
+                tracing::error!("Error while watching blocks for chain: {}", &chain_id);
```
m30m, line 251: you can merge these two

fortuna/src/keeper.rs
0xfirefist: good catch. thanks!

fortuna/src/command/run.rs
```diff
@@ -46,11 +62,70 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
+    config: Config,
+    private_key: String,
+) -> Result<()> {
+    let mut handles = Vec::new();
+    for (chain_id, chain_config) in chains {
+        let chain_eth_config = config
+            .chains
+            .get(&chain_id)
+            .expect("All chains should be present in the config file")
+            .clone();
```
jayantk, line 110: nitpick: I think you could probably set up the parsing logic that creates the hashmap such that the values of the hashmap contain enough information to spawn the keeper threads. That would make this method less error prone, because callers couldn't accidentally pass a mismatched Config and HashMap

fortuna/src/config/run.rs
```diff
@@ -27,4 +31,17 @@ pub struct RunOptions {
     #[arg(long = "provider")]
     #[arg(env = "FORTUNA_PROVIDER")]
     pub provider: Address,
+
+    /// Path to a file containing a 20-byte (40 char) hex encoded Ethereum private key.
+    /// This key is required to submit transactions for entropy callback requests.
+    /// This key should not have any specific privileges.
```
jayantk, line 37: "specific priviledges" is a bit vague. Suggest saying "This key should not be a registered provider"

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,388 @@
...
+/// How many blocks to look back for events that might be missed when starting the keeper
+const BACKLOG_RANGE: u64 = 1000;
+/// How many blocks to fetch events for in a single rpc call
+const BLOCK_BATCH_SIZE: u64 = 100;
+/// How much to wait before polling the next latest block
+const POLL_INTERVAL: Duration = Duration::from_secs(5);
+
+
+/// Get the latest safe block number for the chain. Retry internally if there is an error.
+async fn get_latest_safe_block(chain_config: &BlockchainState) -> BlockNumber {
```
jayantk, line 50: shouldn't this be called chain_state

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,388 @@
...
+            Err(e) => {
+                tracing::error!("Error while getting block number. error: {:?}", e);
+                time::sleep(RETRY_INTERVAL).await;
+            }
+        }
+    }
+}
+
+
+pub async fn run_keeper_threads(
```
jayantk, line 69: can you document this method please? this is a good spot to explain what the different keeper threads are and what they do

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,388 @@
...
+        },
+    },
+    tracing,
+};
+
+
+/// How much to wait before retrying in case of an RPC error
+const RETRY_INTERVAL: Duration = Duration::from_secs(5);
+/// How many blocks to look back for events that might be missed when starting the keeper
+const BACKLOG_RANGE: u64 = 1000;
```
jayantk, line 42: in the future, this may need to be configurable. I think some chains like arbitrum have a blocktime that's < 1 second (because every tx is a block)

fortuna/src/command/run.rs
jayantk: After reading through the code below a bit, one way you could do this is by creating a trait like `EntropyWriter` that's analogous to `EntropyReader` and then adding that to `BlockchainState`. This trait would also help you write tests for the new logic (that's why I created EntropyReader in the first place, actually).

I wouldn't worry about this now though. Just something to think about when writing code in the future.

fortuna/src/keeper.rs
```diff
@@ -0,0 +1,388 @@
...
+                }
+
+                let res = contract
+                    .reveal_with_callback(
+                        event.provider_address,
+                        event.sequence_number,
+                        event.user_random_number,
+                        provider_revelation,
+                    )
+                    .gas(gas_estimate)
```
jayantk, line 190: does this estimate need to be padded a bit to ensure the TX lands? (I really don't know the answer to this)

fortuna/src/keeper.rs
jayantk: some inline comments explaining the logic would help too, e.g., watch_blocks returns a stream of new blocks that get processed by process_new_blocks . In this particular case, it's worth explaining *why* the code works like this. Why subscribe to new blocks vs the events directly? There must be some thinking behind that. Write down the reasoning in the code for future reference.

jayantk: actually generally think all of the functions in this file could use a nice doc comment on top. The current comments are either missing or aren't descriptive enough imo.

jayantk (approved):
Nice, I think this looks pretty good on the whole. I think you probably could refactor this in the future to enable unit testing, but no need to let that block this PR. I left a couple requests for code comments, so please look at those. Feel free to merge without another review after addressing those comments.

fortuna/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name    = "fortuna"
-version = "3.3.4"
+version = "4.3.4"
```
jayantk, line 3: 4.0.0

PR #1370, opened 2024-03-18
feat: pyth pull-based push oracle

guibescos:
It is a program that CPIs into the pyth solana receiver to post updates.
This program enforces that:
-  all the updates for a given price feed get posted to the same account (a PDA of the oracle instance)
- the publish_time of the updates increases monotonically

The meat is in target_chains/solana/programs/pyth-push-oracle/src/lib.rs


target_chains/solana/programs/oracle-instance/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+#[instruction(params : PostUpdateParams, feed_id : FeedId)]
+pub struct UpdatePriceFeed<'info> {
+    #[account(mut)]
+    pub payer:                Signer<'info>,
+    pub pyth_solana_receiver: Program<'info, PythSolanaReceiver>,
+    pub encoded_vaa:          AccountInfo<'info>,
+    pub config:               AccountInfo<'info>,
+    #[account(mut)]
+    pub treasury:             AccountInfo<'info>,
+    #[account(mut, seeds = [&feed_id], bump)]
```
jayantk, line 128: I think we should make the seeds have an arbitrary `protocol_id` or something before `feed_id` so we can have multiple accounts with the same feed in them. Sharding should help isolate solana congestion to the congested app.

target_chains/solana/cli/src/main.rs
```diff
@@ -469,25 +469,32 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+    //     .data(),
+    // };
+    let feed_id = {
+        match from_slice::<byteorder::BE, Message>(merkle_price_update.message.as_ref()).unwrap() {
+            Message::PriceFeedMessage(price_feed_message) => price_feed_message.feed_id,
+            _ => panic!("Invalid message type"),
         }
-        .data(),
     };
+    let update_price_feed_instruction = oracle_instance::instruction::UpdatePriceFeed::populate(payer.pubkey(), encoded_vaa_keypair.pubkey(), feed_id, 0, merkle_price_update.clone());
```
jayantk, line 497: I think you want the CLI to have the ability to post updates to either program. May be worth adding a separate function or flag for the new program.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -440,7 +447,7 @@ fn post_price_update_from_vaa<'info>(
 
     match message {
         Message::PriceFeedMessage(price_feed_message) => {
-            price_update_account.write_authority = payer.key();
+            price_update_account.write_authority = write_authority.key();
```
jayantk, line 450: can you also save the slot when this update was written to the account here?

target_chains/solana/programs/oracle-instance/src/lib.rs
jayantk: (I realize we can deploy multiple instances of this program, one per protocol, but that will be a lot harder to manage IMO. E.g., upgrading all of those instances will be a huge hassle.)

jayantk: We could also call this shard_id or something... I'm not sure what the best name is. It probably depends a bit on how we expect to use this. E.g,. will there be 10 shards of every feed that all work the same, or will there be 1 shard per protocol with only the feeds they care about? I think the former is simpler to manage though may cause problems if multiple protocols end up sharing the same shards.

target_chains/solana/programs/oracle-instance/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            write_authority : ctx.accounts.price_feed_account.to_account_info().clone(),
+            encoded_vaa:          ctx.accounts.encoded_vaa.to_account_info().clone(),
+            config:               ctx.accounts.config.to_account_info().clone(),
+            treasury:             ctx.accounts.treasury.to_account_info().clone(),
+            price_update_account: ctx.accounts.price_feed_account.to_account_info().clone(),
+            system_program:       ctx.accounts.system_program.to_account_info().clone(),
+        };
+
+        let seeds = &[
+            feed_id.as_ref(),
```
jayantk, line 88: reminder to update the seeds here if you change the seeds per the comment below

target_chains/solana/programs/oracle-instance/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+#[error_code]
+pub enum OracleInstanceError {
+    #[msg("Updates must be monotonically increasing")]
+    UpdatesNotMonotonic,
+    #[msg("Price feed id mismatch")]
+    PriceFeedMessageMismatch
+}
+#[program]
+pub mod oracle_instance {
```
jayantk, line 67: i don't love this name, though I also can't think of a name that I really like. The best I have is "pyth-push-oracle", though that does risk being confused with the existing push oracle. Maybe "pyth-push-oracle-v2" (ugh)?

jayantk (commented):
nice this seems pretty straightforward to me. I left a few suggestions.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
ali-bahjati: I recommend doing it on the Oracle Instance.

target_chains/solana/programs/oracle-instance/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            feed_id.as_ref(),
+            &[*ctx.bumps.get("price_feed_account").unwrap()],
+        ];
+        let signer_seeds = &[&seeds[..]];
+        let cpi_context = CpiContext::new_with_signer(cpi_program, cpi_accounts, signer_seeds);
+
+        let old_timestamp = {
+            let price_feed_account_data = ctx.accounts.price_feed_account.try_borrow_data()?;
+            let price_feed_account = PriceUpdateV1::try_deserialize(&mut &price_feed_account_data[..])?;
+            price_feed_account.price_message.publish_time
```
ali-bahjati, line 97: will this be 0 the first time? just want to make sure try_deserialize doesn't fail on uninitialised account

target_chains/solana/common_test_utils/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "common-test-utils"
```
guibescos, line 2: I extracted this here to be able to use it in `pyth-solana-receiver` and `pyth-push-oracle` at the same time

target_chains/solana/programs/oracle-instance/src/lib.rs
guibescos: `shard_id`

target_chains/solana/cli/src/main.rs
guibescos: I didn't update the CLI and instead wrote tests

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: Done

target_chains/solana/programs/oracle-instance/src/lib.rs
guibescos: Done

target_chains/solana/programs/oracle-instance/src/lib.rs
guibescos: pyth-push-oracle

target_chains/solana/programs/oracle-instance/src/lib.rs
guibescos: Fixed

target_chains/solana/programs/pyth-push-oracle/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+                price_feed_account.price_message.feed_id == feed_id,
+                PushOracleError::PriceFeedMessageMismatch
+            );
+        }
+        Ok(())
+    }
+}
+
+#[derive(Accounts)]
+#[instruction(params : PostUpdateParams, shard_id : u8, feed_id : FeedId)]
```
jayantk, line 87: maybe shard_id should be a u16 so that we aren't limited to 255 protocols using push oracle on solana?

(i doubt we'll use 255 but adding 1 byte to be extra safe doesn't seem bad)

target_chains/solana/pyth_solana_receiver_sdk/src/lib.rs
```diff
@@ -1,7 +1,15 @@
...
+    },
+};
 
 pub mod config;
 pub mod error;
 pub mod price_update;
 
 declare_id!("rec5EKMGg6MxZYaMdyBfgwp4d5rB9T1VQH5pJv5LtFJ");
+
+pub const PYTH_PUSH_ORACLE_ID: Pubkey = pubkey!("F9SP6tBXw9Af7BYauo7Y2R5Es2mpv8FP5aNCXMihp6Za");
```
jayantk, line 15: This is confusing to me. You have a separate sdk.rs above for the push oracle which has some useful methods in it. It seems like there should either be one sdk for users of both push/pull, or two different ones. At the moment, it seems like a mix of both options (?)

jayantk (approved):
awesome!

target_chains/solana/pyth_solana_receiver_sdk/src/lib.rs
guibescos: - `pyth-solana-receiver-sdk` is for consuming price updates
- `pyth_solana_receiver::sdk` is for posting price updates 
- `pyth_push_oracle::sdk` is for updating price feeds

guibescos: For this program id, I need to put it here because otherwise there are some circular dependencies that break the code.

PR #1373, opened 2024-03-19
feat: separate write-authority from payer

guibescos:


ali-bahjati (approved):

ali-bahjati:
I think would be good to bump the version when we cut for the release (so the IDL show it correctly)

PR #1374, opened 2024-03-19
feat(express_relay): Sample protocol and monitor

m30m:
Left to do:
- [x] Run the checks in a loop
- [x] More documentation on what's going on
- [x] Deploy and test


express_relay/examples/easy_lend/contracts/EasyLend.sol
```diff
@@ -0,0 +1,341 @@
...
+    mapping(uint256 => Vault) _vaults;
+    address _oracle;
+
+    /**
+     * @notice EasyLend constructor - Initializes a new token vault contract with given parameters
+     *
+     * @param expressRelayAddress: address of the express relay
+     * @param oracleAddress: address of the oracle contract
+     */
+    constructor(address expressRelayAddress, address oracleAddress) {
```
anihamde, line 33: i think we should include the recent updates to allow undercollateralized vaults. it would be good for the protocol that users see in the examples to match the behavior they see on testnet

express_relay/sdk/js/src/examples/simpleSearcher.ts
```diff
@@ -1,9 +1,8 @@
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
-import { checkHex, Client } from "../index";
+import { BidStatusUpdate, checkHex, Client, Opportunity } from "../index";
```
anihamde, line 3: better to pull these directly from types since that's the original dependency?

express_relay/examples/easy_lend/README.md
```diff
@@ -0,0 +1,6 @@
+# EasyLend Protocol
+
+EasyLend is a simplified lending protocol that uses Express Relay for avoiding value leakage on liquidations.
+It uses Pyth price feeds to calculate the asset values and the liquidation thresholds.
+
+This project illustrates how to use the Express Relay SDK for contract integration and publishing opportunities.
```
anihamde, line 6: let's include some notes about what the `monitor.ts` script is, what it does, how it uses recent pyth prices to assess liquidatability, etc.

anihamde: maybe also worth mentioning some notes around how ezlend works and why it potentially allows undercollateralized vault creation (for testing reasons)

anihamde (approved):

PR #1390, opened 2024-03-26
feat(target_chains/ethereum/sdk/solidity): add convertToUint method to the sdk

aditya520:


target_chains/ethereum/sdk/solidity/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/pyth-sdk-solidity",
-  "version": "2.4.1",
+  "version": "2.5.0",
   "description": "Read prices from the Pyth oracle",
   "repository": {
     "type": "git",
```
ali-bahjati, line 6: Please add this new library to the `generate-abi` below.

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+library PythUtils {
+    /// @notice Converts a Pyth price to a uint256 with a target number of decimals
```
jayantk, line 5: can you give an example here to clarify what this means (use ethereum and 18 decimals)

jayantk: Can you also note that this function can lose precision and you need to be careful about that? For example, if the price is 0.003 and you set targetdecimals to 2, then the result will be 0.

jayantk (commented):
thank you for moving this.

Before I approve this, can you please add some tests for this logic ? If we are providing utilities for users to import, then we need to subject them to the same rigorous standards as other on-chain code we provide for them.  You can make a new test file here https://github.com/pyth-network/pyth-crosschain/tree/main/target_chains/ethereum/contracts/forge-test

target_chains/ethereum/sdk/solidity/PythUtils.sol
aditya520: Added the tests [PythTestUtils.t.sol](https://github.com/pyth-network/pyth-crosschain/pull/1390/files#diff-746fac818ac7e04dd794579164481fa11be60cb02c136906a1a216b73e02536f)

ali-bahjati (approved):
Can you please run the `npm run generate-abis` to add the ABI files as well?

ali-bahjati (approved):

target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
```diff
@@ -275,3 +276,40 @@ abstract contract PythTestUtils is Test, WormholeTestUtils, RandTestUtils {
         }
     }
 }
+
+contract PythTestUtilsTest is
```
jayantk, line 280: I think this should be called `PythUtilsTest` ?

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -0,0 +1,33 @@
...
+    /// @param price The Pyth price
+    /// @param expo The Pyth price exponent
+    /// @param targetDecimals The target number of decimals
+    /// @return The price as a uint256
+    /// @dev Function will loose precision if targetDecimals is less than the Pyth price decimals
```
jayantk, line 10: "lose" not "loose"

jayantk: may be worth explicitly noting that the method truncates any digits that cannot be represented in the target number of decimals (which is why you get 0 in the case below)

jayantk (approved):

PR #1394, opened 2024-03-26
[contract_manager] Add logic for tracking fee denominations and dollar values

jayantk:
I'm working on automating the calculation of our various KPIs. One such KPI are fees earned by Pyth. The logical (?) place to put this seems to be in contract manager given that we already have the index of all the contracts and the method to get the fees.

This PR adds the fee denomination to the contract configuration and also adds some logic for calculating the dollar value of the tokens using Hermes. It's a bit odd to have the token pricing code in the "contract manager", but that seemed like the simplest solution. Open to other ideas though.

(the answer is $815 btw)


contract_manager/src/contracts/aptos.ts
```diff
@@ -260,9 +263,13 @@ export class AptosPriceFeedContract extends PriceFeedContract {
...
   }
 
-  async getTotalFee(): Promise<bigint> {
+  async getTotalFee(): Promise<{ amount: bigint, denom?: string }> {
     const client = new CoinClient(this.chain.getClient());
-    return await client.checkBalance(this.stateId);
+    const amount = await client.checkBalance(this.stateId);
+    return {
+      amount,
+      ...(this.denom !== undefined ? {denom: this.denom} : {})
```
ali-bahjati, line 271: I think `denom: this.denom` should be correct type-wise.

contract_manager/src/token.ts
```diff
@@ -0,0 +1,71 @@
...
+  getType(): string {
+    return Token.type;
+  }
+
+  /**
+   * Get the dollar value of 1 token. Returns undefined for tokens that do
+   * not have a configured pricing method.
+   */
+  async getPrice(): Promise<number | undefined> {
+    if (this.pythId) {
```
ali-bahjati, line 34: not for now: maybe later for simplicity we can use Pyth symbol instead of pythId and get the mapping from Hermes too.

contract_manager/store/contracts/EvmPriceFeedContracts.yaml
```diff
@@ -1,30 +1,39 @@
 - chain: evmos
   address: "0x354bF866A4B006C9AF9d9e06d9364217A8616E12"
   type: EvmPriceFeedContract
+  denom: EVMOS
```
ali-bahjati, line 4: I think the denom is chain specific and not contract specific.

contract_manager/src/store.ts
```diff
@@ -143,6 +148,20 @@ export class Store {
     });
   }
 
+  loadAllTokens() {
```
m30m, line 151: the logic here is very similar to `loadAllVaults` and somewhat similar to `loadAllChains`. Possible to reduce the duplicate code.

jayantk: I don't think this is worth it -- the abstracted out code is going to be much harder to understand.

ali-bahjati (approved):

PR #1395, opened 2024-03-27
feat(entropy_contracts): don't call an address if it is an Eoa in reveal callback

0xfirefist:


m30m (approved):
Can you also update the docs to reflect this new logic? if the requester is Eoa, we will not call it back.

PR #1406, opened 2024-03-28
Chore/xc admin ghcr

guibescos:


.github/workflows/push-xc-admin-image.yml
```diff
@@ -6,6 +6,9 @@ on:
 permissions:
```
thmzlt, line 6: I think you also need `packages: write` here.

.github/workflows/push-xc-admin-frontend-image.yml
```diff
@@ -6,8 +6,11 @@ on:
 permissions:
```
thmzlt, line 6: I think you also need `packages: write` here.

thmzlt (approved):

PR #1412, opened 2024-04-02
[wip] [solidity sdk] Solidity multicall / ergonomics

jayantk:
I've been thinking about how to increase adoption amongst lending protocols. One of the obstacles seems to be that people are deploying forks of other protocols (aave, etc) and they do not want to modify the source code. I wanted to experiment with ways to enable these users to integrate Pyth more easily.

My initial thought was that we could put a multicall in front of the unmodified lending protocol's contracts. This however doesn't work because the delegatecall semantics don't quite work the way that you want. Given that, I came up with two approaches that allow for minimal code modification:

1. Make an abstract contract with a multicall function on it that the lending protocol contract inherits from. This works, but has some fallout. In particular, every function that is called from the multicall function must be `payable`, because it will receive the same msg.value as the multicall itself. 
2. Make a solidity guard that performs a pyth price update. This requires a little bit more code modification, but the end result seems nicer and more transparent. 

Both approaches are demonstrated in the snippet below. I think (2) is probably something we could tidy up and ship in the official SDK, but curious what you think.

### DIFF
--- a/target_chains/ethereum/contracts/foundry.toml
+++ b/target_chains/ethereum/contracts/foundry.toml
@@ -1,5 +1,5 @@
 [profile.default]
-solc_version = '0.8.4'
+solc_version = '0.8.21'
 optimizer = true
 optimizer_runs = 200
 src = 'contracts'
--- a/target_chains/ethereum/sdk/solidity/PythMulticall.sol
+++ b/target_chains/ethereum/sdk/solidity/PythMulticall.sol
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "./IPyth.sol";
+
+abstract contract PythMulticall {
+    function pythAddress() private returns (address pythAddress);
+
+    function updateFeedsAndCall(
+        bytes[] calldata priceUpdateData,
+        bytes calldata data
+    ) public payable returns (bytes memory result) {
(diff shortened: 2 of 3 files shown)


target_chains/ethereum/sdk/solidity/PythMulticall.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    // Approach #2: requires more code modification but seems nicer
+    modifier withPyth(bytes[] calldata pythPrices) {
+        updatePythPriceFeeds(pythPrices);
+        _;
+    }
+
+    function updatePythPriceFeeds(
+        bytes[] calldata priceUpdateData
+    ) public payable {
```
aditya520, line 44: Since the method is payable, Every method where `withPyth` is used have to be payable. 

So basically, it just reduces some number of lines.

target_chains/ethereum/contracts/foundry.toml
```diff
@@ -1,5 +1,5 @@
 [profile.default]
-solc_version = '0.8.4'
+solc_version = '0.8.21'
```
ali-bahjati, line 2: This change will modify our binary artifacts and requires a version bump for us.

jayantk: this is only needed for approach 1 fwiw

PR #1417, opened 2024-04-03
feat(contract_manager): implement upgrade evm entropy contracts script

0xfirefist:


contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -0,0 +1,74 @@
+import yargs from "yargs";
```
m30m, line 1: This is 99% similar to the other upgrade script. Please refactor as much as you can

contract_manager/src/contracts/evm.ts
```diff
@@ -541,6 +566,43 @@ export class EvmEntropyContract extends Storable {
     return web3.utils.randomHex(32);
   }
 
+  /**
```
m30m, line 569: this is copy paste from the other class. You can create another BaseEvmContract class and put these shared logics there or just have some utility functions

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +136,59 @@ async function main() {
...
+            const executorAddress =
+              instruction.governanceAction.executorAddress;
+            const callAddress = instruction.governanceAction.callAddress;
+            const calldata = instruction.governanceAction.calldata;
+
+            // currently executor is only being used by the entropy contract
+            const contract = new EvmEntropyContract(chain, callAddress);
+            const owner = await contract.getOwner();
+            console.log(
+              `Executor Address: ${executorAddress.toUpperCase()} is equal to Owner Address: ${owner
```
m30m, line 159: I would only log an error if there is a mismatch. We show the digests since council need to verify them manually but this stage can be done automatically.

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +136,59 @@ async function main() {
...
+                .replace("0x", "")
+                .toUpperCase()}`,
+              executorAddress.toUpperCase() ===
+                owner.replace("0x", "").toUpperCase()
+            );
+
+            const calldataHex = calldata.toString("hex");
+            let newImplementationAddress: string | undefined = undefined;
+            // Method signature for upgradeTo method is 3659cfe6
```
m30m, line 168: Use `web3.eth.abi.encodeFunctionSignature("upgradeTo(address)")`

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +136,59 @@ async function main() {
...
+            if (calldataHex.startsWith("3659cfe6000000000000000000000000")) {
+              newImplementationAddress = calldataHex.replace(
```
m30m, line 170: use `web3.eth.abi.decodeParameter` after removing the function signature

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +136,59 @@ async function main() {
...
+                "3659cfe6000000000000000000000000",
+                "0x"
+              );
+            }
+
+            if (newImplementationAddress === undefined) return;
```
m30m, line 176: log something and `continue` instead of `return`

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +137,71 @@ async function main() {
...
+            if (calldataHex.startsWith(methodSignature)) {
+              newImplementationAddress = web3.eth.abi.decodeParameter(
+                "address",
+                calldataHex.replace(methodSignature, "")
+              ) as unknown as string;
+            }
+
+            if (newImplementationAddress === undefined) {
+              console.log(
+                `No new implementation address found for ${chain.getId()}`
```
m30m, line 189: Some general log would be better here. Like we couldn't parse it

contract_manager/scripts/check_proposal.ts
```diff
@@ -134,6 +137,71 @@ async function main() {
...
+            }
+
+            const calldataHex = calldata.toString("hex");
+            const web3 = new Web3();
+            const methodSignature = web3.eth.abi
+              .encodeFunctionSignature("upgradeTo(address)")
+              .replace("0x", "");
+
+            let newImplementationAddress: string | undefined = undefined;
+            // Method signature for upgradeTo method is 3659cfe6
```
m30m, line 179: I don't think we need this comment anymore

m30m (approved):
some small issues, please fix before merging

PR #1429, opened 2024-04-08
perf: improve tx land rate

guibescos:
I wrote new retry/send logic for the transactions.
It's not perfect but the goal is to unblock xc-admin asap

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
@@ -400,12 +400,17 @@ const Proposal = ({
           squads.wallet.publicKey,
           squads.connection
         )
-        builder.addInstruction({ instruction, signers: [] })
-        const versionedTxs = await builder.buildLegacyTransactions({
+        builder.addInstruction({
+          instruction,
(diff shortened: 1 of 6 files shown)


governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals.tsx
```diff
@@ -400,12 +400,17 @@ const Proposal = ({
...
-        const versionedTxs = await builder.buildVersionedTransactions(
-          DEFAULT_PRIORITY_FEE_CONFIG
-        )
+        builder.addInstruction({
+          instruction,
+          signers: [],
+          computeUnits: 10000,
+        })
+        const transactions = builder.buildLegacyTransactions({
+          computeUnitPriceMicroLamports: 150000,
```
guibescos, line 409: Let's leave this as a magic number for now

ali-bahjati (approved):
Looks good. I think we can later make it cleaner perhaps?

I only have one comment, are we accounting for the computeUnitLimit instruction in tx size limit when we pack instructions to txs?

guibescos:
Yes we intentionally leave some room in every tx to add both the computeUnitLimit and computeUnitPrice

PR #1437, opened 2024-04-10
[solidity sdk] Add zerolend AggregatorV3 adapter to SDK

jayantk:


target_chains/ethereum/sdk/solidity/PythAggregatorV3.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: Apache 2
```
ali-bahjati, line 1: MIT right?

ali-bahjati (approved):
Thanks! Please address the comments before merging.

target_chains/ethereum/sdk/solidity/PythAggregatorV3.sol
jayantk: MIT lets us relicense under other licenses. I think it's better to use our standard license for uniformity.

PR #1439, opened 2024-04-10
refactor(hermes): state->price_feed_metadata downcasting

Reisen:
Continuation of #1432 and #1435

### DIFF
--- a/hermes/src/api.rs
+++ b/hermes/src/api.rs
@@ -26,15 +26,27 @@ mod rest;
 pub mod types;
 mod ws;
 
-#[derive(Clone)]
-pub struct ApiState {
-    pub state:     Arc<State>,
+pub struct ApiState<S = State> {
+    pub state:     Arc<S>,
     pub ws:        Arc<ws::WsState>,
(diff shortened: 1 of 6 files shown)


hermes/src/api.rs
```diff
@@ -26,15 +26,27 @@ mod rest;
...
+impl<S> Clone for ApiState<S> {
+    fn clone(&self) -> Self {
+        Self {
+            state:     self.state.clone(),
+            ws:        self.ws.clone(),
+            metrics:   self.metrics.clone(),
+            update_tx: self.update_tx.clone(),
+        }
+    }
+}
```
ali-bahjati, line 47: Hmmm it is strange that clone doesn't work as expected here. Can you explain more?

ali-bahjati (approved):

hermes/src/api.rs
Reisen: Rust's Clone derive is a bit basic and will automatically add `T: Clone` for each generic. I _could_ mark `State` as `Clone` as it won't actually get cloned, but that introduces the risk a clone happens elsewhere in the future so opted for this.

PR #1443, opened 2024-04-11
feat(hermes): add additional sse features

cctdaniel:
- **add allow_unordered query param**
- **add benchmarks_only query param**
- **update docs**

### DIFF
--- a/hermes/src/api/rest/v2/sse.rs
+++ b/hermes/src/api/rest/v2/sse.rs
@@ -204,6 +204,11 @@ async fn handle_aggregation_event(
         .await?;
     }
 
+    // Check if price_ids is empty after filtering and return None if it is
+    if price_ids.is_empty() {
+        return Ok(None);
+    }
+
     let price_update_data = price_feeds_with_update_data.update_data;
     let encoded_data: Vec<String> = price_update_data
         .into_iter()
--- a/hermes/src/api/rest/v2/sse.rs
+++ b/hermes/src/api/rest/v2/sse.rs
@@ -119,21 +119,10 @@ pub async fn price_stream_sse_handler(
                     )
                     .await
                     {
-                        Ok(price_update) => {
-                            // Check if there is any data to send
-                            if let Some(update) = price_update {
-                                if update.binary.data.is_empty() {
-                                    // No data to send, skip creating an event
-                                    return Ok(Event::default().comment("No data to send"));
-                                }
-                                Ok(Event::default()
-                                    .json_data(update)
-                                    .unwrap_or_else(|e| error_event(e)))
-                            } else {
-                                // No update available, possibly return a different event or handle accordingly
-                                Ok(Event::default().comment("No update available"))
-                            }
-                        }
+                        Ok(Some(update)) => Ok(Event::default()
+                            .json_data(update)
+                            .unwrap_or_else(|e| error_event(e))),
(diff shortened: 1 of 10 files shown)


hermes/src/api/sse.rs
```diff
@@ -99,16 +108,30 @@ pub async fn price_stream_sse_handler(
...
+                        params.benchmarks_only,
                     )
                     .await
                     {
-                        Ok(price_update) => Ok(Event::default().json_data(price_update).unwrap()),
+                        Ok(price_update) => {
+                            // Check if there is any data to send
+                            if price_update.binary.data.is_empty() {
+                                // No data to send, skip creating an event
+                                return Ok(Event::default().comment("No data to send"));
```
ali-bahjati, line 131: What's the behaviour of this one? It sends an empty event? Or nothing?

hermes/src/api/sse.rs
```diff
@@ -99,16 +108,30 @@ pub async fn price_stream_sse_handler(
...
+                            }
+                            Ok(Event::default().json_data(price_update).unwrap())
```
ali-bahjati, line 133: I'm a bit worried about having `unwraps`. can we add `unwrap_or_else(|e| error_event(e))`?

hermes/src/api/sse.rs
```diff
@@ -126,18 +149,51 @@ async fn handle_aggregation_event(
     mut price_ids: Vec<PriceIdentifier>,
     encoding: EncodingType,
     parsed: bool,
+    benchmarks_only: bool,
```
ali-bahjati, line 152: nit: Is it a benchmark price or benchmarks price?

hermes/src/api/sse.rs
```diff
@@ -99,16 +108,30 @@ pub async fn price_stream_sse_handler(
         async move {
             match message {
                 Ok(event) => {
+                    // Check if the event is out-of-order and if unordered updates are not allowed
+                    if let AggregationEvent::OutOfOrder { .. } = event {
+                        if !params.allow_unordered {
+                            return Ok(Event::default().comment("Out-of-order event skipped"));
+                        }
+                    }
```
ali-bahjati, line 116: I think it's cleaner if the handle_aggregation_event does this as well. We want the business logic to be all in one place. Then you can say that `handle_aggregation_event` returns `Result<Option<PriceUpdate>>` to also be more clear when it is going to skip the events.

hermes/src/api/sse.rs
cctdaniel: this will send an SSE event with a comment, no data

hermes/src/api/sse.rs
cctdaniel: agreed, will do

hermes/src/api/rest/v2/sse.rs
```diff
@@ -105,10 +114,26 @@ pub async fn price_stream_sse_handler(
...
+                        params.benchmarks_only,
+                        params.allow_unordered,
                     )
                     .await
                     {
-                        Ok(price_update) => Ok(Event::default().json_data(price_update).unwrap()),
+                        Ok(price_update) => {
+                            // Check if there is any data to send
+                            if let Some(update) = price_update {
+                                if update.binary.data.is_empty() {
```
ali-bahjati, line 125: I think it might be better if we return a None if the list of price_ids becomes empty because of benchmarks.

PR #1444, opened 2024-04-11
feat: jito pusher

guibescos:
I feel like the config could be improve (ex : jito arguments only exist in jito mode).


ali-bahjati (approved):
can you bump the version as well?

PR #1452, opened 2024-04-12
[solana push oracle] Idempotent updates

jayantk:
If someone tries to update a feed with a stale price, ignore the update instead of throwing an error.

### DIFF
--- a/target_chains/solana/programs/pyth-push-oracle/src/lib.rs
+++ b/target_chains/solana/programs/pyth-push-oracle/src/lib.rs
@@ -9,8 +9,13 @@ use {
         price_update::PriceUpdateV2,
         PYTH_PUSH_ORACLE_ID,
     },
-    pythnet_sdk::messages::{FeedId, Message},
-    pythnet_sdk::wire::from_slice
+    pythnet_sdk::{
+        messages::{
+            FeedId,
+            Message,
+        },
+        wire::from_slice,
+    },
 };
 
 pub mod sdk;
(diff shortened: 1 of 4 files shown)


target_chains/solana/programs/pyth-push-oracle/tests/test_update_price_feed.rs
```diff
@@ -161,26 +161,22 @@ async fn test_update_price_feed() {
...
+                DEFAULT_SHARD,
+                feed_id,
+                DEFAULT_TREASURY_ID,
+                merkle_price_updates[0].clone(),
+            ),
+            &vec![&poster],
+            None,
+        )
+        .await
+        .unwrap();
```
jayantk, line 179: note that the account state is checked to be feed_1_recent directly below this already. I changed the update message ^ to feed_1_old to validate that the on-chain state is unchanged.

target_chains/solana/programs/pyth-push-oracle/src/lib.rs
```diff
@@ -64,20 +74,37 @@ pub mod pyth_push_oracle {
...
+        if next_timestamp > current_timestamp {
+            pyth_solana_receiver::cpi::post_update(cpi_context, params)?;
+            {
+                let price_feed_account_data = ctx.accounts.price_feed_account.try_borrow_data()?;
+                let price_feed_account =
+                    PriceUpdateV2::try_deserialize(&mut &price_feed_account_data[..])?;
+
+                require!(
+                    price_feed_account.price_message.feed_id == feed_id,
+                    PushOracleError::PriceFeedMessageMismatch
```
guibescos, line 105: Seems inefficient to deserialize `PriceUpdateV2`

guibescos: Let me check how many CUs we spend doing this

guibescos (approved):

guibescos:
Let's merge since the logic is good and we'll think later about the CU optimizations if needed

PR #1456, opened 2024-04-16
Starknet: parse and verify Wormhole VAAs

Riateche:
- Added a bytes manipulation library that's also going to be useful for Pyth messages parsing.
- Implemented Wormhole methods for setting guardian set and parsing and verifying messages.
- Added Wormhole contract tests.
- Added tests and formatting check to CI.


target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,333 @@
...
+        if num_bytes == 0 {
+            return Result::Ok(0);
+        }
+        let value = u128_byte_reverse(self.read(num_bytes)?)
+            / one_shift_left_bytes_u128(16 - num_bytes);
+        Result::Ok(value)
+    }
+
+    /// Reads and hashes all the remaining data.
+    fn keccak256(ref self: Reader) -> Result<u256, felt252> {
```
ali-bahjati, line 219: I have a feeling that keccak doesn't belong here. Here can be just a function `read_bytes_as_le_array` or something and then the consumer should run keccak itself.

target_chains/starknet/contracts/src/reader.cairo
ali-bahjati, line 1: Would be nice to have all the methods of this function unit tested as well. I know they get tested on WH but we might not use some of them and they will be helpful for future changes too.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,322 @@
...
+        // XXX: storage doesn't work if we use Option here.
+        expiration_time: u64,
+    }
+
     #[storage]
     struct Storage {
-        name: felt252,
+        owner: ContractAddress,
+        current_guardian_set_index: u32,
+        guardian_sets: LegacyMap<u32, GuardianSet>,
```
ali-bahjati, line 78: Can you write a comment on why LegacyMap is used? (I don't have the context but the name has Legacy in it and it brings this question of why). I wonder whether we could use Felt256Dict here.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,322 @@
...
+    struct GuardianSet {
+        num_guardians: usize,
+        // XXX: storage doesn't work if we use Option here.
+        expiration_time: u64,
+    }
+
     #[storage]
     struct Storage {
-        name: felt252,
+        owner: ContractAddress,
```
ali-bahjati, line 76: So the owner here will be itself?

ali-bahjati (commented):
I looked at the reader mostly. I'll review the rest later today or tomorrow.

target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,333 @@
...
+use core::integer::u128_byte_reverse;
+use core::fmt::{Debug, Formatter};
+
+pub const EOF: felt252 = 'unexpected end of input';
+pub const UNEXPECTED_OVERFLOW: felt252 = 'unexpected overflow';
+
+/// A byte array with storage format similar to `core::ByteArray`, but
+/// suitable for reading data from it.
+#[derive(Drop, Clone, Serde)]
+pub struct ByteArray {
```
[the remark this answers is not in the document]
Riateche: AFAIK, the only way to read the data out of `core::ByteArray` is with `.at()`, only 1 byte at a time, so it would be inefficient. I created this utility so that we can slice data into larger chunks.

target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+}
+
+/// Allows to read data from a byte array.
+/// Uses big endian unless specified otherwise.
+/// All methods return `EOF` error if attempted to
+/// read more bytes than is available.
+#[derive(Drop, Clone)]
+pub struct Reader {
```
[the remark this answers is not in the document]
Riateche: I don't think it's practical to avoid Drop in cairo. It's really hard to get anything working without Drop because all arithmetic operations can diverge, so you can't even calculate anything when there is a non-Drop type in scope. The same goes for error propagation and unwraps.

target_chains/starknet/contracts/src/wormhole.cairo
Riateche: My understanding is that in theory there are Legacy (pedersen based) and NonLegacy (poseidon based) storage mapping formats, but NonLegacy mapping is not yet implemented, and there isn't any "non-legacy" map available. The book [just says to use LegacyMap](https://book.cairo-lang.org/ch14-01-contract-storage.html) without any explanation. `Felt252Dict` isn't usable here because it doesn't implement `starknet::Store`. So there is no alternative to `LegacyMap` right now.

target_chains/starknet/contracts/src/wormhole.cairo
Riateche: No, the owner address is specified in the constructor when deploying the contract. Subsequent calls to admin endpoints are only allowed from that address. This is just a placeholder auth until we implement the governance properly.

target_chains/starknet/contracts/src/reader.cairo
Riateche: I agree. I wrote a more flexible hashing implementation that will also be used for Pyth's Merkle tree proofs. Now `Reader` and `Hasher` are cleanly separated, and all byte order changes are now encapsulated in `Hasher`. This also simplified the Wormhole hashing code quite a bit.

target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,333 @@
...
+}
+
+#[generate_trait]
+pub impl ReaderImpl of ReaderTrait {
+    fn new(array: ByteArray) -> Reader {
+        Reader { array, current: 0, num_current_bytes: 0, next: Option::None }
+    }
+
+    /// Reads the specified number of bytes (up to 16) as a big endian unsigned integer.
+    fn read(ref self: Reader, num_bytes: u8) -> Result<u128, felt252> {
```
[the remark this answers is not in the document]
Riateche: Added an assert.

`num_bytes` in other places can be bigger than 16 bytes, so a new type would only be useful in 1 or 2 places I think.

target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,333 @@
+use core::option::OptionTrait;
+use core::array::ArrayTrait;
+use core::keccak::cairo_keccak;
+use core::integer::u128_byte_reverse;
+use core::fmt::{Debug, Formatter};
+
+pub const EOF: felt252 = 'unexpected end of input';
+pub const UNEXPECTED_OVERFLOW: felt252 = 'unexpected overflow';
```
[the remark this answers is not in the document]
Riateche: OK, it wasn't straightforward but I managed to get error enums working without too much loss in ergonomics. I think the new detailed enums are nicer.

target_chains/starknet/contracts/src/reader.cairo
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+}
+
+/// A byte array with storage format similar to `core::ByteArray`, but
+/// suitable for reading data from it.
+#[derive(Drop, Clone, Serde)]
+pub struct ByteArray {
+    // Number of bytes stored in the last item of `self.data` (or 0 if it's empty).
+    num_last_bytes: u8,
```
Reisen, line 26: In core they use `usize` for the equivelent of `num_last_bytes`, is there any performance cost in using smaller ranged integer types? I don't see anything in the docs but curious if you know why they wouldn't also make this optimization.

target_chains/starknet/contracts/src/hash.cairo
```diff
@@ -0,0 +1,122 @@
...
+            ref self.inputs_le, last_le, self.num_last_bytes.into()
+        );
+        u256 { low: u128_byte_reverse(hash_le.high), high: u128_byte_reverse(hash_le.low), }
+    }
+}
+
+#[generate_trait]
+impl HasherPrivateImpl of HasherPrivateTrait {
+    // Adds specified number of bytes to the buffer.
+    fn push_num_bytes(ref self: Hasher, value: u64, num_bytes: u8) {
```
Reisen, line 98: Unless I'm mistaken, this function (and other functions below using similar logic) will push the whole value if num_bytes is 0?

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,370 @@
...
+            let mut result = Result::Ok(());
+            loop {
+                let signature = match signatures_clone.pop_front() {
+                    Option::Some(v) => { v },
+                    Option::None => { break; },
+                };
+
+                match last_index {
+                    Option::Some(last_index) => {
+                        if *(@signature).guardian_index <= last_index {
```
Reisen, line 246: What's the reason for the snapshotting and immediate desnapping in this function? Just curious for my own Cairo knowledge.

target_chains/starknet/contracts/src/reader.cairo
Riateche: All small integers are represented as `felt252`, and `usize` is the same as `u32`, so it has all the same bound checks and conversions. So I think it shouldn't make much difference in performance, but we can do some experiments later to confirm it. I just used the smallest integer type where possible because the book [recommends to choose the type based on max value](https://book.cairo-lang.org/ch02-02-data-types.html).

target_chains/starknet/contracts/src/wormhole.cairo
Riateche: It's not immediate desnapping, it should be read as `*((@signature).guardian_index)`. It forces the compiler to take a snapshot of `signature` when accessing the field. It's a workaround for an issue where the compiler thinks `signature` is moved out, despite `guardian_index` being `Copy`. Not sure if it's a bug or just a limitation of the language.

target_chains/starknet/contracts/src/hash.cairo
Riateche: The implicit assumption is that `value` will only have `num_bytes` set and the unused bytes are always zero, so if `num_bytes == 0` then `value == 0`. I've added asserts for that.

Reisen (approved):

target_chains/starknet/contracts/tests/wormhole.cairo
```diff
@@ -0,0 +1,304 @@
...
+
+// Returns a byte that's not equal to `value`.
+fn corrupted_byte(value: u8, random: usize) -> u8 {
+    let v: u16 = value.into() + 1 + (random % 255).try_into().unwrap();
+    (v % 256).try_into().unwrap()
+}
+
+// Below are actual guardian keys from
+// https://github.com/wormhole-foundation/wormhole-networks/tree/master/mainnetv2/guardianset
+fn guardian_set1() -> Array<felt252> {
```
ali-bahjati, line 198: nit: to be precise this is guardian_set2() and the rest are 3 and 4

target_chains/starknet/contracts/src/hash.cairo
```diff
@@ -0,0 +1,134 @@
...
+        self.push_to_last(value.into(), 1);
+    }
+    fn push_u16(ref self: Hasher, value: u16) {
+        self.push_num_bytes(value.into(), 2);
+    }
+    fn push_u32(ref self: Hasher, value: u32) {
+        self.push_num_bytes(value.into(), 4);
+    }
+    fn push_u64(ref self: Hasher, value: u64) {
+        self.push_num_bytes(value, 8);
```
ali-bahjati, line 38: I am unsure that these will be used in the merkle tree in the future.

target_chains/starknet/contracts/src/hash.cairo
ali-bahjati, line 1: Looking at https://github.com/starkware-libs/cairo/blob/main/corelib/src/keccak.cairo I am wondering if we can use some of their helper functions instead of handling it ourselves.

ali-bahjati (approved):
Very nice! I left some inline comments mostly for my own edification.

As we discussed elsewhere, I think the added complexity here to make sure we don't store anything is not worth it and makes the code prone to mistake.

target_chains/starknet/contracts/src/hash.cairo
```diff
@@ -0,0 +1,134 @@
...
+    }
+    fn push_u128(ref self: Hasher, value: u128) {
+        let divisor = ONE_SHIFT_64.try_into().expect(UNEXPECTED_ZERO);
+        let (high, low) = DivRem::div_rem(value, divisor);
```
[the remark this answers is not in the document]
Riateche: Yep, no bit shifts. The standard library also uses `DivRem` for this.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,377 @@
...
-    fn set_value(ref self: T, name: felt252);
+pub trait IWormhole<T> {
+    fn submit_new_guardian_set(
+        ref self: T, set_index: u32, guardians: Array<felt252>
+    ) -> Result<(), SubmitNewGuardianSetError>;
+    fn parse_and_verify_vm(self: @T, encoded_vm: ByteArray) -> Result<VM, ParseAndVerifyVmError>;
+}
+
+#[derive(Drop, Debug, Clone, Serde)]
+pub struct GuardianSignature {
```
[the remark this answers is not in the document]
Riateche: The `Storage` needs to be inside the inner module. Other structs can be anywhere. In my view, the outer module represent a public interface of the contract, and the inner module is the implementation, so I put the public structs in the former and internal ones in the latter.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,377 @@
...
+        set.expiration_time = now + 86400;
+        self.guardian_sets.write(set_index, set);
     }
 
     #[abi(embed_v0)]
-    impl HelloImpl of super::IWormhole<ContractState> {
-        fn get_value(self: @ContractState) -> felt252 {
-            self.name.read() + 2
+    impl WormholeImpl of IWormhole<ContractState> {
+        fn submit_new_guardian_set(
```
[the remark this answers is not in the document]
Riateche: Yes. We could also make a separate governing contract that will own the existing contract.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,377 @@
...
+            if self.owner.read() != execution_info.caller_address {
+                return Result::Err(SubmitNewGuardianSetError::AccessDenied);
+            }
+            let current_set_index = self.current_guardian_set_index.read();
+            if set_index != current_set_index + 1 {
+                return Result::Err(SubmitNewGuardianSetError::InvalidGuardianSetSequence.into());
+            }
+            store_guardian_set(ref self, set_index, @guardians)?;
+            expire_guardian_set(
+                ref self, current_set_index, execution_info.block_info.unbox().block_timestamp
```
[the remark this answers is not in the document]
Riateche: Yes but it would make another system call for `get_execution_info()` internally so I wanted to avoid that.

target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -1,29 +1,377 @@
...
+            loop {
+                let signature = match signatures_clone.pop_front() {
+                    Option::Some(v) => { v },
+                    Option::None => { break; },
+                };
+
+                match last_index {
+                    Option::Some(last_index) => {
+                        if *(@signature).guardian_index <= last_index {
+                            result = Result::Err(ParseAndVerifyVmError::InvalidSignatureOrder);
```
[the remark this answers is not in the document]
Riateche: Nope, cannot return from inside a loop and cannot use `?` :(

target_chains/starknet/contracts/src/hash.cairo
Riateche: Yeah, we'll only need `push_u8`, `push_u160` and `push_reader`, but I added the rest just for completeness.

target_chains/starknet/contracts/tests/wormhole.cairo
ali-bahjati: my bad i was wrong.

target_chains/starknet/contracts/src/reader.cairo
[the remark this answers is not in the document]
ali-bahjati: Not so important but my comment was whether it is possible to implement the functionality that you've implemented here on core::ByteArray directly.

PR #1421, opened 2024-04-05
refactor(hermes): re-organize in monorepo

Reisen:
This organizes hermes under a new `apps/` directory to kick-start the re-organization of the monorepo. This particular change updates all local references to the `hermes/` directory, but an additional PR will be needed to update [the documentation.](https://github.com/pyth-network/documentation/blob/main/pages/price-feeds/pythnet-price-feeds/hermes.mdx?plain=1#L10)

PR #1464, opened 2024-04-17
refactor(price_pusher): re-organize in monorepo

Reisen:
Continuation of refactoring PRs. See previous:

- #1421 
- #1463 

This also removes the `tsbuildinfo` file and adds it to `.gitignore` as It's a compilation artifact and shouldn't be committed.


apps/price_pusher/.gitignore
```diff
@@ -1,3 +1,4 @@
 docker-compose.yaml
 price-config.yaml
+tsconfig.buildinfo
```
guibescos, line 3: It's already in the root level .gitignore so no needed here

PR #1469, opened 2024-04-18
feat: propagate errors in send transactions

guibescos:
Before this PR this method would succeed if the transaction landed, no matter if it failed or not.
Now it throws an error if the transaction fails and propagates the error message.


target_chains/solana/sdk/js/solana_utils/src/transaction.ts
```diff
@@ -435,57 +436,64 @@ export async function sendTransactions(
...
-    try {
-      const confirmTransactionPromise = connection.confirmTransaction(
-        {
-          signature: txSignature,
-          blockhash: blockhashResult.value.blockhash,
-          lastValidBlockHeight: blockhashResult.value.lastValidBlockHeight,
-        },
-        "confirmed"
-      );
+    const confirmTransactionPromise = connection.confirmTransaction(
```
guibescos, line 439: This diff is displayed very bad by github but it's not that complicated in reality

keyvankhademi (approved):

PR #1472, opened 2024-04-19
feat(fortuna): improve logging

0xfirefist:


apps/fortuna/src/keeper.rs
```diff
@@ -188,28 +194,44 @@ pub async fn process_event(
...
-                    .await;
+                    .gas(gas_estimate);
+
+                let res = contract_call.send().await;
+
+                let pending_tx = match res {
+                    Ok(pending_tx) => pending_tx,
+                    Err(e) => match e {
+                        ContractError::ProviderError { e } => return Err(e.into()),
+                        _ => {
```
m30m, line 218: Please add some comment on what sort of other errors can happen here and why it is ok to ignore them and return Ok(())

m30m (approved):
bump the version too

PR #1474, opened 2024-04-19
sei_devnet_redeploy

aditya520:


governance/xc_admin/packages/xc_admin_common/src/chains.ts
```diff
@@ -140,6 +139,7 @@ export const RECEIVER_CHAINS = {
   linea_sepolia: 50064,
   rol_testnet: 50065,
   morph_testnet: 50066,
+  sei_evm_devnet: 50067,
```
ali-bahjati, line 142: it is fine but we can reuse the previous ones if we are deprecating the previous one.

ali-bahjati (approved):

PR #1479, opened 2024-04-22
refactor(hermes): state->aggregate downcasting

Reisen:
This PR refactors the Aggregates API into the state directory. Aggregates is itself a state API but is currently not in the right location for that, this PR therefore:

- Adds the `Aggregates` abstraction
- Moves the logic under `state/aggregates.rs`
- Removes the `update_tx` from `api.rs` and makes it a part of the now encapsulated `Aggregates` API.
- Introduces async_trait again to deal with the `Send` problem for trait Futures.
- Renames `store` to `state` for consistency.

Followup to:

- #1432 
- #1435 
- #1439


apps/hermes/src/state/aggregate.rs
```diff
@@ -177,124 +198,220 @@ const READINESS_STALENESS_THRESHOLD: Duration = Duration::from_secs(30);
...
-                    )
-                    .await?;
+#[async_trait::async_trait]
+pub trait Aggregates
+where
+    Self: Cache,
+    Self: Benchmarks,
+    Self: PriceFeedMeta,
+{
+    fn subscribe(&self) -> Receiver<AggregationEvent>;
```
ali-bahjati, line 208: clean move. thanks.

ali-bahjati (approved):

PR #1482, opened 2024-04-22
Starknet: update_price_feeds and latest_price_info

Riateche:
- Hermes message parsing and validation
- Saving and reading latest prices
- Basic test

Taking fees is not implemented yet.


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -1,29 +1,384 @@
...
+use super::util::UnwrapWithFelt252;
+
+pub use pyth::{Event, PriceFeedUpdateEvent};
+
 #[starknet::interface]
-trait IPyth<T> {
-    fn get_name(self: @T) -> felt252;
-    fn set_name(ref self: T, name: felt252);
+pub trait IPyth<T> {
+    fn latest_price_info(self: @T, price_id: u256) -> PriceInfo;
```
ali-bahjati, line 10: metacomment about this api: in some target chains we have two separate functions to fetch price and ema. I'm not sure which one is better, there is also a question of efficiency.

we also call the primitive one `unsafe` to make sure people do recency check or use `get_{ema_,}price_no_older_than`. I think would be nice to have `unsafe` and  `no_older_than` here too. (The default one with a valid period seconds was actually not a good choice so please don't add it)

target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -1,29 +1,384 @@
...
+            if reader.read_u8().map_err()? < MINIMUM_ALLOWED_MINOR_VERSION {
+                return Result::Err(UpdatePriceFeedsError::InvalidUpdateData);
+            }
+
+            let trailing_header_size = reader.read_u8().map_err()?;
+            reader.skip(trailing_header_size).map_err()?;
+
+            let update_type: Option<UpdateType> = reader.read_u8().map_err()?.try_into();
+            match update_type {
+                Option::Some(_) => {},
```
ali-bahjati, line 276: I think matching specifically here would be less error prone, match on `Option::Some(WormholeMerkle)`, and else on `_`

ali-bahjati (approved):
Very neat! I left an inline-comment about the contract interface which is important to discuss but can be modified later.

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Changed to `get_price_unsafe` and `get_ema_price_unsafe` and added `publish_time != 0` check, similar to the ethereum contract. We'll add more methods later.

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Added an explicit match here and in similar places. Unfortunately, nested patterns like `Option::Some(WormholeMerkle)` are not supported so it's just a nested match.

ali-bahjati (approved):

PR #1490, opened 2024-04-22
fix(hermes): update cache.rs path in dockerignore

ali-bahjati:
Our dockerignore ignores all the files containing cache in their name and hermes had an exception here. This change was missed in moving Hermes around.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,5 +15,4 @@
 
 .git
 
-hermes/wormhole
-!hermes/src/state/cache.rs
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -15,5 +15,4 @@
 
 .git
 
-hermes/wormhole
```
ali-bahjati, line 18: this is no longer used. the new `build.rs` writes this in the target directory.

PR #1494, opened 2024-04-23
feat(contract_manager): latency script for entropy v2

0xfirefist:


contract_manager/src/contracts/evm.ts
```diff
@@ -589,6 +589,25 @@ export class EvmEntropyContract extends Storable {
     return web3.utils.randomHex(32);
   }
 
+  async requestRandomnessWithCallback(
```
m30m, line 592: This is almost a duplicate of `requestRandomness` function. Please refactor.

contract_manager/scripts/latency_entropy_with_callback.ts
```diff
@@ -0,0 +1,86 @@
...
+  .options({
+    contract: {
+      type: "string",
+      demandOption: true,
+      desc: "Contract to test latency for",
+    },
+    "private-key": COMMON_DEPLOY_OPTIONS["private-key"],
+  });
+
+async function main() {
```
m30m, line 22: There are a lot of common parts between this script and `latency_entropy.ts`. Maybe you can make the request part a standalone function?

0xfirefist: So, the request part (userRandomNumber, providerInfo) in latency_entropy is being used in the reveal part too. 
For latency_with_callback this isn't the case. It is all closely linked

contract_manager/scripts/latency_entropy_with_callback.ts
```diff
@@ -0,0 +1,86 @@
...
+import { hideBin } from "yargs/helpers";
+import { DefaultStore, toPrivateKey } from "../src";
+import { COMMON_DEPLOY_OPTIONS } from "./common";
+import Web3 from "web3";
+
+const parser = yargs(hideBin(process.argv))
+  .usage(
+    "Requests a random number from an entropy contract and measures the\n" +
+      "latency between request submission and fulfillment by the Fortuna keeper service.\n" +
+      "Usage: $0 --contract <entropy_contract_id> --private-key <private-key>"
```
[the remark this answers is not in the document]
0xfirefist: I agree, and it is easy to find the chain id instead of the contract id. made the change

contract_manager/scripts/common.ts
```diff
@@ -181,3 +181,37 @@ export function getSelectedChains(argv: {
...
+/**
+ * Finds the entropy contract for a given EVM chain.
+ * @param chain The EVM chain to find the entropy contract for.
+ * @returns The entropy contract for the given EVM chain.
+ * @throws an error if the entropy contract is not found for the given EVM chain.
+ */
+export function findEntropyContract(chain: EvmChain): EvmEntropyContract {
+  for (const contract of Object.values(DefaultStore.entropy_contracts)) {
+    if (
+      contract instanceof EvmEntropyContract &&
```
m30m, line 194: I don't think this condition is necessary

contract_manager/scripts/common.ts
```diff
@@ -181,3 +181,37 @@ export function getSelectedChains(argv: {
   }
   return selectedChains;
 }
+
+/**
+ * Finds the entropy contract for a given EVM chain.
+ * @param chain The EVM chain to find the entropy contract for.
+ * @returns The entropy contract for the given EVM chain.
+ * @throws an error if the entropy contract is not found for the given EVM chain.
```
m30m, line 189: nit: According to [jsdoc](https://jsdoc.app/tags-throws), The first expression after `@throws` should be type of error or something like `Error`. Currently syntax highlighting is a bit weird:
image

contract_manager/scripts/latency_entropy.ts
```diff
@@ -1,33 +1,32 @@
...
-      "Usage: $0 --contract <entropy_contract_id> --private-key <private-key>"
+      "Usage: $0 --chain <chain-id> --private-key <private-key>"
   )
   .options({
-    contract: {
+    chain: {
       type: "string",
       demandOption: true,
-      desc: "Contract to test latency for",
+      desc: "Chain to upload the contract on. Can be one of the evm chains available in the store",
```
m30m, line 20: wrong desc

contract_manager/scripts/latency_entropy_with_callback.ts
```diff
@@ -0,0 +1,87 @@
...
+  .usage(
+    "Requests a random number from an entropy contract and measures the\n" +
+      "latency between request submission and fulfillment by the Fortuna keeper service.\n" +
+      "Usage: $0 --chain-id <chain-id> --private-key <private-key>"
+  )
+  .options({
+    chain: {
+      type: "string",
+      demandOption: true,
+      desc: "Chain to upload the contract on. Can be one of the evm chains available in the store",
```
m30m, line 21: wrong desc

contract_manager/src/contracts/evm.ts
```diff
@@ -592,19 +592,29 @@ export class EvmEntropyContract extends Storable {
...
+      transactionObject = contract.methods.requestWithCallback(
+        provider,
+        userCommitment
+      );
+    } else {
+      transactionObject = contract.methods.request(
+        provider,
+        userCommitment,
+        false
+      ); // false for useBlockHash
```
m30m, line 615: I think the old way of doing this `const useBlockHash = false;` was more readable. You can define that const in the `else` condition alone.

contract_manager/scripts/latency_entropy_with_callback.ts
m30m: hmm, what if we make a function that returns anything necessary for later?

m30m (approved):

contract_manager/scripts/common.ts
0xfirefist: Yeah, right

PR #1496, opened 2024-04-23
doc(target_chains/starknet): add local deployment script

Riateche:
This script shows how to deploy our contracts. Added it to CI so that we don't forget to update it when we change the constructor signature.


target_chains/starknet/contracts/deploy/local_deploy
```diff
@@ -0,0 +1,150 @@
...
+scarb build
+wormhole_hash=$(starkli declare target/dev/pyth_wormhole.contract_class.json)
+
+# prefunded katana account
+owner=0x6162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03
+
+# deploying wormhole with mainnet guardians
+
+${sleep}
+# guardian set #0 is same as #1
```
ali-bahjati, line 28: iirc the guardian set #0 is only the first guardian in this list.

ali-bahjati (approved):
Nice!

PR #1501, opened 2024-04-23
feat: enable remote stake delegation

guibescos:
I want to add this to be able to delegated Pythnet PGAS stake to new validators.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_cli/src/index.ts
+++ b/governance/xc_admin/packages/xc_admin_cli/src/index.ts
@@ -159,7 +159,11 @@ multisigCommand(
       })
       .instruction();
 
-    await vault.proposeInstructions([proposalInstruction], targetCluster);
+    await vault.proposeInstructions(
+      [proposalInstruction],
+      targetCluster,
+      DEFAULT_PRIORITY_FEE_CONFIG
+    );
   });
 
 multisigCommand(
@@ -183,7 +187,11 @@ multisigCommand(
     })
     .instruction();
 
-  await vault.proposeInstructions([proposalInstruction], targetCluster);
+  await vault.proposeInstructions(
+    [proposalInstruction],
+    targetCluster,
+    DEFAULT_PRIORITY_FEE_CONFIG
+  );
 });
 
 multisigCommand(
@@ -214,7 +222,11 @@ multisigCommand(
       })
       .instruction();
 
-    await vault.proposeInstructions([proposalInstruction], targetCluster);
+    await vault.proposeInstructions(
+      [proposalInstruction],
+      targetCluster,
+      DEFAULT_PRIORITY_FEE_CONFIG
+    );
   });
 
 multisigCommand("upgrade-program", "Upgrade a program from a buffer")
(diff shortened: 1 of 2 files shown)


governance/xc_admin/packages/xc_admin_cli/src/index.ts
```diff
@@ -178,7 +187,11 @@ multisigCommand(
     })
     .instruction();
 
-  await vault.proposeInstructions([proposalInstruction], targetCluster);
+  await vault.proposeInstructions(
+    [proposalInstruction],
+    targetCluster,
+    DEFAULT_PRIORITY_FEE_CONFIG
```
guibescos, line 193: drive-by add priority fees

governance/xc_admin/packages/xc_admin_cli/src/index.ts
```diff
@@ -74,7 +75,7 @@ async function loadVaultFromOptions(options: any): Promise<MultisigVault> {
   const vault: PublicKey = new PublicKey(options.vault);
 
   const squad = SquadsMesh.endpoint(
-    getPythClusterApiUrl(multisigCluster),
+    options.rpcUrlOverride || getPythClusterApiUrl(multisigCluster),
```
keyvankhademi, line 78: It's generally better to use `??` for these conditionals.

governance/xc_admin/packages/xc_admin_common/src/multisig_transaction/SolanaStakingMultisigInstruction.ts
```diff
@@ -54,6 +55,31 @@ export class SolanaStakingMultisigInstruction implements MultisigInstruction {
         case "Authorize":
         case "AuthorizeWithSeed":
         case "Delegate":
+          const decodedDelegate = StakeInstruction.decodeDelegate(instruction);
+          return new SolanaStakingMultisigInstruction(
+            "Delegate",
+            {},
```
keyvankhademi, line 61: Now Authorize and AuthorizeWithSeed with also run the "Delegate" code since there are no breaks. Also "Deactivate" will also run the Delegate code. I think this was unintentional. We should move the Delegate above the Authorize and add break for both "Deactivate" and "Deligate"

guibescos: Great catch. Not adding `break` because it returns inside the case block

governance/xc_admin/packages/xc_admin_cli/src/index.ts
guibescos: Yes

governance/xc_admin/packages/xc_admin_cli/src/index.ts
```diff
@@ -315,7 +340,40 @@ multisigCommand(
...
+    );
+  });
+
+multisigCommand(
+  "delegate-stake",
+  "Delegate a stake account to the given vote account"
+)
+  .requiredOption("-s, --stake-account <pubkey>", "stake account to delegate")
+  .requiredOption("-d, --vote-account <pubkey>", "vote account to delegate to")
+  .action(async (options: any) => {
```
keyvankhademi, line 356: Is there anyway we can make this `option` typed?

PR #1503, opened 2024-04-24
feat(express-relay): Add simulation_failed to bid status

danimhr:


express_relay/sdk/python/express_relay/express_relay_types.py
```diff
@@ -105,13 +105,14 @@ class BidStatus(Enum):
...
     PENDING = "pending"
+    SIMULATION_FAILED = "simulation_failed"
 
 
 class BidStatusUpdate(BaseModel):
     """
     Attributes:
         id: The ID of the bid.
-        bid_status: The status enum, either SUBMITTED, LOST, or PENDING.
+        bid_status: The status enum, either SUBMITTED, LOST, PENDING, SIMULATION_FAILED.
```
m30m, line 115: this is not a scalable comment :)) we can just say the current status of the bid or something like that

m30m (approved):

PR #1525, opened 2024-04-30
feat(target_chains/starknet): add fee configuration

Riateche:


Reisen (approved):
Looks good, just mentioning it was clarified fee address isn't used yet from this PR and is intentional, was uncertain if something was implicit here.

target_chains/starknet/contracts/deploy/local_deploy
```diff
@@ -22,6 +22,8 @@ wormhole_hash=$(starkli declare target/dev/pyth_wormhole.contract_class.json)
 # prefunded katana account
 owner=0x6162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03
 
+fee_token_address=0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7
```
ali-bahjati, line 25: Can you explain it a bit more? There is no native token in Starknet that we need it?

also nitpik: let's name it `fee_contract_address` to match the on-chain name.

Riateche: I'll rename it and add a comment in https://github.com/pyth-network/pyth-crosschain/pull/1527. There is no native token on Starknet. Both ETH and STRK are implemented as ERC20 contracts. This particular contract address is hardcoded as the address used by `starkli balance` cli command. It's also the address of ETH ERC20 contract predeployed in katana (local node) and on testnet.

PR #1527, opened 2024-04-30
feat(target_chains/starknet): fee collection

Riateche:
Fees are now transfered from the caller to the Pyth contract when `update_price_feeds` is called. There is no "native" token on Starknet; both ETH and STRK are only available as ERC20 contracts.

I've also manually tested it on Starknet Sepolia (testnet): 

https://sepolia.starkscan.co/tx/0x02cb85432244bf878168123fc522a8063b5231524d6169dad68c63fa28ff27e8#events


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -66,6 +66,7 @@ pub enum UpdatePriceFeedsError {
     Wormhole: super::wormhole::ParseAndVerifyVmError,
     InvalidUpdateData,
     InvalidUpdateDataSource,
+    InsufficientFee,
```
ali-bahjati, line 69: Maybe `InsufficientFeeAllowance` can make this more clear.

ali-bahjati (approved):

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Renamed.

PR #1529, opened 2024-04-30
fix(fortuna): Fortuna improvements

m30m:
- More logging on failure of deserialization
- Log chain id if the provider registration is failing
- Fix sample config
- Fix dysfunctional rpc address from blast

### DIFF
--- a/apps/fortuna/src/command/run.rs
+++ b/apps/fortuna/src/command/run.rs
@@ -146,7 +146,7 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
             bincode::deserialize::<CommitmentMetadata>(&provider_info.commitment_metadata)
                 .map_err(|e| {
                     anyhow!(
-                        "Failed to deserialize commitment metadata for chain id {}: {}",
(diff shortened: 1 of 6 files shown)


apps/fortuna/src/command/run.rs
```diff
@@ -143,7 +143,14 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
 
         let provider_info = contract.get_provider_info(opts.provider).call().await?;
         let latest_metadata =
-            bincode::deserialize::<CommitmentMetadata>(&provider_info.commitment_metadata)?;
+            bincode::deserialize::<CommitmentMetadata>(&provider_info.commitment_metadata)
+                .map_err(|e| {
+                    anyhow!(
+                        "Failed to deserialize commitment metadata for chain id {}: {}",
```
0xfirefist, line 149: I have tried to followed the log with chain id like this - https://github.com/pyth-network/pyth-crosschain/blob/main/apps/fortuna/src/keeper.rs#L167

For consistency, can we follow that?
There might be some places where it isn't followed. I am gonna change that.

0xfirefist (approved):

PR #1535, opened 2024-05-02
refactor(hermes): state->wormhole/metrics downcasting

Reisen:
This PR refactors the Wormhole API under services, and also wraps up Metrics. The main benefit of this PR is it separates out the runtime part of Wormhole from the Wormhole interface rather than everything being in one file.

This is a followup to the other similar refactors:

- #1432 
- #1435 
- #1439 
- #1479

### DIFF
--- a/apps/hermes/src/state/wormhole.rs
+++ b/apps/hermes/src/state/wormhole.rs
@@ -1,0 +1,268 @@
+use {
+    super::{
+        aggregate::{
+            Aggregates,
+            Update,
+        },
+        State,
+    },
+    crate::network::wormhole::GuardianSet,
+    anyhow::{
+        anyhow,
+        ensure,
+        Result,
+    },
+    chrono::DateTime,
+    pythnet_sdk::{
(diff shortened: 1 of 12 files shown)


apps/hermes/src/state/wormhole.rs
```diff
@@ -0,0 +1,268 @@
...
+                .guardian_set
+                .read()
+                .await
+                .get(&vaa.guardian_set_index)
+                .ok_or_else(|| anyhow!("Unknown guardian set: {}", vaa.guardian_set_index))?,
+            vaa,
+        )?;
+
+        // Finally, store the resulting VAA in Hermes.
+        self.store_vaa(vaa.sequence, vaa_bytes).await;
```
ali-bahjati, line 163: Any reason you removed the thread::spawn here?

ali-bahjati: aha, i see, you spawn it in network (which is actually nicer)

PR #1536, opened 2024-05-02
refactor(target_chains/starknet): remove Result from reader

Riateche:
I'd like to transition from Result to panics for error handling in VAA parsing. This is just the first part, which also serves as an illustration of the changes I want to make.

I think removing Result is better here because:
- Our price update and governance methods should not return a Result:
  - Handling errors gracefully can lead to vulnerabilities. For example, if we write a new guardian set and encounter an error halfway through, we must revert the transaction. If we instead return the value to the caller, it can ignore error and commit the transaction, leaving our contract in an invalid state.
   - A user may fail to handle the error properly, causing bugs or even vulnerabilities in their own contract.
   - Working with writable methods that return Result is inconvenient. For example, `starkli invoke` does not even return the output value, so there is no easy way to check if it returned Ok or Err. However, when a transaction reverts, it's always properly reported.
- Result propagation and error conversion are currently not very ergonomic in cairo, making the code much harder to read. If we're going to unwrap the result at the end anyway, I'm not sure it's worth it.

I think it's better to only use Result in places where we actually want to propagate the error to the user (for example, in simple methods like getting the current price), but keep the code simpler when we don't need that.

Panic payload in cairo is an array of `felt252`. Most typical panics use a single `felt252` that encodes a short error message, which makes them easy to display and also easy to handle programmatically. If we want, we can reconstruct a typed error out of the panic payload in our SDK. Note also that using panics would be pretty much equivalent to how we revert on errors in our Ethereum contract.

Arguments in favor of keeping Result:
- Explicit error type in function signature allows us to declare possible failures for each function.
- We can separate between expected errors and failed internal assertions.
- The user will have an ability to handle the returned error and do something else instead of (or in addition to) reverting.

I'd love to keep using Results everywhere if it was ergonomic, but this is not the case now.

### DIFF
--- a/target_chains/starknet/contracts/src/hash.cairo
+++ b/target_chains/starknet/contracts/src/hash.cairo
@@ -63,16 +63,9 @@ pub impl HasherImpl of HasherTrait {
                 // reader.len() < 8
                 chunk_len = reader.len().try_into().expect(UNEXPECTED_OVERFLOW);
             }
-            match reader.read_num_bytes(chunk_len) {
-                Result::Ok(value) => {
-                    // chunk_len <= 8 so value must fit in u64.
-                    self.push_to_last(value.try_into().expect(UNEXPECTED_OVERFLOW), chunk_len);
-                },
-                Result::Err(err) => {
-                    result = Result::Err(err);
-                    break;
-                },
-            }
+            let value = reader.read_num_bytes(chunk_len);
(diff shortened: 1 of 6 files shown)


ali-bahjati (approved):
Nice. I agree that panics are a safer option here. Even if we don't make mistake now, the chances of making mistake in the future is high with that.

Reisen (approved):

PR #1538, opened 2024-05-02
feat(price_pusher): add more options to evm pusher

ali-bahjati:
This change makes gasLimit configurable and also adds an updateFeeMultiplier which is useful in Hedera as they have an inconsistency between the `value` passed in tx and the `value` on-chain.


aditya520 (approved):
LGTM

PR #1544, opened 2024-05-04
[price_pusher] Sui pusher debugging messages

jayantk:
Something is going wrong in the coin merging logic, and it has something to do with the contents of this error. Print out more info so we can debug better.


apps/price_pusher/src/sui/sui.ts
```diff
@@ -488,18 +488,24 @@ export class SuiPricePusher implements IPricePusher {
...
         ) {
+          /*
           Object.values((e as any).data).forEach((lockedObjects: any) => {
             lockedObjects.forEach((lockedObject: [string, number, string]) => {
               lockedAddresses.add(lockedObject[0]);
             });
           });
           // retry merging without the locked coins
           i--;
+           */
```
jayantk, line 508: something in this block above is what's throwing right now. I think it's ok to actually skip this entire batch of coins in the merging process though, so let's comment it out and see if that fixes the issue.

cctdaniel (approved):

PR #1545, opened 2024-05-04
[price_pusher] Option to ignore gas objects

jayantk:
We have a bunch of locked gas objects on Sui for some reason. These are usually returned in the error message of the failed tx, but they aren't being returned currently for some reason. Add a command line option so we can manually ignore them to get the process started again.

### DIFF
--- a/apps/price_pusher/src/index.ts
+++ b/apps/price_pusher/src/index.ts
@@ -9,6 +9,9 @@ import near from "./near/command";
 import solana from "./solana/command";
 
 yargs(hideBin(process.argv))
+  .parserConfiguration({
+    "parse-numbers": false,
+  })
   .config("config")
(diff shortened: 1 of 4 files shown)


apps/price_pusher/src/index.ts
```diff
@@ -9,6 +9,9 @@ import near from "./near/command";
 import solana from "./solana/command";
 
 yargs(hideBin(process.argv))
+  .parserConfiguration({
+    "parse-numbers": false,
+  })
```
jayantk, line 14: if you don't do this, hexadecimal strings are parsed as numbers automatically by yargs

the javascript ecosystem continues to be a dumpster fire lol

cctdaniel (approved):

PR #1547, opened 2024-05-06
feat(target_chains/starknet): wormhole governance VAA verification

Riateche:
Based on https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/Governance.sol#L190.

### DIFF
--- a/target_chains/starknet/contracts/src/wormhole.cairo
+++ b/target_chains/starknet/contracts/src/wormhole.cairo
@@ -5,7 +5,7 @@ use pyth::util::UnwrapWithFelt252;
 #[starknet::interface]
 pub trait IWormhole<T> {
     fn submit_new_guardian_set(ref self: T, set_index: u32, guardians: Array<felt252>);
-    fn parse_and_verify_vm(self: @T, encoded_vm: ByteArray) -> VM;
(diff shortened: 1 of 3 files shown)


target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -327,4 +377,22 @@ mod wormhole {
         hasher.push_u256(y);
         hasher.finalize() % ONE_SHIFT_160
     }
+
+    #[generate_trait]
+    impl PrivateImpl of PrivateImplTrait {
+        fn verify_governance_vm(self: @ContractState, vm: @VM) {
```
Reisen, line 383: This is very nitpicky but as far as I can tell from the current code, `VM` is only ever produced by parse and verify, might be nice to rename it something like `VeririedVM` because it implies that you only ever have a VM you can work with if you have passed verification which makes code like you've added now clearer in that there won't be a time where governance might have an unverified VM.

Reisen (approved):

target_chains/starknet/contracts/src/wormhole.cairo
Riateche: I agree. Renamed.

PR #1550, opened 2024-05-06
feat(target_chains/starknet): handle wormhole guardian set upgrade VAA

Riateche:
I'm going to remove `owner` and the old `submit_new_guardian_set` later.


target_chains/starknet/contracts/src/wormhole.cairo
```diff
@@ -296,8 +307,23 @@ mod wormhole {
         fn submit_new_guardian_set2(ref self: ContractState, encoded_vm: ByteArray) {
             let vm = self.parse_and_verify_vm(encoded_vm);
             self.verify_governance_vm(@vm);
+            let mut reader = ReaderImpl::new(vm.payload);
+            let header = governance::parse_header(ref reader);
+            if header.action != governance::Action::GuardianSetUpgrade {
```
ali-bahjati, line 312: Maybe add a comment that we do not care about other type of instructions? (and also we do upgrades by changing pyth pointer to wh)

Reisen (approved):

target_chains/starknet/contracts/src/wormhole.cairo
Riateche: This code just replicates the checks in Ethereum Wormhole contract [here](https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/GovernanceStructs.sol#L94). I've added the comment but I've placed it at the interface declaration instead.

PR #1563, opened 2024-05-11
feat(contract_manager): add generate set fee payload script to contract manager

cctdaniel:


contract_manager/scripts/generate_governance_set_fee_payload.ts
```diff
@@ -0,0 +1,41 @@
...
+    exponent: {
+      type: "number",
+      demandOption: true,
+      desc: "The new fee exponent to set",
+    },
+  });
+
+async function main() {
+  const argv = await parser.argv;
+  const chain = argv.chain;
```
m30m, line 28: maybe object destructuring here?

contract_manager/scripts/generate_governance_set_fee_payload.ts
```diff
@@ -0,0 +1,41 @@
...
+  const fee = argv.fee;
+  const exponent = argv.exponent;
+
+  const chain_obj: Chain = DefaultStore.chains[chain];
```
m30m, line 32: Not sure why you aded `: Chain`.
Probably we should check that this chain exists and throw some better error if it does not

m30m (approved):

PR #1572, opened 2024-05-14
chore(governance): add starknet chain id and governance actions, add PayloadByteRepr test

Riateche:
I've added new classes for upgrading and setting wormhole address, similar to how it's done for other chains. There is a bit of code duplication in these classes, but I'm not sure what would be the best way to avoid it. Addresses and contract class hashes are represented as felt252 on Starknet, so I'll represent them as BE uint256 in the payload.

The test is here just to make sure that I get the payload format right when implementing parsing on Starknet. I'll use the payloads for generating test inputs.


governance/xc_admin/packages/xc_admin_common/src/__tests__/PayloadByteRepr.test.ts
```diff
@@ -0,0 +1,107 @@
+import {
```
guibescos, line 1: I recommend instead of this test file you update GovernancePayload.test.ts 
in particular it has a fast check tests that checks randomly generated payloads, so you just need to include the new Starknet payloads there

Riateche: I can't get the GovernancePayload test to work:
```
pyth-crosschain/governance/xc_admin/packages/xc_admin_common main $ npm run test -- GovernancePayload

> xc_admin_common@0.0.0 test
> jest GovernancePayload

 FAIL  src/__tests__/GovernancePayload.test.ts
  ● Test suite failed to run

    src/propose.ts:407:13 - error TS2559: Type '70' has no properties in common with type 'ConfirmOptions'.

    407             MAX_RETRY_SEND
                    ~~~~~~~~~~~~~~

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        2.98 s
Ran all test suites matching /GovernancePayload/i.
npm ERR! Lifecycle script `test` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: xc_admin_common@0.0.0 
npm ERR!   at location: pyth-crosschain/governance/xc_admin/packages/xc_admin_common 
```
Not sure what's happening. Looks like maybe it picks up another version of `@pythnetwork/solana-utils` from somewhere?

governance/xc_admin/packages/xc_admin_common/src/chains.ts
```diff
@@ -67,6 +67,7 @@ export const RECEIVER_CHAINS = {
   parallel: 60048,
   iota: 60049,
   flow_previewnet: 60050,
+  starknet: 60051,
```
ali-bahjati, line 70: These are probably outdated now. Let's rebase it and merge it soon (so we don't use it)

governance/xc_admin/packages/xc_admin_common/src/__tests__/PayloadByteRepr.test.ts
ali-bahjati: It works for me. Can you try a clean install and then running all the builds before this command?

```
rm -rf **/tsconfig.tsbuildinfo **/node_modules
npm ci
npx lerna run build
```

governance/xc_admin/packages/xc_admin_common/src/chains.ts
Riateche: Rebased.

governance/xc_admin/packages/xc_admin_common/src/__tests__/PayloadByteRepr.test.ts
Riateche: Thanks, that helped. I've moved byte data tests to `GovernancePayload.test.ts` and extended ser/de roundtrip test to use new types. However, I've spotted an issue [here](https://github.com/pyth-network/pyth-crosschain/blob/main/governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts#L32-L44): we've assumed that all different UpgradeContract subclasses have different payload sizes, but it's already not true for Aptos and Sui, and Starknet is the same. Maybe I should join Aptos, Sui and Starknet into a single `AptosSuiStarknetUpgradeContract` class? They are already essentially the same.

ali-bahjati: Yeah i think joining them is actually nice. Another idea is to separate them based on number of bytes, 20BytesUpgradeContract and 32Bytes...

ali-bahjati (approved):

governance/xc_admin/packages/xc_admin_common/src/__tests__/PayloadByteRepr.test.ts
Riateche: United them into `UpgradeContract256Bit` and added a comment.

PR #1576, opened 2024-05-14
fix(xc_admin_frontend): inject env variable for RPC in docker build

ali-bahjati:
xc-admin-frontend needs to have some environment variables during build time to make them available on the client-side. This PR adds the most important one.


.github/workflows/push-xc-admin-frontend-image.yml
```diff
@@ -1,5 +1,6 @@
 name: xc_admin_frontend Docker Image
 on:
+  pull_request:
```
ali-bahjati, line 3: only for testing before merging it.

ayazabbas (approved):

PR #1577, opened 2024-05-14
fix(xc_admin/crank_pythnet_relayer): increase compute unit limit 

ali-bahjati:
By default the compute unit limit for a tx is 200k but this is not
enough for all of our instructions. This change adds an instruction
to the pythnet relayer to set it to 1m to make sure all instructions
can be executed without problem.

### DIFF
--- a/governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
+++ b/governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
@@ -183,7 +183,7 @@ async function run() {
             .postInstructions([
               ComputeBudgetProgram.setComputeUnitLimit({ units: 1000000 }),
             ])
-            .rpc({ skipPreflight: false });
+            .rpc({ skipPreflight: true });
         } catch (e) {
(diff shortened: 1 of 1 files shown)


guibescos (approved):

governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
```diff
@@ -177,7 +178,12 @@ async function run() {
             })
             .remainingAccounts(extraAccountMetas)
             .preInstructions(preInstructions)
-            .rpc({ skipPreflight: true });
+            // Use a high compute unit limit to avoid running out of compute units
+            // as some operations can use a lot of compute units.
+            .postInstructions([
+              ComputeBudgetProgram.setComputeUnitLimit({ units: 1000000 }),
+            ])
+            .rpc({ skipPreflight: false });
```
guibescos, line 186: I like skip preflight better because it leaves a trace in the blockchain (as a landed failed tx)

PR #1579, opened 2024-05-15
fix(fortuna): retry blocks

0xfirefist:


apps/fortuna/src/keeper.rs
```diff
@@ -627,26 +701,38 @@ pub async fn watch_blocks(
 
         let latest_safe_block = get_latest_safe_block(&chain_state).in_current_span().await;
         if latest_safe_block > *last_safe_block_processed {
+            let mut from = latest_safe_block.checked_sub(100).unwrap_or(0);
```
m30m, line 704: make this 100 a constant

apps/fortuna/src/keeper.rs
```diff
@@ -157,6 +167,16 @@ pub struct BlockRange {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequestState {
+    /// Fulfilled means that the request was either revealed or we are sure we
+    /// will not be able to reveal it.
+    Fulfilled,
+    /// We have already processed the request but couldn't fulfill it and we are
+    /// unsure if we can fulfill it or not.
+    Processed,
```
m30m, line 177: where are we using this value?

apps/fortuna/src/keeper.rs
```diff
@@ -499,6 +544,34 @@ pub async fn process_single_block_batch(
             Ok(events) => {
                 tracing::info!(num_of_events = &events.len(), "Processing",);
                 for event in &events {
```
m30m, line 546: can you add a TODO here that this is handling all events in a single block sequentially? We might need to parallelise this later on.

apps/fortuna/src/keeper.rs
```diff
@@ -499,6 +544,34 @@ pub async fn process_single_block_batch(
...
+                    {
+                        match state {
+                            RequestState::Fulfilled => {
+                                tracing::info!(
+                                    sequence_number = &event.sequence_number,
+                                    "Skipping already fulfilled request",
+                                );
+                                continue;
+                            }
+                            RequestState::Processed => {
```
0xfirefist, line 560: @m30m here

m30m (approved):

PR #1580, opened 2024-05-15
feat(target_chains/starknet): handle SetFee governance instruction

Riateche:
- Implemented payload parsing for governance header and SetFee variant
- Added a test to check that fee actually changes to the requested value

Generating price updates from scratch is a bit tricky, so I opted to get real price updates from mainnet and just re-sign their wormhole VAAs with the test guardian set.


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -190,9 +193,17 @@ mod pyth {
             let wormhole = IWormholeDispatcher { contract_address: self.wormhole_address.read() };
             let vm = wormhole.parse_and_verify_vm(data);
             self.verify_governance_vm(@vm);
-            //...
+            let data = governance::parse_instruction(vm.payload);
+            if data.target_chain_id != wormhole.chain_id() {
```
ali-bahjati, line 197: we allow chain id of 0 as well.

target_chains/starknet/tools/test_vaas/src/bin/generate_wormhole_vaas.rs
```diff
@@ -77,7 +77,7 @@ fn main() {
                 consistency_level: 6,
                 payload: PayloadKind::Binary(
                     GuardianSetUpgrade {
-                        chain_id: 1,
+                        chain_id: 60051,
```
ali-bahjati, line 80: Normally chainId of 0 is used here to mention all target chains.

ali-bahjati: Let's double check that wormhole governance also supports 0.

ali-bahjati (commented):
Please see my inline comments.

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Fixed.

target_chains/starknet/tools/test_vaas/src/bin/generate_wormhole_vaas.rs
Riateche: Changed to 0 for this one. Note that we also use mainnet VAAs in tests and those also have chain_id == 0.

ali-bahjati (approved):

PR #1587, opened 2024-05-20
chore(target_chains/starknet): automate test data generation

Riateche:
Now, instead of manually copying output from `test_vaas` tool to the cairo tests, we can simply run a command to automatically generate `data.cairo` that contains all the generated data. I've also expanded comments for each item. Main test files (`pyth.cairo` and `wormhole.cairo`) should be more readable now. I've also added a CI check to ensure that `generate_test_data.rs` and `data.cairo` stay in sync. There are no significant code changes, just a lot of moving code around.


Reisen:
These are great test VAA's and I'd call these golden tests even, am +1'ing but want to ask if you think we could rip these out and apply these to all the contracts rather than just this one? It's something we wanted a little while ago and can probably take advantage of for modifying other contracts (or introducing new ones).

Reisen (approved):

Riateche:
Yes, I think we can use this implementation for other tests as well. Apart from calldata encoding, there is nothing specific to starknet in these tests.

PR #1588, opened 2024-05-21
feat(entropy): Entropy inspection

m30m:


apps/fortuna/src/command/inspect.rs
```diff
@@ -0,0 +1,113 @@
...
+    println!("Initial request number: {}", current_request_number);
+    let last_request_number = current_request_number.saturating_sub(num_requests);
+    if multicall_exists {
+        println!("Using multicall");
+        let mut multicall = Multicall::new(
+            rpc_provider.clone(),
+            Some(ethers::contract::MULTICALL_ADDRESS),
+        )
+        .await?;
+        let batch_size = 100;
```
jayantk, line 63: suggest putting this in InspectOptions with a default value, just in case it needs to be changed later.

apps/fortuna/src/config.rs
```diff
@@ -66,6 +68,9 @@ pub enum Options {
     /// Request a random number from the contract.
     RequestRandomness(RequestRandomnessOptions),
 
+    /// Inspect recent requests and find unfulfilled requests with callback.
+    Inspect(InspectOptions),
```
jayantk, line 72: question about this: how do I tell that the request didn't get a callback from the result? It seems like the output simply prints out all of the requests that have happened.

contract_manager/src/contracts/evm.ts
```diff
@@ -622,6 +633,52 @@ export class EvmEntropyContract extends Storable {
     };
   }
 
+  async getRequest(
```
jayantk, line 636: nit: a 1 sentence doc comment on these methods would be nice. (esp for getUserRandomNumber, would be nice to explain how to get the `block` argument if you don't have it)

jayantk (approved):

apps/fortuna/src/config.rs
m30m: any request still stored in the contract is not fulfilled yet. We just iterate through the requests and find the ones where the callback flag is also true and print them out.

PR #1589, opened 2024-05-21
feat(target_chains/starknet): governance data source transfer

Riateche:
I had to move sequence storing to `verify_governance_vm()` to make sure that we store the correct last sequence after the transfer. The new version is closer to our ethereum contract.


ali-bahjati (approved):
Nice!

PR #1592, opened 2024-05-22
feat(target_chains/starknet): pyth contract upgrade

Riateche:


ali-bahjati (approved):
Nice! Can you explain somewhere how the upgrade works? Based on your code I see that we need to have a vaa of the class-hash of the new contract. I can assume that there should be a method to upload the contract and it will be accessible by its hash (which is smart compared to other networks!)

ali-bahjati (approved):

Riateche:
I've added a comment. Yes, the new contract class must be declared in advance. Anyone who has the compiled contract artifact can declare it (but only once per network), and the class hash uniquely identifies all properties of the implementation.

PR #1596, opened 2024-05-23
[fortuna] Adjust gas oracle logic to eliminate ethers.rs priority fee lower bound

jayantk:
So this was a fun one. It turns out that the default fee estimation logic in ethers.rs has a lower bound on the priority fee of 3 gwei. This value is way too high for the L2s that entropy is deployed on.

(see this logic here https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/rpc/provider.rs#L452
which calls this to estimate the fee https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/utils/mod.rs#L496
which uses this constant as a lower bound https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/utils/mod.rs#L88 )

Unfortunately, there isn't a way to configure this behavior in ethers.rs , so I've ended up copy/pasting the relevant bits and dividing the constants by 100k which puts them more in the range of l2 behavior.

### DIFF
--- a/apps/fortuna/src/chain/eth_gas_oracle.rs
+++ b/apps/fortuna/src/chain/eth_gas_oracle.rs
@@ -1,9 +1,20 @@
-use ethers::types::{U256, I256};
-use ethers::providers::Middleware;
-use ethers::prelude::GasOracle;
-use ethers::prelude::gas_oracle::GasOracleError;
-use ethers::prelude::gas_oracle::Result;
-use axum::async_trait;
+use {
+    axum::async_trait,
+    ethers::{
+        prelude::{
+            gas_oracle::{
+                GasOracleError,
+                Result,
+            },
+            GasOracle,
+        },
+        providers::Middleware,
+        types::{
+            I256,
+            U256,
+        },
+    },
+};
 
 // The default fee estimation logic in ethers.rs includes some hardcoded constants that do not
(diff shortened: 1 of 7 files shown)


apps/fortuna/src/keeper.rs
```diff
@@ -223,7 +223,7 @@ pub async fn run_keeper_threads(
             .await
             .expect("Chain config should be valid"),
     );
-    let keeper_address = contract.client().inner().inner().signer().address();
+    let keeper_address = contract.client().inner().inner().inner().signer().address();
```
jayantk, line 226: (these are added because you have to unwrap the middleware 1 more time now)

apps/fortuna/src/chain/eth_gas_oracle.rs
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+        rewards
+    };
+
+    // Return the median.
+    values[values.len() / 2]
+}
+
+fn base_fee_surged(base_fee_per_gas: U256) -> U256 {
+    if base_fee_per_gas <= U256::from(40_000u64) {
```
cctdaniel, line 152: I think you were gonna use `SURGE_THRESHOLD_1` here? same for the ones below

cctdaniel (approved):

PR #1605, opened 2024-05-24
[fortuna] Fix config and CLI args

jayantk:
I spent about an hour yesterday trying to set up a local fortuna instance to test my changes because (1) the instructions are out of date, and (2) the configuration is quite confusing. 

This PR updates the configuration / CLI args such that everything goes in a single config file. Once you have that file, you can simply run the commands without CLI args and they just work.

**Review suggestion** start with the config.sample.yaml and then look at config.rs. The rest of the changes are mechanical. There are a couple other minor updates in there which are all called out with comments.


apps/fortuna/src/command/register_provider.rs
```diff
@@ -28,42 +37,65 @@ pub struct CommitmentMetadata {
...
-    let fee_in_wei = opts.fee;
+    let fee_in_wei = chain_config.fee;
     let commitment = chain.reveal()?;
     // Store the random seed and chain length in the metadata field so that we can regenerate the hash
     // chain at-will. (This is secure because you can't generate the chain unless you also have the secret)
     let commitment_metadata = CommitmentMetadata {
         seed:         random,
         chain_length: commitment_length,
     };
+    let uri = get_register_uri(&provider_config.uri, &chain_id)?;
```
jayantk, line 90: drive-by bugfix: the previous implementation registered different URLs if you used register-provider vs. setup-provider

apps/fortuna/src/command/run.rs
```diff
@@ -184,14 +189,15 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
 
-    if let Some(keeper_private_key) = opts.load_keeper_private_key()? {
-        spawn(run_keeper(
-            chains.clone(),
-            config.clone(),
-            keeper_private_key,
-            metrics_registry.clone(),
-        ));
-    }
+    let keeper_private_key = config.keeper.private_key.load()?.ok_or(anyhow!(
```
jayantk, line 192: drive-by: disable option to run service without the keeper. I think we don't really care about the webservice only version anymore so this seems safer. (but open to debate on this)

apps/fortuna/src/command/run.rs
```diff
@@ -223,6 +232,16 @@ async fn setup_chain_state(
         )
     })?;
 
+    let last_prior_commitment = provider_commitments.last();
```
jayantk, line 235: drive-by bugfix: the previous implementation would happily let you specify commitments for future sequence numbers. That of course makes no sense and indicates a configuration problem.

apps/fortuna/src/config.rs
```diff
@@ -155,65 +134,86 @@ pub struct EthereumConfig {
...
+impl SecretString {
+    pub fn load(&self) -> Result<Option<String>> {
+        match &self.value {
+            Some(v) => return Ok(Some(v.clone())),
+            _ => {}
+        }
+
+        match &self.file {
+            Some(v) => {
+                return Ok(Some(fs::read_to_string(v)?.trim().to_string()));
```
jayantk, line 212: the trim here is possibly controversial but i lost like 20 mins on this because my hex file had a newline at the end

m30m: it's not. I also spent some time yesterday fixing this in another place and ended up adding a trim there too.

apps/fortuna/src/command/register_provider.rs
```diff
@@ -28,42 +37,65 @@ pub struct CommitmentMetadata {
...
+    chain_id: &ChainId,
+    chain_config: &EthereumConfig,
+) -> Result<()> {
+    let private_key_string = provider_config
+        .private_key
+        .load()?
+        .ok_or(anyhow!(
+            "Please specify a provider private key in the config"
+        ))?
+        .clone();
```
m30m, line 59: this `clone` seems unnecessary

apps/fortuna/src/command/run.rs
m30m: we deliberately did this to avoid running the keeper in multiple clusters since they would compete for fulfilling callbacks and we will have nonce issues.

apps/fortuna/src/command/run.rs
```diff
@@ -223,6 +232,16 @@ async fn setup_chain_state(
...
+    if last_prior_commitment.is_some()
+        && last_prior_commitment
+            .unwrap()
+            .original_commitment_sequence_number
+            >= provider_info.original_commitment_sequence_number
+    {
+        return Err(anyhow!("The current hash chain for chain id {} has configured commitments for sequence numbers greater than the current on-chain sequence number. Are the commitments configured correctly?", &chain_id).into());
```
m30m, line 242: the `.into` is not necessary

m30m (approved):

apps/fortuna/src/command/run.rs
jayantk: ok i'll revert this

PR #1610, opened 2024-05-24
[fortuna] Withdraw fees

jayantk:
Add logic to withdraw fees. Couple related features here:
1. The provider configuration now includes the fee manager and the provider config job sets it on-chain
2. CLI command for withdrawing fees as either the provider or keeper
3. Background task for the keeper to withdraw fees when its wallet gets low


apps/fortuna/src/keeper.rs
```diff
@@ -837,3 +846,68 @@ pub async fn track_provider(
...
+        .await
+        .map_err(|e| anyhow!("Error while getting balance. error: {:?}", e))?;
+
+    let provider_info = contract
+        .get_provider_info(provider_address)
+        .call()
+        .await
+        .map_err(|e| anyhow!("Error while getting provider info. error: {:?}", e))?;
+
+    let fees = provider_info.accrued_fees_in_wei;
```
m30m, line 887: we should add some check that the fee_manager is actually the keeper. Otherwise, we may get confusing revert errors

apps/fortuna/src/keeper.rs
```diff
@@ -837,3 +846,68 @@ pub async fn track_provider(
         })
         .set(end_sequence_number as i64);
 }
+
+// TODO: fields
```
m30m, line 850: no fields necessary here. We only want the `chain_id` and if you add `in_current_span` when you are spawning this, it should be enough.

m30m (approved):

apps/fortuna/src/keeper.rs
```diff
@@ -837,3 +846,68 @@ pub async fn track_provider(
...
+
+    let provider_info = contract
+        .get_provider_info(provider_address)
+        .call()
+        .await
+        .map_err(|e| anyhow!("Error while getting provider info. error: {:?}", e))?;
+
+    let fees = provider_info.accrued_fees_in_wei;
+
+    let min_balance = U256::from(12345);
```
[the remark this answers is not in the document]
jayantk: whoops

PR #1612, opened 2024-05-25
feat(xc_admin_frontend): add more details to summary view

cprussin:
This PR extends the summary to include publisher and price account details for addPublisher and delPublisher instructions.

screenshot-2024-05-24-183617

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/utils.ts
+++ b/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/utils.ts
@@ -1,5 +1,5 @@
 import { PythCluster } from '@pythnetwork/client'
-import { AccountMeta } from '@solana/web3.js'
+import { PublicKey } from '@solana/web3.js'
 import { MultisigAccount, TransactionAccount } from '@sqds/mesh/lib/types'
 import {
   ExecutePostedVaa,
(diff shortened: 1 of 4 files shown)


cprussin:
For some reason, Vercel isn't posting the preview URL, even though the preview built (probably because I'm working on a fork...)

In any case, here's a preview URL that's particularly extreme, with a ton of `addPublisher` and `delPublisher` commands:

https://xc-admin-frontend-2dsxc5l6p-pyth-web.vercel.app/?vercelToolbarCode=uKzMsd55JQiCTaw&tab=proposals&proposal=4feQoLU6WCXbNiS8wGw69qD5rEVH51JWyKwMxCknw37q

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/utils.ts
```diff
@@ -35,66 +35,89 @@ export const getProposalStatus = (
...
+      return {
+        name: 'delPublisher',
+        priceAccount:
+          parsedInstruction.accounts.named['priceAccount'].pubkey.toBase58(),
+        pub: (parsedInstruction.args['pub'] as PublicKey).toBase58(),
+      } as const
+    default:
+      return {
+        name: parsedInstruction.name,
+      } as const
```
guibescos, line 121: What is the point of the "as const"

cprussin: In case you're curious, here's a good illustration in the typescript playground of the difference between using the const assertion vs not: https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAJiAyiAtgUygCwJZgOYwC8MAFAIYBOeAXDNBbngJREB8MA3gLABQMdAd2xRgmUpWade-fsDIQ0MAOQAjMgC8ltChgCuFMJxhgy6Wqo1KANHVQYc+WgFYYAXwDc0mXDQAzMroANlDaegZGJmbKEHZYjACigQrWtuhx+IkKtFAUuooeXq68RTy8oJCwCMhpDngA6sKYAMLg0ETiVLT0jCyE7Nx8gsKiHZIDMjByCspqmqFQ+oYcxqZo5rMpMTWMzm4w8pOtUJ6D-D7+QSEwOgvhy5Fr0bG1mWibzwlJjzl5ewfl0BO-BKriAA.

If you take a look at the `.d.ts` tab on the right side, you'll see how typescript infers the types.  If you see how `doSomethingWithConst` gets inferred with literals, it essentially allows any consuming code that matches `name` to `baz` to automatically infer correctly that `something` is 5 and `somethingElse` is `undefined`.

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/utils.ts
```diff
@@ -35,66 +35,89 @@ export const getProposalStatus = (
...
+      )
+    } else if (governanceAction instanceof PythGovernanceActionImpl) {
+      return [{ name: governanceAction.action } as const]
+    } else if (governanceAction instanceof SetDataSources) {
+      return [{ name: governanceAction.actionName } as const]
+    } else {
+      return [{ name: 'unknown' } as const]
+    }
+  } else {
+    return [{ name: instruction.name } as const]
```
guibescos, line 94: When instruction is not a `WormholeMultisigInstruction`, you should also generate a summary if it's `addPublisher` or `delPublisher`

cprussin: Oh interesting, I derived this logic by looking at when we display more details for `addPublisher` and `delPublisher` -- in particular I looked at where leverage the mappings from keys to symbol & publisher name -- and I noticed this was only happening [in the component to render `WormholeMultisigInstruction` types](https://github.com/pyth-network/pyth-crosschain/blob/c0a4ab04ddc0c590f0dd5f699d3b88512ac0e88f/governance/xc_admin/packages/xc_admin_frontend/components/InstructionViews/WormholeInstructionView.tsx#L225-L236).  Does this mean we should update the details pages to expand that info for other instruction types too?

guibescos: Here's an example of a proposal where this comes into play: https://proposals.pyth.network/?tab=proposals&proposal=8zU6pr28Ta5zLwn5zPH2dWjHTAVGhg4VnjwpHeGALmFg

guibescos:
This is great, please add it for non wormhole proposals as well

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/utils.ts
cprussin: ah yes, I missed it [here](https://github.com/pyth-network/pyth-crosschain/blob/c0a4ab04ddc0c590f0dd5f699d3b88512ac0e88f/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/Proposal.tsx#L633-L642)!  Apologies on the oversight.  I will update this to account for non-wormhole instructions.

cprussin: Updated!

guibescos (approved):

PR #1614, opened 2024-05-27
feat(target_chains/starknet): update if necessary and get no older than

Riateche:
The stale check is different from the ethereum contract that essentially does `abs(block_timestamp - publish_time) < age`, which seems incorrect to me. Not sure what's the best way to handle the invalid `block_timestamp < publish_time` case, but it shouldn't depend on `age`.

### DIFF
--- a/target_chains/starknet/contracts/src/pyth.cairo
+++ b/target_chains/starknet/contracts/src/pyth.cairo
@@ -10,10 +10,13 @@ pub use pyth::{
     DataSourcesSet, FeeSet,
 };
 pub use errors::{GetPriceUnsafeError, GovernanceActionError, UpdatePriceFeedsError};
-pub use interface::{IPyth, IPythDispatcher, IPythDispatcherTrait, DataSource, Price};
+pub use interface::{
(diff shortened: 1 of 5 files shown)


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -229,6 +255,23 @@ mod pyth {
...
+            let mut i = 0;
+            while i < required_publish_times.len() {
+                let item = required_publish_times.at(i);
+                let latest_time = self.latest_price_info.read(*item.price_id).publish_time;
+                if latest_time < *item.publish_time {
+                    self.update_price_feeds(update);
+                    break;
+                }
+                i += 1;
+            }
```
ali-bahjati, line 272: in the evm implementation we revert here if it's unsuccessful. I think would be nice to have it here too. The rationale is that if we revert, there is a chance that it gets reverted on simulation and then users won't pay any fees.

ali-bahjati (approved):
Please address my inline comment before merging.

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Fixed.

PR #1615, opened 2024-05-27
fix-bn.js dependency issue

aditya520:


ali-bahjati (approved):
Since this is a bugfix, please bump the last number.

PR #1618, opened 2024-05-27
chore(target_chains/sui): vendor Wormhole dependency into the repo

ali-bahjati:
This change vendors the Wormhole dependency for movement devnet into this repo by just copying it. Unfortunately our proper solution, git-subtree, does't work properly on pull-requests; so we decided to go with a copy paste and tracking the commit because it is easy and we are not going to modify this codebase often.

Please review this PR commit by commit.


target_chains/sui/vendor/wormhole_movement_devnet/wormhole/Move.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "Wormhole"
+version = "0.2.0"
+
+[dependencies.Sui]
+git = "https://github.com/MystenLabs/sui.git"
+subdir = "crates/sui-framework/packages/sui-framework"
+rev = "041c5f2bae2fe52079e44b70514333532d69f4e6"
+published-at = "0x23a373b70e6e23a39e4846fa6896fa12beb08da061b3d4ec856bc8ead54f1e22"
```
Reisen, line 9: I think this should be under `[package]`, and `_` under `[addresses]` should also be updated to the same address.

Reisen (approved):

PR #1619, opened 2024-05-27
feat(express-relay): Add api key

danimhr:
This PR adds API key compatibility for web sockets and HTTP requests. There is also an implementation of the server's "get_bids" API.


express_relay/sdk/python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "express-relay"
-version = "0.5.0"
+version = "0.6.0"
```
m30m, line 3: let's sync these again with the server side.

m30m (approved):

PR #1622, opened 2024-05-28
feat(target_chains/starknet): parce price feeds

Riateche:


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -490,6 +476,98 @@ mod pyth {
             let event = ContractUpgraded { new_class_hash: new_implementation };
             self.emit(event);
         }
+
+        fn parse_price_feed_updates_internal(
```
ali-bahjati, line 480: Can you add some docs for this function explaining what it does? Specifically the fact that it has a side-effect of updating the contract state during the parsing.

ali-bahjati (approved):
Very nice!

target_chains/starknet/contracts/src/pyth.cairo
Riateche: Renamed to `update_price_feeds_internal` and added a comment.

PR #1624, opened 2024-05-29
[fortuna] Add subsampling of hash chains

jayantk:
Update PebbleHashChain to subsample the hashes stored in memory. This change trades off increased compute at reveal time for reduced memory usage.

It looks like the constraint on registering a big sequence now is actually the startup time. It takes about 5 seconds to compute a chain of 10M numbers. I think 5s is probably fine, but we probably can't 10x that.


apps/fortuna/src/command/register_provider.rs
```diff
@@ -65,18 +65,19 @@ pub async fn register_provider_from_config(
...
         &chain_config.contract_addr,
         &random,
         commitment_length,
+        provider_config.chain_sample_interval,
     )?;
 
     // Arguments to the contract to register our new provider.
     let fee_in_wei = chain_config.fee;
-    let commitment = chain.reveal()?;
+    let commitment = chain.reveal_ith(0)?;
```
jayantk, line 80: drive-by: I deleted the `reveal` method which wasn't really used anywhere

apps/fortuna/src/command/run.rs
```diff
@@ -36,6 +36,10 @@ use {
             BlockNumber,
         },
     },
+    futures::{
+        future::join_all,
+        FutureExt,
+    },
```
jayantk, line 42: only new changes are here

apps/fortuna/src/command/run.rs
```diff
@@ -157,10 +161,23 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
     ))?;
     let (tx_exit, rx_exit) = watch::channel(false);
 
-    let mut chains: HashMap<ChainId, BlockchainState> = HashMap::new();
+    let mut tasks = Vec::new();
     for (chain_id, chain_config) in &config.chains {
-        let state =
-            setup_chain_state(&config.provider.address, &secret, chain_id, chain_config).await;
+        tasks.push(
+            setup_chain_state(
```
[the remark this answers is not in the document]
m30m: Example:
```rust
use std::time::Instant;
use futures::future::join_all;
use tokio::spawn;
pub async fn long_task() {
    let mut z:i32 = 3;
    for _ in 0..100000000 {
        z.overflowing_mul(z);
    }
}

#[tokio::main]
async fn main() {
    let current_time = Instant::now();
    let mut tasks = Vec::new();
    for _ in 0..5 {
        // tasks.push(long_task()); // This will run the tasks sequentially
        tasks.push(spawn(long_task())); // This will run the tasks concurrently
    }
    join_all(tasks).await;
    println!("Duration: {:?}", Instant::now().duration_since(current_time));
}
```

m30m (approved):

PR #1625, opened 2024-05-29
fix(target_chains/fuel): upgrade to fuel testnet from beta-5

cctdaniel:
changes:
- len -> len()
- have to explicitly add `pub` before struct fields to declare it as public fields
- function changes


ali-bahjati (approved):
Nice. Thank you!

PR #1635, opened 2024-05-30
fix(xc_admin_frontend):  product no price account

guibescos:
Recently, a product account was wrongly initialized and didn't have any price accounts, causing the website to crash.
I think it's better to handle it more gracefully, by just ignoring such a product.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/hooks/usePyth.ts
+++ b/governance/xc_admin/packages/xc_admin_frontend/hooks/usePyth.ts
@@ -112,14 +112,7 @@ export const usePyth = (): PythHookData => {
           switch (base?.type) {
             case AccountType.Product:
               const parsed = parseProductData(allPythAccounts[i].account.data)
-
-              if (!parsed.priceAccountKey) {
-                productRawConfigs[allPythAccounts[i].pubkey.toBase58()] = {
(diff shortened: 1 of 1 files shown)


cprussin (approved):
The change makes sense.

I'm not sure I agree with the idea of just hiding these.  Maybe this isn't the right tool for it, but I guess it seems like whatever tools we have that give you insights into what's happening on the network should be able to handle any state the network could get into and shouldn't omit anything.  Obviously you know better than I do and maybe this is something that would make more sense to consider when I work on the transparency app instead.

cprussin:
Also worth noting that we could possibly use a tool like https://zod.dev/ (you can think about this as a typescript version of serde's deserialization functionality) to catch error states earlier and to ensure that typescript is actually reliable.

guibescos:
> The change makes sense.
> 
> I'm not sure I agree with the idea of just hiding these. Maybe this isn't the right tool for it, but I guess it seems like whatever tools we have that give you insights into what's happening on the network should be able to handle any state the network could get into and shouldn't omit anything. Obviously you know better than I do and maybe this is something that would make more sense to consider when I work on the transparency app instead.

I merged for now... but you're right, this fix doesn't seem right

PR #1644, opened 2024-05-31
feat(target_chains/starknet): support two fee tokens

Riateche:
The user just needs to set allowance for either of the tokens before pushing the update.

### DIFF
--- a/target_chains/starknet/contracts/src/pyth.cairo
+++ b/target_chains/starknet/contracts/src/pyth.cairo
@@ -97,9 +97,9 @@ mod pyth {
     #[storage]
     struct Storage {
         wormhole_address: ContractAddress,
-        fee_token_address: ContractAddress,
+        fee_token_address1: ContractAddress,
         fee_token_address2: ContractAddress,
-        single_update_fee: u256,
+        single_update_fee1: u256,
         single_update_fee2: u256,
         data_sources: LegacyMap<usize, DataSource>,
         num_data_sources: usize,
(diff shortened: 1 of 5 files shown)


target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -98,7 +98,9 @@ mod pyth {
     struct Storage {
         wormhole_address: ContractAddress,
         fee_token_address: ContractAddress,
```
ali-bahjati, line 100: i think renaming the existing one to `fee_token_address1` (or `_1`) makes it more verbose.

target_chains/starknet/contracts/src/pyth.cairo
```diff
@@ -708,4 +736,22 @@ mod pyth {
...
+        single_update_fee: u256,
+    ) -> bool {
+        let total_fee = single_update_fee * num_updates.into();
+        let fee_contract = IERC20CamelDispatcher { contract_address: fee_token };
+        if fee_contract.allowance(caller, contract) < total_fee {
+            return false;
+        }
+        if !fee_contract.transferFrom(caller, contract, total_fee) {
+            panic_with_felt252(UpdatePriceFeedsError::InsufficientFeeAllowance.into());
+        }
```
ali-bahjati, line 754: This can fail if there is allowance but not enough balance right? I think in this situation we don't want to panic if the user has balance with the other token, right?

Riateche: Good point! The ERC20 contract actually just panics with u256_sub_overflow in this case. I've added a check of balance and a test for this case.

ali-bahjati (approved):

PR #1650, opened 2024-06-03
[contract_manager] Add some nice-to-have features

jayantk:
This PR fixes a couple minor things I noticed while trying to do the entropy contract upgrade. Comments inline explaining each change.


contract_manager/scripts/check_proposal.ts
```diff
@@ -140,7 +140,7 @@ async function main() {
       if (instruction.governanceAction instanceof EvmExecute) {
         // Note: it only checks for upgrade entropy contracts right now
         console.log(
-          `Verifying EVMExecute Contract on ${instruction.governanceAction.targetChainId}`
+          `Verifying EVMExecute on ${instruction.governanceAction.targetChainId}`
```
jayantk, line 143: The previous output for verifying the entropy upgrade proposal was very confusing and made it look like you were upgrading the executor contract. The new output is very explicit:

```
Verifying EVMExecute on blast_s2_testnet
blast_s2_testnet  call upgradeTo(address) with arguments (0x014393b4c3c28054D533204309c52c77373Bea24) on EvmEntropyContract at address:98046bd286715d3b0bc227dd7a956b83d8978603 from executor:5f3c61944ceb01b3eaef861251fb1e0f14b848fb.
blast_s2_testnet    new implementation address:0x014393b4c3c28054D533204309c52c77373Bea24 has code digest:0x00796fddbb5c438e486af88939daaaade9560d659124099113b05c9fb5c82fb1
```

contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -27,6 +28,17 @@ const parser = yargs(hideBin(process.argv))
     },
   });
 
+// Override these URLs to use a different RPC node for mainnet / testnet.
+// TODO: extract these RPCs to a config file (?)
+const RPCS: Record<string, string> = {
+  "mainnet-beta": "https://api.mainnet-beta.solana.com",
+  testnet: "https://api.testnet.solana.com",
+};
+1;
```
m30m, line 37: seems not necessary :D

contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -27,6 +28,17 @@ const parser = yargs(hideBin(process.argv))
...
+function registry(cluster: PythCluster): string {
+  return RPCS[cluster as string];
```
m30m, line 39: you can remove the `as string` here and the `Record<string,string>` from the `const RPCS` to simplify the typing a bit.

contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -45,41 +57,56 @@ async function main() {
...
+      );
     }
   }
 
   console.log("Using vault at for proposal", vault.getId());
   const wallet = await loadHotWallet(argv["ops-key-path"]);
   console.log("Using wallet ", wallet.publicKey.toBase58());
-  await vault.connect(wallet);
-  const proposal = await vault.proposeWormholeMessage(payloads);
+  await vault.connect(wallet, registry);
```
m30m, line 108: You can remove the await since the function is not async

contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -27,6 +28,17 @@ const parser = yargs(hideBin(process.argv))
...
   });
 
+// Override these URLs to use a different RPC node for mainnet / testnet.
+// TODO: extract these RPCs to a config file (?)
+const RPCS: Record<string, string> = {
+  "mainnet-beta": "https://api.mainnet-beta.solana.com",
+  testnet: "https://api.testnet.solana.com",
+};
+1;
+function registry(cluster: PythCluster): string {
```
m30m, line 38: Unrelated to this PR but I just noticed that you didn't add the new `registry` parameter into the JSDocs in `governance.ts`. It would be great to add it there.

contract_manager/src/chains.ts
```diff
@@ -447,12 +447,20 @@ export class EvmChain extends Chain {
...
+        gas,
+        gasPrice: gasPrice.toString(),
+      });
+    } catch (e) {
+      // RPC errors often have useful information in the non-primary message field. Log the whole error
+      // to simplify identifying the problem.
+      console.log(`Error deploying contract: ${JSON.stringify(e)}`);
+      throw e;
+    }
+    return deployedContract!.options.address;
```
m30m, line 463: you can move this return inside the try and remove the non-null assertion + remove `let deployedContract = undefined;`

contract_manager/scripts/upgrade_evm_entropy_contracts.ts
```diff
@@ -45,41 +57,56 @@ async function main() {
 
   console.log("Using cache file", cacheFile);
 
-  const payloads: Buffer[] = [];
+  // Try to deploy on every chain, then collect any failures at the end. This logic makes it simpler to
+  // identify deployment problems (e.g., not enough gas) on every chain where they occur.
+  const payloads: (Buffer | undefined)[] = [];
```
m30m, line 62: this type automatically gets converted to `Buffer[]` because of how the indexing works in typescript. Maybe you can use some other falsy value like false.
Another way is to have a allSuccessful flag which we can flip and remove the for loop in the end.

m30m (approved):
thanks for all the improvements

PR #1651, opened 2024-06-03
feat(fortuna): Traced client for better observability

m30m:


apps/fortuna/src/chain/traced_client.rs
```diff
@@ -0,0 +1,135 @@
...
+        let latency = start.elapsed().as_secs_f64();
+        self.metrics.latency.get_or_create(label).observe(latency);
+        res
+    }
+}
+
+impl TracedClient {
+    pub fn new(
+        chain_id: ChainId,
+        url: &str,
```
jayantk, line 125: nit: suggest passing the `Url` type here

jayantk (approved):
very nice

PR #1653, opened 2024-06-04
feat(apps/hermes): new client package for hermes

cctdaniel:
- created a new hermes client package
- used `zod` for schema validation
- moved hermes server to a new `server` folder in `apps/hermes`
- removed `oracle-swap` and `send-usd` example since they are moved to https://github.com/pyth-network/pyth-examples

will deprecate `price-service-client` once this package is published

### DIFF
--- a/apps/hermes/client/js/package.json
+++ b/apps/hermes/client/js/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@pythnetwork/hermes-client",
   "version": "1.0.0",
-  "description": "Pyth Hermes client",
+  "description": "Pyth Hermes Client",
   "author": {
     "name": "Pyth Data Association"
   },


governance/xc_admin/packages/proposer_server/package.json
```diff
@@ -21,11 +21,14 @@
...
     "@solana/web3.js": "^1.76.0",
     "@sqds/mesh": "^1.0.6",
     "@types/cors": "^2.8.17",
     "cors": "^2.8.5",
-    "ts-node": "^10.9.1",
-    "@pythnetwork/xc-admin-common": "*"
+    "ts-node": "^10.9.1"
+  },
+  "devDependencies": {
+    "@types/express": "^4.17.21"
```
cctdaniel, line 32: for some reason I was getting this error even though this package wasnt changed and hence had to add this

```
    ✖  @pythnetwork/xc-admin-proposer-server:build
       > @pythnetwork/xc-admin-proposer-server@0.0.0 build
       > tsc

       src/index.ts(1,44): error TS7016: Could not find a declaration file for module 'express'. '/Users/danielchew/Code/pyth-crosschain/node_modules/express/index.js' implicitly has an 'any' type.
         Try `npm i --save-dev @types/express` if it exists or add a new declaration (.d.ts) file containing `declare module 'express';`
       npm ERR! Lifecycle script `build` failed with error:
       npm ERR! Error: command failed
       npm ERR!   in workspace: @pythnetwork/xc-admin-proposer-server@0.0.0
       npm ERR!   at location: /Users/danielchew/Code/pyth-crosschain/governance/xc_admin/packages/proposer_server
```

apps/hermes/client/js/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  rules: {
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-empty-function": "off",
```
ali-bahjati, line 9: Let's remove these exceptions if we can.

apps/hermes/client/js/package.json
ali-bahjati, line 1: we should add this new package to the top level `package.json` too (to add it to the workspace)

apps/hermes/client/js/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "jest": "^29.4.0",
+    "prettier": "^2.6.2",
+    "ts-jest": "^29.0.5",
+    "typescript": "^4.6.3",
+    "yargs": "^17.4.1"
+  },
+  "dependencies": {
+    "@types/ws": "^8.5.3",
+    "eventsource": "^2.0.2",
+    "isomorphic-ws": "^4.0.1",
```
ali-bahjati, line 55: we won't need ws related dependencies anymore, right?

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,237 @@
...
+  private timeout: DurationInMs;
+  private httpRetries: number;
+  private logger: Logger;
+
+  /**
+   * Custom handler for web socket errors (connection and message parsing).
+   *
+   * Default handler only logs the errors.
+   */
+  onWsError: (error: Error) => void;
```
ali-bahjati, line 38: This is unused.

apps/hermes/client/js/README.md
```diff
@@ -0,0 +1,89 @@
+# Hermes Client
+
+[Pyth Network](https://pyth.network/) provides real-time pricing data in a variety of asset classes, including cryptocurrency, equities, FX and commodities.
+These prices are available either via HTTP or WebSocket from [Hermes](/apps/hermes).
```
ali-bahjati, line 4: HTTP or Streaming

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,237 @@
...
+   * This will return an EventSource that can be used to listen to streaming updates.
+   *
+   * @param ids Array of hex-encoded price feed IDs for which streaming updates are requested.
+   * @param encoding Optional encoding type. If specified, updates are returned in the specified encoding. Default is hex.
+   * @param parsed Optional boolean to specify if the parsed price update should be included in the response. Default is false.
+   * @param allow_unordered Optional boolean to specify if unordered updates are allowed to be included in the stream. Default is false.
+   * @param benchmarks_only Optional boolean to specify if only benchmark prices that are the initial price updates at a given timestamp (i.e., prevPubTime != pubTime) should be returned. Default is false.
+   * @returns An EventSource instance for receiving streaming updates.
+   */
+  async getStreamingPriceUpdates(
```
ali-bahjati, line 210: maybe `subscribePriceUpdates` or `getPriceUpdatesStream`?

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,237 @@
...
+      url.searchParams.append("parsed", String(parsed));
+    }
+    return await this.httpRequest(url.toString());
+  }
+
+  /**
+   * Fetch streaming price updates for a set of price feed IDs.
+   * This endpoint can be customized by specifying the encoding type, whether the results should include parsed updates,
+   * and if unordered updates or only benchmark updates are allowed.
+   * This will return an EventSource that can be used to listen to streaming updates.
```
ali-bahjati, line 201: What is the behaviour if invalid ID is passed? Will it throw an error or?

ali-bahjati (commented):
Nice! I left some comments. I think it would be good if it gets more eyes.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,237 @@
...
+  ): Promise<any> {
+    const controller = new AbortController();
+    const { signal } = controller;
+    options = { ...options, signal }; // Merge any existing options with the signal
+
+    // Set a timeout to abort the request if it takes too long
+    const timeout = setTimeout(() => controller.abort(), this.timeout);
+
+    try {
+      const response = await fetch(url, options);
```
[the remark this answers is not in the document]
cctdaniel: yup we don't afaik

apps/hermes/client/js/src/HermesConnection.ts
cctdaniel: it will return error 404 `Price ids not found: xxxx` as per how it is if you directly access the hermes endpoint

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+      }
+      return await response.json();
+    } catch (error) {
+      clearTimeout(timeout);
+      if (
+        retries > 0 &&
+        !(error instanceof Error && error.name === "AbortError")
+      ) {
+        // Wait for a backoff period before retrying
+        await new Promise((resolve) => setTimeout(resolve, backoff));
```
cprussin, line 68: Don't do this, but a cool esoteric trick is that in Node you can actually do this:

```ts
import { promisify } from 'util';

await promisify(setTimeout)(backoff);
```

The reason this works is that Node will actually promisify [using the `nodejs.util.promisify.custom` Symbol attached to a function](https://github.com/nodejs/node/blob/0281e2cbf0531ade992265b7b734dd3f1ecffe8a/lib/internal/util.js#L401-L410) if it exists, and [`setTimeout` defines that symbol](https://github.com/nodejs/node/blob/0281e2cbf0531ade992265b7b734dd3f1ecffe8a/lib/timers.js#L172-L180) using [a promisifed version of `setTimeout`](https://github.com/nodejs/node/blob/0281e2cbf0531ade992265b7b734dd3f1ecffe8a/lib/timers/promises.js#L52-L93).

There are actually good reasons to prefer this (or just use `import { setTimeout } from 'timers/promises'` directly) -- the promisified version does some things like handle  abort controllers, does not wrap the underlying `setTimeout` and is thus arguably marginally more performant, etc.  However, I don't think that either the `promisify` trick or the `timers/promises` module are portable so I wouldn't use it in a place like this.  Still, it's a cool trick to be aware of.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param query Optional query string to filter the price feeds. If provided, the results will be filtered to all price feeds whose symbol contains the query string. Query string is case insensitive. Example : bitcoin
+   * @param filter Optional filter string to filter the price feeds. If provided, the results will be filtered by asset type. Possible values are crypto, equity, fx, metal, rates. Filter string is case insensitive. Available values : crypto, fx, equity, metals, rates
+   * @returns Array of PriceFeedMetadata objects.
+   */
+  async getPriceFeeds(
+    query?: string,
+    filter?: string
+  ): Promise<PriceFeedMetadata[]> {
+    const url = new URL(`${this.baseURL}/v2/price_feeds`);
```
cprussin, line 92: Nice use of URL but rather than interpolating it's preferable to do `new URL('/v2/price_feeds', this.baseURL)` since doing it that will will enable [relative references to resolve properly](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references)

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.timeout = config?.timeout || DEFAULT_TIMEOUT;
+    this.httpRetries = config?.httpRetries || DEFAULT_HTTP_RETRIES;
+  }
+
+  private async httpRequest<ResponseData>(
+    url: string,
+    options?: RequestInit,
+    retries = this.httpRetries,
+    backoff = 300
+  ): Promise<ResponseData> {
```
cprussin, line 46: If you follow my advice above about using a schema validator, then rather than casting, here you would pass a schema and add a schema parse before returning the result.  Then you don't have to typecast unsafely.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Number of times a HTTP request will be retried before the API returns a failure. Default: 3.
+   *
+   * The connection uses exponential back-off for the delay between retries. However,
+   * it will timeout regardless of the retries at the configured `timeout` time.
+   */
+  httpRetries?: number;
+};
+
+export class HermesConnection {
```
cprussin, line 24: Minor nit but IMO there's not much of a good reason to make this a class.  There isn't any internal state being tracked, and form a PLT perspective, state management is really the only thing that classes actually exist for.

This is personal preference but for things that aren't stateful, I prefer doing this:

```tsx
import { doAThing } from "hermes-client";

const sharedOptions = { ... };

doAThing({ ...sharedOptions, ...moreOptions });
```

Even though it's is more verbose, I actually greatly prefer it.  There's much less indirection and it's much easier to follow what's happening at the call site, rather than trying to find options that get passed in to some constructor somewhere which may be far away and may not be obvious at the call site actually have influence over the behavior.

This is totally opinion and feel free to disregard.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+      return await response.json();
+    } catch (error) {
+      clearTimeout(timeout);
+      if (
+        retries > 0 &&
+        !(error instanceof Error && error.name === "AbortError")
+      ) {
+        // Wait for a backoff period before retrying
+        await new Promise((resolve) => setTimeout(resolve, backoff));
+        return this.httpRequest(url, options, retries - 1, backoff * 2); // Exponential backoff
```
cprussin, line 69: This isn't exponential, this is linear.  To make it exponential this should be `Math.pow(2, this.httpRetries - retries) * initialBackoff`.

I would also suggest making the initial backoff a random number within a range of a few hundred ms, as this will help prevent a "thundering herd" problem where some network outage causes all clients to retry connecting at exactly the same moment, and then if opening the connections fails, they all back off at the exact same rate and so the backoff doesn't actually help.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @returns Array of PriceUpdate objects containing the latest updates.
+   */
+  async getLatestPriceUpdates(
+    ids: HexString[],
+    encoding?: EncodingType,
+    parsed?: boolean
+  ): Promise<PriceUpdate[]> {
+    const url = new URL(`${this.baseURL}/v2/updates/price/latest`);
+    // Append parameters to the URL search parameters
+    ids.forEach((id) => url.searchParams.append("ids[]", id));
```
cprussin, line 119: It's preferable to use `for ... of` over `forEach`, see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md.  I know it feels less "functional" but in reality `forEach` isn't functional anyways since the only reason to use it is to trigger side effects.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+      allow_unordered:
+        allow_unordered !== undefined ? String(allow_unordered) : undefined,
+      benchmarks_only:
+        benchmarks_only !== undefined ? String(benchmarks_only) : undefined,
+    };
+    Object.entries(params).forEach(([key, value]) => {
+      if (value !== undefined) {
+        url.searchParams.append(key, value);
+      }
+    });
```
cprussin, line 195: I suggest making all these optional params an `options` object argument rather than positional, i.e.:

```ts
async getPriceUpdatesStream(ids: HexString[], options?: {
  encoding?: EncodingType,
  parsed?: boolean,
  allow_unordered?: boolean,
  benchmarks_only?: boolean
}): Promise<EventSource> {
  ...
  if (options) {
    for (const [key, value] of Object.entries(options)) {
      if (value !== undefined) {
        url.searchParams.append(key, typeof value === "boolean" ? String(value) : value);
      }
    }
  }
  ...
}
```

And the same for all the other apis for the `encoding` and `parsed` params.  You can wrap up the block to append the params from the options into a nice little function too so you don't have to duplicate it to all of the methods.

apps/hermes/client/js/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from "./types";
+
+export {
+  DurationInMs,
+  DurationInSeconds,
+  HermesConnection,
+  HermesConnectionConfig,
+  HexString,
+  UnixTimestamp,
+} from "./HermesConnection";
```
cprussin, line 10: I have generally moved away from using barrel files.  The reason is that there are many environments where barrel files don't tree-shake correctly ([nextjs itself had this problem for many years](https://github.com/vercel/next.js/issues/12557) and may still have it, I'm honestly not sure and the thread about it is extremely long and split across many discussions and tickets so it's hard to say without just testing it) and it adds overhead to the bundler and require resolution that node doesn't compile away very effectively.

It obviously doesn't matter much for a case like this where it's a single layer of indirection and there's nothing that _could_ be shaken anyways, but patterns get copied and libraries get extended over the years, so I advise just avoiding the pattern to start with.

[Package exports are your friend](https://nodejs.org/api/packages.html#exports)

apps/hermes/client/js/src/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export type AssetType = components["schemas"]["AssetType"];
+export type BinaryPriceUpdate = components["schemas"]["BinaryPriceUpdate"];
+export type EncodingType = components["schemas"]["EncodingType"];
+export type GetVaaCcipInput = components["schemas"]["GetVaaCcipInput"];
+export type GetVaaCcipResponse = components["schemas"]["GetVaaCcipResponse"];
+export type GetVaaResponse = components["schemas"]["GetVaaResponse"];
+export type ParsedPriceUpdate = components["schemas"]["ParsedPriceUpdate"];
+export type PriceFeedMetadata = components["schemas"]["PriceFeedMetadata"];
+export type PriceIdInput = components["schemas"]["PriceIdInput"];
+export type PriceUpdate = components["schemas"]["PriceUpdate"];
```
cprussin, line 13: Similarly to the comment about avoiding barrel files, I also generally try to avoid `types.ts` files except in very specific scenarios.  These types nearly always are directly tied to API methods and very very often have to be imported alongside those APIs, and putting them in a separate file usually just means you have to re-export things, or the call site has to import from different places, or you have to use barrel files.  I would just move these  to the `HermesConnection.ts` file.

apps/hermes/client/js/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "../../../../tsconfig.base.json",
```
cprussin, line 2: In case you're interested [I also have a canned tsconfig](https://www.cprussin.net/modules/_cprussin_tsconfig.html) which makes typescript drastically more strict and reliable [the definition of the tsconfig is here](https://github.com/cprussin/cprussin-monorepo/blob/main/packages/tsconfig/src/base.json).

cprussin (changes requested):
Most of my comments are just little suggestions and not super important.

IMO the most important thing is I strongly, strongly recommend using a runtime schema validator rather than relying on generated types and unsafe casting.

Feel free to disregard the other comments as you feel it makes sense.  Happy to expand on anything if it helps, a lot of this is somewhat esoteric knowledge I've picked up over the years and have become opinionated on but of course other opinions are valid too.

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.baseURL = endpoint;
+    this.timeout = config?.timeout || DEFAULT_TIMEOUT;
+    this.httpRetries = config?.httpRetries || DEFAULT_HTTP_RETRIES;
+  }
+
+  private async httpRequest<ResponseData>(
+    url: string,
+    options?: RequestInit,
+    retries = this.httpRetries,
+    backoff = 300
```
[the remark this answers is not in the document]
cctdaniel: great point

apps/hermes/client/js/src/HermesConnection.ts
cctdaniel: thanks for the tips

apps/hermes/client/js/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /home/node/
+USER 1000
+
+COPY --chown=1000:1000 apps/hermes/client/js apps/hermes/client/js
+
+RUN npx lerna run build --scope="@pythnetwork/hermes-client" --include-dependencies
+
+WORKDIR /home/node/apps/hermes/client/js
+
+ENTRYPOINT ["npm"]
```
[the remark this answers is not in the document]
cctdaniel: removed

apps/hermes/client/js/src/HermesConnection.ts
```diff
@@ -0,0 +1,222 @@
...
+    });
+
+    if (options) {
+      this.appendUrlSearchParams(url, options);
+    }
+
+    return new EventSource(url.toString());
+  }
+
+  private appendUrlSearchParams(url: URL, params: Record<string, any>) {
```
cprussin, line 215: You should not use `any` here, it's very unsafe.  In general you shouldn't use `any` _ever_, it effectively bails out of the type system completely.  You can use `unknown` if you want to allow any type -- and this will ensure typescript can make sure that your code is safe for an arbitrary type -- but in this case I'd just type it as `EncodingType | boolean` since that's all you've tested it to work for anyways.

apps/hermes/client/js/src/index.ts
```diff
@@ -0,0 +1,18 @@
+export {
```
cprussin, line 1: Might as well just get rid of this file now and move `HermesClient.ts` to `index.ts` or just point the package.json at `HermesClient.ts` instead.  This file serves no purpose and just adds indirection.

apps/hermes/client/js/src/HermesConnection.ts
cctdaniel: thanks for the tip!

apps/hermes/client/js/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "eslint": "^8.14.0",
+    "jest": "^29.4.0",
+    "prettier": "^2.6.2",
+    "ts-jest": "^29.0.5",
+    "typescript": "^4.6.3",
+    "yargs": "^17.4.1"
+  },
+  "dependencies": {
+    "eventsource": "^2.0.2",
+    "openapi-zod-client": "^1.18.1",
```
cprussin, line 55: This should be a dev dependency

apps/hermes/client/js/src/HermesClient.ts
```diff
@@ -0,0 +1,228 @@
...
+   */
+  async getPriceFeeds(options?: {
+    query?: string;
+    filter?: string;
+  }): Promise<PriceFeedMetadata[]> {
+    const url = new URL("/v2/price_feeds", this.baseURL);
+    if (options) {
+      this.appendUrlSearchParams(url, options);
+    }
+    return await this.httpRequest<PriceFeedMetadata[]>(
```
cprussin, line 106: You shouldn't have to pass type params to the calls to `httpRequest` any more now that you have schemas and type inference -- and in fact if removing the type params fails then it means something isn't working correctly with inference so I'd suggest not shipping it

cprussin (approved):
LGTM

.github/workflows/ci-hermes-server.yml
```diff
@@ -1,11 +1,11 @@
...
+name: Check Hermes Server
 
 on:
   pull_request:
-    paths: [apps/hermes/**]
+    paths: [apps/hermes/server/**]
   push:
     branches: [main]
-    paths: [apps/hermes/**]
+    paths: [apps/hermes/server/**]
```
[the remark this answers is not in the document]
ali-bahjati: Let's do it in a later PR.

apps/hermes/client/js/src/HermesClient.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  private async httpRequest<ResponseData>(
+    url: string,
+    schema: z.ZodSchema<ResponseData>,
+    options?: RequestInit,
+    retries = this.httpRetries,
+    backoff = 100 + Math.floor(Math.random() * 100), // Adding randomness to the initial backoff to avoid "thundering herd" scenario where a lot of clients that get kicked off all at the same time (say some script or something) and fail to connect all retry at exactly the same time too
+    externalAbortController?: AbortController
+  ): Promise<ResponseData> {
+    const controller = externalAbortController || new AbortController();
```
ali-bahjati, line 58: probably `??` here too?

ali-bahjati (approved):

.github/workflows/ci-hermes-server.yml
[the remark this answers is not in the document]
Reisen: Agreed with both above comments, agreed we should use the pattern from the doc and to do it in another PR.

PR #1654, opened 2024-06-04
feat(governance): add SetFeeInToken directive

Riateche:
It's needed for Starknet because there are two fee tokens. We add it as a separate action because extending SetFee action would be incompatible with existing contracts.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
+++ b/governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
@@ -15,6 +15,7 @@ export const TargetAction = {
   SetValidPeriod: 4,
   RequestGovernanceDataSourceTransfer: 5,
   SetWormholeAddress: 6,
+  SetFeeInToken: 7,
 } as const;
 
 export const EvmExecutorAction = {
@@ -43,6 +44,8 @@ export function toActionName(
         return "RequestGovernanceDataSourceTransfer";
       case 6:
         return "SetWormholeAddress";
+      case 7:
+        return "SetFeeInToken";
     }
   } else if (
     deserialized.moduleId == MODULE_EVM_EXECUTOR &&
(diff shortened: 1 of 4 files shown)


governance/xc_admin/packages/xc_admin_common/src/governance_payload/SetFee.ts
```diff
@@ -42,3 +42,62 @@ export class SetFee extends PythGovernanceActionImpl {
...
+    targetChainId: ChainName,
+    readonly newFeeValue: bigint,
+    readonly newFeeExpo: bigint,
+    readonly token: Buffer
+  ) {
+    super(targetChainId, "SetFeeInToken");
+  }
+
+  static decode(data: Buffer): SetFeeInToken | undefined {
+    console.log("SetFeeInToken.decode");
```
ali-bahjati, line 79: missed print

governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
```diff
@@ -176,10 +179,13 @@ export abstract class PythGovernanceActionImpl implements PythGovernanceAction {
   ): Buffer {
     const headerBuffer = this.header().encode();
 
-    const payloadBuffer = Buffer.alloc(payloadLayout.span);
-    payloadLayout.encode(payload, payloadBuffer);
+    // XXX: there is no way to get the length before the encoding.
```
ali-bahjati, line 182: I needed to think a lot why this is happening. I don't like this very much. i see that in `SetDataSources` we haven't defined the whole thing as a layout and manually decode/encode there. Maybe we can do the same?

governance/xc_admin/packages/xc_admin_common/src/governance_payload/SetFee.ts
Riateche: Fixed, thanks

governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
Riateche: I tried to avoid this but I guess it makes sense to use the approach we already use in `SetDataSources` (I hadn't noticed that before). Updated.

ali-bahjati (approved):

PR #1657, opened 2024-06-05
feat(target_chain/fuel): make abis public

cctdaniel:


ali-bahjati (approved):
It seems to be more than ABIs. Do we need the contract binaries as well? (I mean the`pyth-contract.bin` files)

cctdaniel:
> It seems to be more than ABIs. Do we need the contract binaries as well? (I mean the`pyth-contract.bin` files)

I want to err on the side of transparency here unless there's a good reason we don't want them on the repo

PR #1658, opened 2024-06-05
feat(target_chains/solana): relax rust sdk dependencies

ali-bahjati:
Some downstream consumers are using older versions of anchor and solana and cannot easily upgrade their code. Downgrading the requirement on the SDK won't have no impact on us as our lock file is unchanged and previous versions are being used. (To be more precise, we use anchor 0.28.0 because the main program depends on Wormhole which is locked to it)

### DIFF
--- a/target_chains/solana/Cargo.lock
+++ b/target_chains/solana/Cargo.lock
@@ -1030,7 +1030,7 @@ dependencies = [
  "pyth-sdk",
  "pyth-sdk-solana",
  "pyth-solana-receiver",
- "pyth-solana-receiver-sdk 0.3.0",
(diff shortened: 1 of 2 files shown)


guibescos:
This change is challenging because we also need to make `pythnet-sdk` compatible with anchor 0.26.0, but that's hard because 
anchor 0.26.0 uses borsh 0.9.0 while anchor 0.28.0 uses borsh 0.10.0.
Our manual implementation of `BorshSerialize` and `BorshDeserialize` for `PostedMessageUnreliableData` in pythnet_sdk breaks with borsh 0.9.0.

PR #1659, opened 2024-06-05
[contract_manager] Add a script to list wormhole contracts

jayantk:
We have a couple of these for other contract types, and this was useful for berachain debugging so let's merge it.

### DIFF
--- a/contract_manager/scripts/list_wormhole_contracts.ts
+++ b/contract_manager/scripts/list_wormhole_contracts.ts
@@ -1,0 +1,40 @@
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+import { DefaultStore, EvmWormholeContract } from "../src";
+
+const parser = yargs(hideBin(process.argv))
+  .usage("Usage: $0")
+  .options({
+    testnet: {
+      type: "boolean",
+      default: false,
+      desc: "Fetch testnet contracts instead of mainnet",
+    },
+  });
+
+async function main() {
+  const argv = await parser.argv;
+  const entries = [];
+  for (const contract of Object.values(DefaultStore.wormhole_contracts)) {
+    if (contract instanceof EvmWormholeContract) {
+      if (contract.getChain().isMainnet() === argv.testnet) continue;
+
+      try {
+        const index = await contract.getCurrentGuardianSetIndex();
+        const chainId = await contract.getChainId();
+        entries.push({
+          chain: contract.getChain().getId(),
+          contract: contract.address,
+          guardianSetIndex: index,
+          chainId: chainId,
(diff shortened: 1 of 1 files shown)


anihamde (approved):

contract_manager/scripts/list_wormhole_contracts.ts
```diff
@@ -0,0 +1,40 @@
...
+      desc: "Fetch testnet contracts instead of mainnet",
+    },
+  });
+
+async function main() {
+  const argv = await parser.argv;
+  const entries = [];
+  for (const contract of Object.values(DefaultStore.wormhole_contracts)) {
+    if (contract instanceof EvmWormholeContract) {
+      if (contract.getChain().isMainnet() === argv.testnet) continue;
```
cprussin, line 20: This is a minor nit but I definitely prefer:

```
if (contract instanceof EvmWormholeContract && contract.getChain().isMainnet() !== argv.testnet) {
```

on the line above -- the reason is that I generally try to avoid control-flow-breaking syntax like `continue`, because it makes it much harder to look through the code and understand that an early exit could have happened somewhere that isn't represented by the branch condition.

contract_manager/scripts/list_wormhole_contracts.ts
```diff
@@ -0,0 +1,40 @@
...
+
+      try {
+        const index = await contract.getCurrentGuardianSetIndex();
+        const chainId = await contract.getChainId();
```
cprussin, line 24: In general it's preferable to do `const [index, chainId] = await Promise.all([ contract.getCurrentGuardianSetIndex(), contract.getChainId() ]);` so that the async calls can be parallelized

contract_manager/scripts/list_wormhole_contracts.ts
```diff
@@ -0,0 +1,40 @@
...
+          contract: contract.address,
+          guardianSetIndex: index,
+          chainId: chainId,
+        });
+        console.log(`Fetched contract for ${contract.getId()}`);
+      } catch (e) {
+        console.error(`Error fetching contract for ${contract.getId()}`, e);
+      }
+    }
+  }
```
cprussin, line 36: You can restructure this to parallelize out all the contracts too by doing something like:

```ts
const contracts = await Promise.all(Object.values(
    DefaultStore.wormhole_contracts)
        .filter(contract => contract instanceof EvmWormholeContract && contract.getChain().isMainnet() !== argv.testnet)
        .map(async (contract) => {
            const [index, chainId] = await Promise.all(...);
            ...
        })
));
```

cprussin (approved):

PR #1665, opened 2024-06-06
refactor(target_chains/starknet): rename ByteArray to ByteBuffer

Riateche:
Using the same name as the core library type (that's even included in prelude) can be confusing to users. It's probably better to use a different name. Naming suggestions are welcome.


ali-bahjati (approved):
ByteBuffer sounds good!

PR #1667, opened 2024-06-06
feat(target_chains/starknet): add all possible derives

Riateche:
I'm being liberal with Copy and Default impls because they are needed when using the type in a map or storage.


ali-bahjati (approved):
Nice! i hope that the compiler be smart enough with the copy lol.

PR #1676, opened 2024-06-07
feat(target_chains/starknet): add JS SDK

Riateche:
This is a pretty minimalistic library. I've used Aptos SDK as a template.


target_chains/starknet/sdk/js/package.json
ali-bahjati, line 1: can you also add this package to the package.json in the root of the repo?

ali-bahjati (approved):
Nice!

I recommend adding a CI to make sure ABIs are up-to-date in later PRs.

target_chains/starknet/sdk/js/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  rules: {
+    "@typescript-eslint/no-explicit-any": "off",
```
ali-bahjati, line 7: do we need this anywhere? would be nice if we remove it

target_chains/starknet/sdk/js/README.md
```diff
@@ -0,0 +1,114 @@
...
+);
+
+// Approve fee withdrawal.
+strkErc0Contract.connect(account0);
+let tx = await strkErc0Contract.approve(pythContract.address, fee);
+await provider.waitForTransaction(tx.transaction_hash);
+
+// Create a transaction and submit to your contract using the price update data.
+pythContract.connect(account0);
+tx = await pythContract.update_price_feeds(pythUpdate);
```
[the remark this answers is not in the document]
Riateche: Starknet-js reads the ABI object to understand the native type of every argument. The conversion from JS objects to Starknet calldata depends on that type. In this case, we're working with a struct declared by us, so starknet-js expects us to supply an object with the corresponding property names (`num_last_bytes` and `data`). It simply reads these properties from our object and converts them (in case of `data`, we represent each item as a string containing a HEX number, and starknet-js is smart enough to convert it to `felt252`).

target_chains/starknet/sdk/js/package.json
Riateche: Added to `pnpm-workspace.yaml`.

target_chains/starknet/sdk/js/.eslintrc.js
Riateche: Removed.

ali-bahjati (approved):

PR #1678, opened 2024-06-08
Migrate to pnpm

cprussin:
This PR migrates us from npm to pnpm.

I tried as painstakingly as possible to ensure all dependencies were the same (using [pnpm import](https://pnpm.io/cli/import) helps a lot), but it's simply not possible to guarantee, in large part because there were a ton of dependencies that were not explicitly declared and only worked due to npm hoisting.

Based on some various issues I had to fix, it's clear that some dependencies have skewed.  All packages are passing tests/typescripts/lint etc so it's reasonable to think the version changes are minor, but I will really need everyone's help running some simple manual tests for things and making sure that all the packages work.

I was able to test the xc-admin-frontend package but I don't know enough about the other packages to be able to manually test them with any kind of confidence.

I've also tried to update all documentation as well as any tasks to use `pnpm exec`, `pnpm run`, `pnpm dlx`, etc instead of `npx`, `npm run`, etc.  However I almost certainly missed a bunch of places and might have applied it to places where it shouldn't be applied, will appreciate any eyes on that as well!


.envrc
ali-bahjati, line 1: This gives me an error `zsh: command not found: nix-shell`. Not a big deal, but can you change it in a way that people without nix won't be affected?

contract_manager/package.json
```diff
@@ -23,20 +23,32 @@
...
     "@coral-xyz/anchor": "^0.29.0",
+    "@cosmjs/cosmwasm-stargate": "^0.32.3",
+    "@cosmjs/stargate": "^0.32.3",
     "@injectivelabs/networks": "^1.14.6",
     "@mysten/sui.js": "^0.49.1",
+    "@pythnetwork/client": "^2.21.0",
+    "@pythnetwork/contract-manager": "workspace:*",
     "@pythnetwork/cosmwasm-deploy-tools": "*",
     "@pythnetwork/entropy-sdk-solidity": "*",
     "@pythnetwork/price-service-client": "*",
```
ali-bahjati, line 34: why are these not changed?
and also why for some we've `workspace:^` and for some `workspace:*`?

ali-bahjati (approved):
Very nice, thank you!

.envrc
cprussin: Ah yes for sure, apologies, I actually didn't even consider folks having direnv installed without having nix installed!  I will fix

contract_manager/package.json
cprussin: I actually didn't manually add these, they came in by running `pnpm add`.  However, they are documented at https://pnpm.io/workspaces#publishing-workspace-packages -- basically, they affect how the dependency versions get modified when publishing.  Upon publishing, uses of `workspace:*` will be replaced with an exact version with no prefix (i.e. `1.0.0`), where `workspace:^` will be replaced with a carat-prefixed dependency (i.e. `^1.0.0`).

If we're using semver correctly then presumably `workspace:^` is the more semantically correct way to do this, and probably should be the default.  I'm not sure what the heuristic is that the pnpm cli uses to select which to use when running `pnpm add` but IMO we should leave it as-is for this release and should audit these later.

PR #1679, opened 2024-06-10
fix(apps/hermes/client): use camelCase for streaming params

cctdaniel:


package-lock.json
```diff
@@ -54903,7 +54903,7 @@
     },
     "target_chains/ethereum/sdk/js": {
       "name": "@pythnetwork/pyth-evm-js",
-      "version": "1.51.0",
+      "version": "1.52.0",
```
cctdaniel, line 54906: this seemed to be changed because previously when `@pythnetwork/pyth-evm-js` was bumped to 1.52.0 but the `package-lock.json` wasnt updated

ali-bahjati (approved):

PR #1681, opened 2024-06-10
fix: delete removed packages

guibescos:
These packages where deleted here https://github.com/pyth-network/pyth-crosschain/pull/1653


cprussin (approved):
Oh nice, thank you, I didn't notice these got removed!

PR #1683, opened 2024-06-11
fix(typescript): add missing dependencies for build

ali-bahjati:
lerna build was not working due to some missed typescript dependencies.


cprussin (approved):
Apologies on missing this!

PR #1686, opened 2024-06-12
chore(docker): fix & rewrite Node Dockerfiles

cprussin:
The new versions of these are designed to properly take advantage of pnpm and to also minimize output size.

The builder image no longer is built on layers starting with just the package.json files.  The reason for this is that with pnpm's content-addressed store and the docker buildkit cache mounts, rebuilding the layers adds only trivial overhead if dependencies haven't changed, and removing the Docker layering to attempt to cache images with just node modules reduces complexity significantly.

The runner image is now based on alpine and is drastically smaller.  We use `pnpm export` (or, in the case of nextjs, we use standalone builds) to populate the runner images with only the necessary packages and dependencies to further reduce image sizes.


governance/xc_admin/Dockerfile
```diff
@@ -1,18 +0,0 @@
-# Defined in Dockerfile.lerna
```
cprussin, line 1: @guibescos fyi I moved this to `governance/xc_admin/packages/crank_executor/Dockerfile` which seemed like the more correct home

ali-bahjati: Hmmmm this one is a bit tricky. crank-executor, crank-pythnet-relayer and proposer-server are all 3 different services that use this image. Is there a way to keep them all three?

ali-bahjati: I think it's good to just have them as a multi service docker in `/svc/`?

ali-bahjati (approved):
Looks very nice! thanks.

I think we can address xc-admin dockerfile in a later PR.

governance/xc_admin/Dockerfile
cprussin: As a pattern I generally prefer to avoid multi-service images for 2 reasons:

1. It means the image isn't optimized for any specific use case, which is one of the main superpowers of containerization in the first place
2. Doing things with multi-service images is a crutch that folks can use to avoid building good tooling to make it easy and cheap to spin up purpose-built images.  I don't think that crutch is good in the long run and I'd rather feel the pain of not having those tools and be compelled to add tools to make it easier.

That's my opinion, thoughts?

governance/xc_admin/packages/xc_admin_frontend/package.json
```diff
@@ -33,6 +33,7 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-hot-toast": "^2.4.0",
+    "sharp": "^0.33.4",
```
guibescos, line 36: why is this needed

cprussin: Sharp should be installed  for any production environment running nextjs (outside e.g.g Vercel where it is already part of the environment) and is used for image optimizations, but [it's actually mandatory for standalone mode](https://nextjs.org/docs/messages/sharp-missing-in-production) which the new `Dockerfile` uses to significantly reduce the size of the resulting image.

apps/price_pusher/docker-compose.beta.sample.yaml
```diff
@@ -1,9 +1,10 @@
 services:
   price-pusher:
-    # Use this to build the price pusher from source. A dependency of the pusher docker
-    # image is the pyth-crosschain monorepo lerna docker image. Build lerna image
-    # using the following command from the repo root:
-    # `docker buildx build -f Dockerfile.lerna -t lerna .`
+    # Use this to build the price pusher from source. The pusher docker image
```
ali-bahjati, line 3: I think we should update these.

ali-bahjati (approved):
This one is very clever. Nice!

apps/price_pusher/docker-compose.beta.sample.yaml
cprussin: Updated @ali-bahjati , see https://github.com/pyth-network/pyth-crosschain/pull/1686/commits/fe0e96f4f53f21d5518e282719e96515d4eeea5a

PR #1692, opened 2024-06-13
feat(target_chains/fuel): add minimal js sdk

ali-bahjati:
This change is duplicate of #1690 but is going to merge on main and fixes a couple of TS issues in the monorepo to make the CI work.


governance/multisig_wh_message_builder/package.json
ali-bahjati, line 1: changes are in attempt to fix the issues here. the new versions are matching versions as xc_admin_cli which has similar logic. We will soon remove this package.

target_chains/ethereum/contracts/package.json
```diff
@@ -31,8 +31,8 @@
     "@matterlabs/hardhat-zksync-deploy": "^0.6.2",
     "@matterlabs/hardhat-zksync-solc": "^0.3.14",
     "@nomiclabs/hardhat-etherscan": "^3.1.7",
-    "@openzeppelin/contracts": "^4.5.0",
-    "@openzeppelin/contracts-upgradeable": "^4.5.2",
+    "@openzeppelin/contracts": "=4.8.1",
+    "@openzeppelin/contracts-upgradeable": "=4.8.1",
```
ali-bahjati, line 35: The contracts might get changed in each upgrade (and actually were). This is the version we were using before the migration to pnpm. The newer versions (although semver compatible) would result in compilation failure.

target_chains/sui/cli/package.json
```diff
@@ -16,7 +16,7 @@
     "@pythnetwork/contract-manager": "workspace:*",
     "@pythnetwork/price-service-client": "^1.4.0",
     "@pythnetwork/price-service-sdk": "^1.2.0",
-    "@pythnetwork/xc-admin-common": "^0.0.0",
+    "@pythnetwork/xc-admin-common": "workspace:*",
```
ali-bahjati, line 19: missed in the migration.

cctdaniel (approved):

.github/workflows/ci-typescript.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          run_install: true
+          run_install: "args: [--frozen-lockfile]"
```
ali-bahjati, line 33: This avoids unexpected behaviours.

cprussin: :thinking: this shouldn't be necessary since [it's on by default in CI environments](https://pnpm.io/cli/install#--frozen-lockfile).

If we find it actually is necessary then we should add it in all the github actions where we use `pnpm/action-setup`

ali-bahjati: idk why but enabling this actually gave me the error to understand what's wrong. apparently before that it was installing without any check.

cprussin: hmm could be because of being on pnpm 8.x and not 9.x, I'll see if upgrading to 9.x resolves this...

PR #1693, opened 2024-06-13
feat(express-relay): Add permit2

danimhr:
This PR aims to add permit2 compatibility to express relay sdks

### DIFF
--- a/express_relay/sdk/js/src/index.ts
+++ b/express_relay/sdk/js/src/index.ts
@@ -59,6 +59,19 @@ export function checkTokenQty(token: {
   };
 }
 
+class ChainConfig {
+  wethAddress: Address;
+  opportunityAdapterAddress: Address;
+  constructor(wethAddress: Address, opportunityAdapterAddress: Address) {
+    this.wethAddress = wethAddress;
+    this.opportunityAdapterAddress = opportunityAdapterAddress;
+  }
+}
+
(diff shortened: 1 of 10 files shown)


express_relay/sdk/js/src/index.ts
```diff
@@ -59,6 +59,19 @@ export function checkTokenQty(token: {
   };
 }
 
+class ChainConfig {
+  wethAddress: Address;
+  opportunityAdapterAddress: Address;
```
anihamde, line 64: does the `chainId` not also fit in here?

express_relay/sdk/js/src/index.ts
```diff
@@ -303,6 +318,35 @@ export class Client {
...
+  getPermittedTokens(
+    sellTokens: TokenAmount[],
+    bidAmount: bigint,
+    value: bigint,
+    wethAddress: Address
+  ) {
+    const permittedTokens = sellTokens.map((token) => ({
+      ...token,
+    }));
+    const wethAmount = bidAmount + value;
```
anihamde, line 330: weth can also be a sellToken, so this may be better named `wethAmountBidValue` or something to distinguish it from weth in the sellTokens

express_relay/sdk/js/src/index.ts
danimhr: Yes, but we are not using it in the sdk.

PR #1697, opened 2024-06-13
Upgrade pnpm

cprussin:
This PR does a few related changes, all related to locking the node / pnpm versions:

- Upgrades pnpm to ~9.3.0 (latest)~ 9.2.0 since it doesn't look like Vercel supports 9.3.0 yet
- Locks the pnpm and node versions via the `engines` field of `package.json`
- Sets the `package-manager-strict`, ~`package-manager-strict-version`~ options for pnpm to enforce that the correct pnpm version is in use *Edit*: disabling `package-manager-strict-version` since corepack doesn't appear to work correctly on some light testing with @cctdaniel  and we have no other way to easily lock the pnpm version on dev machines
- Configures github actions to use the node version from the `engines` field
- Adds an `.nvmrc` file
- Downgrades the node versions used in `Dockerfile.node` to match the versions used in Vercel, the `package.json` `engines` field, and the `.nvmrc` file
- Upgrades nix shell configuration
- Moves nix config over to using flakes


cctdaniel (approved):
seeing `ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)` in vercel build logs

cprussin:
Yeah, looks like vercel's node and pnpm versions are abit behind so I'll have to relax these a bit

ali-bahjati (approved):
Nice!

PR #1703, opened 2024-06-14
fix(target_chains/fuel): update js sdk to properly gen and export types

ali-bahjati:
This change adds type generation to the build path to have them in the resulting library and also fixes the issues around type exports by updating the typescript and using `export type *`. The problem was that the normal `export *` adds a similar export to the produced js file but there is no module to be exported from in our case.


target_chains/fuel/sdk/js/src/types/factories/PythContractAbi__factory.ts
ali-bahjati, line 1: drive-by: generating types again

cprussin (approved):
LGTM but @cctdaniel and I spoke and I think it's very unwise to be 1) generating type declarations into a src dir and 2) re-exporting generated types.  It's really not how typescript packaging is intended to work and I think we're very likely going to run into weird limitations because we're kind of abusing typescript here.

I think it would be wise to eventually do one of two things:

1) Move the generated types to a separate package
2) Move the generated types to a separate export using [subpath exports](https://nodejs.org/api/packages.html#subpath-exports)

I don't think we need to do either for this PR, but I'm noting this because I think we should resist copying this pattern to another package when the opportunity comes

cprussin:
Note the eslint error in CI `Parsing error: DeprecationError: 'originalKeywordKind' has been deprecated since v5.0.0 and can no longer be used. Use 'identifierToKeywordKind(identifier)' instead` is happening because you upgraded typescript but didn't upgrade the `@typescript-eslint/parser`.  The 5.x range of the `@typescript-eslint/*` packages are compatible with typescript 4.x.

cctdaniel (approved):

ali-bahjati:
@cprussin I totally hear you on this. Please feel free to modify the structure here. This PR just solves the bugs that exists which prevented me to run contract manager scripts totally irrelevant to this change.

PR #1705, opened 2024-06-16
fix(target_chains/fuel): audit fix

cctdaniel:
- **fix PN-01: Guardians are assumed trusted and no input checks are done**
- **refactor and add comments**

### DIFF
--- a/target_chains/fuel/contracts/pyth-interface/src/data_structures/price.sw
+++ b/target_chains/fuel/contracts/pyth-interface/src/data_structures/price.sw
@@ -1,6 +1,6 @@
 library;
 
-use std::bytes::Bytes;
+use std::{block::timestamp, bytes::Bytes};
 
 use ::errors::PythError;
 use ::utils::absolute_of_exponent;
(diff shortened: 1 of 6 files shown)


target_chains/fuel/contracts/pyth-interface/src/data_structures/price.sw
```diff
@@ -175,6 +175,12 @@ impl PriceFeed {
         //convert publish_time from UNIX to TAI64
         publish_time += TAI64_DIFFERENCE;
 
+        // If a quorum of Wormhole Guardians were to be compromised or malicious they could sign a price update far out into the future (say u64::max()) , preventing any further updates since no new price can have a later `publish_time`, so we check here that the `publish_time` is not a future timestamp
+        require(
+            timestamp() >= publish_time,
+            PythError::PriceFeedPublishTimeInTheFuture,
+        );
```
ali-bahjati, line 182: If the guardians get compromised they can really do anything (modify the price) and this prevention here won't have a significant impact.
This check is fine if the timestamp() on Fuel sequencer never falls behind. Maybe we ask Fuel guys before merging it in.

ali-bahjati (approved):

PR #1708, opened 2024-06-17
[fortuna] Automated fee adjustment based on gas prices

jayantk:
This should save us from having to keep dialing up/down the blast fees. I've tried to set up the logic so that it ensures the provider is always profitable and doesn't adjust the fees too often, but we'll probably have to tune this. See inline comment explaining the update logic.

### DIFF
--- a/apps/fortuna/config.sample.yaml
+++ b/apps/fortuna/config.sample.yaml
@@ -12,6 +12,12 @@ chains:
     # How much to charge in fees
     fee: 1500000000000000
 
+    # Configuration for dynamic fees under high gas prices. The keeper will set
+    # on-chain fees to make between [min_profit_pct, max_profit_pct] of the max callback
+    # cost in profit per transaction.
+    min_profit_pct: 10
+    max_profit_pct: 100
+
     # Historical commitments -- delete this block for local development purposes
     commitments:
       # prettier-ignore
--- a/apps/fortuna/config.sample.yaml
+++ b/apps/fortuna/config.sample.yaml
@@ -15,7 +15,8 @@ chains:
     # Configuration for dynamic fees under high gas prices. The keeper will set
     # on-chain fees to make between [min_profit_pct, max_profit_pct] of the max callback
     # cost in profit per transaction.
-    min_profit_pct: 10
+    min_profit_pct: 0
+    target_profit_pct: 20
     max_profit_pct: 100
 
     # Historical commitments -- delete this block for local development purposes
--- a/apps/fortuna/src/config.rs
+++ b/apps/fortuna/src/config.rs
@@ -147,7 +147,13 @@ pub struct EthereumConfig {
 
     /// The minimum percentage profit to earn as a function of the callback cost.
     /// For example, 20 means a profit of 20% over the cost of the callback.
+    /// The fee will be raised if the profit is less than this number.
     pub min_profit_pct: u64,
+
+    /// The target percentage profit to earn as a function of the callback cost.
+    /// For example, 20 means a profit of 20% over the cost of the callback.
+    /// The fee will be set to this target whenever it falls outside the min/max bounds.
+    pub target_profit_pct: u64,
 
     /// The maximum percentage profit to earn as a function of the callback cost.
     /// For example, 100 means a profit of 100% over the cost of the callback.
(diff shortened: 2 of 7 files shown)


apps/fortuna/src/keeper.rs
```diff
@@ -940,3 +959,195 @@ pub async fn withdraw_fees_if_necessary(
...
+///
+/// These conditions are designed to prevent the
+pub async fn adjust_fee_if_necessary(
+    contract: Arc<InstrumentedSignablePythContract>,
+    provider_address: Address,
+    legacy_tx: bool,
+    gas_limit: u64,
+    min_profit_pct: u64,
+    max_profit_pct: u64,
+    min_fee: u128,
```
m30m, line 1018: maybe rename this to `fixed_min_fee` to avoid confusion

apps/fortuna/src/keeper.rs
```diff
@@ -940,3 +959,195 @@ pub async fn withdraw_fees_if_necessary(
...
+    let provider_fee: u128 = provider_info.fee_in_wei;
+    if is_chain_active
+        && ((provider_fee > target_fee_max && can_reduce_fees) || provider_fee < target_fee)
+    {
+        tracing::info!(
+            "Adjusting fees. Current: {:?} Target: {:?}",
+            provider_fee,
+            target_fee
+        );
+        let contract_call = contract.set_provider_fee_as_fee_manager(provider_address, target_fee);
```
m30m, line 1077: I think the what we set the value to should be something between target_fee(which can be renamed to target_fee_min) and target_fee_max. Otherwise, small changes in one direction can trigger updates continuously.

apps/fortuna/src/keeper.rs
```diff
@@ -940,3 +959,195 @@ pub async fn withdraw_fees_if_necessary(
 
     Ok(())
 }
+
```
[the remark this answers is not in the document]
jayantk: i would rather do this separately and refactor all of the similar functions here so we don't end up in an inconsistent state

apps/fortuna/src/keeper.rs
jayantk: ah good point

apps/fortuna/config.sample.yaml
```diff
@@ -12,6 +12,13 @@ chains:
     # How much to charge in fees
     fee: 1500000000000000
 
+    # Configuration for dynamic fees under high gas prices. The keeper will set
+    # on-chain fees to make between [min_profit_pct, max_profit_pct] of the max callback
+    # cost in profit per transaction.
+    min_profit_pct: 0
+    target_profit_pct: 20
```
m30m, line 19: I'm not sure if we want this level of customisation. We can set this as the average of min_profit_pct and max_profit_pct.

apps/fortuna/src/keeper.rs
```diff
@@ -288,6 +291,23 @@ pub async fn run_keeper_threads(
         .in_current_span(),
     );
 
+    // Spawn a thread that periodically adjusts the provider fee.
+    spawn(
+        adjust_fee_wrapper(
+            contract.clone(),
+            chain_state.provider_address.clone(),
+            ADJUST_FEE_INTERVAL,
+            chain_eth_config.legacy_tx,
```
m30m, line 300: maybe just pass the whole config object down?

apps/fortuna/src/config.rs
```diff
@@ -145,6 +145,22 @@ pub struct EthereumConfig {
...
+
+    /// The target percentage profit to earn as a function of the callback cost.
+    /// For example, 20 means a profit of 20% over the cost of the callback.
+    /// The fee will be set to this target whenever it falls outside the min/max bounds.
+    pub target_profit_pct: u64,
+
+    /// The maximum percentage profit to earn as a function of the callback cost.
+    /// For example, 100 means a profit of 100% over the cost of the callback.
+    /// The fee will be lowered if it is more profitable than specified here.
+    /// Must be larger than min_profit_pct.
```
m30m, line 161: we should probably check this `min_profit_pct < target_profit_pct < max_profit_pct` inequality somewhere. I can imagine a misconfiguration resulting in constant fee updates and draining the fee manager because of gas usage.

m30m (approved):
It would be great if you can add a metric inside `track_provider` function to gauge the fee. Otherwise, it would be difficult to monitor how the system is working

PR #1712, opened 2024-06-18
[xc-admin-frontend] Unify publisher keys file schema with observer and balance tracker

mcamou:
Deployment of this in the k8s node needs to be coordinated with observer, [balance tracker](https://github.com/pyth-network/balance-tracker/pull/4) and the deduplication of key files in the deployment repo.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
@@ -16,12 +16,15 @@ import '../mappings/signers.json'
 
 const readPublisherKeyToNameMapping = (filename: string) => {
   if (fs.existsSync(filename)) {
-    return YAML.parse(fs.readFileSync(filename, 'utf8')).reduce((acc, rec) => {
-      acc[rec.key] = rec.name
-      return acc
-    })
+    return YAML.parse(fs.readFileSync(filename, 'utf8')).reduce(
+      (acc: { [key: string]: string }, rec: { key: string; name: string }) => {
+        acc[rec.key] = rec.name
+        return acc
+      },
+      {}
+    )
   } else {
-    return {} as { string: string }
+    return {}
   }
 }
 
--- a/governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
@@ -20,15 +20,17 @@ const keyToNameMappingSchema = z.array(
 )
 
 const readPublisherKeyToNameMapping = async (filename: string) => {
+  let data = ''
   try {
-    await fs.promises.access(filename)
-    const arr = keyToNameMappingSchema
-      .parse(YAML.parse(await fs.promises.readFile(filename, 'utf8')))
-      .map((key, name) => [key, name])
-    return Object.fromEntries(arr)
+    data = await fs.promises.readFile(filename, 'utf8')
   } catch {
     return {}
   }
+
+  const yaml = YAML.parse(data)
+
+  const arr = await keyToNameMappingSchema.parseAsync(yaml)
+  return Object.fromEntries(arr.map((key, name) => [key, name]))
 }
 
 export const getServerSideProps: GetServerSideProps = async () => {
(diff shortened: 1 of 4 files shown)


governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
```diff
@@ -13,26 +14,28 @@ import { StatusFilterProvider } from '../contexts/StatusFilterContext'
 import { classNames } from '../utils/classNames'
 import '../mappings/signers.json'
 
+const readPublisherKeyToNameMapping = (filename: string) => {
+  if (fs.existsSync(filename)) {
```
cprussin, line 18: I would use the promisified version of this instead of using the sync version, the sync version is blocking and using the promisified version and turning this into an async function will enable you to use `Promise.all` to do the io in parallel

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
```diff
@@ -13,26 +14,28 @@ import { StatusFilterProvider } from '../contexts/StatusFilterContext'
...
+    return YAML.parse(fs.readFileSync(filename, 'utf8')).reduce(
```
cprussin, line 19: In general there's not much reason to use `reduce`, it's much harder to grok for anyone who doesn't have a good FP background, and there are clearer ways to do nearly everything it does, see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md.

In this case I'd probably do this as:

```ts
return Object.fromEntries(arr.map(({ key, name }) => [key, name]))
```

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
```diff
@@ -13,26 +14,28 @@ import { StatusFilterProvider } from '../contexts/StatusFilterContext'
...
+      (acc: { [key: string]: string }, rec: { key: string; name: string }) => {
```
cprussin, line 20: Note that the type casting here removes typescript's ability to actually guarantee anything about the types.  IMO this shouldn't even exist in typescript.

Typically any time I load data from an external source I use [zod](https://zod.dev/) to parse the data so I can guarantee that my typescript types are actually correct and can handle invalid data accordingly, rather than typecasting and assuming the types are accurate.

cprussin (changes requested):
Minor suggestions here, let me know if you need any specific help with any of these!

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
mcamou: yeah I thought of that but since it's only done once during startup and these are fairly small files (<30K) I thought it would be clearer to just take the minimal I/O hit.

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
mcamou: OK, the last time I used TS it was on a team with a functional background so it's where I will usually go. I'll change it.

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
mcamou: yeah I know... the original code didn't have any typechecking either so I put in the cast to shut the compiler up. Since this is a file that we control (and that at least 2 other projects will use) I think it's fairly safe to do this, WDYT?

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
mcamou: OK, this one is done.

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
cprussin: In general with modern node, there's not much of a good reason to prefer sync over async.  It's semantically incorrect and doesn't produce any better ergonomics since you could very easily use the async version here and just add `await` on line 34 and 35 and achieve pretty much the exact ergonomics you have here without the semantic incorrectness (it would still be blocking but at least it would be blocking in an obvious way at the callsite).

I don't know any good reason to use the sync functions nowadays.

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
cprussin: as a side note I also think the amount of things people do in typescript to silence the compiler without actually fixing the underlying type safety issues is a huge problem and it's a big reason why typescript isn't nearly as reliable as something like rust.... good discipline pays off IMHO.

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
cprussin: Oh actually I just remembered that `exists` has been [deprecated in node](https://nodejs.org/api/fs.html#fsexistspath-callback), one of the reasons is that doing this causes race conditions (obviously not in this case, but in the general case, a file could be removed between the call to `exists` and the call to `readFile`).

The more "correct" way to do this is to use `readFile` in a try-catch and handle `ENOENT` as a "file does not exist" error

governance/xc_admin/packages/xc_admin_frontend/pages/index.tsx
mcamou: I *could* be persuaded to add some rudimentary validation and throwing exceptions (or `os.exit(1)`) on failure.

cprussin: You could do that, but it will be more work and less flexible than just using zod.  We need zod for this app anyways so it wouldn't be like it's a dependency I don't plan to introduce regardless.

cprussin (approved):

PR #1714, opened 2024-06-19
feat(apps/price_pusher)!: switch to structured logger via pino

ali-bahjati:
This change switches from `console` to `pino` for logging that is a structured logger and allows us to have json formatted loggers. In adition to just that change, pino does an amazing job at logging the errors by including all their data + the stacktrace.

This change does minimal change to the log level and the `console.log` is changed to info or debug level based on my judgement.

This change is marked as breaking because it breaks any reliability/ops based on the logs.

The attached images show the new log outputs in both pretty and json formats, as well as error logs.

image
image


apps/price_pusher/src/logger.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type LogFormat = "json" | "pretty";
+
+export function createLogger(logLevel?: Level, logFormat?: LogFormat): Logger {
+  return pino({
+    level: logLevel ?? "info",
+    transport:
+      logFormat === "pretty"
+        ? {
+            target: "pino-pretty",
```
cprussin, line 14: I generally don't think `pino-pretty` should be installed directly into the code.  It's really designed to take the log stream and format it by piping and it's just as easy for any downstream consumer to pipe their log streams themselves if they want to, and the authors themselves [recommend against including it as a production dependency](https://github.com/pinojs/pino-pretty?tab=readme-ov-file#programmatic-integration).

I would instead remove the `logFormat` option and just document in the README or something how someone could use `pino-pretty` if they wanted pretty logs.

apps/price_pusher/src/logger.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export function createPriceServiceConnectionLogger(
+  logger: Logger
+): LoggerInterface {
+  return {
+    trace: logger.trace.bind(logger),
+    debug: logger.debug.bind(logger),
+    // Use debug level here because the info level is too noisy
+    info: logger.debug.bind(logger),
```
cprussin, line 31: Instead of doing this, I would recommend simply setting the log level on the child logger to be higher, and maybe exposing an option to pass the log level for price service connection logs.  By grouping the debug and info logs in the same level, you remove the ability for someone to ever be able to selectively separate them.

Presumably you can pass a `pino` instance directly as a `Logger` from `ts-log` since typescript isn't nominal and they are structurally compatible?

cprussin (changes requested):
Pino is awesome!  Great call moving this over.

Mostly LGTM but I'd definitely recommend removing `pino-pretty` from the production dependencies.

I'm not familiar with the interfaces enough here but is there an option to pass a logger instance to any entrypoint that a library would consume?  If not, I strongly recommend adding one -- it's very common for apps to have their own `pino` (or `bunyan`) instances and to want to use a child logger for a library.

apps/price_pusher/src/logger.ts
ali-bahjati: Done. I removed it entirely and updated the readme.

apps/price_pusher/src/logger.ts
ali-bahjati: Makes sense to me. With this change i removed additional complexity of added `logger.ts`

cprussin (approved):

apps/price_pusher/src/evm/command.ts
```diff
@@ -89,50 +91,62 @@ export default {
...
       {
-        logger: {
-          // Log only warnings and errors from the price service client
-          info: () => undefined,
-          warn: console.warn,
-          error: console.error,
-          debug: () => undefined,
-          trace: () => undefined,
-        },
+        // Swtich to warn level if log level is info to reduce the noise
```
jayantk, line 103: i find this confusing -- this is saying that the internal price service logger only logs at "warning" if we have info logs coming out of here?

jayantk: also this code block seems to be repeated in a few places. Maybe worth extracting this out as a factory method for PriceServiceConnection (?)

jayantk (approved):

apps/price_pusher/src/evm/command.ts
ali-bahjati: I'll update the comment. I think it's not that much given the whole command.ts is almost the same for all. I might refactor them entirely later.

PR #1716, opened 2024-06-20
feat(contract_manager): more support for starknet contract

Riateche:


contract_manager/src/contracts/starknet.ts
```diff
@@ -65,29 +67,146 @@ export class StarknetPriceFeedContract extends PriceFeedContract {
...
+    }
+  }
+
+  async executeUpdatePriceFeed(
     senderPrivateKey: PrivateKey,
     vaas: Buffer[]
   ): Promise<TxResult> {
-    throw new Error("Method not implemented.");
+    // We need the account address to send transactions.
+    throw new Error("Use executeUpdatePriceFeedWithAddress instead");
```
ali-bahjati, line 130: Isn't there any way to fetch the address from private key?

Riateche: It's possible to derive it but it depends on the wallet application used to create the account and on the constructor arguments. Looks like the mainstream way of handling it is just to pass the account address explicitly.

ali-bahjati: Since it's our own tooling, can't we just rely on that one?
if we had time, a better option would've been refactoring this code to have better interfaces.

ali-bahjati (approved):
I'm approving this but I think if we go back and revisit it later. This is a sign of bad interface design for us and not a problem in this PR specifically.

PR #1718, opened 2024-06-20
Update PythUtils.sol

Iam0TI:
created a Math Library to combine Pyth prices in Solidity.can easily combine 2 Pyth Prices into 1 (SOL:USD/ETH:USD=SOL/ETH).


ali-bahjati:

PR #1719, opened 2024-06-21
Math Library for Combining Pyth Prices on Solidity

holps-7:
**Description**
This is an implementation to replicate the Math library from Solana that can combine 2 Pyth Prices into 1 (SOL:USD/ETH:USD=SOL/ETH)

**Functions replicated in Solidity for Ethereum's combine library**
- getPriceInQuote
- div
- normalize
  - price and confidence are normalized to be between `MIN_PD_V_I64` and `MAX_PD_V_I64`. 
- scaleToExponent
- toUnsigned
  - Helper function to convert signed integers to unsigned and a sign bit, which simplifies some of the computations above.

- Added additional check to Return 0 values instead of throwing error so that it can be handled gracefully by the caller


ali-bahjati:

PR #1722, opened 2024-06-21
Create a Math Library to Combine Pyth Prices in Solidity

Ayushjhax:
As of today, only developers using Pyth in Rust language can easily combine 2 Pyth Prices into 1 (SOL:USD/ETH:USD=SOL/ETH).

The goal of this pull request is to create a Math Library in Solidity to combine Pyth prices. This library will provide the necessary functions to perform operations on Pyth prices, such as:

**getPriceInQuote:** Calculates the price of an asset in a different quote currency.
**getBorrowValuationPrice:** Calculates the valuation of a collateral position.
**getPriceInQuoteNoOld:** Calculates the price of an asset in a different quote currency, ensuring the result is not older than the specified age.
**getBorrowValuationPriceNoOld:** Calculates the valuation of a collateral position, ensuring the result is not older than the specified age.

The implementation of this **Math Library** is inspired by the Rust-based Pyth SDK, ensuring compatibility between the two implementations.

This pull request will allow Solidity developers to easily combine Pyth prices, expanding the use of the Pyth Network in the Ethereum ecosystem.


ali-bahjati:

PR #1723, opened 2024-06-22
Add remaining evm apis to api reference

cprussin:
This PR fills out all the remaining EVM apis for the API reference app.

The API reference app should be good to go for EVM after this PR, although there are a few UX cleanups I want to get to (especially cleaning up the arguments sidebar design).


apps/api-reference/src/apis/evm/get-ema-price-no-older-than.ts
```diff
@@ -0,0 +1,65 @@
+import Btc from "cryptocurrency-icons/svg/color/btc.svg";
+import Eth from "cryptocurrency-icons/svg/color/eth.svg";
```
cctdaniel, line 2: can we use the ones we are currently using on our website for standardization purposes?

apps/api-reference/src/apis/evm/common.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const HERMES_URL = "https://hermes.pyth.network";
+
+export const getLatestPriceFeed = async (feedId: string) => {
+  const url = new URL("/api/latest_price_feeds", HERMES_URL);
+  url.searchParams.set("ids[]", feedId);
+  url.searchParams.set("target_chain", "evm");
+  url.searchParams.set("binary", "true");
+  return safeFetch(priceFeedSchema, url);
+};
```
cctdaniel, line 43: `/api/latest_price_feeds` is v1 and its deprecated, we should use `/v2/updates/price/latest` instead

apps/api-reference/src/components/EvmApi/networks.ts
```diff
@@ -0,0 +1,18 @@
...
+export const getContractAddress = (networkId: number) =>
+  isSupportedNetwork(networkId)
+    ? NETWORK_TO_CONTRACT_ADDRESS[networkId]
+    : undefined;
+
+const NETWORK_TO_CONTRACT_ADDRESS = {
+  [mainnet.id]: "0x4305FB66699C3B2702D4d05CF36551390A4c69C6",
+  [avalanche.id]: "0x4305FB66699C3B2702D4d05CF36551390A4c69C6",
+  [arbitrum.id]: "0xff1a0f4744e8582DF1aE09D5611b887B6a12925C",
+  [sepolia.id]: "0xDd24F84d36BF92C65F92307595335bdFab5Bbd21",
```
cctdaniel, line 12: sepolia is not a network in its own, it usually refers to testnet, for e.g. eth sepolia, arbitrum sepolia, base sepolia are all evms but they have different contract addresses so we shouldnt assume that they are all deployed on the same address

cctdaniel (commented):
looks good, left some minor comments

apps/api-reference/src/apis/evm/common.ts
cprussin: ah, good to know, will update!

apps/api-reference/src/components/EvmApi/networks.ts
cprussin: "sepolia" here comes out of wagmi/chains which itself re-exports viem's chains definitions which uses `sepolia` as the name for Ethereum sepolia: https://github.com/wevm/viem/blob/main/src/chains/definitions/sepolia.ts

Here I'm not naming it "sepolia", I'm just using the definition out of viem.

I can rename the import declaration though to make it a bit clearer if you think that would help.

cprussin: Also note this object doesn't match a network _name_ to a contract address, it maps a network _id_ to the contract address.  The names aren't really important, they're just local variables.

aditya520: We will be replacing this `contract-manager` once it gets published.

apps/api-reference/src/apis/evm/get-ema-price-no-older-than.ts
cprussin: I think that's a good idea but because the icons are not packaged up nicely I'm going to wait until I can migrate the main website into the monorepo to avoid duplicating all the svgs and related code all over the place.

Once I can move that into the monorepo I'll extract the icons as a package and re-use it here.

apps/api-reference/src/apis/evm/common.ts
cprussin: Done, thanks!

PR #1728, opened 2024-06-24
fix(apps/hermes/client): allow baseURL with suffixes

mubaris:


apps/hermes/client/js/src/HermesClient.ts
```diff
@@ -100,7 +100,7 @@ export class HermesClient {
     query?: string;
     filter?: string;
   }): Promise<PriceFeedMetadata[]> {
-    const url = new URL("/v2/price_feeds", this.baseURL);
+    const url = new URL(`${this.baseURL}/v2/price_feeds`);
```
cctdaniel, line 103: A more robust approach would be to ensure that `this.baseURL` always ends with a slash before appending the path:

e.g.
```
const baseURL = this.baseURL.endsWith('/') ? this.baseURL : `${this.baseURL}/`;
const url = new URL('v2/price_feeds', baseURL);
```

The URL constructor ensures that the protocol and host parts of the base URL are preserved correctly, even if the base URL has additional path components.

If you could also make the changes for `getLatestPriceUpdates` and `getPriceUpdatesAtTimestamp` that would be appreciated!

cctdaniel (changes requested):
Thanks for catching the bug and opening a PR, left a comment and happy to look again once that's addressed.

apps/hermes/client/js/src/HermesClient.ts
mubaris: Have updated it

cctdaniel (approved):

cctdaniel:
thanks for your contribution! can you bump the package version?

mubaris:
done!

Issue #1704, opened 2024-06-14
Inaccurate Logging for Price Update Thresholds in `apps/price_pusher/src/controller.ts`


ericxor61a0c0d:
The console.log messages in the price update system reports `"None of the above values passed the threshold. No push needed."` as a catch-all. This message is misleading as it suggests a normal non-update condition. However, critical failures in data polling, corruption, staleness... will also lead to this message.

https://github.com/pyth-network/pyth-crosschain/blob/5952b9df12083fc9cd342361e7cc5e4bd3390d0a/apps/price_pusher/src/controller.ts#L72

PR #1730, opened 2024-06-25
refactor(apps/price_pusher): crash on RPC failures

ali-bahjati:
This PR updates the price pusher code to cause a crash when the RPC has issues. The RPC issues can be either the target network or Hermes.

This also solves #1704.

### DIFF
--- a/apps/price_pusher/src/solana/solana.ts
+++ b/apps/price_pusher/src/solana/solana.ts
@@ -35,14 +35,14 @@ export class SolanaPriceListener extends ChainPriceListener {
     const blockTime = await this.pythSolanaReceiver.connection.getBlockTime(
       slot
     );
-    if (blockTime !== undefined || blockTime < Date.now() / 1000 - 30) {
+    if (blockTime === null || blockTime < Date.now() / 1000 - 30) {
       throw new Error("Solana connection is unhealthy");
     }
   }
 
   async start() {
     // Frequently check the RPC connection to ensure it is healthy
-    setInterval(() => this.checkHealth.bind(this), 5000);
+    setInterval(this.checkHealth.bind(this), 5000);
 
     await super.start();
   }
@@ -125,6 +125,7 @@ export class SolanaPricePusher implements IPricePusher {
       this.logger.info({ signatures }, "updatePriceFeed successful");
     } catch (err: any) {
       this.logger.error(err, "updatePriceFeed failed");
+      return;
     }
   }
 }
--- a/apps/price_pusher/src/solana/solana.ts
+++ b/apps/price_pusher/src/solana/solana.ts
@@ -28,6 +28,25 @@ export class SolanaPriceListener extends ChainPriceListener {
     super(config.pollingFrequency, priceItems);
   }
 
+  // Checking the health of the Solana connection by checking the last block time
+  // and ensuring it is not older than 30 seconds.
+  private async checkHealth() {
+    const slot = await this.pythSolanaReceiver.connection.getSlot();
+    const blockTime = await this.pythSolanaReceiver.connection.getBlockTime(
+      slot
+    );
+    if (blockTime !== undefined || blockTime < Date.now() / 1000 - 30) {
+      throw new Error("Solana connection is unhealthy");
+    }
+  }
+
+  async start() {
+    // Frequently check the RPC connection to ensure it is healthy
+    setInterval(() => this.checkHealth.bind(this), 5000);
+
+    await super.start();
+  }
+
   async getOnChainPriceInfo(priceId: string): Promise<PriceInfo | undefined> {
     try {
       const priceFeedAccount =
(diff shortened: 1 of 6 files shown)


apps/price_pusher/src/sui/sui.ts
```diff
@@ -383,25 +363,20 @@ export class SuiPricePusher implements IPricePusher {
...
-      return ref;
+    const objectResponse = await provider.getObject({ id: ref.objectId });
+    if (objectResponse.data !== undefined) {
+      return {
+        digest: objectResponse.data!.digest,
+        objectId: objectResponse.data!.objectId,
+        version: objectResponse.data!.version,
+      };
+    } else {
+      throw new Error("Failed to refresh object reference");
```
m30m, line 379: I'm not sure how often this may happen, but if it fails frequently and causes crashes, there is a higher chance that we lock all the gas price objects

ali-bahjati: During my experiments I think I've resolved the issue of locked gases. It seems that it happens when we reuse a gas coin upon crash/restarts. I changed our deployment configuration to never allow to live instances at the same time and wait 10s before starting a new instance to make sure older ones are settled. No locking issues since this change.

I'll also run this code for sui for a day or so before merging to see how it works.

apps/price_pusher/src/solana/solana.ts
```diff
@@ -28,6 +28,25 @@ export class SolanaPriceListener extends ChainPriceListener {
     super(config.pollingFrequency, priceItems);
   }
 
+  // Checking the health of the Solana connection by checking the last block time
+  // and ensuring it is not older than 30 seconds.
+  private async checkHealth() {
```
guibescos, line 33: should we have a health check for Jito?

ali-bahjati: i throw an error on the failures there. Didn't do on Solana because it might fail due to landing issues.

apps/price_pusher/src/solana/solana.ts
```diff
@@ -28,6 +28,25 @@ export class SolanaPriceListener extends ChainPriceListener {
...
+    const slot = await this.pythSolanaReceiver.connection.getSlot();
+    const blockTime = await this.pythSolanaReceiver.connection.getBlockTime(
+      slot
+    );
+    if (blockTime !== undefined || blockTime < Date.now() / 1000 - 30) {
```
guibescos, line 38: this boolean is wrong
`if (blockTime && blockTime < Date.now() / 1000 - 30)`

apps/price_pusher/src/solana/solana.ts
```diff
@@ -28,6 +28,25 @@ export class SolanaPriceListener extends ChainPriceListener {
...
+      throw new Error("Solana connection is unhealthy");
+    }
+  }
+
+  async start() {
+    // Frequently check the RPC connection to ensure it is healthy
+    setInterval(() => this.checkHealth.bind(this), 5000);
```
guibescos, line 45: this needs to be `setInterval(this.checkHealth.bind(this), 5000);`

guibescos (changes requested):
I debugged it a bit, found couple of bugs

apps/price_pusher/src/sui/sui.ts
```diff
@@ -337,7 +317,7 @@ export class SuiPricePusher implements IPricePusher {
     ignoreGasObjects: string[],
     logger: Logger
   ): Promise<SuiObjectRef[]> {
-    const signerAddress = await signer.toSuiAddress();
+    const signerAddress = signer.toSuiAddress();
```
ali-bahjati, line 320: drive-by: linter

apps/price_pusher/src/sui/sui.ts
```diff
@@ -208,12 +194,6 @@ export class SuiPricePusher implements IPricePusher {
       provider,
       logger,
       priceServiceConnection,
-      pythPackageId,
-      pythStateId,
-      wormholePackageId,
-      wormholeStateId,
-      endpoint,
-      keypair,
```
ali-bahjati, line 216: drive-by: linter

apps/price_pusher/src/solana/solana.ts
```diff
@@ -125,6 +125,7 @@ export class SolanaPricePusher implements IPricePusher {
       this.logger.info({ signatures }, "updatePriceFeed successful");
     } catch (err: any) {
       this.logger.error(err, "updatePriceFeed failed");
+      return;
```
guibescos, line 128: this return doesn't seem to do anything

guibescos (approved):

apps/price_pusher/src/solana/solana.ts
ali-bahjati: lol it was there before and i removed it. added it back to minimize the changes.

PR #1733, opened 2024-06-25
PRNG Library with Pyth Entropy Integration

baileynottingham:
This pull request introduces a new Pseudorandom Number Generation (PRNG) library designed to work seamlessly with Pyth Entropy. The library provides a suite of functions for generating, manipulating, and converting random numbers from a Pyth Entropy seed.

Key Features:
- Utilizes Pyth Entropy as a secure seed source
- Includes functions for generating random bytes, integers, and permutations
- Provides utility functions for expanding randomness and type conversions
- Fully documented with NatSpec comments for easy integration and usage


target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,132 @@
...
+/// @title PRNG Library
+/// @notice A library for pseudorandom number generation and related utility functions
+/// @dev This PRNG library is designed to work with Pyth Entropy as the seed source.
+///      Pyth Entropy provides secure, rapid random number generation for blockchain applications,
+///      enabling responsive UX for NFT mints, games, and other use cases requiring randomness.
+library PRNG {
+    /// @notice Generate a random byte between 0 and 255
+    /// @param seed The Pyth Entropy seed (bytes32)
+    /// @return A random uint8 value
+    function randomByte(bytes32 seed) internal pure returns (uint8) {
```
jayantk, line 13: @baileynottingham can you think of a way to ensure that the user of this library doesn't accidentally pass the same Entropy seed to multiple of these random functions? E.g., the following code would be easy to write but also a security mistake:

```
bytes32 entropySeed = ... // get from entropy
unit8 rand = PRNG.randomByte(entropySeed);
uint256 rand2 = PRNG.randInt(entropySeed);
```

In most programming languages, a PRNG has internal state to track the current seed. Each time the user calls a function requesting randomness, the PRNG automatically updates that state.

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,132 @@
...
+                (uint256(keccak256(abi.encode(seed, i))) % (length - i));
+            (permutation[i], permutation[j]) = (permutation[j], permutation[i]);
+        }
+        return permutation;
+    }
+
+    /// @notice Convert bytes32 to uint256
+    /// @param value The bytes32 value to convert
+    /// @return The converted uint256 value
+    function toUint(bytes32 value) internal pure returns (uint256) {
```
jayantk, line 87: Instead of type conversion functions, I would rather see `randUint`, `randUint64` similar to the existing `randInt` method. The user will know how to do type conversions in solidity (imo).

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,132 @@
...
+    ) internal pure returns (uint256) {
+        require(max > min, "Max must be greater than min");
+        return (uint256(keccak256(abi.encode(seed))) % (max - min + 1)) + min;
+    }
+
+    /// @notice Expand a seed into an array of random bytes32 values
+    /// @param seed The Pyth Entropy seed (bytes32)
+    /// @param numHashes The number of random bytes32 values to generate
+    /// @return An array of random bytes32 values
+    function expandRandomness(
```
jayantk, line 52: (if the PRNG had internal state, then this method could be replaced by the user simply calling `randBytes32` repeatedly)

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
baileynottingham: Thank you for your input. I have updated the code to be a contract, allowing it to store an internal seed and nonce. The `nextBytes32()` function now returns a value and updates the nonce. This function is utilized internally by all other functions, ensuring that the nonce is consistently updated. As a result, the following code works correctly and returns a different number each time:
`
PRNG prng = new PRNG(pythEntropySeed);
uint256 randomNumber1 = prng.randUint();
uint256 randomNumber2 = prng.randUint();
uint256 randomNumber3= prng.randUint();
uint256 randomNumber4 = prng.randUint();
`

jayantk (commented):
I like the approach here. I left a couple minor comments. One ask though is for some unit tests.

I realize it's hard to test random code. I think the best approach is probably to hardcode the behavior you expect for a couple cases, just to validate that (1) the code runs, and (2) simple stuff like calling nextUint() repeatedly returns different values.

You can put your unit tests here https://github.com/pyth-network/pyth-crosschain/tree/main/target_chains/ethereum/contracts/forge-test and they will run with the rest of the contract tests.

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    /// @notice Generate a random uint64 value
+    /// @return A random uint64 value
+    function randUint64() internal returns (uint64) {
+        return uint64(uint256(nextBytes32()));
+    }
+
+    /// @notice Generate a random uint256 value within a specified range
+    /// @param min The minimum value (inclusive)
+    /// @param max The maximum value (inclusive)
```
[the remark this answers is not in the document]
baileynottingham: Updated max to be exclusive and added your note in the docstring

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,88 @@
...
+    ) internal returns (uint256) {
+        require(max > min, "Max must be greater than min");
+        return (randUint() % (max - min + 1)) + min;
+    }
+
+    /// @notice Generate a sequence of random bytes
+    /// @param length The number of random bytes to generate (max 32)
+    /// @return A bytes array of random values
+    function randomBytes(uint256 length) internal returns (bytes memory) {
+        require(length <= 32, "Length must be 32 or less");
```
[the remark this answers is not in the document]
baileynottingham: Yeah after thinking about this again, I agree it is redundant now with the nextBytes32() method. I went ahead and deleted it

target_chains/ethereum/entropy_sdk/solidity/PRNG.sol
```diff
@@ -0,0 +1,88 @@
...
+        bytes32 randomness = nextBytes32();
+        bytes memory result = new bytes(length);
+        assembly {
+            mstore(add(result, 32), randomness)
```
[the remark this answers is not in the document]
baileynottingham: Not applicable anymore since this is removed

baileynottingham:
Tests have been added!

jayantk (approved):

PR #1736, opened 2024-06-26
feat(target_chains/starknet): generate env file from deploy script

Riateche:
This is useful:
- to simplify local deployment instructions for examples.
- to set up CI for examples.


ali-bahjati (approved):
I think maybe getting an argument could make the script more verbose than using stdout and piping the normal stdout to stderr but lgtm.

Riateche:
Implementing it with named arguments like `deploy.sh --create-env file.env` would be nice, but it quickly gets complicated (I would also like to have deploy mode as an argument, etc.), and unfortunately doing this in bash is a pain.

PR #1738, opened 2024-06-27
feat(apps/hermes/client): allow authorization headers

cctdaniel:


apps/hermes/client/js/src/examples/HermesClient.ts
```diff
@@ -28,8 +28,29 @@ const argv = yargs(hideBin(process.argv))
   })
   .parseSync();
 
+function extractAuthorizationHeadersFromUrl(urlString: string): {
```
ali-bahjati, line 31: i recommend naming it `extractBasicAuthorizationHeadersFromUrl`. Also, can you add a doc to this document to say what it does?

ali-bahjati (approved):
Very nice!

PR #1746, opened 2024-06-29
Create a Math Library for Combining Pyth Prices on Solidity

0xHaku:
# Overview
This pull request includes the implementation of the `get_collateral_valuation_price`, `get_borrow_valuation_price`, and `get_price_in_quote` functions in Solidity, along with their supporting functions `affine_combination` and `fraction`. \
These functions were ported from their [original Rust implementations](https://github.com/pyth-network/pyth-sdk-rs/blob/main/pyth-sdk/src/price.rs) to Solidity to ensure compatibility and functionality within our Solidity-based smart contract environment.

# Changes Made
## get_collateral_valuation_price
Ported the logic to calculate the valuation of a collateral position based on the amount deposited, using a linear interpolation between initial and final discount rates.
Ensured the function handles the correct scaling and precision of rates and amounts.

## get_borrow_valuation_price
Ported the logic to calculate the valuation of a borrow position based on the amount borrowed, using a linear interpolation between initial and final premium rates.
Ensured the function maintains correct scaling and precision for the premiums.

## get_price_in_quote
Implemented the function to convert the price of one token into the price of another token (quote).
Ensured the function accurately handles conversions and maintains precision.

# Supporting Functions
Implemented `affine_combination` to perform affine combination of two prices.
Implemented `fraction` to create fractions of prices with appropriate precision handling.

# Documentation
In-code comments were added to explain the logic and assumptions behind each function.

# Additional Notes
Ensure that Solidity version compatibility is checked before merging.
Peer review and additional testing are recommended to validate the correctness of the ported functions.


target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        (uint64 basePrice, int64 baseSign) = toUnsigned(base.price);
+        (uint64 otherPrice, int64 otherSign) = toUnsigned(other.price);
+
+        uint64 midprice = (basePrice * PD_SCALE) / otherPrice;
+        int32 midpriceExpo = base.expo - other.expo + PD_EXPO;
+
+        uint64 otherConfidencePct = (other.conf * PD_SCALE) / otherPrice;
+        uint128 conf = ((uint128(base.conf) * PD_SCALE) / otherPrice) +
```
anihamde, line 234: i think solidity handles casting uints up, but make sure? generally need some tests to make sure this all works

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+
+    /**
+     * @dev Multiply a price by a constant.
+     * @param self The price
+     * @param c The constant
+     * @param e The exponent of the constant
+     * @return The resulting price
+     */
+    function cmul(
+        PythStructs.Price memory self,
```
anihamde, line 284: again rename this to something clearer, like `price`

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+
+    /**
+     * @dev Multiply two prices, propagating uncertainty.
+     * @param self The first price
+     * @param other The second price
+     * @return The product of the two prices
+     */
+    function mul(
+        PythStructs.Price memory self,
+        PythStructs.Price memory other
```
anihamde, line 308: `price1`, `price2`

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+        return div(x_as_price, y_as_price);
+    }
+
+    /**
+     * @dev Helper function to return the minimum of two uint64 values.
+     * @param a The first value
+     * @param b The second value
+     * @return The minimum of a and b
+     */
+    function min(uint64 a, uint64 b) internal pure returns (uint64) {
```
anihamde, line 446: don't use this, use Math.min from openzeppelin

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+        return a < b ? a : b;
+    }
+
+    /**
+     * @dev Helper function to return the minimum of two uint256 values.
+     * @param a The first value
+     * @param b The second value
+     * @return The minimum of a and b
+     */
+    function min(uint256 a, uint256 b) internal pure returns (uint256) {
```
anihamde, line 456: same as above

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+        return a < b ? a : b;
+    }
+}
```
anihamde, line 459: This generally looks good to me, main thing is it needs a bunch of tests. See [here](https://github.com/pyth-network/pyth-sdk-rs/blob/05ddf2beccac5bccdb8c31999f1940e4589b2c66/pyth-sdk/src/price.rs#L645) for examples, i'd recommend using [forge](https://book.getfoundry.sh/reference/forge/forge) for unit testing

target_chains/ethereum/sdk/solidity/Price.sol
```diff
@@ -0,0 +1,459 @@
...
+     * @param self The original price
+     * @param deposits Quantity of token deposited in the protocol
+     * @param deposits_endpoint Deposits right endpoint for the affine combination
+     * @param rate_discount_initial Initial discounted rate at 0 deposits (units given by discount_exponent)
+     * @param rate_discount_final Final discounted rate at deposits_endpoint deposits (units given by discount_exponent)
+     * @param discount_exponent The exponent to apply to the discounts
+     * @return The valuation price of the collateral
+     */
+    function get_collateral_valuation_price(
+        PythStructs.Price memory self,
```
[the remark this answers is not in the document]
0xHaku: While "price" might be a better fit according to general function naming conventions, using "self" makes sense if we assume the use of the `using PriceLib for PythStructs.Price;` syntax (as the first argument is omitted). 
Which one do you prefer?

target_chains/ethereum/sdk/solidity/Price.sol
0xHaku: @anihamde 
I referred to `forge-test/PythTestUtils.t.sol` for how to write forge-test. 
This test code seems to be testing the code stored in the` @pythnetwork/pyth-sdk-solidity` library. 
However, my code is not included in the package, so I cannot use this notation. 
Should I import the test target file by traversing the local directory, or should I incorporate the implementation into the package and then import the target file? 
Please advise on the correct procedure.

anihamde: You can just write unit tests for this library. I would recommend [using a Harness](https://book.getfoundry.sh/tutorials/best-practices#test-harnesses) to access the methods of the library. You could write a Harness to inherit `PythStructs.Price`, and then do `using PriceLib for PriceHarness` to write all the test cases. Let me know if that makes sense?

0xHaku: @anihamde 
I have created a test contract that encompasses the harness and performed a series of tests, so please review it.

target_chains/ethereum/sdk/solidity/PythPrice.sol
```diff
@@ -0,0 +1,438 @@
...
+            revert DivisionByZero();
+        }
+
+        (uint64 basePrice, int64 baseSign) = toUnsigned(base.price);
+        (uint64 otherPrice, int64 otherSign) = toUnsigned(price2.price);
+
+        uint64 midprice = (basePrice * PD_SCALE) / otherPrice;
+        int32 midpriceExpo = base.expo - price2.expo + PD_EXPO;
+
+        uint64 otherConfidencePct = (price2.conf * PD_SCALE) / otherPrice;
```
anihamde, line 235: maybe some comments here to explain the math, similar to what's in the rust sdk? o/w it's close to impossible for a first-time reader of the code to understand what is going on here

target_chains/ethereum/sdk/solidity/PythPrice.sol
```diff
@@ -0,0 +1,438 @@
...
+        PythStructs.Price memory base = normalize(price);
+        price2 = normalize(price2);
+
+        (uint64 basePrice, int64 baseSign) = toUnsigned(base.price);
+        (uint64 otherPrice, int64 otherSign) = toUnsigned(price2.price);
+
+        uint64 midprice = basePrice * otherPrice;
+        int32 midpriceExpo = base.expo + price2.expo;
+
+        uint64 conf = base.conf * otherPrice + price2.conf * basePrice;
```
anihamde, line 318: same comment here as in `div` function: some comments to explain the math would be great

target_chains/ethereum/contracts/forge-test/Pyth.Price.t.sol
```diff
@@ -0,0 +1,281 @@
...
+        });
+        PythStructs.Price memory price2 = PythStructs.Price({
+            price: 50,
+            conf: 2,
+            expo: -8,
+            publishTime: 1100
+        });
+        PythStructs.Price memory result = harness.mul(price1, price2);
+        assertEq(result.price, 5000, "Multiplication result should be 5000");
+        assertEq(result.expo, -16, "Exponents should be added");
```
anihamde, line 112: assert the confidence is as expected?

target_chains/ethereum/contracts/forge-test/Pyth.Price.t.sol
```diff
@@ -0,0 +1,281 @@
...
+        });
+        PythStructs.Price memory result = harness.affineCombination(
+            0,
+            y1,
+            10,
+            y2,
+            5,
+            -9
+        );
+        assertEq(result.price, 10 ** 7, "Affine combination should be 150");
```
anihamde, line 197: this assert message is wrong. i think this test case is wrong too? should be 200 for y2, i think?

anihamde (commented):
This looks better. I would say it still needs a lot more tests, and there are some existing issues I pointed out. In particular on the point about needing more tests, I would like to see tests around `getBorrowValuationPrice`, `getCollateralValuationPrice`, but also would be good to see a lot more tests for the primitive ops like `mul`, `div`, etc. See how many tests are [here](https://github.com/pyth-network/pyth-sdk-rs/blob/c34d0100363c30945cf53972fab78e543f030596/pyth-sdk/src/price.rs#L645-L2254) for reference--doesn't need to be that many, but really would be good to check edge cases and make sure behavior is as intended. This shouldn't be too hard, but it is important

ali-bahjati:
Combining prices is an easy task and the tricky part is combining confidence; however confidence is widely used to make this PR a high priority. This PR hasn't been updated for a while and therefore closing it.

PR #1747, opened 2024-06-29
Express Relay: directly submit Bids

anihamde:
This version of the Express Relay SDKs submits Bids directly instead of submitting OpportunityBids. This reduces the dependency on the server's version of the OpportunityAdapter contract and OpportunityAdapterFactory address.


express_relay/sdk/js/package.json
```diff
@@ -38,6 +38,7 @@
     "openapi-client-axios": "^7.5.5",
     "openapi-fetch": "^0.8.2",
     "viem": "^2.16.2",
+    "web3": "^4.10.0",
```
m30m, line 41: no need to another library, viem should have all the functionality

express_relay/sdk/js/src/examples/simpleSearcher.ts
```diff
@@ -47,25 +47,26 @@ class SimpleSearcher {
...
+    const bid = await this.client.signBid(
       opportunity,
       bidParams,
-      checkHex(argv.privateKey)
+      checkHex(argv.privateKey),
+      false
     );
     try {
-      const bidId = await this.client.submitOpportunityBid(opportunityBid);
+      const bidId = await this.client.submitBid(bid as Bid);
```
m30m, line 67: why `as Bid` is necessary here? This usually hints to a problem

express_relay/sdk/js/src/index.ts
```diff
@@ -338,18 +340,53 @@ export class Client {
...
    * @returns Signed opportunity bid
    */
-  async signOpportunityBid(
+  async signBid(
     opportunity: Opportunity,
     bidParams: BidParams,
-    privateKey: Hex
-  ): Promise<OpportunityBid> {
+    privateKey: Hex,
+    returnOpportunityBid: boolean = false
```
m30m, line 388: this set of changes mean we want to deprecate the serverside opportunity bid endpoint or at least it's not going to support all sdk versions. So we shouldn't provide this option to the users to choose between them. signOpportunityBid should just return a bid and the submitOpportunityBid should be removed.

express_relay/sdk/js/src/examples/simpleSearcher.ts
```diff
@@ -102,7 +103,7 @@ const argv = yargs(hideBin(process.argv))
   .option("bid", {
     description: "Bid amount in wei",
     type: "string",
-    default: "20000000000000000",
+    default: "600000000000000000",
```
danimhr, line 106: This is a really big number, why should this be the default value?

express_relay/sdk/js/src/index.ts
```diff
@@ -397,12 +441,7 @@ export class Client {
       types,
       primaryType: "PermitBatchWitnessTransferFrom",
       message: {
-        permitted: getPermittedTokens(
-          opportunity.sellTokens,
-          bidParams.amount,
-          opportunity.targetCallValue,
-          checkAddress(opportunityAdapterConfig.weth)
-        ),
+        permitted: permitted,
```
danimhr, line 444: no need to add `permitted` as value

express_relay/sdk/js/src/types.ts
```diff
@@ -162,3 +162,107 @@ export type BidResponse = components["schemas"]["SimulatedBid"];
 export type BidsResponse = {
   items: BidResponse[];
 };
+
+export const executeOpportunityAbi = {
```
danimhr, line 166: Maybe add this to abi.ts?

express_relay/sdk/python/express_relay/client.py
```diff
@@ -25,6 +27,7 @@
     TokenAmount,
     OpportunityBidParams,
     OpportunityAdapterConfig,
+    EXECUTION_PARAMS_TYPESTRING,
```
danimhr, line 30: is it const or type? If it's type we should rename it to have consistent naming

m30m (approved):

PR #1755, opened 2024-07-02
feat(contract_manager): basic support for express relay contracts

m30m:


guibescos (approved):

contract_manager/src/contracts/evm_abis.ts
ali-bahjati, line 1: this is more of a personal preference but i would've preferred separate files for different ABIs for more clarity.

ali-bahjati (approved):
Nice!

PR #1757, opened 2024-07-04
fix(apps/price_pusher/solana): make priceUpdates per bundle configurable

ali-bahjati:
We are using a fixed priceUpdate per bundle constant but it might need to be changed as the price proof size increases. This change makes it configurable.


ayazabbas (approved):

guibescos (approved):
lgtm. can you update `config.solana.mainnet.sample.json` too

PR #1761, opened 2024-07-10
fix(price-pusher): randomize price feed ids

guibescos:
Solana devnet pusher doesn't work great.
Since we have no Jito, we need to land many transactions sequentially. And until now it sequentially follows the order in `price-config.yaml`. In periods of bad landing rate it crashes before getting to the end of the sequence. Therefore price feeds towards the end of the list never get updated.

By randomizing the order of price updates to avoid the imbalance between the top of the list and the bottom of the list in terms of frequency of updates.

### DIFF
--- a/apps/price_pusher/src/solana/solana.ts
+++ b/apps/price_pusher/src/solana/solana.ts
@@ -93,7 +93,12 @@ export class SolanaPricePusher implements IPricePusher {
       return;
     }
 
-    const shuffledPriceIds = priceIds.sort(() => Math.random() - 0.5);
+    const shuffledPriceIds = priceIds
+      .map((x) => {
+        return { element: x, key: Math.random() };
+      })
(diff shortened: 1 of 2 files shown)


apps/price_pusher/src/solana/solana.ts
```diff
@@ -93,10 +93,11 @@ export class SolanaPricePusher implements IPricePusher {
       return;
     }
 
+    const shuffledPriceIds = priceIds.sort(() => Math.random() - 0.5);
```
ali-bahjati, line 96: Hmmmm is this key computed only once? :? I'm wondering if it would crash if it's recalibrated each time

ali-bahjati (approved):

apps/price_pusher/src/solana/solana.ts
guibescos: I don't understand your comment

PR #1764, opened 2024-07-12
feat(target_chains/starknet): specify guardian set index in wormhole constructor

Riateche:
This mitigates a potential issue where if we deploy wormhole contract from the 0th guardian set and then upgrade it, all previous sets will be considered valid for 1 day. Now we can deploy the contract with the latest guardian set only.


ali-bahjati (approved):
Nice!

PR #1778, opened 2024-07-17
feat: add publisher stake caps to validator

guibescos:
`PublisherStakeCapsMessage` is a new type of message, that contains how much stake each publisher should have to maximize their rewards in the context of CIP. 
This message is different from other messages in that there's only one message of this type in the Merkle tree, and also it is potentially really big (since it contains the list of all publishers and their cap).

https://forum.pyth.network/t/community-proposal-for-the-design-of-the-community-integrity-pool-cip/1571

I gets computed in https://github.com/pyth-network/pythnet/pull/302 , and will be consumed by the reward distribution logic on Solana.

### DIFF
--- a/pythnet/pythnet_sdk/src/messages.rs
+++ b/pythnet/pythnet_sdk/src/messages.rs
@@ -119,15 +119,15 @@ pub struct PriceFeedMessage {
 #[cfg(feature = "quickcheck")]
 impl Arbitrary for PriceFeedMessage {
     fn arbitrary(g: &mut quickcheck::Gen) -> Self {
-        let mut id = [0u8; 32];
-        for item in &mut id {
+        let mut feed_id = [0u8; 32];
+        for item in &mut feed_id {
             *item = u8::arbitrary(g);
         }
 
         let publish_time = i64::arbitrary(g);
 
         PriceFeedMessage {
-            id,
+            feed_id,
             price: i64::arbitrary(g),
             conf: u64::arbitrary(g),
             exponent: i32::arbitrary(g),
@@ -156,15 +156,15 @@ pub struct TwapMessage {
 #[cfg(feature = "quickcheck")]
 impl Arbitrary for TwapMessage {
     fn arbitrary(g: &mut quickcheck::Gen) -> Self {
-        let mut id = [0u8; 32];
-        for item in &mut id {
+        let mut feed_id = [0u8; 32];
+        for item in &mut feed_id {
             *item = u8::arbitrary(g);
         }
 
         let publish_time = i64::arbitrary(g);
 
         TwapMessage {
-            id,
+            feed_id,
             cumulative_price: i128::arbitrary(g),
             cumulative_conf: u128::arbitrary(g),
             num_down_slots: u64::arbitrary(g),
--- a/pythnet/pythnet_sdk/src/messages.rs
+++ b/pythnet/pythnet_sdk/src/messages.rs
@@ -54,6 +54,10 @@ pub enum Message {
     PublisherStakeCapsMessage(PublisherStakeCapsMessage),
 }
 
+/// PublisherStakeCapsMessage is a global message that aggregates data from all price feeds
+/// we can't associate it with a specific feed, so we use a feed id that is not used by any price feed
+pub const PUBLISHER_STAKE_CAPS_MESSAGE_FEED_ID: FeedId = [0u8; 32];
+
 impl Message {
     pub fn publish_time(&self) -> i64 {
         match self {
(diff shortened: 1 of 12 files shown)


pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -50,20 +51,23 @@ use {
...
             Self::TwapMessage(msg) => msg.publish_time,
+            Self::PublisherCapsMessage(msg) => msg.publish_time,
         }
     }
 
     pub fn feed_id(&self) -> FeedId {
         match self {
             Self::PriceFeedMessage(msg) => msg.feed_id,
             Self::TwapMessage(msg) => msg.feed_id,
+            Self::PublisherCapsMessage(_) => [0u8; 32],
```
guibescos, line 70: the new type of message breaks this

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -30,7 +31,7 @@ use {
 /// some of the methods for PriceFeedMessage and TwapMessage are not used by the oracle
 /// for the same reason. Rust compiler doesn't include the unused methods in the contract.
 /// Once we start using the unused structs and methods, the contract size will increase.
-#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
guibescos, line 34: Is there anyway we can keep the Copy with a variable length publisher cap list?

guibescos: I really don't like this since it means we'll have to put `.clone()` in a lot of places

ali-bahjati: no it's not because Copy does a bit-wise copy

pythnet/pythnet_sdk/src/messages.rs
```diff
@@ -50,20 +51,23 @@ use {
...
             Self::TwapMessage(msg) => msg.publish_time,
+            Self::PublisherStakeCapsMessage(msg) => msg.publish_time,
         }
     }
 
     pub fn feed_id(&self) -> FeedId {
         match self {
             Self::PriceFeedMessage(msg) => msg.feed_id,
             Self::TwapMessage(msg) => msg.feed_id,
+            Self::PublisherStakeCapsMessage(_) => [0u8; 32],
```
Riateche, line 70: Maybe we should change the function to return `Option<FeedId>` since there is no feed id for this kind of message.

target_chains/solana/programs/pyth-push-oracle/src/lib.rs
```diff
@@ -83,7 +83,7 @@ pub mod pyth_push_oracle {
                 .map_err(|_| PushOracleError::DeserializeMessageFailed)?;
         let next_timestamp = match message {
             Message::PriceFeedMessage(price_feed_message) => price_feed_message.publish_time,
-            Message::TwapMessage(_) => {
+            _ => {
```
Riateche, line 86: I think it's safer to list both message types explicitly here.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -437,7 +437,7 @@ fn post_price_update_from_vaa<'info>(
             price_update_account.price_message = price_feed_message;
             price_update_account.posted_slot = Clock::get()?.slot;
         }
-        Message::TwapMessage(_) => {
+        _ => {
```
Riateche, line 440: I think it's safer to list both message types explicitly here.

Riateche (changes requested):

target_chains/solana/programs/pyth-push-oracle/src/lib.rs
guibescos: yes

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: yes

pythnet/pythnet_sdk/src/messages.rs
guibescos: addressed

keyvankhademi (approved):

pythnet/pythnet_sdk/src/messages.rs
ali-bahjati: I think the api won't be clean in this way and at some point we should fix it. It's fine now as we are not using this on the validator. Let's also add a TODO to mention that we should fix it. i also recommend using something other than [0; 32] for it to distinguish it from empty/default.

guibescos: [0;32] is good because it's guaranteed to not conflict with any price feed ids

PR #1779, opened 2024-07-18
feat(remote_executor): enable other SVM networks

cctdaniel:


governance/remote_executor/programs/remote-executor/src/state/posted_vaa.rs
```diff
@@ -10,7 +10,13 @@ use {
...
 // The current chain's wormhole bridge owns the VAA accounts
 impl Owner for AnchorVaa {
-    #[cfg(not(feature = "pythtest"))]
+    // do something like this
+    // #[cfg(feature = "eclipse-testnet")]
+    // fn owner() -> Pubkey {
+    //     Pubkey::from_str("HDwcJBJXjL9FpJ7UBsYBtaDjsBUhuLCUYoz3zr8SWWaQ").unwrap()
+    // }
+
+    // #[cfg(not(feature = "pythtest"))]
```
guibescos, line 19: also it's weird that we need to have both a feature flag and an env variable, is there a way to unify both?

governance/remote_executor/programs/remote-executor/src/state/governance_payload.rs
```diff
@@ -188,7 +212,7 @@ pub mod tests {
     fn test_check_deserialization_serialization() {
         // No instructions
         let payload = ExecutorPayload {
-            header:       super::GovernanceHeader::executor_governance_header(),
+            header:       super::GovernanceHeader::executor_governance_header(*CHAIN_ID),
```
guibescos, line 215: maybe weird to have this test depend on an env variable

guibescos (approved):
lgtm

governance/remote_executor/programs/remote-executor/src/state/posted_vaa.rs
```diff
@@ -19,6 +19,11 @@ impl Owner for AnchorVaa {
     fn owner() -> Pubkey {
```
ali-bahjati, line 19: so now the default is pythtest?

I recommend having a feature for each (including pythtest) and then also not having any default so the code fails to compile when it is not explicitly defined (to make sure we don't make mistakes). Also it's worth adding a short comment to the readme or on Cargo.toml file describing why they exists.

ali-bahjati (approved):

PR #1781, opened 2024-07-18
chore(xc_admin_frontend) remove reference to deprecated s3 bucket

cprussin:


cprussin:
Going to merge this without review since it's trivial

PR #1782, opened 2024-07-18
fix(xc_admin_frontend): fix OG image url

cprussin:
In #1781 I accidentally used the wrong base path for the OG image


cprussin:
Merging immediately since this is trivial

PR #1789, opened 2024-07-22
More API reference feedback

cprussin:
- Remove summaries from sidebar
- Allow selecting from all supported networks
- Make feed ID a dropdown (allow valid hex strings that aren't valid price feeds so people can see errors)
- Add link to price feed list from price feed ID autocomplete dropdown
- Put results inline


apps/api-reference/public/currency-icons/$pac.svg
cprussin, line 1: Note to reviewer -- all of these icons are copied from [the cryptocurrency-icons node module](https://github.com/spothq/cryptocurrency-icons).  Ideally we will get rid of these and find a way to serve icons from that package directly but since that package has no barrel export, we'd either need to come up with our own, or write a script / symlink or similar to copy the icons from the package into our public folder.  I'm not really happy with either solution so I just did this for now as a stopgap and I'll think about a better solution later.

aditya520:
Screenshot 2024-07-22 at 11 23 15 AM
Can we add a horizontal scroll bar to the error tab.

aditya520 (approved):

PR #1795, opened 2024-07-24
chore(target_chains/fuel): update sdk to use the latest abi

cctdaniel:


target_chains/fuel/sdk/js/package.json
```diff
@@ -20,7 +20,6 @@
     "access": "public"
   },
   "scripts": {
-    "test": "jest --passWithNoTests",
```
cctdaniel, line 23: we dont use this anyway so removing it

ali-bahjati (approved):

PR #1798, opened 2024-07-25
Create a new Pyth price service client in Rust

aoikurokawa:
## Description
- Add rs client library in order to interact with Hermes server.

## Additional Notes
I will make an another PR to handle Websocket stuff.


ali-bahjati:
Hi @aoikurokawa 
Thanks a lot for your contribution!! Would it be possible for you to create this client based on the v2 apis and not the v1 apis? (and that also means using http streaming instead of websockets for future)

aoikurokawa:
> Hi @aoikurokawa Thanks a lot for your contribution!! Would it be possible for you to create this client based on the v2 apis and not the v1 apis? (and that also means using http streaming instead of websockets for future)

Hello @ali-bahjati 
Thank you for the response. I rewrote the client based on v2 instead of v1.
I appreciate your reviewing.

PR #1799, opened 2024-07-26
fix(target_chains/fuel): address audit

cctdaniel:


target_chains/fuel/contracts/pyth-interface/src/data_structures/price.sw
```diff
@@ -175,6 +175,11 @@ impl PriceFeed {
         //convert publish_time from UNIX to TAI64
         publish_time += TAI64_DIFFERENCE;
 
+        require(
+            publish_time <= timestamp(),
+            PythError::FuturePriceNotAllowed,
+        );
+
```
cctdaniel, line 182: added a check here to ensure that future publish_time is not allowed

ali-bahjati (approved):

PR #1806, opened 2024-07-29
fix(target_chains/fuel): resolves issue with incorrect abi paths

martines3000:
Use `/out/release` instead of `/out/debug` for the ABI and BIN files of the Pyth contract.


cctdaniel (approved):
thank you for your contribution!

PR #1808, opened 2024-07-31
fix(target_chains/fuel): remove test from prepublishOnly

cctdaniel:
gh action for `Publish Pyth JS packages` is failing due to `test` command being removed from `pyth-fuel-js` so removing it from `prepublishOnly` script

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -199,7 +199,7 @@ importers:
         version: 8.57.0
       jest:
         specifier: ^29.4.0
-        version: 29.7.0(@types/node@20.14.2)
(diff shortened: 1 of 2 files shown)


cprussin (changes requested):
I think this is failing because in `prepublishOnly` we call `pnpm run test`: https://github.com/pyth-network/pyth-crosschain/blob/8013c7db2cd380fbd4dc60c18c44e5b111e0b39d/target_chains/fuel/sdk/js/package.json#L28.  Unless I'm mistaken about why it's failing, IMO it's better to remove it from that than to add a dummy script that adds unused dependencies and does nothing useful.

cprussin (approved):

PR #1810, opened 2024-08-01
feat(target_chains/ton): initialize ton contracts

cctdaniel:
initialize ton contracts with `npm create ton@latest`


ali-bahjati (approved):
Shall we add it to the monorepo as well?

PR #1811, opened 2024-08-05
feat(target_chains/ethereum): remove getPrice from IPyth

ali-bahjati:
getPrice method uses a configured default staleness from the Pyth contract which is normally set to 60 seconds. While this is enough for some types protocols it's very high for latency-sensitive protocols and protocols should use set their own desired staleness. This change removes this method and getEmaPrice and leaves protocols with getPriceUnsafe or getPriceNoOlderThan.

The contract will still support the getPrice method to avoid breaking existing users of it.

### DIFF
--- a/target_chains/ethereum/sdk/solidity/AbstractPyth.sol
+++ b/target_chains/ethereum/sdk/solidity/AbstractPyth.sol
@@ -19,6 +19,7 @@ abstract contract AbstractPyth is IPyth {
         bytes32 id
     ) public view virtual returns (bool exists);
 
+    /// @notice This function is deprecated and is only kept for backward compatibility.
     function getValidTimePeriod()
         public
         view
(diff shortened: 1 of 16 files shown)


jayantk (approved):

cprussin (approved):

target_chains/ethereum/sdk/solidity/AbstractPyth.sol
```diff
@@ -23,18 +23,19 @@ abstract contract AbstractPyth is IPyth {
         public
         view
         virtual
-        override
```
aditya520, line 26: Can we add the same @notice to `getValidTimePeriod()` as well?

aditya520 (approved):

aditya520 (approved):

PR #1812, opened 2024-08-05
feat: publisher-caps-program

guibescos:
Add some stuff to pythnet_sdk that I need

### DIFF
--- a/.github/workflows/ci-pre-commit.yml
+++ b/.github/workflows/ci-pre-commit.yml
@@ -32,6 +32,11 @@ jobs:
           profile: minimal
           toolchain: nightly-2024-03-26
           components: rustfmt, clippy
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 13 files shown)


.github/workflows/ci-pre-commit.yml
```diff
@@ -32,6 +32,11 @@ jobs:
           profile: minimal
           toolchain: nightly-2024-03-26
           components: rustfmt, clippy
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly-2024-08-04
+          components: rustfmt, clippy
```
keyvankhademi, line 39: did we need to use the newer version? It feels weird maintaining 3 different toolchain versions for fmt and clippy.

keyvankhademi (approved):

.github/workflows/ci-pre-commit.yml
guibescos: yeah old version broke some things

PR #1817, opened 2024-08-08
feat(target_chains/fuel): update testnet address

cctdaniel:


ali-bahjati (approved):
Do you mind updating contract-manager too?

PR #1818, opened 2024-08-08
fix(apps/hermes): filter coming soon feeds from v2/price_feeds

cctdaniel:


apps/hermes/server/src/network/pythnet.rs
```diff
@@ -432,7 +440,7 @@ async fn fetch_price_feeds_metadata(
                     // TODO: Add stricter type checking for attributes
                     let attributes = prod_acct
                         .iter()
-                        .filter(|(key, _)| !key.is_empty())
+                        .filter(|(key, _)| !key.is_empty() && !key.contains('\0'))
```
cctdaniel, line 443: for some reason feeds like `Equity.US.AAPL/USD` returns `"25/C\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000": "",` in the `attributes` object so filtering null bytes here so that this doesnt show up in the result

ali-bahjati: Hmmmm but why such a thing is returned 🤔 maybe there is a bug in our prod_attr implementation.
i recommend reverting this change and investigating it separately.

ali-bahjati (approved):

PR #1819, opened 2024-08-08
feat: support initializing stake accounts

guibescos:
Adding some logic to create proposals that create new stake accounts from the PGAS treasury.


ali-bahjati (approved):
It's very nice!!

PR #1820, opened 2024-08-08
feat: initial implementation of new staking app

cprussin:
This is a low-fidelity effectively unstyled build out of the new staking app, against mock on-chain APIs.


apps/staking/src/api.ts
```diff
@@ -0,0 +1,205 @@
...
+  if (publisher) {
+    if (publisher.positions?.staked) {
+      publisher.positions.staked -= amount;
+      publisher.positions.cooldown =
+        (publisher.positions.cooldown ?? 0n) + amount;
+    }
+  } else {
+    throw new Error(`Invalid publisher key: "${publisherKey}"`);
+  }
+};
```
cprussin, line 205: @keyvankhademi @guibescos here are the mock APIs which we will need to replace with correct on-chain calls.

I think the general signatures are reasonable except perhaps `loadData`.  I opted to make that a single API that returns everything, rather than splitting up governance/publishers/publisher stakes or similar, because by making it a single API it makes it much more trivial to just refresh the whole data set every time any transfer is completed and not have to worry about how things affect each other (for instance if I withdraw a bunch of tokens, how does that affect the counts that show up in the governance & integrity sections).  It's not that hard to manage those things, but it's marginally messier and adds more logic on the frontend.  So, if there's a good reason to split up that API then I can update accordingly, but if not I recommend just keeping it all in one API.

PR #1823, opened 2024-08-14
feat(governance/remote_executor): add mantis testnet

cctdaniel:


governance/remote_executor/cli/src/main.rs
```diff
@@ -87,9 +87,11 @@ fn main() -> Result<()> {
             let wormhole_config_data =
                 Config::try_from_slice(&rpc_client.get_account_data(&wormhole_config)?)?;
 
+            let guardian_set_data_offset = if cli.chain == 26 { 0 } else { 8 }; // Pythnet's guardian set account has no discriminator
```
guibescos, line 90: Pythnet's wormhole is pre-wormhole rewrite so accounts don't have 8 bytes discriminators at the beginning.
On the other hand our wormhole receiver that we will post to every svm, does.

governance/remote_executor/cli/src/main.rs
```diff
@@ -167,9 +169,18 @@ fn main() -> Result<()> {
...
 
+            let executor_key = Pubkey::find_program_address(
+                &[EXECUTOR_KEY_SEED.as_bytes(), &payer.pubkey().to_bytes()],
+                &ID,
+            )
+            .0;
             let payload = ExecutorPayload {
                 header:       GovernanceHeader::executor_governance_header(cli.chain),
-                instructions: vec![],
+                instructions: vec![InstructionData::from(&system_instruction::transfer(
```
guibescos, line 179: I added an instruction here in the test vaa to give us more confidence. It transfer 1 lamport from  one of the executor keys.

target_chains/solana/cli/src/cli.rs
```diff
@@ -57,10 +57,12 @@ pub enum Action {
...
-        fee:     u64,
+        fee:                  u64,
         #[clap(short = 'e', long, parse(try_from_str = Pubkey::from_str), help = "Source emitter")]
-        emitter: Pubkey,
+        emitter:              Pubkey,
         #[clap(short = 'c', long, help = "Source chain")]
-        chain:   u16,
+        chain:                u16,
+        #[clap(short = 'a', long, parse(try_from_str = Pubkey::from_str), help = "Governance authority")]
+        governance_authority: Pubkey,
```
guibescos, line 66: Having this will avoid what happened on eclipse mainnet. This should be set to the Pythian council from the very beginning.

guibescos (approved):

PR #1828, opened 2024-08-16
(refactor) api reference

aditya520:


apps/api-reference/next.config.js
```diff
@@ -9,6 +9,18 @@ export default {
...
+  redirects: async () => [
+    {
+      source: "/home",
+      destination: "/",
+      permanent: true,
+    },
+    {
+      source: "/price-feeds/evm",
+      destination: "/price-feeds/evm/parsePriceFeedUpdates",
+      permanent: true,
```
cprussin, line 21: We shouldn't use `permanent` for either of these two redirects; a permanent redirect is cached by the browser and never re-checked (at least, while the browser considers the cache live, which is around 6 months for Chrome and the spec does not require browsers to cap at all).  As a result, doing this will prevent us from being able to add pages to these URLs in the future, if we ever wanted them to not be redirects.

apps/api-reference/next.config.js
```diff
@@ -9,6 +9,18 @@ export default {
     },
   },
 
+  redirects: async () => [
+    {
+      source: "/home",
+      destination: "/",
+      permanent: true,
+    },
```
cprussin, line 17: I'm not sure I understand why this redirect is here, `/home` doesn't really mean anything?  I'm assuming this has to do with redirecting the old docs pages URLs and if that's the case, it's better to add the redirect to the docs app, where the URL used to be, rather than adding it here.

cprussin (changes requested):
Most important is to not use `permanent`

cprussin (approved):

PR #1829, opened 2024-08-16
refactor(apps/price_pusher): Use viem instead of web3 for evm pusher

ali-bahjati:
Web3 library is not very widely used anymore due to its complex and magical internal design. Some users were reporting issues like strange timeouts or unsupported RPC calls via web3 (and the deprecated HDWalletProvider by truffle that we use). This change refactors the code to use Viem. The experience with Viem is nice: it has strong types and good utilities. The error handling is also very well designed. The downside is that it has a steep learning curve to get it right.

This change refactors the code based on the PR reviews to make it simpler as well.

Lastly the version is bumped as a major release because the behaviour and logs have changed and it might affect production environments. It also signals the users to test it out properly before using it which is good because all the failure cases might not be handled.


apps/price_pusher/src/evm/evm.ts
```diff
@@ -188,31 +206,40 @@ export class EvmPricePusher implements IPricePusher {
...
     let gasPriceToOverride = undefined;
 
     if (this.lastPushAttempt !== undefined) {
       if (this.lastPushAttempt.nonce <= lastExecutedNonce) {
         this.lastPushAttempt = undefined;
       } else {
-        gasPriceToOverride = Math.ceil(
-          this.lastPushAttempt.gasPrice * this.overrideGasPriceMultiplier
-        );
+        gasPriceToOverride =
```
jayantk, line 234: the ceil may have been important here -- does it cause a problem if the result here isn't an integer?

jayantk: oh is this caught already by the call to `.round` below? Maybe this is fine.

apps/price_pusher/src/evm/pyth-abi.ts
```diff
@@ -0,0 +1,656 @@
+const IPyth = [
```
jayantk, line 1: this is a bit annoying because we have to update it when the contract API changes but i guess it's fine...

One thought is that we could update the the pyth EVM SDK to export an object like this, then import that here. That would also help Pyth users who use viem.

jayantk (approved):
looks good to me

aditya520 (approved):
Thank you for this. We need viem now.

apps/price_pusher/src/utils.ts
```diff
@@ -18,11 +18,11 @@ export function removeLeading0x(id: HexString): HexString {
...
 
-export function addLeading0x(id: HexString): HexString {
+export function addLeading0x(id: HexString): `0x${string}` {
   if (id.startsWith("0x")) {
-    return id;
+    return id as `0x${string}`;
   }
-  return "0x" + id;
+  return ("0x" + id) as `0x${string}`;
 }
```
cprussin, line 26: More idiomatic would be using a type guard:

```tsx
const addLeading0x = (id: HexString): `0x${string}` =>
    hasLeading0x(id) ? id : `0x${id}`;

const hasLeading0x = (input: string): input is `0x${string}` =>
    input.startsWith('0x');
```

apps/price_pusher/src/evm/evm.ts
```diff
@@ -61,53 +84,44 @@ export class EvmPriceListener extends ChainPriceListener {
...
+  private onPriceFeedUpdate(
+    logs: WatchContractEventOnLogsParameter<typeof PythAbi, "PriceFeedUpdate">
+  ) {
+    for (const log of logs) {
+      const priceId = removeLeading0x(log.args.id!);
+
+      const priceInfo: PriceInfo = {
+        conf: log.args.conf!.toString(),
+        price: log.args.price!.toString(),
+        publishTime: Number(log.args.publishTime!),
```
cprussin, line 103: In general I strongly advise against nearly ever using type casts, including `!`, in typescript.  It's not hard to just add an explicit check (e.g. something like `if (log.args.conf === undefined) { throw new MissingLogArgError('conf'); }` and adding the explicit check avoids possible scenarios where the runtime code does not behave in ways that the type system claims to guarantee.  As an added bonus, the runtime checks can help make it obvious when there's some structure in the code that could be cleaned up / improved / types that can be improved / better schema validation that could be happening.

There are a few scenarios where type casts are necessary due to limitations around how typescript does dependent typing but it's rare and this isn't one of the cases.

apps/price_pusher/src/evm/evm.ts
```diff
@@ -329,11 +448,38 @@ export class EvmPricePusher implements IPricePusher {
 }
 
 export class PythContractFactory {
```
cprussin, line 450: Is this export actually consumed?  I know this wasn't introduced in this code but IMO this pattern significantly overcomplicates things.  It adds a significant amount of abstraction and indirection for what is fundamentally just some constants for default `getContract` config or perhaps a function or two wrapping it and adding default config.

I don't think this is an appropriate use of `class`.  It isn't stateful and really has no good benefit to being encapsulated in this way.  I'd suggest considering replacing this (perhaps in a follow up) with some simple types / objects and functions and reducing the amount of abstractions.

apps/price_pusher/src/evm/evm.ts
```diff
@@ -223,42 +250,121 @@ export class EvmPricePusher implements IPricePusher {
...
+            gas:
+              this.gasLimit !== undefined
+                ? BigInt(Math.round(this.gasLimit))
+                : undefined,
+          }
+        );
+
+      this.logger.debug({ request }, "Simulated request successfully");
+
+      const hash = await this.pythContract.write.updatePriceFeedsIfNecessary(
```
cprussin, line 274: In viem you can use the response of a simulation as the argument to `write` i.e.: https://viem.sh/docs/contract/simulateContract.html#pairing-with-writecontract

apps/price_pusher/src/evm/evm.ts
```diff
@@ -223,42 +250,121 @@ export class EvmPricePusher implements IPricePusher {
...
+          return;
+        }
+
+        if (err.walk((e) => e instanceof InsufficientFundsError)) {
+          this.logger.error(
+            { err },
+            "Wallet doesn't have enough balance. In a rare case, there might be issues with gas price " +
+              "calculation in the RPC."
+          );
+          throw new Error("Please top up the wallet");
```
cprussin, line 315: Minor nit but I suggest adding new error classes for something like this, it makes it much easier for consumers to disambiguate errors:

```ts
class PleaseTopUpWalletError extends Error {
  override message = "Please top up the wallet";
};
```

Although in this case I'd probably just re-throw the `InsufficientFundsError` myself...

apps/price_pusher/src/evm/evm.ts
```diff
@@ -318,6 +413,30 @@ export class EvmPricePusher implements IPricePusher {
...
+      } else if (receipt.status === "reverted") {
+        this.logger.info(
+          { hash, receipt },
+          "Price update reverted or its transaction did not land. " +
+            "This is an expected behaviour in high frequency or multi-instance setup."
+        );
+      } else {
+        throw new Error(
+          "This codepath should be unreachable. Please file a bug report."
+        );
```
cprussin, line 433: How is this reachable in typescript?  Is `receipt.status` not an enum?

cprussin (approved):
Overall lgtm, most of my comments are just suggestions for more idiomatic typescript code, feel free to disregard

apps/price_pusher/src/evm/chains.ts
```diff
@@ -0,0 +1,347 @@
+import { defineChain } from "viem";
+import * as viemChains from "viem/chains";
+import { Chain } from "viem/chains";
+
+// Get the chain object from the chainId or return an unknown chain object
+// if the chainId is not found in the list of chains. Normally the unknown
+// chain object should work for most cases, but some networks may have special
+// requirements (like custom fee model, etc) that may require a custom chain.
+export function getChain(chainId: number): Chain {
```
jxom, line 9: have you seen https://viem.sh/docs/utilities/extractChain#extractchain?

apps/price_pusher/src/evm/evm.ts
```diff
@@ -362,51 +505,42 @@ export class PythContractFactory {
...
-      mnemonic: {
-        phrase: this.mnemonic,
-      },
-      providerOrUrl: this.createWeb3Provider() as ProviderOrUrl,
+  getAccount(): Account {
+    return mnemonicToAccount(this.mnemonic);
+  }
+
+  createWalletClient(): WalletClient {
+    return createWalletClient({
```
jxom, line 532: You can also extend your Wallet Client with Public Actions if you just want to use a single Client (and not define both a Wallet & Public Client): https://viem.sh/docs/clients/wallet#optional-extend-with-public-actions

jxom: The Public vs. Wallet Client separation is only really useful if you're interfacing with a JSON-RPC Account via an EIP-1193 Provider (ie. `window.ethereum`). This is because some of these Wallets may not expose some "public" RPC methods via their EIP-1193 Provider, so you'll also need to interface with a HTTP/WS RPC Provider as well. 

On the otherhand, if you're only using a Local Account like here (Mnemonic, PK, etc), then you don't run into these issues as you're always interfacing with a singular HTTP/WS RPC Provider.

apps/price_pusher/src/evm/chains.ts
ali-bahjati: Thanks for referencing it. I saw it in the code but didn't see the example that showed the `Object.values(chains)` will work and ironically when i used it to iterate upon it gave me error. Now I tried it out and it worked.

The `extractChain` itself is not useful here though, because it uses enum of ids and in our case we want to support the unknown chain config in case it's not there as we do have some chains that are not in that list.

apps/price_pusher/src/evm/evm.ts
ali-bahjati: Unfortunately in contrary to the documentation the simulate call result is not compatible with the type required there, the account type being passed is different. @jxom fyi.

apps/price_pusher/src/evm/evm.ts
ali-bahjati: This is a strange typing from Viem. They should not be optional at all. I'll add strict defined checks here.

apps/price_pusher/src/utils.ts
ali-bahjati: Thanks!

apps/price_pusher/src/evm/evm.ts
cprussin: fwiw I usually have a helper in most codebases, something like:

```ts
const assertDefined = <T>(value: T | undefined): T => {
    if (value === undefined) {
        throw new Error('Assertion failed: value was undefined');
    } else {
        return value;
    }
};
```

Then instead of using `someVar!` in other code, I can use `assertDefined(someVar)`.  Nearly as easy to type, but also more obvious to read, and not subject to making the code unsafe.

ali-bahjati: Thanks. Apparently Viem has a strict mode that handles types but with the contract instance even if I set it the types won't get resolved. @jxom too

apps/price_pusher/src/evm/pyth-abi.ts
ali-bahjati: @cprussin The JSON file is autogenerated from the contract itself.

@jayantk I'll make a separate PR for it because I need to do some clean up there as well.

apps/price_pusher/src/evm/evm.ts
```diff
@@ -188,31 +206,40 @@ export class EvmPricePusher implements IPricePusher {
...
-        (await this.web3.eth.getGasPrice())
-    );
+    const fees = await this.publicClient.estimateFeesPerGas();
+
+    this.logger.debug({ fees }, "Estimated fees");
+
+    let gasPrice =
+      Number(await this.customGasStation?.getCustomGasPrice()) ||
+      Number(fees.gasPrice) ||
+      Number(fees.maxFeePerGas);
```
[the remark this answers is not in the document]
ali-bahjati: I think this is equivalent to to the gas price itself. If we add a limit, we add a failure chance in spikes of gas price.

PR #1844, opened 2024-08-25
feat(staking): add low-res implementation for new requirements

cprussin:
This PR adds low-res / unstyled implementation for the following new feature requirements that we've discovered:

- Must claim rewards before withdrawing
- Must claim rewards before unstaking
- Add swr to loadData
- Calculate APY in the frontend
- Show how APY will change due to staking tokens
- Show historical APY in a spark chart
- Add functionality to select a stake account if multiple exist for the given wallet
- If wallet is a publisher, show self highlighted (& adjust apy calculation)
- Claim within 1 year
- Show next locked token unlock period & amount
- Show account history

### DIFF
--- a/apps/staking/src/api.ts
+++ b/apps/staking/src/api.ts
@@ -1,58 +1,37 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
+// TODO remove these disables when moving off the mock APIs
+/* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/no-non-null-assertion */
 
 import type { WalletContextState } from "@solana/wallet-adapter-react";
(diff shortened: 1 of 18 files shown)


apps/staking/src/api.ts
```diff
@@ -1,58 +1,37 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
```
cprussin, line 1: @keyvankhademi especially take a look at this file as the requirements for the API have changed significantly and there are many new parts to account for all the new requirements.

cprussin:
Merging without review since this is pre-prod and some folks who need to review the UX can't access the preview build

PR #1848, opened 2024-08-26
(fix) Change fortuna error code

aditya520:


jayantk (approved):
thanks

if you want to deploy this change now, you also need to bump the minor version number in Cargo.toml then update the fortuna version number in the deployments repo

PR #1850, opened 2024-08-27
initial sdk prototype

anihamde:


express_relay/sdk/js/src/examples/simpleSearcher.ts
```diff
@@ -75,15 +100,84 @@ class SimpleSearcher {
...
       );
+      console.log(`Successful bid. Bid id ${bidId}`);
     } catch (error) {
-      console.error(error);
-      this.client.websocket?.close();
+      console.error(`Failed to bid: ${error}`);
+    }
+  }
+
+  async start() {
```
m30m, line 165: looking at the logic split between svm and evm I suggest creating a separate simpleSvmSearcher.ts file. The current format makes it a bit confusing to follow what's happening based on chains

express_relay/sdk/js/src/types.ts
```diff
@@ -153,6 +156,41 @@ export type Bid = {
...
+   */
+  permissionKey: string;
+  /**
+   * @description Amount of bid in wei.
+   * @example 10
+   * // TODO: this is not covering the case where 2**53-1 < bid < 2**64
+   */
+  amount: number;
+  /**
+   * @description Transaction object, as a base58-encoded string.
```
m30m, line 181: base64

express_relay/sdk/js/src/index.ts
```diff
@@ -591,4 +609,139 @@ export class Client {
...
+      Keypair.fromSecretKey(secretKey)
+    );
+    if (this.clientOptions.svmEndpoints === undefined) {
+      throw new Error("SVM endpoints not provided");
+    }
+    const connectionSvm = new Connection(
+      new URL(this.clientOptions.svmEndpoints[chainId]).toString(),
+      "confirmed"
+    );
+    const { blockhash } = await connectionSvm.getLatestBlockhash();
```
m30m, line 728: let's move this logic to the simple searcher

express_relay/sdk/js/src/index.ts
```diff
@@ -591,4 +609,139 @@ export class Client {
       return response.data;
     }
   }
+
+  /**
```
m30m, line 613: Ideally we can provide multiple functions to help here:
constructPermissionInstruction
createSvmBid: calls the above and adds it to the provided tx and returns an BidSvm object. needs searcher private key for signing and assumes the latestBlockhash is already inside the tx

express_relay/sdk/js/src/index.ts
```diff
@@ -591,4 +610,142 @@ export class Client {
...
+      [anchor.utils.bytes.utf8.encode("metadata")],
+      svmConstants.expressRelayProgram
+    )[0];
+
+    const ixSubmitBid = await expressRelay.methods
+      .submitBid({
+        deadline,
+        bidAmount,
+      })
+      .accountsPartial({
```
m30m, line 685: you can use `accounts` instead of `accountsPartial` and remove anything that's not predetermined which makes the logic simpler

anihamde: i don't want to upload a new IDL for each new program ID (e.g. if we have deployments on a different SVM chain). Instead i'd rather include the new program ID for the new chain in SVM_CONSTANTS and manually construct all the accounts in the client

express_relay/sdk/js/src/index.ts
```diff
@@ -591,4 +616,136 @@ export class Client {
...
+
+  /**
+   * Signs and submits an SVM bid to the express relay auction server
+   * @param bid The SVM bid to sign and submit
+   * @param secretKeys The secret keys that need to sign this transaction, apart from the relayer signer key.
+   * @returns The id of the submitted bid, you can use this id to track the status of the bid
+   */
+  async signAndSubmitSvmBid(
+    bid: BidSvm,
+    secretKeys: Uint8Array[]
```
m30m, line 740: why not an array of keypairs?

express_relay/sdk/js/src/examples/simpleSearcherSvm.ts
```diff
@@ -0,0 +1,182 @@
...
+      dummyExecutable,
+      permission,
+      bidAmount,
+      new anchor.BN(Math.round(Date.now() / 1000 + DAY_IN_SECONDS)),
+      this.chainId
+    );
+
+    try {
+      const { blockhash } = await this.connectionSvm.getLatestBlockhash();
+      bid.transaction.recentBlockhash = blockhash;
```
m30m, line 111: I would rather have the signing here and outside the sdk. Similar to the RPC argument, I can see searchers implementing a bunch of logging and monitoring around the signing procedures and not willing to pass the underlying keypair around so much

express_relay/sdk/js/src/types.ts
```diff
@@ -153,6 +156,29 @@ export type Bid = {
...
+   * @description The execution environment for the bid.
+   */
+  env: "evm";
+};
+/**
+ * Represents a raw SVM bid on acquiring a permission key
+ */
+export type BidSvm = {
+  /**
+   * @description Transaction object, as a base64-encoded string.
```
m30m, line 169: it's not a base64-string anymore

express_relay/sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@pythnetwork/express-relay-evm-js",
-  "version": "0.8.1",
+  "name": "@pythnetwork/express-relay-js",
+  "version": "0.1.0",
```
m30m, line 3: not sure what the best practice is, but I'd rather keep incrementing the versions

m30m (approved):

express_relay/sdk/js/src/index.ts
```diff
@@ -591,4 +614,99 @@ export class Client {
...
+    const ixSubmitBid = await this.constructSubmitBidInstruction(
+      searcher,
+      router,
+      permissionKey,
+      bidAmount,
+      deadline,
+      chainId
+    );
+
+    tx.instructions.unshift(ixSubmitBid);
```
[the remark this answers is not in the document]
anihamde: no particular reason, just preference as to where the instruction lies in the tx by default

compute budget instructions' order in the tx doesn't matter afaik

PR #1851, opened 2024-08-27
fix(hermes): Bump pyth-sdk-solana version to fix attribute iteration bug

jayantk:
Tested locally and it works


ali-bahjati (approved):
can you bump the version of Hermes as well?

PR #1855, opened 2024-08-28
(feat) add hyperlink api reference

aditya520:


apps/api-reference/src/components/EvmApi/run-button.tsx
```diff
@@ -88,7 +88,18 @@ export const RunButton = <ParameterName extends string>(
       {status.type === StatusType.Results && (
         <div>
           <h3 className="mb-2 text-lg font-bold">Results</h3>
-          <Code language="json">{stringifyResponse(status.data)}</Code>
+          {props.type === EvmApiType.Write && status.data && typeof status.data === 'object' && 'hash' in status.data && typeof status.data.hash === 'string' ? (
+            <>
+              <Code language="json">{`Tx Hash: ${status.data.hash}`}</Code>
+              {'link' in status.data && typeof status.data.link === 'string' && (
+                <a href={status.data.link} target="_blank" rel="noopener noreferrer" className="text-sm text-blue-500 hover:underline">
```
cprussin, line 95: I'm thinking that it might be a better UI to just not show the tx hash in a `<Code>` at all; it's a bit strange to use a code block for something that isn't code or json.  Thoughts?

I also suggest using the `InlineLink` component instead of using `a` directly as the `InlineLink` component will apply styling that is consistent with links elsewhere in the app.

Finally, passing `rel="noopener noreferrer"` is [no longer needed](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) as modern browsers will automatically apply `noopener` any link that has `target="_blank"` and `noreferrer` is not really all that meaningful to something like us (in fact it may be even more advantageous for us to expose that we're referring folks to other apps)

apps/api-reference/src/components/EvmApi/run-button.tsx
```diff
@@ -175,10 +186,15 @@ const useRunButton = <ParameterName extends string>({
             })
               .then(({ request }) => writeContract(config, request))
               .then((result) => {
-                setStatus(Results(result));
+                const output: {hash: string, link?: string} = {hash: result};
+                if (config.chains[0].blockExplorers?.default !== undefined) {
+                  const blockExplorerTxLink = `${config.chains[0].blockExplorers.default.url}/tx/${result}`;
+                  output.link = blockExplorerTxLink;
+                }
+                setStatus(Results(output));
```
cprussin, line 194: Minor nit but marginally more idiomatic here would be something like this:

```tsx
.then((hash) => {
    const explorer = config.chains[0].blockExplorers?.default;
    setStatus(Results({
        hash,
        link: explorer ? new URL(`/tx/${hash}`, explorer.url).toString() : undefined
    }))
})
```

apps/hermes/server/Cargo.lock
```diff
@@ -3121,9 +3121,9 @@ dependencies = [
 
 [[package]]
 name = "pyth-sdk-solana"
-version = "0.9.0"
+version = "0.10.2"
```
cprussin, line 3124: Presumably this change is unintentional right?

cprussin (approved):
LGTM, minor stylistic nits but not really that important and you can disregard if you prefer

apps/api-reference/src/components/EvmApi/run-button.tsx
```diff
@@ -88,7 +88,18 @@ export const RunButton = <ParameterName extends string>(
       {status.type === StatusType.Results && (
         <div>
           <h3 className="mb-2 text-lg font-bold">Results</h3>
-          <Code language="json">{stringifyResponse(status.data)}</Code>
+          {props.type === EvmApiType.Write && status.data && typeof status.data === 'object' && 'hash' in status.data && typeof status.data.hash === 'string' ? (
```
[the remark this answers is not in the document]
aditya520: This looks much cleaner. I will keep it mind to use it now.

apps/hermes/server/Cargo.lock
aditya520: Yes. I will revert it.

PR #1860, opened 2024-08-30
feat: add new pyth staking sdk

keyvankhademi:


target_chains/solana/sdk/js/solana_utils/src/transaction.ts
```diff
@@ -419,7 +419,7 @@ export async function sendTransactions(
     signers?: Signer[] | undefined;
   }[],
   connection: Connection,
-  wallet: Wallet,
```
guibescos, line 422: will this break things that use sendTransactions in a non-browser context?

keyvankhademi: I highly doubt it. AnchorWallet is a subset of NodeWallet.

apps/staking/src/api.ts
```diff
@@ -1,11 +1,15 @@
...
 
+import { getCurrentEpoch, getPositionState, PositionState, PythStakingClient } from "@pythnetwork/pyth-staking-sdk";
 import type { WalletContextState } from "@solana/wallet-adapter-react";
-import type { Connection } from "@solana/web3.js";
+import { PublicKey, type Connection } from "@solana/web3.js";
 
 export type StakeAccount = {
-  publicKey: `0x${string}`;
+  // Why was this prefixed with 0x?
+  // publicKey: `0x${string}`;
```
guibescos, line 11: I think this makes sense in EVM land but not in Solana

apps/staking/src/api.ts
```diff
@@ -140,19 +144,68 @@ type AccountHistory = {
...
   // While using mocks we need to clone the MOCK_DATA object every time
   // `loadData` is called so that swr treats the response as changed and
   // triggers a rerender.
-  return { ...MOCK_DATA[context.stakeAccount.publicKey]! };
+
+  const pythStakingClient = new PythStakingClient({ connection: context.connection, wallet: context.wallet });
+  const p = new PublicKey(context.stakeAccount.publicKey);
+  const stakeAccountCustody = await pythStakingClient.getStakeAccountCustody(p);
+
+  const ownerATAAccount = await pythStakingClient.getOwnerPythATAAccount();
```
guibescos, line 170: getOwnerPythAtaAccount

guibescos: https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,204 @@
...
+      .createPosition(
+        {
+          voting: {},
+        },
+        new BN(amount.toString())
+      )
+      .accounts({
+        stakeAccountPositions,
+      })
+      .rpc();
```
guibescos, line 169: Transaction sending seems inconsistent. Here you use anchor's rpc and below you use `sendTransactions` from `solana-utils`

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,204 @@
...
+  public async getStakeAccountCustody(
+    stakeAccountPositions: PublicKey
+  ): Promise<Account> {
+    return getAccount(
+      this.connection,
+      getStakeAccountCustodyAddress(stakeAccountPositions)
+    );
+  }
+
+  public async initializePool({
```
guibescos, line 109: Not entirely convinced about having these things in this sdk since they're not called by users of the app

governance/pyth_staking_sdk/src/staking/accounts.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.owner = new PublicKey(buffer.slice(i, i + 32));
+    let numberOfPositions = Math.floor(
+      (buffer.length - POSITIONS_ACCOUNT_SIZE) / POSITION_BUFFER_SIZE
+    );
+    i += 32;
+    this.positions = [];
+    for (let j = 0; j < numberOfPositions; j++) {
+      if (buffer[i] === 1) {
+        this.positions.push(
+          coder.types.decode("position", buffer.subarray(i + 1))
```
guibescos, line 51: this felt hacky in the previous sdk and still does...

apps/staking/src/api.ts
keyvankhademi: Done

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: fixed

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: i added them for tests but i agree with your point. But let's deal with them later, i don't want to remove the tests right now.

governance/pyth_staking_sdk/src/staking/accounts.ts
keyvankhademi: Is there a better way to deserialize this?

governance/pyth_staking_sdk/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+  ? {
+      [K in keyof T]: T[K] extends From
+        ? To
+        : T[K] extends From | null
+        ? To | null
+        : T[K] extends Record<string, unknown> | Array<unknown>
+        ? Convert<T[K], From, To>
+        : T[K];
+    }
+  : T;
```
cprussin, line 20: Nice, very clever, I like this

governance/pyth_staking_sdk/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+  amount: bigint;
+}[];
+
+export enum PositionState {
+  UNLOCKED = 0,
+  LOCKING = 1,
+  LOCKED = 2,
+  PREUNLOCKING = 3,
+  UNLOCKING = 4,
+}
```
cprussin, line 53: Is there any reason to be assigning the ints here?  Generally the actual enum values should be opaque unless they're actually significant for some reason (and honestly usually if they are significant [a const object can be used as an enum](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) and the semantics are more appropriate)

governance/pyth_staking_sdk/src/utils/bn.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const convertBNToBigInt = <T>(obj: T): ConvertBNToBigInt<T> => {
+  if (obj instanceof BN) {
+    return BigInt(obj.toString()) as any;
+  }
+  if (Array.isArray(obj)) {
+    return obj.map(convertBNToBigInt) as any;
+  }
+  if (typeof obj !== "object" || obj === null || obj.constructor !== Object) {
+    return obj as any;
```
cprussin, line 12: what's the point of typecasting to `any`?  `any` is an absolutely toxic type, I would be really concerned about any code that requires it to work.

Would it be more appropriate to type this as something along these lines:

```ts
const ConvertibleToBigInt = BN | BN[] | Record<string, ConvertableToBigInt>;

const convertBNToBigInt = <T extends ConvertableToBigInt>(obj: T): ConvertBNToBigInt<ConvertableToBigInt> => {
    ...
}
```

?

governance/pyth_staking_sdk/src/utils/bn.ts
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      newObj[key] = convertBNToBigInt(obj[key]);
+    }
+  }
+  return newObj;
+};
+
+export const convertBigIntToBN = <T>(obj: T): ConvertBigIntToBN<T> => {
+  if (typeof obj === "bigint") {
+    return new BN(obj.toString()) as any;
```
cprussin, line 28: Same comment about avoiding `any`, probably `T` needs to be restricted to avoid these cases.  Added benefit is that restricting `T` will also enable typescript to prevent you from calling these functions on values that aren't valid for converting to BN...

governance/pyth_staking_sdk/src/utils/clock.ts
```diff
@@ -0,0 +1,17 @@
+import { BN } from "@coral-xyz/anchor";
+import { Connection } from "@solana/web3.js";
+import { EPOCH_DURATION } from "../constants";
+
+export const getCurrentSolanaTimestamp = async (connection: Connection) => {
+  const slot = await connection.getSlot();
+  const blockTime = await connection.getBlockTime(slot);
+  return new BN(blockTime!);
```
cprussin, line 8: Same comment about avoiding `!`, also should we be returning bigint here instead of BN to be consistent with other APIs?

governance/pyth_staking_sdk/src/utils/position.ts
```diff
@@ -0,0 +1,55 @@
...
+  epoch: bigint;
+}): bigint => {
+  const { stakeAccountPositions, targetWithParameters, positionState, epoch } =
+    options;
+
+  return stakeAccountPositions.data.positions
+    .filter((p) => p && getPositionState(p, epoch) === positionState)
+    .filter((p) => {
+      if (targetWithParameters.voting) {
+        return !!p?.targetWithParameters.voting;
```
cprussin, line 44: Same comment about avoiding `!!`, his line in particular reads unnecessarily confusingly to me too since it's not really obvious at quick glance that it's a boolean being returned and you get a bit of "symbol overload" on `!!p?.`

governance/pyth_staking_sdk/src/utils/position.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+      return (
+        p?.targetWithParameters?.integrityPool?.publisher &&
+        targetWithParameters?.integrityPool?.publisher?.equals(
+          p?.targetWithParameters?.integrityPool?.publisher
+        )
+      );
+    })
+    .map((p) => p!.amount)
```
cprussin, line 53: You can avoid the type cast by using [inferred type predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#inferred-type-predicates0)

governance/pyth_staking_sdk/src/utils/vesting.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+  } else if (vestingSchedule.periodicVesting) {
+    return getPeriodicUnlockSchedule({
+      balance: vestingSchedule.periodicVesting.initialBalance,
+      numPeriods: vestingSchedule.periodicVesting.numPeriods,
+      periodDuration: vestingSchedule.periodicVesting.periodDuration,
+      startDate: vestingSchedule.periodicVesting.startDate,
+    });
+  }
+  assertNever(vestingSchedule);
```
cprussin, line 27: Why is this needed?  Can the `vestingSchedule` be typed such that one of `fullyVested`, `periodicVestingAfterListing`, or `periodicVesting` is guaranteed to be set, or perhaps use an enum to represent the states so you can get exhaustivity checks from TS?

governance/pyth_staking_sdk/src/utils/vesting.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  return unlockSchedule;
+};
+
+// Utility function to assert a value is never
+function assertNever(value: never): never {
+  throw new Error(`Unexpected value: ${value}`);
+}
```
cprussin, line 58: There's no real reason to wrap this in a function, IMO it just adds indirection to what's actually happening and it's easier to just read a throw at the call site.

What I do often do, however, is to create a custom error type to make it really easy to know what error I'm catching without doing string matching:

```ts
class UnexpectedValueError extends Error {
    constructor(value: unknown) {
        super(`Unexpected value: ${value}`);
    }
}
```

cprussin (approved):
I have a ton of little suggestions and tips that you might find valuable but none of this is super important; you can feel free to disregard if you don't like the tips and any of it could be done in a follow up.

The big stuff looks great to me

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
+    total: stakeAccountCustody.amount,
+    governance: {
+      warmup: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.LOCKING, epoch: currentEpoch}),
+      staked: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.LOCKED, epoch: currentEpoch}),
+      cooldown: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.PREUNLOCKING, epoch: currentEpoch}),
+      cooldown2: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.UNLOCKED, epoch: currentEpoch}),
+    },
+    unlockSchedule,
+    locked: unlockSchedule.reduce((sum, { amount }) => sum + amount, 0n),
+    walletAmount: ownerAtaAccount.amount,
```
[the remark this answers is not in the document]
keyvankhademi: ATA means Associated Token Account. All the users Pyth tokens are in their PYTH ATA.

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
-  } else {
-    throw new Error(`Invalid publisher key: "${publisherKey}"`);
-  }
+
+  const pythStakingClient = new PythStakingClient({ connection: context.connection, wallet: context.wallet });
+  const p = new PublicKey(context.stakeAccount.address);
+
+  await pythStakingClient.stakeToPublisher({
+    stakeAccountPositions: p,
+    publisher: new PublicKey(publisherKey),
```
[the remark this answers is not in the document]
keyvankhademi: Yes we should use PublicKey everywhere. I wanted to fix these but apparently i forgot.

apps/staking/src/components/AccountHistory/index.tsx
```diff
@@ -135,7 +135,7 @@ const useAccountHistoryData = () => {
   const apiContext = useApiContext();
 
   const { data, isLoading, ...rest } = useSWR(
-    `${apiContext.stakeAccount.publicKey}/history`,
+    `${apiContext.stakeAccount.address.toBase58()}/history`,
```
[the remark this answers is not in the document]
keyvankhademi: A pattern that I really like is to have functions for caching keys. Something along the lines of `const getHistoryKey = (stakeAccountAddress: Pubkey) => `${stakeAccountAddress}/history`.

apps/staking/src/components/AccountSummary/index.tsx
```diff
@@ -136,8 +136,8 @@ export const AccountSummary = ({
           name="Available Rewards"
           amount={availableRewards}
           description="Rewards you have earned but not yet claimed from the Integrity Staking program"
-          action={<ClaimButton disabled={availableRewards === 0n} />}
```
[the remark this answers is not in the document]
keyvankhademi: good catch. I wanted to test claim, i forgot to revert back.

apps/staking/src/components/WalletProvider/index.tsx
```diff
@@ -30,7 +30,7 @@ export const WalletProvider = ({
   rpc,
 }: Props) => {
   const endpoint = useMemo(
-    () => rpc ?? clusterApiUrl(WalletAdapterNetwork.Mainnet),
+    () => rpc ?? clusterApiUrl(WalletAdapterNetwork.Devnet),
```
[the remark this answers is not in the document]
keyvankhademi: I thought we already do this and this was just the default. I think it's better to have default devnet.

governance/pyth_staking_sdk/keypairs/alice.json
```diff
@@ -0,0 +1,6 @@
+[
+  183, 145, 202, 243, 78, 73, 124, 200, 67, 201, 221, 160, 44, 228, 7, 237, 159,
+  162, 229, 154, 26, 254, 210, 186, 13, 234, 77, 153, 59, 16, 132, 85, 188, 161,
+  67, 10, 200, 220, 80, 95, 43, 11, 222, 172, 144, 237, 178, 104, 39, 96, 83,
+  131, 107, 198, 0, 181, 234, 6, 113, 123, 228, 84, 138, 58
+]
```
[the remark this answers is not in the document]
keyvankhademi: I guess it's better to put them all in the `test/` folder.

apps/staking/src/components/WalletProvider/index.tsx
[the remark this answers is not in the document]
cprussin: No, I never added that in since all the data was demo data, I'll go back and add it in a separate PR

governance/pyth_staking_sdk/src/constants.ts
```diff
@@ -0,0 +1,18 @@
+import { PublicKey } from "@solana/web3.js";
+
+export const EPOCH_DURATION = 60n * 60n * 24n * 7n; // 1 week
+
+export const POSITION_BUFFER_SIZE = 200;
+export const POSITIONS_ACCOUNT_SIZE = 8 + 32;
```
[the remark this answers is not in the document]
keyvankhademi: 8 is account discriminator size and 32 is the account data size

governance/pyth_staking_sdk/src/pdas.ts
```diff
@@ -0,0 +1,37 @@
...
+  stakeAccountPositionsAddress: PublicKey
+) => {
+  return PublicKey.findProgramAddressSync(
+    [Buffer.from("stake_metadata"), stakeAccountPositionsAddress.toBuffer()],
+    STAKING_PROGRAM_ADDRESS
+  )[0];
+};
+
+export const getPoolConfigAddress = () => {
+  return PublicKey.findProgramAddressSync(
```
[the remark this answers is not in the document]
keyvankhademi: I honestly don't know why they have sync and async. Their async one is deprecated in favor of sync. Doesn't make sense to me. What is actually does is a for loop of 256 to find a valid pubkey that is usually found in the first few iterations. So it is not actually blocking anything.

governance/pyth_staking_sdk/src/constants.ts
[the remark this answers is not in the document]
cprussin: right -- I'm just saying it would be more obvious to name them explicitly:

```ts
const ACCOUNT_DISCRIMINATOR = 8;
const ACCOUNT_DATA_SIZE = 32;
const POSITIONS_ACCOUNT_SIZE = ACCOUNT_DISCRIMINATOR + ACCOUNT_DATA_SIZE;
```

governance/pyth_staking_sdk/src/pdas.ts
```diff
@@ -0,0 +1,37 @@
...
+    [Buffer.from("pool_config")],
+    INTEGRITY_POOL_PROGRAM_ADDRESS
+  )[0];
```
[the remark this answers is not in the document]
keyvankhademi: no, these are pure functions with deterministic outputs.

governance/pyth_staking_sdk/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "@cprussin/tsconfig/base.json",
+  "include": ["**/*.ts", "**/*.json"],
+  "exclude": ["node_modules", "dist"],
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "strict": true,
+    "verbatimModuleSyntax": false,
```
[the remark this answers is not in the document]
keyvankhademi: It was giving me an error, I'll show it to you.

governance/pyth_staking_sdk/src/pdas.ts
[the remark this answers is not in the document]
cprussin: Ah, totally makes sense!  Thanks for the context.

Guessing it used to do something off thread and they changed it so it no longer does...

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+      this.provider
+    );
+    this.publisherCapsProgram = new Program(
+      PublisherCapsIdl as PublisherCaps,
+      this.provider
+    );
+  }
+
+  async setGlobalConfig(config: GlobalConfig) {
+    const globalConfigAnchor = convertBigIntToBN(config);
```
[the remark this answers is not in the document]
keyvankhademi: You were totally right. We should not reduce the quality of our app because of a third party library that is not up to date.

governance/pyth_staking_sdk/src/staking/accounts.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { Position, PositionAnchor } from "../types";
+import type { Staking } from "../../types/staking";
+import { POSITION_BUFFER_SIZE, POSITIONS_ACCOUNT_SIZE } from "../constants";
+import { convertBNToBigInt } from "../utils/bn";
+
+export type StakeAccountPositions = {
+  address: PublicKey;
+  data: {
+    owner: PublicKey;
+    positions: (Position | null)[];
```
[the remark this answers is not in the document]
keyvankhademi: It's because of the deserialization, it can be null on chain so have to accommodate in the type. However we can then filter the null and convert to a new type that doesn't have nulls. But I think it's a bit overkill.

governance/pyth_staking_sdk/src/utils/bn.ts
keyvankhademi: I 100% agree with you. I tried for hours trying to avoid any with no luck. Your suggestion doesn't work because our object can literally have anything and anything is ConvertibleToBigInt. For example, `a: {b: BN, c: Pubkey}` here Pubkey should be `ConvertibleToBigInt` because of `Record<string, ConvertableToBigInt>`.

governance/pyth_staking_sdk/keypairs/alice.json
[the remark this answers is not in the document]
guibescos: Same with the program binaries actually

governance/pyth_staking_sdk/src/constants.ts
guibescos, line 6: Technically just the header right?

governance/pyth_staking_sdk/src/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const STAKING_PROGRAM_ADDRESS = new PublicKey(
+  "pytS9TjG1qyAZypk7n8rw8gfW9sUaqqYyMhJQ4E7JCQ"
```
guibescos, line 9: I think the IDLs have these addresses already, see if you can get them from there

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+      IntegrityPoolIdl as IntegrityPool,
+      this.provider
+    );
+    this.publisherCapsProgram = new Program(
+      PublisherCapsIdl as PublisherCaps,
+      this.provider
+    );
+  }
+
+  async setGlobalConfig(config: GlobalConfig) {
```
guibescos, line 64: why setGlobalConfig instead of initGlobalConfig. This is not really a setter tbh since it only works once anyway

governance/pyth_staking_sdk/src/pdas.ts
```diff
@@ -0,0 +1,37 @@
+import { PublicKey } from "@solana/web3.js";
+import {
+  INTEGRITY_POOL_PROGRAM_ADDRESS,
+  STAKING_PROGRAM_ADDRESS,
+} from "./constants";
+
+export const getConfigAddress = () => {
```
guibescos, line 7: seems inconsistent that this one doesn't return just the Pubkey but also the bump

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+        },
+        new BN(amount.toString())
+      )
+      .accounts({
+        stakeAccountPositions,
+      })
+      .instruction();
+
+    const transactions =
+      await TransactionBuilder.batchIntoVersionedTransactions(
```
guibescos, line 247: you could refator this into a function that takes in some instruction and does this step + `sendTransactions`

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  public async stakeToPublisher(options: {
+    stakeAccountPositions: PublicKey;
+    publisher: PublicKey;
+    amount: bigint;
+  }) {
+    const { stakeAccountPositions, publisher, amount } = options;
+    this.integrityPoolProgram.methods
+      .delegate(new BN(amount.toString()))
```
guibescos, line 338: should this use `convertBigIntToBN`

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+    return getUnlockSchedule({
+      vestingSchedule,
+      pythTokenListTime: config.pythTokenListTime!,
+    });
+  }
+
+  public async advanceDelegationRecord(options: {
+    stakeAccountPositions: PublicKey;
+  }) {
+    // TODO: optimize to only send transactions for publishers that have positive rewards
```
guibescos, line 371: agree

governance/pyth_staking_sdk/src/utils/vesting.ts
```diff
@@ -0,0 +1,58 @@
+import type { UnlockSchedule, VestingSchedule } from "../types";
+
+export const getUnlockSchedule = (options: {
+  pythTokenListTime: bigint;
```
guibescos, line 4: You can probably handle the case where pythTokenListTime is undefined by returning an empty list maybe?

governance/pyth_staking_sdk/src/staking/accounts.ts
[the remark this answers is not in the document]
guibescos: Agree maybe we should filter out the nulls after deserializing

governance/pyth_staking_sdk/src/staking/accounts.ts
```diff
@@ -0,0 +1,59 @@
...
+  };
+};
+
+export class StakeAccountPositionsAnchor {
+  public address: PublicKey;
+  public data: PositionAccountDataAnchor;
+
+  constructor(address: PublicKey, data: Buffer, idl: Staking) {
```
guibescos, line 20: This problem was already in the old code, but isn't it weird that the idl is an argument, when we could import the idl in this file?

governance/pyth_staking_sdk/src/utils/position.ts
```diff
@@ -0,0 +1,55 @@
...
+  position: Position,
+  current_epoch: bigint
+): PositionState => {
+  if (current_epoch < position.activationEpoch) {
+    return PositionState.LOCKING;
+  }
+  if (!position.unlockingStart) {
+    return PositionState.LOCKED;
+  }
+  const has_activated = position.activationEpoch <= current_epoch;
```
guibescos, line 18: snake case feels weird

governance/pyth_staking_sdk/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@pythnetwork/staking-sdk",
+  "version": "0.0.0",
+  "description": "Pyth staking SDK",
+  "main": "src/index.ts",
+  "scripts": {
+    "build": "tsc",
+    "test": "jest"
```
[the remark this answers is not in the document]
cprussin: I can probably help with that  but also that's not so important, the important thing is to make sure that lint, prettier check, tsc, and unit test are being run as part of `test`, otherwise you won't get CI

apps/staking/src/api.ts
```diff
@@ -1,17 +1,16 @@
 // TODO remove these disables when moving off the mock APIs
 /* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/no-non-null-assertion */
 
+import { getAmountByTargetAndState, getCurrentEpoch, PositionState, PythStakingClient } from "@pythnetwork/staking-sdk";
+import type { StakeAccountPositions } from "@pythnetwork/staking-sdk/src/staking/accounts";
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
 export const getStakeAccounts = async (
-  _connection: Connection,
-  _wallet: AnchorWallet,
-): Promise<StakeAccount[]> => {
-  await new Promise((resolve) => setTimeout(resolve, MOCK_DELAY));
-  return MOCK_STAKE_ACCOUNTS;
+  connection: Connection,
+  wallet: AnchorWallet,
+): Promise<StakeAccountPositions[]> => {
+  const pythStakingClient = new PythStakingClient({ connection, wallet });
```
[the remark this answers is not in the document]
keyvankhademi: It doesn't technically hold any state, but there are some stuff in the constructor that most likely should not be instantiated on every request.

I agree that the `PythStakingClient` should be instantiated once in the ui and reused, that was the whole purpose of making it a class imo. I didn't do it right now because i was trying to make this PR just to be about the sdk with minimal changes to the ui app.

```
  constructor(config: PythStakingClientConfig) {
    this.connection = config.connection;
    this.wallet = config.wallet;
    this.provider = new AnchorProvider(this.connection, this.wallet, {
      skipPreflight: true,
    });
    this.stakingProgram = new Program(StakingIdl as Staking, this.provider);
    this.integrityPoolProgram = new Program(
      IntegrityPoolIdl as IntegrityPool,
      this.provider
    );
    this.publisherCapsProgram = new Program(
      PublisherCapsIdl as PublisherCaps,
      this.provider
    );
  }
```

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
+  return await pythStakingClient.getAllStakeAccountPositions(wallet.publicKey);
```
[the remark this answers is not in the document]
keyvankhademi: it is completely redundant. it won't have any effect tho since the function will not be called unless the outer promise is awaited.

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
+  const stakeAccountPositions = context.stakeAccount;
+
+  const stakeAccountCustody = await pythStakingClient.getStakeAccountCustody(stakeAccountPositions.address);
+  const publishers = await pythStakingClient.getPublishers();
+  const ownerAtaAccount = await pythStakingClient.getOwnerPythAtaAccount();
+  const currentEpoch = await getCurrentEpoch(context.connection);
+
+  const unlockSchedule = await pythStakingClient.getUnlockSchedule({
+    stakeAccountPositions: stakeAccountPositions.address
+  });
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

apps/staking/src/api.ts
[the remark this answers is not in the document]
keyvankhademi: Yes this is correct. SPL tokens are not stored in the users solana wallet. They are stored in their Associated Token Account (ATA) which is unique per (wallet, mint).

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
+  return {
+    lastSlash: undefined,
+    availableRewards: 0n,
+    expiringRewards: undefined,
+    total: stakeAccountCustody.amount,
+    governance: {
+      warmup: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.LOCKING, epoch: currentEpoch}),
+      staked: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.LOCKED, epoch: currentEpoch}),
+      cooldown: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.PREUNLOCKING, epoch: currentEpoch}),
+      cooldown2: getAmountByTargetAndState({stakeAccountPositions, targetWithParameters: {voting: {}}, positionState: PositionState.UNLOCKED, epoch: currentEpoch}),
```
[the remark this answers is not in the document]
keyvankhademi: done

apps/staking/src/api.ts
```diff
@@ -140,96 +139,135 @@ type AccountHistory = {
...
 };
 
 export const stakeGovernance = async (
   context: Context,
   amount: bigint,
 ): Promise<void> => {
-  await new Promise((resolve) => setTimeout(resolve, MOCK_DELAY));
-  MOCK_DATA[context.stakeAccount.publicKey]!.governance.warmup += amount;
+  const pythStakingClient = new PythStakingClient({ connection: context.connection, wallet: context.wallet });
+  const p = new PublicKey(context.stakeAccount.address);
```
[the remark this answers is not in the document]
keyvankhademi: yes i forgot to fix this. my intention is to use PublicKey everywhere.

apps/staking/src/components/AccountSummary/index.tsx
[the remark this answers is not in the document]
keyvankhademi: fixed.

apps/staking/src/components/AccountSummary/index.tsx
```diff
@@ -136,8 +136,8 @@ export const AccountSummary = ({
...
-          {...(expiringRewards.amount > 0n && {
+          action={<ClaimButton  />}
+          {...(!!expiringRewards && expiringRewards.amount > 0n && {
```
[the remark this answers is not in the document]
keyvankhademi: makes sense. fixed.

governance/pyth_staking_sdk/keypairs/alice.json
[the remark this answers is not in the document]
keyvankhademi: fixed.

apps/staking/src/api.ts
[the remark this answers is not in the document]
cprussin: that's ok @keyvankhademi just merge this as is and I'll handle putting it into context in a follow up PR.

apps/staking/src/api.ts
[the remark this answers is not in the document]
cprussin: The promise will be called and dispatched even if not awaited so the behavior is no different.  In the past, eslint would ban `return await` because it [created extra microtasks](https://eslint.org/docs/latest/rules/no-return-await) but that's [no longer true](https://v8.dev/blog/fast-async), so it's actually identical functionally to returning the promise.  I just don't like the semantics much of `return await`.

governance/pyth_staking_sdk/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@types/node": "^22.0.0",
+    "@typescript-eslint/parser": "^8.3.0",
+    "eslint": "^9.8.0",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.5.4",
+    "@cprussin/eslint-config": "^3.0.0",
+    "@cprussin/jest-config": "^1.4.1",
+    "@cprussin/prettier-config": "^2.1.1",
+    "@cprussin/tsconfig": "^3.0.1"
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/constants.ts
```diff
@@ -0,0 +1,18 @@
+import { PublicKey } from "@solana/web3.js";
+
+export const EPOCH_DURATION = 60n * 60n * 24n * 7n; // 1 week
```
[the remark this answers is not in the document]
keyvankhademi: good idea

governance/pyth_staking_sdk/src/constants.ts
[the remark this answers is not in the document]
keyvankhademi: Fixed.

governance/pyth_staking_sdk/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "@cprussin/tsconfig/base.json",
+  "include": ["**/*.ts", "**/*.json"],
+  "exclude": ["node_modules", "dist"],
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "strict": true,
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
+import { AnchorProvider, BN, Program } from "@coral-xyz/anchor";
+import { Connection, PublicKey } from "@solana/web3.js";
+import { type Staking } from "../types/staking";
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/package.json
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/tsconfig.json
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+              memcmp: {
+                offset: 8,
+                bytes: user.toBase58(),
+              },
+            },
+          ],
+        }
+      );
+    return res.map((account) => {
+      const stakeAccountPositionsAnchor = new StakeAccountPositionsAnchor(
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+  public async getStakeAccountPositions(
+    stakeAccountPositions: PublicKey
+  ): Promise<StakeAccountPositions> {
+    const account =
+      await this.stakingProgram.provider.connection.getAccountInfo(
+        stakeAccountPositions
+      );
+    const stakeAccountPositionsAnchor = new StakeAccountPositionsAnchor(
+      stakeAccountPositions,
+      account!.data,
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/staking/accounts.ts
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/staking/accounts.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.address = address;
+    this.data = new PositionAccountDataAnchor(data, idl);
+  }
+
+  public toStakeAccountPositions(): StakeAccountPositions {
+    return {
+      address: this.address,
+      data: {
+        owner: this.data.owner,
+        positions: this.data.positions.map(convertBNToBigInt),
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/staking/accounts.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+    };
+  }
+}
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/types.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/utils/clock.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/utils/position.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/utils/position.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/utils/vesting.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/utils/vesting.ts
keyvankhademi: n/a anymore

governance/pyth_staking_sdk/src/constants.ts
keyvankhademi: let's deal with this later

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pdas.ts
keyvankhademi: will deal with this later

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -0,0 +1,410 @@
...
+  }) {
+    const { stakeAccountPositions, publisher, amount } = options;
+    this.integrityPoolProgram.methods
+      .delegate(new BN(amount.toString()))
+      .accounts({
+        owner: this.wallet.publicKey,
+        publisher,
+        stakeAccountPositions,
+      })
+      .rpc();
```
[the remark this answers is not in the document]
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: fixed.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: we have to fix it soon. Currently there are 8 transactions which is probably not acceptable.

governance/pyth_staking_sdk/src/utils/vesting.ts
keyvankhademi: handled it in the outer function.

governance/pyth_staking_sdk/src/staking/accounts.ts
keyvankhademi: that was my initial thought too, but when importing the idl it doesn't work for some reason.

governance/pyth_staking_sdk/src/utils/position.ts
keyvankhademi: lol rust habits

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: done.

PR #1866, opened 2024-09-03
feat: update hermes to support publisher stake caps

guibescos:
This PR adds a new endpoint `/api/get_publisher_stake_caps_update_data` for consumption of publisher caps messages


apps/hermes/server/src/api.rs
```diff
@@ -150,6 +150,10 @@ where
         .route("/api/get_price_feed", get(rest::get_price_feed))
         .route("/api/get_vaa", get(rest::get_vaa))
         .route("/api/get_vaa_ccip", get(rest::get_vaa_ccip))
+        .route(
```
ali-bahjati, line 153: we have deprecated the old api endpoints and don't add things to them. could you please move it to `v2` and use something like `v2/updates/publisher_stake_caps/latest`?

apps/hermes/server/src/api/types.rs
```diff
@@ -271,9 +273,28 @@ impl From<PriceFeedUpdate> for ParsedPriceUpdate {
...
+    pub publisher: Pubkey,
+    pub cap:       u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct GetPublisherStakeCapsUpdateDataResponse {
+    pub binary: BinaryUpdate,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub parsed: Option<Vec<ParsedPublisherStakeCapsUpdate>>,
+}
```
ali-bahjati, line 293: This is only used in a single api right? please move it there.

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -403,6 +417,41 @@ where
...
+            .collect::<Result<Vec<_>>>()?;
+
+        let update_data = construct_update_data(messages.into_iter().map(|m| m.into()).collect())?;
+        Ok(PublisherStakeCapsWithUpdateData {
+            publisher_stake_caps,
+            update_data,
+        })
+    }
+
     async fn get_price_feed_ids(&self) -> HashSet<PriceIdentifier> {
```
ali-bahjati, line 455: Please also fix this method.

ali-bahjati (changes requested):
The code looks good, I left some comments about some structural changes in the code. I think that publishers cap is not very compatible with the terminologies in Hermes and we need a larger refactor (probably later) to handle them.

apps/hermes/server/src/api.rs
guibescos: done

apps/hermes/server/src/api/types.rs
guibescos: done

ali-bahjati (approved):

PR #1874, opened 2024-09-08
chore(staking): initialize pyth staking client only once

cprussin:
This PR restructures things a bit so that the staking client can be initialized only once and stored in React context


apps/staking/src/api.ts
```diff
@@ -210,12 +195,12 @@ export const loadData = async (context: Context): Promise<Data> => {
     walletAmount: ownerAtaAccount.amount,
     integrityStakingPublishers: publishers.map(({ pubkey: publisher }) => ({
       apyHistory: [], // TODO
-      isSelf: false, // TODO
+      isSelf: publisher === client.wallet.publicKey,
```
cprussin, line 198: fyi @keyvankhademi

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -274,30 +274,26 @@ export class PythStakingClient {
     return sendTransaction([instruction], this.connection, this.wallet);
   }
 
-  public async stakeToPublisher(options: {
```
cprussin, line 277: @keyvankhademi FYI I updated the api signatures for a few functions to make them less verbose, I don't think it's less clear though but PTAL and let me know if you're OK with these changes

keyvankhademi: In a typed language like strict ts or rust the type of a variable also brings information when working with the code. I think the `address` in `stakeAccountPositionsAddress: PublicKey` is redundant because the type PublicKey already shows that this is an address. Having said that I'm not necessarily opposed to this change. Feel free to keep it this way.

keyvankhademi (approved):

governance/pyth_staking_sdk/src/pyth-staking-client.ts
cprussin: Yeah, I agree with you, but in this case it felt weird to name it `stakeAccountPositions` because my expectation from that variable name is that it's a list of positions.

Would `stakeAccount` be a more correct name?

apps/staking/src/api.ts
keyvankhademi: This is actually incorrect, a publisher stake account is not necessarily the owner of the stake account. I will create a function in the sdk to retrieve the publishers stake accounts.

cprussin: ah, ok, thanks -- I'll undo this line then

PR #1875, opened 2024-09-08
feat(staking): sort, filter, and paginate publishers

cprussin:


apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -98,46 +119,270 @@ export const OracleIntegrityStaking = ({
...
+              calculateApy(b.poolCapacity, b.poolUtilization, false)
+            );
+          }
+          case SortField.NumberOfFeeds: {
+            return Number(a.numFeeds - b.numFeeds);
+          }
+          case SortField.PoolUtilization: {
+            return Number(
+              a.poolUtilization / a.poolCapacity -
+                b.poolUtilization / b.poolCapacity,
```
keyvankhademi, line 178: you can re-write a/b < c/d as a*d < c*d to avoid divisions thus being less error prone.

keyvankhademi (approved):

PR #1880, opened 2024-09-10
feat(staking): add more real data

keyvankhademi:


apps/staking/src/app/api/publishers-ranking/route.tsx
```diff
@@ -0,0 +1,10 @@
+import { NextResponse } from "next/server";
+
+export async function GET() {
+  const publisherRankingsResponse = await fetch(
+    "https://www.pyth.network/api/publishers-ranking?cluster=pythnet",
```
cprussin, line 5: This endpoint is just a thin proxy to web-api, see https://github.com/pyth-network/pyth-web/blob/main/pages/api/publishers-ranking.tsx.  We should just fetch to `https://web-api.pyth.network/publishers_ranking?cluster=pythnet` instead to avoid the extra hop.

Also make sure to add `{ cache: 'no-store' }` or [next will cache the response](https://nextjs.org/docs/app/api-reference/functions/fetch#optionscache) and since there's no datestamp in the query params, we'd get stale ranking data in the UI.

Note that web-api itself is also fairly slim and is really just hitting a redis cache [or calling a pretty simple clickhouse query](https://github.com/pyth-network/web-api/blob/main/web_api/db.py#L1035-L1064) so IMO we should ideally move away from using web-api too and just wrap this up in a node lib, but we'll need to get access control to clickhouse from Vercel set up so that's a problem for another day......

apps/staking/src/app/api/publishers-ranking/route.tsx
```diff
@@ -0,0 +1,10 @@
...
+  );
+
+  const publisherRankings = (await publisherRankingsResponse.json()) as JSON;
+  return NextResponse.json(publisherRankings);
+}
```
cprussin, line 10: In this case, since the endpoint is going to be a straight up proxy, you can actually avoid having the endpoint entirely by using [nextjs rewrites](https://nextjs.org/docs/app/api-reference/next-config-js/rewrites#rewriting-to-an-external-url)

governance/pyth_staking_sdk/src/utils/pool.ts
```diff
@@ -0,0 +1,34 @@
+import { PublicKey } from "@solana/web3.js";
+
+import type { PoolDataAccount, PublisherData } from "../types";
+
+export const extractPublisherData = (
+  poolData: PoolDataAccount,
+): PublisherData => {
+  return poolData.publishers
+    .filter((publisher) => !publisher.equals(PublicKey.default))
```
guibescos, line 9: This feels wrong if there were gaps in the publishers (which can't happen in the program though)

apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -517,14 +518,7 @@ const Publisher = ({
...
         <PublisherTableCell className="text-center">
-          <div>
-            {calculateApy(
-              publisher.poolCapacity,
-              publisher.poolUtilization,
-              publisher.isSelf,
-            )}
-            %
-          </div>
+          <div>{publisher.lastApy}%</div>
```
cprussin, line 521: Nice but you should also change https://github.com/pyth-network/pyth-crosschain/blob/main/apps/staking/src/components/OracleIntegrityStaking/index.tsx#L177-L180 to use this new field

Note as I mentioned above we still need a method to calculate the apy for https://github.com/pyth-network/pyth-crosschain/blob/main/apps/staking/src/components/OracleIntegrityStaking/index.tsx#L646-L686

apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -517,14 +518,7 @@ const Publisher = ({
           </Meter>
         </PublisherTableCell>
         <PublisherTableCell className="text-center">
-          <div>
-            {calculateApy(
```
guibescos, line 521: I think the APY we display here should be the forecasted APY for the next epoch, if there are no more changes

apps/staking/src/hooks/use-stake-account.tsx
```diff
@@ -109,13 +118,16 @@ const useStakeAccountState = () => {
           signTransaction: wallet.signTransaction,
         },
       });
+      // TODO: use env var to support mainnet
+      const hermesClient = new HermesClient("https://hermes-beta.pyth.network");
```
cprussin, line 122: IMO this should be in a separate react context, it doesn't really make sense to be in this context as it's not associated to the wallet state (I should rename this hook as it's no longer just about staking accounts, it's now really a hook for the wallet state, where the list of stake accounts & the selected stake account is part of the wallet state -- and the Pyth Staking Client is here because it is only valid in certain wallet states).

governance/pyth_staking_sdk/src/utils/pool.ts
```diff
@@ -0,0 +1,34 @@
...
+    .map((publisher, index) => ({
+      pubkey: publisher,
+      stakeAccount:
+        poolData.publisherStakeAccounts[index] === undefined ||
+        poolData.publisherStakeAccounts[index].equals(PublicKey.default)
+          ? null
```
cprussin, line 15: Minor preference but I suggest just using `undefined` instead of mixing undefined and null; the fact that javascript has both is a wart and it's very very easy to misuse them in subtle ways.  For all intents and purposes they are typically used interchangeably, so best to just standardize on one and always use it.  Due to the semantics of missing values, you generally can't avoid `undefined`, but `null` is generally only introduced in user code, so I generally suggest just standardizing on using `undefined` everywhere.

If you need a true ternary state, I will always recommend using the [fake sum types](https://www.typescriptlang.org/play/#code/C4TwDgpgBAIghsOUC8UDeAoK2oDc4A2ArhAFxQB2RAtgEYQBOGAvgNwYYRXVQDKiwCABVw0TDigA5APbB+cBoIAmAGiw4AMtLhKAlhQDmaiQFEGDaQ2M5eRAMZ2IAZycsOd6RSfA+A6KnEcGTlERQglcgAKAEoUAD4oSLQoUEhyeUERSAA6YPkwpSg4JygPLx9maOtsLR19AyjY5ASklNF0vyyIbNq9QyKSsu8oSuqoMwsGKMZJ8iIKAGsKaQB3CiaW5NSyXwRhUWyJywHSz28VKBnj0fVsWwdnJyilBDhyeEQNxK323cyD+6OFwnIbAC4vRAjaIsdgYbZ-fxQABKEGARAYFC6AB5ttIAGYIgDaCwgIHxbUg5IyEAAunFYRhQVAnAALVbUlCJbx7Dp7L6BbBOFa6YB2FlcvzZbaxAUSUrFaDUrq5WT5ZTkBio9EUKAAImCuwKAEJdew5Tg7AqEcrevUNVqMXrbYYTWbzfKnIrOgcjlMoJq0Y6AAYAQR1VwYUGkDnRmoiUAAJGhuYJshHmEG3ebLZ7rQD7ECnv6HTqg71wnhCCRyEmU90IXBsvhiBAM1mRm5GWdpARugRpAZIqz2X5ItSVSEFMoYtFouwhj2+wOh2yVtSx5LnYPZ3Ou15F9l+4Ph2vR+PfZEKBAVuNzJZIrqiOLpCzdTvdwve4flyf1+PAY8rTNtWUAAKwAAxQrOrBAA) pattern I shared with you earlier, it's far more explicit, easier to understand the intent, and less error-prone.

cprussin (changes requested):
Most important changes are:

1. Calculate upcoming epoch APY rather than last epoch APY
2. Move some stuff to `Promise.all`
3. Validate publisher ranking response using schema instead of typecasting
4. Remove the `publishers-ranking` endpoint and instead use nextjs rewrites; also rewrite to web-api directly instead of rewriting to www.pyth.network

Everything else is optional / minor

apps/staking/src/components/OracleIntegrityStaking/index.tsx
keyvankhademi: done

apps/staking/src/api.ts
```diff
@@ -46,6 +56,7 @@ type Data = {
     poolUtilization: bigint;
     numFeeds: number;
     qualityRanking: number;
+    lastApy: number;
```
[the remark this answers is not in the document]
keyvankhademi: done

apps/staking/src/api.ts
```diff
@@ -144,22 +155,29 @@ export const getStakeAccounts = async (
...
-    client.getPublishers(),
+    client.getPoolDataAccount(),
     client.getOwnerPythAtaAccount(),
     getCurrentEpoch(client.connection),
     client.getUnlockSchedule(stakeAccount.address),
   ]);
 
+  const publishers = extractPublisherData(poolData);
+
+  const publisherRankingsResponse = await fetch("/api/publishers-ranking");
```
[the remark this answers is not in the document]
keyvankhademi: done.

apps/staking/src/api.ts
```diff
@@ -179,6 +197,17 @@ export const loadData = async (
       epoch: currentEpoch,
     });
 
+  const publisherCaps = await hermesClient.getLatestPublisherCaps({
+    parsed: true,
+  });
```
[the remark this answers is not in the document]
keyvankhademi: Done.

apps/staking/src/app/api/publishers-ranking/route.tsx
keyvankhademi: done.

apps/staking/src/components/OracleIntegrityStaking/index.tsx
keyvankhademi: both use calculateApy now

apps/staking/src/api.ts
```diff
@@ -144,22 +155,29 @@ export const getStakeAccounts = async (
...
     client.getOwnerPythAtaAccount(),
     getCurrentEpoch(client.connection),
     client.getUnlockSchedule(stakeAccount.address),
   ]);
 
+  const publishers = extractPublisherData(poolData);
+
+  const publisherRankingsResponse = await fetch("/api/publishers-ranking");
+  const publisherRankings =
+    (await publisherRankingsResponse.json()) as PublisherRankings;
```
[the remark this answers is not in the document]
keyvankhademi: done.

apps/staking/src/app/api/publishers-ranking/route.tsx
keyvankhademi: used it in rewrite

apps/staking/src/api.ts
```diff
@@ -179,6 +197,17 @@ export const loadData = async (
...
     });
 
+  const publisherCaps = await hermesClient.getLatestPublisherCaps({
+    parsed: true,
+  });
+
+  const getPublisherCap = (publisher: PublicKey) =>
+    BigInt(
+      publisherCaps.parsed?.[0]?.publisher_stake_caps.find(
+        ({ publisher: p }) => p === publisher.toBase58(),
```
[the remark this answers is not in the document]
keyvankhademi: not too expensive but fixed it.

governance/pyth_staking_sdk/src/utils/pool.ts
keyvankhademi: yes there are a lot of assumptions based on that.

cprussin (approved):

PR #1887, opened 2024-09-11
feat(staking): create a stake account on initial deposit

cprussin:
This PR includes a fairly significant refactor to restructure some hooks:

- `useStakeAccount` has been replaced by a very similar hook called `useApi`.  The biggest difference here, apart from some renaming, is that the states in `useApi` expose wrapped functions from `api.ts`, bound to the current stake account & client.  So there's no need to interact with the stake account or staking client or `api.ts` anywhere else in the app any more.
- `useTransfer` is replaced by a more generic hook `useAsync`, which takes an async callback and returns a function to trigger execution and the current execution satate.  I did this because, once moving everything to `useApi`,  everything specific to transferring, including busting the swr cache, got moved to the `useApi` wrapper functions, and what was left was just really a generic mechanism for running and tracking state of an async function.
- `useAccountHistory` and `useDashboardData` have been replaced by `useData`, which takes `useSwr` and wraps it to bind the output to enum types.  Both of the original hooks were doing mostly the same thing but with marginally different ways to set up their fetch functions; with `useApi` those things are now abstracted in the API states.

With those in place, I was able to use the states in `useApi` to abstract away how to handle depositing & fetching data when a stake account exists from when it doesn't, and I was able to leverage the type system to guarantee that an API method can't be called in an invalid state, which cleaned up a lot of other code too and removed the possibility of some class of runtime exceptions.


apps/staking/src/api.ts
```diff
@@ -144,73 +144,114 @@ export const getStakeAccounts = async (
...
+            publisher.publicKey,
+            PositionState.UNLOCKED,
+          ),
+        },
+      })),
+    };
+  }
+};
+
+const loadPublisherData = async (client: PythStakingClient) => {
```
cprussin, line 241: @keyvankhademi FYI, this stuff is separated out because there are now two separate flows for fetching dashboard data -- one for when there is a stake account and one for when there isn't.  The publisher data is common between the two.

apps/staking/src/api.ts
```diff
@@ -221,6 +262,13 @@ export const loadAccountHistory = async (
   return mkMockHistory();
 };
 
+export const createStakeAccountAndDeposit = async (
```
cprussin, line 265: fyi @keyvankhademi

apps/staking/src/components/AccountSummary/index.tsx
```diff
@@ -130,16 +132,25 @@ export const AccountSummary = ({
               actionDescription="Move funds from your account back to your wallet"
               actionName="Withdraw"
               max={availableToWithdraw}
-              transfer={withdraw}
-              isDisabled={availableToWithdraw === 0n}
+              {...(api.type === ApiStateType.Loaded && {
+                transfer: api.withdraw,
+              })}
```
cprussin, line 137: Note to reviewer: in this PR I made the `<TransferButton>` component render a disabled button if the max is `0n` or if `transfer` is undefined.  That makes it a lot easier to just use it all over and let it disable itself correctly if there's no function to call (generally meaning the API is in a state where this transfer is impossible) or if there's no tokens to transfer.  So you'll see here and a few other places where I removed some checks that control the button states, since those are now handled by the button itself.

keyvankhademi (approved):

PR #1899, opened 2024-09-11
feat(staking): add create staking account

keyvankhademi:


cprussin (approved):

governance/pyth_staking_sdk/eslint.config.js
```diff
@@ -6,6 +6,7 @@ export default [
     rules: {
       "n/no-unpublished-import": "off",
       "unicorn/no-null": "off",
+      "unicorn/prefer-node-protocol": "off",
```
guibescos, line 9: I don't understand this

guibescos (approved):
lgtm

governance/pyth_staking_sdk/eslint.config.js
keyvankhademi: it asking to import `node:crypto` instead of `crypto` which doesn't work for our setup.

PR #1900, opened 2024-09-12
feat(xc-admin): add price store instructions and executor support

Riateche:


governance/xc_admin/packages/xc_admin_common/src/price_store.ts
```diff
@@ -0,0 +1,236 @@
...
+enum InstructionId {
+  Initialize = 0,
+  SubmitPrices = 1,
+  InitializePublisher = 2,
+}
+
+export function createPriceStoreInstruction(
+  data: PriceStoreInstruction
+): TransactionInstruction {
+  if (data.type == "Initialize") {
```
ali-bahjati, line 50: i recommend using switch here, but if not triple equals (`===`)

governance/xc_admin/packages/xc_admin_common/src/price_store.ts
```diff
@@ -0,0 +1,236 @@
...
+        payerKey: instruction.keys[0].pubkey,
+        authorityKey,
+      },
+    };
+  } else if (instructionId == InstructionId.InitializePublisher) {
+    if (instruction.data.length < 35) {
+      throw new Error("instruction data is too short");
+    }
+    const publisherKey = new PublicKey(instruction.data.subarray(3, 35));
+    if (instruction.keys.length != 5) {
```
ali-bahjati, line 165: de we need to check remaining accounts to be valid?

ali-bahjati (approved):
Nice!

governance/xc_admin/packages/xc_admin_common/src/price_store.ts
Riateche: I've added a check that accounts and instruction data match the expected values.

governance/xc_admin/packages/xc_admin_common/src/price_store.ts
Riateche: Refactored to use switch.

ali-bahjati (approved):

PR #1910, opened 2024-09-13
feat(apps/price_pusher): add fuel price pusher

cctdaniel:


apps/price_pusher/.gitignore
```diff
@@ -1,3 +1,4 @@
 docker-compose.yaml
 price-config.yaml
 lib
+mnemonic
```
cctdaniel, line 4: many of the sample testnet config files use `/mnemonic` as the private key file so adding it here to prevent people from accidentally comitting it

ali-bahjati: thanks

apps/price_pusher/src/fuel/fuel.ts
```diff
@@ -0,0 +1,131 @@
...
+import { addLeading0x, DurationInSeconds } from "../utils";
+import { Logger } from "pino";
+import { Provider, Contract, hexlify, arrayify, Wallet, BN } from "fuels";
+import {
+  PYTH_CONTRACT_ABI,
+  FUEL_ETH_ASSET_ID,
+} from "@pythnetwork/pyth-fuel-js";
+
+function tai64ToUnix(tai64: BN): number {
+  const result = BigInt(tai64.toString()) - BigInt(2n ** 62n) - 10n;
```
ali-bahjati, line 17: what is this actually? what's 2^62 and 10? can you add a comment?

ali-bahjati (approved):

PR #1911, opened 2024-09-13
feat(target_chains/fuel): update fuel toolchains to the latest version

cctdaniel:


target_chains/fuel/contracts/pyth-contract/src/main.sw
```diff
@@ -848,57 +847,62 @@ impl PythGovernance for Contract {
...
+        GovernanceAction::RequestGovernanceDataSourceTransfer => {
+            // RequestGovernanceDataSourceTransfer can be only part of AuthorizeGovernanceDataSourceTransfer message
+            // The `revert` function only accepts u64, so as
+            // a workaround we use require.
+            require(false, PythError::InvalidGovernanceMessage);
+        },
+        _ => {
+            // The `revert` function only accepts u64, so as
+            // a workaround we use require.
+            require(false, PythError::InvalidGovernanceMessage);
```
cctdaniel, line 905: there is a compiler issue with fuel where it fails to compile with `Too many arguments, cannot handle.: Immediate12TooLarge` error and the solution was to make the contract functions call internal functions

PR #1913, opened 2024-09-13
feat(staking): add merge positions after claim rewards

keyvankhademi:


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -441,7 +441,7 @@ export class PythStakingClient {
 
     // anchor does not calculate the correct pda for other programs
     // therefore we need to manually calculate the pdas
-    return Promise.all(
+    const advanceDelegationRecordInstructions = await Promise.all(
```
guibescos, line 444: should they all be in the Promise.all ?

guibescos (approved):
makes sense to me

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: I need to have them separated, it's more cleaner this way. The performance overhead is also negligible.

PR #1915, opened 2024-09-13
feat(staking): add reward expiry

keyvankhademi:


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -609,6 +622,24 @@ export class PythStakingClient {
...
+        this.getDelegationRecord(stakeAccountPositions, pubkey),
+      ),
+    );
+
+    let lowestEpoch: bigint | undefined;
+    for (const record of delegationRecords) {
+      if (lowestEpoch === undefined || record.lastEpoch < lowestEpoch) {
+        lowestEpoch = record.lastEpoch;
+      }
+    }
```
keyvankhademi, line 637: @guibescos I'm assuming that when there is no stake (i.e reward is zero) the position will get removed after merging positions. Therefore, this will only account for publishers that we have an stake with.

cprussin (approved):

PR #1918, opened 2024-09-13
feat(staking): implement some feedback items

cprussin:
- Fix alignment of stake account indicator in header
- Update language of opt-out modal
- Round APY in publishers table to 2 digits
- Show estimated APY even if amount exceeds wallet balance
- Separate main account from other accounts in stake account menu
- Store last selected stake account in local storage

### DIFF
--- a/apps/staking/src/components/Home/index.tsx
+++ b/apps/staking/src/components/Home/index.tsx
@@ -16,6 +16,10 @@ import { Error as ErrorPage } from "../Error";
 import { Loading } from "../Loading";
 import { NoWalletHome } from "../NoWalletHome";
 
+const ONE_SECOND_IN_MS = 1000;
+const ONE_MINUTE_IN_MS = 60 * ONE_SECOND_IN_MS;
+const REFRESH_INTERVAL = 1 * ONE_MINUTE_IN_MS;
+
 export const Home = () => {
   const isSSR = useIsSSR();
 
@@ -30,6 +34,8 @@ const MountedHome = () => {
     case ApiStateType.LoadingStakeAccounts: {
       return <Loading />;
     }
+    case ApiStateType.WalletDisconnecting:
+    case ApiStateType.WalletConnecting:
     case ApiStateType.NoWallet: {
       return <NoWalletHome />;
     }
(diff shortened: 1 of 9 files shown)


apps/staking/src/components/WalletButton/index.tsx
```diff
@@ -41,13 +46,18 @@ import {
...
 import { useLogger } from "../../hooks/use-logger";
 import { usePrimaryDomain } from "../../hooks/use-primary-domain";
 import { AccountHistory } from "../AccountHistory";
 import { Button } from "../Button";
 import { ModalDialog } from "../ModalDialog";
 import { TruncatedKey } from "../TruncatedKey";
 
+const ONE_SECOND_IN_MS = 1000;
+const ONE_MINUTE_IN_MS = 60 * ONE_SECOND_IN_MS;
+const REFRESH_INTERVAL = 1 * ONE_MINUTE_IN_MS;
```
cprussin, line 59: @keyvankhademi it's a bit clunky because this is duplicated in two places since the dashboard data needs to be loaded now in the wallet menu too so we can display which account is the main account.  So we have to do the `useData` call on the dashboard data.  But it won't trigger a separate network call because swr will re-use the global cache, and this will go away when I refactor the dashboard data into use-api eventually, so I don't think it's a big issue to leave the duplication for now.

apps/staking/src/components/Home/index.tsx
```diff
@@ -16,6 +16,10 @@ import { Error as ErrorPage } from "../Error";
 import { Loading } from "../Loading";
 import { NoWalletHome } from "../NoWalletHome";
 
+const ONE_SECOND_IN_MS = 1000;
```
keyvankhademi, line 19: we can have these in a global constants file to avoid repetition.

cprussin: Yeah, I'll refactor them in eventually, but it will go away when I clean up use-api anyways

apps/staking/src/components/WalletButton/index.tsx
keyvankhademi: oh ignore my last comment then, i just saw this.

keyvankhademi (approved):

PR #1931, opened 2024-09-18
feat(staking): include warmup token in pool utilization

keyvankhademi:


keyvankhademi:
I was also thinking that maybe we should include the warmup in `poolUtilization` and remove `poolUtilizationDelta`.

cprussin (approved):

cprussin:
> I was also thinking that maybe we should include the warmup in `poolUtilization` and remove `poolUtilizationDelta`.

Agreed

guibescos (approved):

PR #1932, opened 2024-09-18
feat(staking): make wallet optional in staking sdk

keyvankhademi:


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -68,7 +68,9 @@ export class PythStakingClient {
   constructor(config: PythStakingClientConfig) {
     this.connection = config.connection;
     this.wallet = config.wallet;
-    this.provider = new AnchorProvider(this.connection, this.wallet, {
+
+    // {} as AnchorWallet is a workaround for AnchorProvider requiring a wallet
+    this.provider = new AnchorProvider(this.connection, this.wallet ?? {} as AnchorWallet, {
```
guibescos, line 73: does {} === undefined?

guibescos (approved):

governance/pyth_staking_sdk/src/pyth-staking-client.ts
keyvankhademi: no, but the check is for `this.wallet` not the `{}` in the provider.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -604,6 +625,7 @@ export class PythStakingClient {
   }
 
   public async getClaimableRewards(stakeAccountPositions: PublicKey) {
+    this.assertWallet();
```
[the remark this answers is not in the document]
keyvankhademi: it's using `this.wallet.publicKey` but since it's only the simulation i think we can replace this with any pubkey.

PR #1933, opened 2024-09-18
feat(staking): make sdk ready for release

keyvankhademi:


governance/pyth_staking_sdk/scripts/update-package-json.js
```diff
@@ -0,0 +1,22 @@
...
+ * instead of the TypeScript source files.
+ */
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const distPackageJsonPath = path.join(__dirname, "..", "dist", "package.json");
+
+const packageJson = JSON.parse(fs.readFileSync(distPackageJsonPath, "utf8"));
+
+packageJson.exports = {
```
guibescos, line 18: there must be a better way to deal with this 😃

guibescos: Let's see what @cprussin says

cprussin: I actually have a package for doing this: https://www.npmjs.com/package/@cprussin/transform-package-json

You essentially have to do some transformation like this to the package.json to dual export packages or to use nextjs's transpileModules on a package that gets published, so I think it's fine however you want to do this, and what you have here is perfectly ok

cprussin (approved):

PR #1938, opened 2024-09-19
feat(staking): add default instead of assertWallet

keyvankhademi:
Actually I think this approach is better.


guibescos (approved):
It's weird that you can call the methods that send transactions and I won't return an error saying this is a read-only

keyvankhademi:
I like this idea. will do it in another PR soon.

PR #1939, opened 2024-09-19
feature(target_chains/ton): add upgrade standard

cctdaniel:
due to the limit of guardians payload size, we had to split this one into 2 steps:
1. create a governance vaa with the hash of the new code as the payload signed by the guardians and store it in the storage
2. send a `OP_UPGRADE_CONTRACT` tx to perform the actual upgrade (this will check that the new code hash is the same as the previously stored hash)


Riateche (approved):
LGTM

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -254,8 +254,25 @@ int apply_decimal_expo(int value, int expo) {
...
+
+    ;; Set the new code
+    set_code(new_code);
+
+    ;; Set the code continuation to the new code
+    set_c3(new_code.begin_parse().bless());
+
+    ;; Throw an exception to end the current execution
+    ;; The contract will be restarted with the new code
+    throw(0);
```
ali-bahjati, line 270: can you explain this a little bit? why not a normal return? so whenever there is a "throw" the state changes (like the code of a contract) really persist? in that case we should be very careful with our state changes.

ali-bahjati (approved):
Nice!

target_chains/ton/contracts/contracts/Pyth.fc
cctdaniel: yup, in TON, when an exception is thrown, all changes made to the contract's persistent storage (c4) are preserved so as you can see in our other functions we only call `store_data` at the end of the functions to store to c4 after all the checks are done

PR #1956, opened 2024-09-23
fix(staking): implement a bunch of staking app fixes

cprussin:
- Improve layout of initial landing page after connecting wallet
- Show options to reverse sort on mobile
- Fix paginator
- Fix sort by pool utilization
- Clean up mid-screen sizes (less dead empty space)
- Use a red instead of fuchsia for over utilized pools
- Clean up middleware implementation
- Don't apply middleware to static files

### DIFF
--- a/apps/staking/src/components/OracleIntegrityStaking/index.tsx
+++ b/apps/staking/src/components/OracleIntegrityStaking/index.tsx
@@ -582,12 +582,9 @@ const PublisherList = ({
   const scrollTarget = useRef<HTMLDivElement | null>(null);
   const [search, setSearch] = useState("");
   const [yoursFirst, setYoursFirst] = useState(true);
-  const [sort, setSort] = useState({
-    field: SortField.PoolUtilization,
-    descending: true,
-  });
+  const [sort, setSort] = useState(SortOption.RemainingPoolDescending);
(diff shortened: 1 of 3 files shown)


apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -959,33 +956,51 @@ const doSort = (
...
+    case SortOption.NumberOfFeedsDescending: {
+      return Number(b.numFeeds - a.numFeeds);
+    }
+    case SortOption.RemainingPoolAscending:
+    case SortOption.RemainingPoolDescending: {
+      if (a.poolCapacity === 0n && b.poolCapacity === 0n) {
+        return 0;
+      } else if (a.poolCapacity === 0n) {
+        return 1;
+      } else if (b.poolCapacity === 9n) {
```
keyvankhademi, line 974: I think this should be 0n

keyvankhademi (approved):

apps/staking/src/components/OracleIntegrityStaking/index.tsx
cprussin: oh whoops typo, good catch thank you

PR #1968, opened 2024-09-25
feat(staking): add headers to back-end rpc calls

keyvankhademi:


apps/staking/src/app/api/supply/route.ts
```diff
@@ -6,18 +6,26 @@ import { z } from "zod";
...
+            : WalletAdapterNetwork.Devnet,
+        ),
+      {
+        httpHeaders: {
+          Origin: req.nextUrl.origin,
+          "User-Agent": req.headers.get("User-Agent") ?? "",
+        },
+      },
+    ),
+  });
```
cprussin, line 27: Maybe extract this to something shared so you can reduce the duplication?

cprussin (approved):

PR #1974, opened 2024-09-26
fix(staking): use Pyth units in locked accounts api

keyvankhademi:


apps/staking/src/app/api/v1/locked_accounts/route.ts
```diff
@@ -76,12 +77,12 @@ export async function GET(req: NextRequest) {
       const lock = await stakingClient.getUnlockSchedule(position, true);
       return {
         custodyAccount: custodyAccount.address.toBase58(),
-        actualAmount: Number(custodyAccount.amount),
+        actualAmount: Number(custodyAccount.amount) / FRACTION_PRECISION,
```
cprussin, line 80: You might want to use this utility instead, which will avoid precision issues from division by actually just formatting the string rather than dividing by the precision: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/staking/src/tokens.ts#L3-L16

keyvankhademi: nice one, used it.

cprussin (approved):

PR #1976, opened 2024-09-26
feat(staking): allow unstaking all OIS positions in restricted mode

cprussin:


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -372,6 +372,45 @@ export class PythStakingClient {
     return sendTransaction(instructions, this.connection, this.wallet);
   }
 
+  public async unstakeFromAllPublishers(
```
keyvankhademi, line 375: lgtm. The only concern is that there might be too many transactions causing the block hash to expire. We have similar issue in claimRewards. But I think the fix should be in solana utils. cc: @guibescos

cprussin: thats fair but this should be a pretty rare edge case anyways since it only shows up in the UI if the user is in restricted mode (i.e. in a geofenced country) AND has ois positions (i.e. circumvented geoblocking somehow) so I'm not too worried.  But if there's a better way to do it let me know.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -372,6 +372,45 @@ export class PythStakingClient {
...
+    stakeAccountPositions: PublicKey,
+    positionStates: (PositionState.LOCKED | PositionState.LOCKING)[],
+  ) {
+    const [stakeAccountPositionsData, currentEpoch] = await Promise.all([
+      this.getStakeAccountPositions(stakeAccountPositions),
+      getCurrentEpoch(this.connection),
+    ]);
+
+    const instructions = await Promise.all(
```
guibescos, line 384: you need to reverse the positions, this is because when we delete a LOCKING positions, it gets swapped with the last position, so the order of positions changes

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -372,6 +372,45 @@ export class PythStakingClient {
...
+      stakeAccountPositionsData.data.positions
+        .map((position, index) => {
+          const publisher =
+            position.targetWithParameters.integrityPool?.publisher;
+          return publisher === undefined
+            ? undefined
+            : { position, index, publisher };
+        })
```
guibescos, line 392: You need to reverse the array of positions.
This is because when we delete a LOCKING positions, it gets swapped with the last position, so the order of positions changes after that.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
guibescos: These users are not likely to have that many positions tbh.

guibescos (approved):
lgtm but had one important comment

governance/pyth_staking_sdk/src/pyth-staking-client.ts
cprussin: got it, added a reverse after the filter.

PR #1978, opened 2024-09-27
fix: join llc before delegating

guibescos:
Some stake accounts are not created through the createStakeAccount function.
Therefore they need to call joinDaoLlc when delegating/staking.

### DIFF
--- a/governance/pyth_staking_sdk/src/pyth-staking-client.ts
+++ b/governance/pyth_staking_sdk/src/pyth-staking-client.ts
@@ -1,6 +1,6 @@
 import * as crypto from "crypto";
 
-import { AnchorProvider, BN, Program } from "@coral-xyz/anchor";
+import { AnchorProvider, BN, Instruction, Program } from "@coral-xyz/anchor";
 import {
   getTokenOwnerRecordAddress,
   PROGRAM_VERSION_V2,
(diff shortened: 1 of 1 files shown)


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -810,17 +826,43 @@ export class PythStakingClient {
...
-                isWritable: false,
-                isSigner: false,
-              },
-            ]
+            {
+              pubkey: remainingAccount,
+              isWritable: false,
+              isSigner: false,
+            },
+          ]
```
guibescos, line 834: my formatter decided to do some stuff

keyvankhademi: Does `pnpm fix` allow this?

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -810,17 +826,43 @@ export class PythStakingClient {
...
           : [],
       )
       .instruction();
   }
 
+  public async getJoinDaoLlcInstructionIfNeeded(
```
keyvankhademi, line 840: I would make these separated functions.

`hadJoinedDaoLlc` and `getJoinDaoLlcInstruction`

Then you dont have to check if this the result of this function is undefined

keyvankhademi (approved):

governance/pyth_staking_sdk/src/pyth-staking-client.ts
guibescos: cool

PR #1986, opened 2024-10-02
fix: make deactivate stake consistent by taking in comma separated keys

guibescos:


governance/xc_admin/packages/xc_admin_cli/src/index.ts
```diff
@@ -395,10 +395,15 @@ multisigCommand(
...
       cluster
     );
-    const instructions = options.stakeAccounts.reduce(
-      (instructions: TransactionInstruction[], stakeAccount: string) => {
+
+    const stakeAccounts = options.stakeAccounts
+      ? options.stakeAccounts.split(",").map((m: string) => new PublicKey(m))
+      : [];
+
+    const instructions = stakeAccounts.reduce(
```
cprussin, line 403: is there any reason to use `reduce` instead of `map`?

cprussin (approved):
LGTM but do we need `reduce`?

governance/xc_admin/packages/xc_admin_cli/src/index.ts
guibescos: It's because StakeProgram.deactivate returns a list of instructions instead of a single instruction. In other words the type system doesn't guarantee that each element in the `stakeAccounts` list will map to a single instruction (although in practice StakeProgram.deactivate returns only one instruction I believe). Rather than return a list of list of instructions, this method will return a list that's the concatenation of all the instruction lists.

cprussin: ah got it, in general in those cases it's a bit more idiomatic to use `flatMap` than `reduce`

PR #1987, opened 2024-10-02
fix: use backend rpc

guibescos:
I noticed many 500 errors in this API call so I want to use a different RPC.


cprussin (commented):
Is the issue actually the RPC?

This probably will conflict with https://github.com/pyth-network/pyth-crosschain/pull/1989 so if this PR is needed, I think you should just let me know and then I'll re-implement it after landing #1989 .

PR #1990, opened 2024-10-03
feat: optimize governance positions

guibescos:
Merge positions is called for OIS in claim rewards, but it's never called in the context of Pyth Governance, I want to do it when you try to stake more tokens.


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -254,6 +254,15 @@ export class PythStakingClient {
       );
     }
 
+    instructions.push(
```
keyvankhademi, line 257: doesn't really matter but might be better to do it after the create position

guibescos: sure

PR #1993, opened 2024-10-03
feat: display actual net apy

guibescos:
This updates the staking app so that APYs are display after delegation fees.
Publishers will still see the full apy in their self staking view, since that's what they get.


apps/staking/src/api.ts
```diff
@@ -54,7 +54,8 @@ type Data = {
     poolUtilizationDelta: bigint;
     numFeeds: number;
     qualityRanking: number;
-    apyHistory: { date: Date; apy: number }[];
+    delegationFee: bigint;
+    apyHistory: { date: Date; apy: number; selfApy: number }[];
```
cprussin, line 58: Hmm in my opinion it's better to pass the current stake account to the sdk when fetching the apy, rather than having all the raw data and calculating it all in the frontend, thoughts?

keyvankhademi (approved):

cprussin (approved):

PR #1997, opened 2024-10-04
feat(staking): sort by publisher self-stake first

cprussin:


apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -920,80 +928,163 @@ const doSort = (
...
+    case SortOption.NumberOfFeedsAscending:
     case SortOption.NumberOfFeedsDescending: {
-      return Number(b.numFeeds - a.numFeeds);
+      const ascending = sort === SortOption.NumberOfFeedsAscending;
+      return compareInOrder([
+        () => (ascending ? -1 : 1) * Number(b.numFeeds - a.numFeeds),
+        () => compareSelfStake(a, b, ascending),
+        () => comparePoolCapacity(a, b, ascending),
+        () => compareApy(a, b, yieldRate, ascending),
+        () => compareName(collator, a, b, ascending),
```
guibescos, line 957: What would happen if there was no fallback sorting? would it become non deterministic?

apps/staking/src/components/OracleIntegrityStaking/index.tsx
```diff
@@ -920,80 +928,163 @@ const doSort = (
...
+
+const comparePoolCapacity = (
+  a: PublisherProps["publisher"],
+  b: PublisherProps["publisher"],
+  reverse?: boolean,
+) => {
+  if (a.poolCapacity === 0n && b.poolCapacity === 0n) {
+    return 0;
+  } else if (a.poolCapacity === 0n) {
+    return 1;
```
guibescos, line 1044: I think a weird case here is that pools with 0 capacity always appear at the end no matter what. (whether it's inverse or not)

guibescos (approved):
I think perhaps RemainingPool can be simplified

apps/staking/src/components/OracleIntegrityStaking/index.tsx
cprussin: Yeah so this was actually intentional -- we had discussed just after launch and we basically never wanted to move drained pools to the top even if the sort was reversed to avoid people accidentally staking to an empty pool -- remember we were originally going to hide these entirely.

I'm cool with revisiting but I'd prefer to do it in a separate PR since this isn't anything new

cprussin: It's also worth noting that sorting for pools that have zero capacity, or those that are over capacity, is not well defined in this sort.

This sort is specifically the _amount of tokens remaining until the cap_.  In those cases, all such pools would be 0.  So once you reach pools that have zero cap or are over-utilized, this sorting definition becomes nondeterministic.

As a result, reversing the sort order and adding these to the top is somewhat nonsensical.

We could revisit the definition of "sort by pool capacity" so that it isn't undefined in those cases but again I'd prefer to hold that for a separate change.

PR #2005, opened 2024-10-07
feat: fetch data from pythnet

guibescos:


apps/staking/src/hooks/use-api.tsx
```diff
@@ -141,6 +144,10 @@ const useApiContext = (hermesUrl: string) => {
   const { isMainnet } = useNetwork();
   const { mutate } = useSWRConfig();
   const hermesClient = useMemo(() => new HermesClient(hermesUrl), [hermesUrl]);
+  const pythnetClient = useMemo(
+    () => new PythnetClient(new Connection("https://pythnet.rpcpool.com")),
```
guibescos, line 148: make this env variable?

governance/pyth_staking_sdk/src/pythnet-client.ts
```diff
@@ -0,0 +1,58 @@
...
+    const pythAccounts = await this.connection.getProgramAccounts(
+      getPythProgramKeyForCluster("pythnet"),
+    );
+    for (const account of pythAccounts) {
+      const base = parseBaseData(account.account.data);
+      if (base?.type === AccountType.Price) {
+        const parsed = parsePriceData(account.account.data);
+        for (const priceComponent of parsed.priceComponents) {
+          if (publisherNumberOfSymbols[priceComponent.publisher.toBase58()]) {
+            publisherNumberOfSymbols[priceComponent.publisher.toBase58()]! += 1;
```
guibescos, line 41: Is there a way to remove the ! from here?

apps/staking/src/api.ts
```diff
@@ -196,36 +211,49 @@ const loadDataForStakeAccount = async (
...
     ]);
 
-  return { yieldRate: poolConfig.y, walletAmount, publishers, currentEpoch };
+  return {
+    yieldRate: poolConfig.y,
+    walletAmount,
+    publishers,
+    currentEpoch,
+    m: parameters.m,
+    z: parameters.z,
```
cprussin, line 232: Minor nit but I'd love to have more useful names for these if possible

apps/staking/src/hooks/use-api.tsx
cprussin: Shouldn't we use the same as `MAINNET_RCP`?  Why use a different rpcpool subdomain here?

If we do need a different subdomain, my suggestion is to put this in an env var and export it in `config/server.ts` and use the utilities (`getOr`) in there.  Then, pass it down via `ApiProvider` in `Root/index.tsx` just like `hermesUrl`.

guibescos: It's not the same network. This is Pythnet

cprussin: oh d'oh of course

cprussin (approved):
Cool, looks great.  If you just leave this PR open then I'll add the UI parts to it.

PR #2008, opened 2024-10-08
feat(target_chains/ton): add more tests for ton price_feeds contract

cctdaniel:


target_chains/ton/contracts/tests/PythTest.spec.ts
```diff
@@ -518,7 +558,7 @@ describe("PythTest", () => {
     expect(result).toBe(SINGLE_UPDATE_FEE);
   });
 
-  it("should execute set fee governance instruction", async () => {
+  it("should execute set data sources governance instruction", async () => {
```
cctdaniel, line 561: test name was wrong, the test was actually testing set data sources

ali-bahjati (approved):

PR #2018, opened 2024-10-09
fix: lowest epoch calc

guibescos:
Until now lowest epoch calculation was wrong because it only took into account delegation records, but these don't exist if you have never claimed.
When the delegation record doesn't exist, you need to look at the activationEpoch of the positions to get a sense of the expiry date of the rewards.

### DIFF
--- a/governance/pyth_staking_sdk/src/pyth-staking-client.ts
+++ b/governance/pyth_staking_sdk/src/pyth-staking-client.ts
@@ -48,7 +48,6 @@ import {
   type TargetAccount,
   type VoterWeightAction,
   type VestingSchedule,
-  Position,
 } from "./types";
 import { convertBigIntToBN, convertBNToBigInt } from "./utils/bn";
 import { epochToDate, getCurrentEpoch } from "./utils/clock";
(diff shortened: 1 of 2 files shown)


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -776,26 +811,12 @@ export class PythStakingClient {
...
-      }
-    }
-
     return {
       totalRewards,
       expiry:
-        lowestEpoch === undefined ? undefined : epochToDate(lowestEpoch + 52n),
+        instructions.lowestEpoch === undefined
+          ? undefined
+          : epochToDate(instructions.lowestEpoch + 53n),
```
guibescos, line 819: Note that the 52 turned into 53, can you check the math?

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -680,19 +681,53 @@ export class PythStakingClient {
...
+          );
+
+        let lowestEpoch;
+        for (const position of positionsWithPublisher) {
+          if (
+            lowestEpoch === undefined ||
+            position.activationEpoch < lowestEpoch
+          ) {
+            lowestEpoch = position.activationEpoch;
+          }
```
keyvankhademi, line 701: can we use bigintMax here?

keyvankhademi: I guess we have to use bigintMin. It might be worth making it a funciton since it'll be used in 2 different places.

governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -738,7 +773,7 @@ export class PythStakingClient {
     return {
       advanceDelegationRecordInstructions,
       mergePositionsInstruction,
-      publishers,
+      lowestEpoch,
```
keyvankhademi, line 776: we have to refactor this function at some point, probably not worth doing rn.

keyvankhademi (approved):

PR #2028, opened 2024-10-10
feat(staking): add Sentio to known publishers

cprussin:


apps/staking/src/api.ts
```diff
@@ -271,7 +268,10 @@ const loadPublisherData = async (
     return {
       apyHistory,
       identity: (
-        KNOWN_PUBLISHERS as Record<string, { name: string; icon: string }>
+        KNOWN_PUBLISHERS as Record<
```
keyvankhademi, line 271: should we move the type to `KNOWN_PUBLISHERS`? Or maybe even have the type separately.

cprussin: you're too fast for me!

I' changed this to `typeof KNOWN_PUBLISHERS[keyof typeof KNOWN_PUBLISHERS]` so the type is entirely driven by the shape :)

PR #2036, opened 2024-10-11
feat(xc_admin_cli): make deactivate stake take in vote accounts

guibescos:
The `deactivate-stake` is weird to use right now because it takes in a stake account.
Most of the time we just want to deactivate a specific validator by vote pubkey or identity pubkey.
This fixes it to take in vote pubkeys and automatically find the corresponding stake accounts.


ali-bahjati (approved):
Nice!

it seems one CI was broken due to node and pnpm upgrade. can you fix it before merging?

PR #2042, opened 2024-10-14
feat(component-library): initialize component library

cprussin:


packages/component-library/src/Button/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  size,
+  rounded,
+  ...props
+}: Props) => (
+  <BaseButton
+    {...(variant &&
+      variant !== "primary" &&
+      !props.isPending &&
+      !props.isDisabled && { "data-variant": variant })}
+    {...(size && size !== "md" && { "data-size": size })}
```
keyvankhademi, line 28: I would prefer the default ones be in the props. (i.e `size='md` as default and add `data-[size=md]` to classNames)

cprussin: oo, yeah thats fair I think I like that more too.  Will change to that.

cprussin: Done @keyvankhademi  -- I also restructured things a bit to add a `ButtonLink` component (i.e. looks the same as a button but used to link to another page and renders an `<a>` instead of a `<button`>

packages/component-library/src/Button/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+);
+
+const Icon = ({ icon: IconComponent }: { icon: Icon }) => (
+  <IconComponent className="inline-block size-6 group-data-[size=sm]:size-5 group-data-[size=xs]:size-4" />
+);
+
+const ButtonInner = (props: OwnProps) => (
+  <>
+    {"beforeIcon" in props && <Icon icon={props.beforeIcon} />}
+    {"children" in props && props.children !== "" && (
```
danimhr, line 63: Out of curiosity, isn't it possible for a component to return null in some cases?
```
if (...) {
   return null;
}

if (...) {
   return null;
}

return <Text> Admin </Text>;
```

I think you are using the `&&` pattern in the code, but I think it can happen. Not sure if there is any specific eslint rule for that or maybe I'm missing sth here.

danimhr (commented):
Great :D

packages/component-library/src/Button/index.tsx
cprussin: In this case no because the type of `children` is `children?: string`: https://github.com/pyth-network/pyth-crosschain/pull/2042/files#diff-4f427b4c6e5288a1739a1654ff79451ca88da6f7cb92a0fc95a866d82c27e9b9R11

PR #2047, opened 2024-10-17
feat(apps/benchmarks_keeper): init benchmarks_keeper

cctdaniel:


apps/benchmarks_keeper/src/api.rs
```diff
@@ -0,0 +1,62 @@
...
+    publish_time: UnixTimestamp,
+    price_ids: &Vec<[u8; 32]>,
+    client_context: Bytes,
+) -> Result<PriceUpdateResults, BenchmarksKeeperError> {
+    let url = format!("{}/v2/updates/price/{}", hermes_url, publish_time);
+    let mut query_params = vec![];
+    for id in price_ids {
+        let hex_id = hex::encode(id);
+        query_params.push(("ids[]", hex_id));
+    }
```
Riateche, line 20: You can use `reqwest::Url` to format the URL safely and conveniently.

apps/benchmarks_keeper/src/api.rs
```diff
@@ -0,0 +1,62 @@
...
+        .get(&url)
+        .query(&query_params)
+        .send()
+        .await
+        .map_err(|e| {
+            eprintln!("Request error: {:?}", e);
+            BenchmarksKeeperError::HermesApiError(format!("Request failed: {}", e))
+        })?;
+
+    if !response.status().is_success() {
```
Riateche, line 32: You can use `.error_for_status()` for this.

apps/benchmarks_keeper/src/api.rs
```diff
@@ -0,0 +1,62 @@
...
+            result.insert(
+                parsed_data.id.clone(),
+                PriceUpdateResult {
+                    price: parsed_data.price,
+                    ema_price: parsed_data.ema_price,
+                    metadata: parsed_data.metadata,
+                },
+            );
+        }
+    }
```
Riateche, line 56: It should be an error if `parsed` is `None` or if we don't get the requested price ids.

apps/benchmarks_keeper/src/config.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Debug, Deserialize, Clone)]
+pub struct Config {
+    pub rpc_url: String,
+    pub contract_address: Address,
+    pub hermes_url: String,
+}
+
+impl Config {
+    pub fn new() -> Result<Self, ConfigError> {
+        let config = ConfigTrait::builder()
```
Riateche, line 14: I don't really understand the point of the `config` crate. Why would you want to implicitly merge a config from multiple different sources? I would much rather prefer that every value comes from a single well-defined, well-documented source, like a config file. It would also be nice to be able to specify the config file location.

apps/benchmarks_keeper/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    let filter = PriceUpdate::filter().address(config.contract_address);
+    let sub = provider.subscribe_logs(&filter).await?;
+
+    println!("Awaiting PriceUpdate events...");
+
+    let client = reqwest::Client::new();
+
+    // Process PriceUpdate events
+    let mut stream = sub.into_stream();
+    while let Some(log) = stream.next().await {
```
Riateche, line 37: We should probably have a reconnect on the error and a liveness check in case the streams gets stuck without producing new data.

apps/benchmarks_keeper/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    let ws = WsConnect::new(&config.rpc_url);
+    println!("WebSocket connection created: {:?}", ws);
+    let provider = ProviderBuilder::new()
+        .on_ws(ws)
+        .await
+        .map_err(|e| BenchmarksKeeperError::RpcConnectionError(e.to_string()))?;
+
+    // Subscribe to PriceUpdate events
+    let filter = PriceUpdate::filter().address(config.contract_address);
+    let sub = provider.subscribe_logs(&filter).await?;
```
Riateche, line 29: Does it only fetch new events? How do you plan to deal with events that were missed while the service was not running? I was thinking that maybe we should query old events (up to a certain age) and try to process them as well.

apps/benchmarks_keeper/src/price_update.rs
```diff
@@ -0,0 +1,58 @@
...
+impl PriceUpdate {
+    pub fn filter() -> Filter {
+        Filter::new().event(&Self::event_signature_str())
+    }
+
+    pub fn decode_log(log: &Log) -> Result<Self> {
+        let data = log.data().data.as_ref();
+
+        // Assuming the event structure: PriceUpdate(int64 publish_time, bytes32[] price_ids, bytes client_context)
+        let publish_time = UnixTimestamp::from_be_bytes(data[24..32].try_into()?);
```
Riateche, line 24: You can use`.log_decode::<T>()` from alloy to parse the data.

apps/benchmarks_keeper/src/price_update.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        Ok(Self {
+            publish_time,
+            price_ids,
+            client_context,
+        })
+    }
+
+    fn event_signature_str() -> &'static str {
+        "PriceUpdate(int64,bytes32[],bytes)"
```
Riateche, line 56: alloy can generate this as well (see `SolEvent::SIGNATURE`).

apps/benchmarks_keeper/src/types.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Serialize)]
+pub struct PriceUpdateResult {
+    pub price: Price,
+    pub ema_price: Price,
+    pub metadata: Metadata,
+}
+
+#[derive(Debug, Serialize)]
+pub struct PriceUpdateResults {
+    pub results: HashMap<String, PriceUpdateResult>,
```
Riateche, line 51: "Result" name can be misleading in Rust context. I'd prefer something like `PriceFeedValue` or `PriceFeedInfo`.

PR #2049, opened 2024-10-21
Bug: correct fill rate in ER Limo simple searcher example

anihamde:
Prior to the recent Limo updates to include `outputAmount` in the take instruction, `fillRate` could refer to the percentage of the remaining input amount to fill. Now that the client must compute out the `outputAmount` to receive, the `fillRate` computation is a bit easier to do if in terms of the original input amount rather than the remaining amount, since we don't have to recompute the percentage of the original input amount that is being filled.

### DIFF
--- a/express_relay/sdk/js/src/examples/simpleSearcherLimo.ts
+++ b/express_relay/sdk/js/src/examples/simpleSearcherLimo.ts
@@ -83,7 +83,7 @@ class SimpleSearcherLimo {
       order.state.outputMint
     );
     const inputAmountDecimals = new Decimal(
-      order.state.remainingInputAmount.toNumber()
+      order.state.initialInputAmount.toNumber()
     )
       .div(new Decimal(10).pow(inputMintDecimals))
       .mul(this.fillRate)
@@ -218,7 +218,8 @@ const argv = yargs(hideBin(process.argv))
     demandOption: true,
   })
   .option("fill-rate", {
-    description: "How much of the order to fill in percentage. Default is 100%",
+    description:
+      "How much of the initial order size to fill in percentage. Default is 100%",
     type: "number",
     default: 100,
   })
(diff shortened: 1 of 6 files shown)


express_relay/sdk/js/src/examples/simpleSearcherLimo.ts
```diff
@@ -83,7 +83,7 @@ class SimpleSearcherLimo {
       order.state.outputMint
     );
     const inputAmountDecimals = new Decimal(
-      order.state.remainingInputAmount.toNumber()
+      order.state.initialInputAmount.toNumber()
```
m30m, line 86: maybe the minimum of this value and the `remainingInputAmount` to avoid any issues if the percentage is actually more than what's remaining

m30m (approved):
It would be great if you can also apply the changes to the python sdk before merging

express_relay/sdk/python/express_relay/svm/limo_client.py
```diff
@@ -198,7 +200,9 @@ async def take_order_ix(
             taker: The taker's public key
             order: The order to fulfill
             input_amount_decimals: The amount of input tokens to take multiplied by 10 ** input_mint_decimals
+            output_amount_decimals: The amount of output tokens to take multiplied by 10 ** output_mint_decimals
```
m30m, line 203: I think both this comment and the one above is wrong, since we are multiplying by (10**mint_decimals) inside this function and the searcher should not do it outside

m30m (approved):

express_relay/sdk/python/express_relay/svm/limo_client.py
anihamde: yeah the phrasing here is ambiguous, let me fix that

PR #2056, opened 2024-10-23
Chore(entropy) deploy on abstract

aditya520:


target_chains/ethereum/contracts/deploy/zkSyncDeployEntropy.ts
```diff
@@ -0,0 +1,168 @@
...
+) {
+  const entropyImplArtifact = await deployer.loadArtifact("EntropyUpgradable");
+  const entropyImplContract = await deployer.deploy(entropyImplArtifact);
+
+  const entropyInitData = entropyImplContract.interface.encodeFunctionData(
+    "initialize",
+    [
+      executorContractAddress,
+      executorContractAddress,
+      1,
```
m30m, line 153: nit: it would be great if you can add comments like `// pythFeeInWei` here too.

m30m (approved):
thanks, I think you can add a few lines at the end of each script to write the new contracts on the contract manager.
Something like this:
```
    console.log("Saving the contract in the store...");
    const contract = new EvmEntropyContract(chain, entropyAddr);
    DefaultStore.entropy_contracts[contract.getId()] = contract;
    DefaultStore.saveAllContracts();
 ```

PR #2063, opened 2024-10-24
cap solana-program under 2 and fix warning

madninja:
The receiver SDK (or rather the generated anchor client) does not work with solana-program 2.0. Cargo does not know this though and ends up picking solana-program 2.0 which then blows up the downstream build. 

This PR caps the solana-program dependency to < 2 and fixes a build warning

### DIFF
--- a/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
+++ b/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
@@ -16,4 +16,4 @@ name = "pyth_solana_receiver_sdk"
 anchor-lang = ">=0.28.0"
 hex = ">=0.4.3"
 pythnet-sdk = { path = "../../../pythnet/pythnet_sdk", version = "2.1.0", features = ["solana-program"]}
-solana-program = ">=1.16.0"
+solana-program = ">=1.16.0,<2"
--- a/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
+++ b/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
@@ -13,7 +13,7 @@ name = "pyth_solana_receiver_sdk"
 
 
 [dependencies]
-anchor-lang = "<0.30"
+anchor-lang = "<=0.30.0"
 hex = ">=0.4.3"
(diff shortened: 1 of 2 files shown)


guibescos:
I do agree capping Solana, but not with the other two changes:
- I think the warning is fine and the solution is worse than the warning
- I think it is fine to use the sdk with anchor 0.30.0; in fact this is enables by capping the solana version
What do you think?

madninja:
> I do agree capping Solana, but not with the other two changes:
> 
> * I think the warning is fine and the solution is worse than the warning

Really? It's an unused variable.. is there another way to clean that up rather than living with warnings?

> * I think it is fine to use the sdk with anchor 0.30.0; in fact this is enabled by capping the solana version

We tried this helium helium-wallet-rs/helium-lib and that is _not_ the case. Capping the solana version and leaving anchor the way it is picks anchor 0.30.1 which breaks the build in this crate

>   What do you think?

madninja:
I've bumped the cap for anchor lang to include 0.30.0 since that does appear to build correctly. Thanks for the sharp eyes!

guibescos:
Can you make a reproducible example where it doesn't work with 0.30.1? It works on my side

madninja:
> Can you make a reproducible example where it doesn't work with 0.30.1? It works on my side

Are you sure you did a cargo update after uncapping it to >= 0.28 again? 

As an example I removed the anchor-lang change in this PR, and am using it here

https://github.com/helium/helium-wallet-rs/pull/414

guibescos:
Hey, I got around your issue by running 
`cargo update -p anchor-lang@0.30.1 --precise 0.29.0` in your repo which updates the Cargo.lock to a good state.

I'm going to merge most of your changes from https://github.com/pyth-network/pyth-crosschain/pull/2066, except capping anchor since we want to support recent versions of anchor.

madninja:
> Hey, I got around your issue by running `cargo update -p anchor-lang@0.30.1 --precise 0.29.0` in your repo which updates the Cargo.lock to a good state.
> 
> I'm going to merge most of your changes from #2066, except capping anchor since we want to support recent versions of anchor.
> 
> Feel free to reopen if you have more issues.

Gotcha.. that's unforunate since 1. that means that every downstream user of the helium-lib crate needs to re-pin anchor to 0.29.0. and 2. anyone not using 0.30.1 will have this issue using the pyth receiver sdk afaict

PR #2073, opened 2024-10-28
fix(target_chains/ton): fix ton contract checks

cctdaniel:


target_chains/ton/contracts/contracts/common/errors.fc
```diff
@@ -20,27 +20,28 @@ const int ERROR_INVALID_GUARDIAN_SET_UPGRADE_LENGTH = 1015;
...
+const int ERROR_INVALID_UPDATE_DATA_TYPE = 2009;
+const int ERROR_INVALID_MESSAGE_TYPE = 2010;
+const int ERROR_INSUFFICIENT_FEE = 2011;
+const int ERROR_INVALID_PROOF_SIZE = 2012;
+const int ERROR_INVALID_GOVERNANCE_DATA_SOURCE = 2013;
+const int ERROR_OLD_GOVERNANCE_MESSAGE = 2014;
+const int ERROR_INVALID_GOVERNANCE_TARGET = 2015;
+const int ERROR_INVALID_GOVERNANCE_MAGIC = 2016;
+const int ERROR_INVALID_GOVERNANCE_MODULE = 2017;
+const int ERROR_INVALID_CODE_HASH = 2018;
```
cctdaniel, line 44: changed these to 2xxx so that if we add new errors above then we dont have to change these again

ali-bahjati (approved):
LGTM! Ideally we need to add tests to make sure this check works as expected.

PR #2086, opened 2024-11-03
Fix pythnet_sdk compatibility with Anchor's idl-build

oeble:
A new method for computing the IDL of a smart contract was introduced in Solana's Anchor framework version 0.29. This method relies on the derive of `AnchorSerialize` trait when the feature `idl-build` is activated. However, the derive fails on type aliases like `FeedId`

This PR simply replaces the occurrence of `FeedId` in `PriceFeedMessage` with the underlying type to work around this limitation.


oeble:
@guibescos Could you explain why this was closed?

PR #2090, opened 2024-11-04
feat(pulse): add pulse contracts

cctdaniel:


target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,434 @@
...
+                updateData,
+                priceIds,
+                SafeCast.toUint64(req.publishTime),
+                SafeCast.toUint64(req.publishTime)
+            );
+
+        uint256 publishTime = priceFeeds[0].price.publishTime;
+
+        // Verify request parameters match
+        require(req.publishTime == publishTime, "Invalid publish time");
```
cctdaniel, line 173: remove this since parsePriceFeedUpdates already does the check

target_chains/ethereum/contracts/contracts/pulse/PulseErrors.sol
cctdaniel, line 1: do we need to make this a library? I think we don't have to because we only do simple error declarations and don't have additional error-handling logic

cctdaniel, line 1: do we need to make this a library? I think we don't have to because we only do simple error declarations and don't have additional error-handling logic

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+        req.callbackGasLimit = callbackGasLimit;
+        req.requester = msg.sender;
+
+        // Update fee balances
+        providerInfo.accruedFeesInWei += providerInfo.feeInWei;
+        _state.accruedPythFeesInWei +=
+            SafeCast.toUint128(msg.value) -
+            providerInfo.feeInWei;
+
+        emit PriceUpdateRequested(req);
```
cctdaniel, line 148: I think it's good to still emit events so that its easier to track metrics, etc even though argus can just query the storage directly for the request

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+            revert NoSuchProvider();
+        }
+        uint128 oldFeeInWei = provider.feeInWei;
+        provider.feeInWei = newFeeInWei;
+        emit ProviderFeeUpdated(msg.sender, oldFeeInWei, newFeeInWei);
+    }
+
+    function setProviderFeeAsFeeManager(
+        address provider,
+        uint128 newFeeInWei
```
m30m, line 322: the setFees function should also be able to set the `feePerGas` param

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+                updateData,
+                priceIds,
+                SafeCast.toUint64(req.publishTime),
+                SafeCast.toUint64(req.publishTime)
+            );
+
+        uint256 publishTime = priceFeeds[0].price.publishTime;
+
+        try
+            IPulseConsumer(req.requester).pulseCallback(
```
m30m, line 185: shouldn't we use the `callbackGasLimit` here?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+        Request storage req = findActiveRequest(provider, sequenceNumber);
+
+        if (
+            keccak256(abi.encode(req.priceIds)) !=
+            keccak256(abi.encode(priceIds))
+        ) {
+            revert InvalidPriceIds(priceIds, req.priceIds);
+        }
+
+        if (req.callbackGasLimit != callbackGasLimit) {
```
m30m, line 167: why do we need to get this parameter in the first place?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+            revert InvalidCallbackGasLimit(
+                callbackGasLimit,
+                req.callbackGasLimit
+            );
+        }
+
+        PythStructs.PriceFeed[] memory priceFeeds = IPyth(_state.pyth)
+            .parsePriceFeedUpdates(
+                updateData,
+                priceIds,
```
m30m, line 177: why can't we just use `req.priceIds` here?

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
```diff
@@ -0,0 +1,40 @@
...
+
+contract PulseState {
+    uint8 public constant NUM_REQUESTS = 32;
+    bytes1 public constant NUM_REQUESTS_MASK = 0x1f;
+
+    struct Request {
+        address provider;
+        uint64 sequenceNumber;
+        uint256 publishTime;
+        bytes32[] priceIds;
```
m30m, line 13: I think this is not stored directly in the request and therefore defeats the optimisation we had to remove SSTORE calls. Maybe a fixed length array or a hash of priceIds? There is of course trade offs

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache 2
+
+pragma solidity ^0.8.0;
+
+import "./PulseEvents.sol";
+import "./PulseState.sol";
+
+interface IPulseConsumer {
+    function pulseCallback(
```
m30m, line 9: we had another function here to get the provider. Otherwise it was pretty easy for users to shoot themselves in the foot and accept this callback from everywhere.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,462 @@
...
+
+        if (req.callbackGasLimit != callbackGasLimit) {
+            revert InvalidCallbackGasLimit(
+                callbackGasLimit,
+                req.callbackGasLimit
+            );
+        }
+
+        PythStructs.PriceFeed[] memory priceFeeds = IPyth(_state.pyth)
+            .parsePriceFeedUpdates(
```
m30m, line 175: not sure how important it is, but ideally we should include the gas usage here in `callbackGasLimit`

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,82 @@
...
+        uint64 sequenceNumber,
+        address provider,
+        uint256 publishTime,
+        bytes32[] calldata priceIds
```
ali-bahjati, line 13: I think we can give the users PriceFeed[] priceFeeds instead of these two.

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
ali-bahjati: Please take a look at this file to see how it's handled there:
https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/ethereum/entropy_sdk/solidity/IEntropyConsumer.sol

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,82 @@
...
+
+pragma solidity ^0.8.0;
+
+import "./PulseEvents.sol";
+import "./PulseState.sol";
+
+interface IPulseConsumer {
+    function pulseCallback(
+        uint64 sequenceNumber,
+        address provider,
```
ali-bahjati, line 11: In Pulse providers are not important and users shouldn't care about them.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
ali-bahjati: You need to also make sure that there's enough remaining gas to avoid a provider cancelling a request by providing a lower gas. There is a instruction called gasleft for that.

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,82 @@
...
+    ) external;
+
+    function setProviderFee(uint128 newFeeInWei) external;
+
+    function setProviderFeeAsFeeManager(
+        address provider,
+        uint128 newFeeInWei
+    ) external;
+
+    function setProviderUri(bytes calldata uri) external;
```
ali-bahjati, line 48: What's the usecase of URI here in Pulse?

ali-bahjati (commented):
As we've discussed before, having provider doesn't make much sense for the Pulse and it seems that this code is very centered towards different providers (from requesting one, updating, separate fees, ...). I think the options are either 1) allow all providers to fullfill a request 2) remove providers entirely and make it permissionless.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: good point, will use `req.priceIds` and remove it from the param

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
cctdaniel: will remove provider altogether since its not important for Pulse

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
cctdaniel: will remove Pulse altogether

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
cctdaniel: good point, I think `priceIdsHash` makes more sense as it has no size limit compared to using fixed length array

cctdaniel:
> As we've discussed before, having provider doesn't make much sense for the Pulse and it seems that this code is very centered towards different providers (from requesting one, updating, separate fees, ...). I think the options are either 1) allow all providers to fullfill a request 2) remove providers entirely and make it permissionless.

yeah this makes sense, I have gone with option 2 and removed providers entirely

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,48 @@
...
+        bytes[] calldata updateData,
+        bytes32[] calldata priceIds
+    ) external payable;
+
+    // Getters
+    function getFee(
+        uint256 callbackGasLimit
+    ) external view returns (uint128 feeAmount);
+
+    function getPythFeeInWei() external view returns (uint128 pythFeeInWei);
```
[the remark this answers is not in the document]
cctdaniel: this is just getting the base pyth fee in wei as the function name stated -- similar to entropy

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,264 @@
...
+        require(admin != address(0), "admin is zero address");
+        require(pythAddress != address(0), "pyth is zero address");
+
+        _state.admin = admin;
+        _state.accruedFeesInWei = 0;
+        _state.pythFeeInWei = pythFeeInWei;
+        _state.pyth = pythAddress;
+        _state.currentSequenceNumber = 1;
+
+        if (prefillRequestStorage) {
```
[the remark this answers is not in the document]
cctdaniel: this is also referenced from the entropy contract -- which is to optimize for gas costs, by pre-filling the storage slots with non-zero values during initialization we pay the high gas cost once during deployment and hence subsequent writes to these slots will cost less gas

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,264 @@
...
+                priceIds,
+                req.requester,
+                "low-level error (possibly out of gas)"
+            );
+        }
+
+        clearRequest(sequenceNumber);
+    }
+
+    function emitPriceUpdate(
```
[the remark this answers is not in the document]
cctdaniel: agreed but apart from debugging this is also useful for off-chain indexing/integration so i think the benefits outweigh the costs

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,264 @@
...
+        );
+    }
+
+    function getFee(
+        uint256 callbackGasLimit
+    ) public view override returns (uint128 feeAmount) {
+        uint128 baseFee = _state.pythFeeInWei;
+        uint256 gasFee = callbackGasLimit * tx.gasprice;
+        feeAmount = baseFee + SafeCast.toUint128(gasFee);
+    }
```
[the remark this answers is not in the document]
cctdaniel: yeah that's a good point, i think not allowing a publish time more than 1 min in the future is a reasonable solution

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,264 @@
...
+
+        _state.accruedFeesInWei += SafeCast.toUint128(msg.value);
+
+        emit PriceUpdateRequested(req);
+    }
+
+    function executeCallback(
+        uint64 sequenceNumber,
+        bytes[] calldata updateData,
+        bytes32[] calldata priceIds
```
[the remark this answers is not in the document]
cctdaniel: hmm i think we should still emit the event with the priceId and the trade-off of higher gas costs for event emission is necessary because the system can't function without Argus having access to the price IDs, we can keep the storage the hash still so that it remains optimized, what do you think?

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
cctdaniel: went with using fixed-length array of `priceIds` instead otherwise with hash alone its not possible to do storage polling for argus

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache 2
+
+pragma solidity ^0.8.0;
+
+contract PulseState {
+    uint8 public constant NUM_REQUESTS = 32;
+    bytes1 public constant NUM_REQUESTS_MASK = 0x1f;
+    uint8 public constant MAX_PRICE_IDS = 10;
```
jayantk, line 8: Is this enough?

I take it the problem is that you end up with higher gas usage if this limit is increased?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,282 @@
...
+                updateData,
+                priceIds,
+                SafeCast.toUint64(req.publishTime),
+                SafeCast.toUint64(req.publishTime)
+            );
+
+        // Check if enough gas remains for the callback plus 50% overhead for cross-contract call (uses integer arithmetic to avoid floating point)
+        if (gasleft() < (req.callbackGasLimit * 3) / 2) {
+            revert InsufficientGas();
+        }
```
jayantk, line 104: why do you need this check? Why not just run the rest of the code, and then either you run out of gas or you don't?

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,66 @@
...
+        PythStructs.PriceFeed[] memory priceFeeds
+    ) external;
+}
+
+interface IPulse is PulseEvents {
+    // Core functions
+    /**
+     * @notice Requests price updates with a callback
+     * @dev The msg.value must cover both the Pyth fee and gas costs
+     * Note: The actual gas required for execution will be 1.5x the callbackGasLimit
```
jayantk, line 22: This comment will be confusing to users as they don't know how to determine the fee. I think the comment should say msg.value should be set to `getFee(callbackGasLimit)`

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    struct State {
+        address admin;
+        uint128 pythFeeInWei;
+        uint128 accruedFeesInWei;
+        address pyth;
+        uint64 currentSequenceNumber;
+        address feeManager;
+        Request[32] requests;
```
jayantk, line 26: NUM_REQUESTS?

jayantk (approved):
I have a couple comments on this, but I think this is fine to merge and then we can iterate on top.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,282 @@
...
+        } catch {
+            // Out of gas or other low-level errors
+            emit PriceUpdateCallbackFailed(
+                sequenceNumber,
+                msg.sender,
+                priceIds,
+                req.requester,
+                "low-level error (possibly out of gas)"
+            );
+        }
```
jayantk, line 131: I don't think you want these catches either. If either of these errors happens for a transient reason, it means the callback can't be called in the future. 

In entropy if a callback fails, we leave the request open. It could get fulfilled later, or it could just sit there forever.

jayantk: (this is actually important. In practice, we have callbacks that error on the first try and then succeed on a later retry)

jayantk: Oh you also need to use the Check-Effects-Interactions pattern here, meaning the `clearRequest` call (the effect) needs to happen before cross-contract calls (interactions)

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
cctdaniel: yup, 10 seems like a reasonable estimate per request

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: we could remove this check, the initial thought behind was to revert early so that user only pays for operations up to the check which is a much cheaper failure case (especially if the callback is expensive)

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: > In entropy if a callback fails, we leave the request open. It could get fulfilled later, or it could just sit there forever.
i looked at the entropy code and it seems like the request is cleared before calling the callback, what does leaving the request open means here?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,282 @@
...
+        // Parse price feeds first to measure gas usage
+        PythStructs.PriceFeed[] memory priceFeeds = IPyth(_state.pyth)
+            .parsePriceFeedUpdates(
+                updateData,
+                priceIds,
+                SafeCast.toUint64(req.publishTime),
+                SafeCast.toUint64(req.publishTime)
+            );
+
+        // Check if enough gas remains for the callback plus 50% overhead for cross-contract call (uses integer arithmetic to avoid floating point)
```
ali-bahjati, line 101: i think 50% overhead is too much, the overhead is mostly constant.

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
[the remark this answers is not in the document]
ali-bahjati: hmmm so what is a base fee? asking because this interface is what integrators will see and it's not documented.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+    }
+
+    function requestPriceUpdatesWithCallback(
+        uint256 publishTime,
+        bytes32[] calldata priceIds,
+        uint256 callbackGasLimit
+    ) external payable override returns (uint64 requestSequenceNumber) {
+        require(publishTime <= block.timestamp + 60, "Too far in future");
```
ali-bahjati, line 48: I recommend documenting the reasoning behind any unintuitive behaviour. 
Please add why this check is there (for a future reader of the code)

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -0,0 +1,66 @@
...
+     * to account for cross-contract call overhead + some gas for some other operations in the function before the callback
+     * @param publishTime The minimum publish time for price updates
```
ali-bahjati, line 24: document the constraints here for an external user to avoid surprises.

target_chains/ethereum/contracts/contracts/pulse/PulseState.sol
ali-bahjati: document why 10 is chosen and the consequences of it. we need to update the IPulse to also say at most 10 feeds are supported.

ali-bahjati (approved):

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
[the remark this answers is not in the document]
cctdaniel: good point, will document

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: i think it just provides better UX by failing early with a clear error message, also actual gas paid is gas used

there might be a situation where:
- user provides exactly `callbackGasLimit` to `pulseCallback`
- `gasleft()` is also `callbackGasLimit`
- we don't have enough gas for events/cleanup
- transaction fails with generic out-of-gas error

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: the reason why we have this try-catch is because @m30m said that in entropy it's difficult to know if its the user error or our error so logging this will provide better UX, we can document that the callback contract should not fail, and if it fails we wont retry

PR #2091, opened 2024-11-04
Add `ignore_invalid_price_ids` flag to Hermes v2 APIs

tejasbadadare:
Ticket: https://www.notion.so/pyth-network/Hermes-ignore-non-existent-feeds-as-an-option-61cfe2e9279a40bcb807ca7183cc951f?pvs=4

## Purpose
Today when a user requests latest price updates with a list of price_ids, the request returns an error if any of those IDs is non-existent. Since price feeds come and go, some protocols want different behavior to avoid issues. The fix is to provide a `ignore_invalid_price_ids` flag that simply ignores invalid provided price IDs and only returns data for the valid ones.

## Implementation details
- The price ID validation is currently done by the `rest::verify_price_ids_exist(state, price_ids)` function, which is used by both deprecated v1 APIs and v2 APIs. 
- This PR modifies the function to something like `validate_price_ids(state, price_ids, remove_invalid) -> valid_price_ids`. Now the function returns a Vec of valid IDs that the caller can operate on. 
  - If `remove_invalid` is true (sourced from `params.ignore_invalid_price_ids`), then the resulting Vec won't contain any invalid price IDs.
  - If `remove_invalid` is false and there are invalid IDs in the passed in `price_ids`, then an Error will be returned, just like the original function would.
- This minimizes disruption to the existing codebase, since the APIs are written in a way that expects the code to panic if `verify_price_ids_exist` returns an Error. With the new `validate_price_ids` function, downstream code can always operate on the resulting list without worrying about invalid IDs, like before.

- Misc: Fixed some small gotchas I encountered following the setup guide in the README
## Testing
Manually tested, there's no coverage for the REST webserver module in Hermes.


apps/hermes/server/README.md
```diff
@@ -21,9 +21,10 @@ To set up and run a Hermes node, follow the steps below:
    using [Beacon](https://github.com/pyth-network/beacon), a highly available rewrite for spy, for production purposes.
 1. **Install Rust nightly-2023-07-23**: If you haven't already, you'll need to install Rust. You can
    do so by following the official instructions. Then, run the following command to install the required
-   nightly version of Rust:
+   nightly version of Rust and set it as the default:
```
keyvankhademi, line 24: I think this repo doesn't have `rust-toolchain` file and the issue is rooted in this. I think instead of asking user to set this specific rust version to repo, let's add the `rust-toolchain` file so the correct version is automatically picked up by the compiler.

apps/hermes/server/src/api/rest.rs
```diff
@@ -94,25 +94,45 @@ impl IntoResponse for RestError {
...
+    let available_ids = Aggregates::get_price_feed_ids(state).await;
+
+    // Find any price IDs that don't exist in the valid set
+    let not_found_ids: Vec<PriceIdentifier> = price_ids
         .iter()
-        .filter(|id| !all_ids.contains(id))
-        .cloned()
-        .collect::<Vec<_>>();
+        .filter(|id| !available_ids.contains(id))
+        .copied()
```
keyvankhademi, line 116: what was the reason behind replacing `cloned` with `copied`? Was it just for the performance?

apps/hermes/server/src/api/rest.rs
```diff
@@ -94,25 +94,45 @@ impl IntoResponse for RestError {
...
+        } else {
+            // Return error with list of missing IDs
+            Err(RestError::PriceIdsNotFound {
+                missing_ids: not_found_ids,
+            })
+        }
+    } else {
+        // All IDs are valid, return them unchanged
+        Ok(price_ids.to_vec())
     }
```
keyvankhademi, line 137:
```suggestion
   if !not_found_ids.is_empty() && !remove_invalid {
            Err(RestError::PriceIdsNotFound {
                missing_ids: not_found_ids,
            })
        }
    }
    Ok(price_ids
                .iter()
                .filter(|id| available_ids.contains(id))
                .copied()
                .collect())
   
```

apps/hermes/server/src/api/rest/v2/latest_price_updates.rs
```diff
@@ -57,6 +57,10 @@ pub struct LatestPriceUpdatesQueryParams {
     /// If true, include the parsed price update in the `parsed` field of each returned feed. Default is `true`.
     #[serde(default = "default_true")]
     parsed: bool,
+
+    /// If true, invalid price IDs in the `ids` parameter are ignored. Default is `false`.
+    #[serde(default)]
+    ignore_invalid_price_ids: bool,
```
keyvankhademi, line 63: I suggest somehow specifying that it only affects V2 routes, either by name or comment.

apps/hermes/server/src/api/rest/v2/sse.rs
```diff
@@ -73,6 +73,10 @@ pub struct StreamPriceUpdatesQueryParams {
     /// If true, only include benchmark prices that are the initial price updates at a given timestamp (i.e., prevPubTime != pubTime).
     #[serde(default)]
     benchmarks_only: bool,
+
+    /// If true, invalid price IDs in the `ids` parameter are ignored. Default is `false`.
+    #[serde(default)]
+    ignore_invalid_price_ids: bool,
```
keyvankhademi, line 79: I thought you said websockets are not affected by this?

keyvankhademi: I guess this counts as SSE not WS? :D

apps/hermes/server/src/api/rest.rs
ali-bahjati: I think being more explicit with copied is actually nice.

apps/hermes/server/src/api/rest.rs
```diff
@@ -94,25 +94,45 @@ impl IntoResponse for RestError {
...
-    let missing_ids = price_ids
+    let available_ids = Aggregates::get_price_feed_ids(state).await;
+
+    // Find any price IDs that don't exist in the valid set
+    let not_found_ids: Vec<PriceIdentifier> = price_ids
         .iter()
-        .filter(|id| !all_ids.contains(id))
-        .cloned()
-        .collect::<Vec<_>>();
+        .filter(|id| !available_ids.contains(id))
```
ali-bahjati, line 115: nitpicking: you can use iterator's [partition](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) to partition the ids and avoid doing this logic two times.

ali-bahjati (approved):

apps/hermes/server/src/api/rest.rs
tejasbadadare: Oh awesome, just what I was looking for!

apps/hermes/server/src/api/rest/v2/sse.rs
tejasbadadare: Yeah :)

apps/hermes/server/README.md
tejasbadadare: It is actually there in the `server/` dir, but for some reason cargo wouldnt use the right toolchain until I set it as default, not sure why

apps/hermes/server/src/api/rest/v2/latest_price_updates.rs
tejasbadadare: Good idea

apps/hermes/server/src/api/rest.rs
tejasbadadare: Yeah, when i looked into `cloned` vs `copied`, copied looked more performant so just went with that

apps/hermes/server/src/api/rest.rs
tejasbadadare: Went with Ali's `partition` suggestion here

apps/hermes/server/README.md
tejasbadadare: I'll go ahead and remove though, rust-toolchain does indeed solve this

cctdaniel (approved):
you need to bump the version [here](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/hermes/client/js/package.json#L3) otherwise when you release a new version it will fail

apps/hermes/server/src/api/rest.rs
```diff
@@ -94,25 +94,183 @@ impl IntoResponse for RestError {
...
+        .copied()
+        .partition(|id| available_ids.contains(id));
+
+    if invalid_ids.is_empty() {
+        // All IDs are valid
+        Ok(valid_ids)
+    } else if remove_invalid {
+        // Return only valid IDs
+        Ok(valid_ids)
+    } else {
```
keyvankhademi, line 124:
```suggestion
    if invalid_ids.is_empty() ||  remove_invalid {
        // Return all valid IDs
        Ok(valid_ids)
    } else {
```

apps/hermes/server/src/api/rest/v2/timestamp_price_updates.rs
```diff
@@ -67,6 +67,9 @@ pub struct TimestampPriceUpdatesQueryParams {
     /// If true, include the parsed price update in the `parsed` field of each returned feed. Default is `true`.
     #[serde(default = "default_true")]
     parsed: bool,
+
+    /// If true, invalid price IDs in the `ids` parameter are ignored. Only applicable to the v2 APIs. Default is `false`.    #[serde(default)]
```
keyvankhademi, line 71: I think `#[serde(default)]` is commented by mistake

keyvankhademi (approved):

apps/hermes/server/src/api/rest/v2/timestamp_price_updates.rs
tejasbadadare: oof good catch!

PR #2092, opened 2024-11-05
fix(target_chains/ton): audit fixes

cctdaniel:


target_chains/ton/contracts/contracts/common/storage.fc
```diff
@@ -37,8 +35,6 @@ global int governance_data_source_index; ;; u32
         .end_cell();
 
     cell data_sources_cell = begin_cell()
-        .store_dict(data_sources)
-        .store_uint(num_data_sources, 32)
```
ali-bahjati, line 41: we are removing these just because they are not needed?

ali-bahjati (approved):

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -17,7 +17,7 @@ cell store_price(int price, int conf, int expo, int publish_time) {
     .end_cell();
 }
 
-slice read_and_verify_header(slice data) {
+slice read_and_verify_header(slice data) impure {
```
ali-bahjati, line 20: why did it turn impure? because we use a global variable? strange (because it's const)

cctdaniel: they should be marked impure because it can throw exceptions using `throw_unless` and `throw_if` according to [FunC docs here](https://docs.ton.org/v3/documentation/smart-contracts/func/docs/functions#impure-specifier) otherwise it might be incorrectly optimized away

target_chains/ton/contracts/contracts/common/storage.fc
cctdaniel: yup and they are not used anywhere

PR #2097, opened 2024-11-05
chore(hermes): Remove git from build, vendor WH & Google protos

tejasbadadare:
The Hermes `build.rs` currently pulls the [wormhole](https://github.com/wormhole-foundation/wormhole.git) and [googleapis](https://github.com/googleapis/googleapis.git) git repositories in order to get the WH protobuf definitions. (The WH protos depend on googleapi protos, so those are downloaded as well.)

Aside from a slow build, this causes some issues with LSP. TLDR: The build doesn't clean up the .git lock which breaks incremental builds, since subsequent builds fail to `git init`. On my machine, this manifests as `cargo watch` and `rust-analyzer` not working since they continuously and incrementally build.

Talking to @ali-bahjati , we concluded that vendoring in the required protobufs is preferable to constantly downloading them from git, which is what this PR does.

Both repos are Apache 2.0 licensed (as are we) so we should be good to include them directly. 
The build is much faster now :)

_Notes:_
- Not all google protos from the repo are included, just the ones specified in `service.proto`. This works fine as the WH protos only depend on a select few google ones.
- Added the protos dir to the precommit exclusions and cleaned up the regex


apps/hermes/server/proto/vendor/google/api/README.md
```diff
@@ -0,0 +1,36 @@
...
+logging, and the like.
+
+Google refers to this configuration colloquially as the "service config",
+and the `service.proto` file in this directory is the entry point for
+understanding these.
+
+## Using these protos
+
+To be honest, we probably open sourced way too much of this (basically by
+accident). There are a couple files in here you are most likely to be
```
tejasbadadare, line 14: lmaoo

apps/hermes/server/.gitignore
```diff
@@ -7,4 +7,6 @@ src/network/p2p.proto
 tools/
 
 # Ignore Wormhole cloned repo
+# (Legacy -- we no longer clone the Wormhole repo, the proto files are included in this project under `proto/vendor`.
```
ali-bahjati, line 10: let's remove it entirely.

ali-bahjati (approved):

.pre-commit-config.yaml
```diff
@@ -3,11 +3,26 @@ repos:
     rev: v3.2.0
     hooks:
       - id: trailing-whitespace
-        exclude: target_chains/sui/vendor/|patches/
+        exclude: >
+          (?x)^(
```
ali-bahjati, line 7: Big 🧠 stuff

.pre-commit-config.yaml
```diff
@@ -3,11 +3,26 @@ repos:
...
+            target_chains/sui/vendor/|
+            patches/|
+            apps/hermes/server/proto/
```
ali-bahjati, line 10: probably make it `/proto/vendor`

PR #2098, opened 2024-11-06
fix: timeout for tx confirmation

m30m:


jayantk (approved):
wow there are so many dumb ways for these transactions to fail

PR #2099, opened 2024-11-06
feat(target_chains/ton): add parse_price_feed_updates

cctdaniel:


target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -156,50 +156,88 @@ int parse_pyth_payload_in_wormhole_vm(slice payload) impure {
     return payload~load_uint(160);  ;; Return root_digest
 }
 
-
-() update_price_feeds(int msg_value, slice data) impure {
-    load_data();
+tuple parse_price_feeds_from_data(slice data) {
```
cctdaniel, line 159: tuple is how we deal with arrays in func

ali-bahjati (commented):
LGTM! I'll approve once all comments are addressed.

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -176,29 +168,217 @@ int parse_pyth_payload_in_wormhole_vm(slice payload) impure {
...
+;; - Easier to traverse and read individual price feeds
+;; - Cleaner separation of data
+;; - More predictable in terms of cell structure
+cell create_price_feed_cell_chain(tuple price_feeds) {
+    cell result = null();
+
+    int i = price_feeds.tlen() - 1;
+    while (i >= 0) {
+        tuple price_feed = price_feeds.at(i);
+        int price_id = price_feed.at(0);
```
ali-bahjati, line 265: why is it 0? i think this should be i as well. can you add a test to catch this as well?

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -176,29 +168,217 @@ int parse_pyth_payload_in_wormhole_vm(slice payload) impure {
...
+        int i = 0;
+        repeat(price_ids.tlen()) {
+            int requested_id = price_ids.at(i);
+            (_, int found?) = found_price_ids.udict_get?(256, requested_id);
+            throw_unless(ERROR_PRICE_FEED_NOT_FOUND_WITHIN_RANGE, found?);
+            i += 1;
+        }
+    }
+
+    return price_feeds;
```
ali-bahjati, line 245: the returned price_feeds list does not follow the order of the given price feed ids and it gets difficult for the caller because now they need to loop each time to find the price id they asked for. Can you change the code to ensure they have the same order?

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -176,29 +168,217 @@ int parse_pyth_payload_in_wormhole_vm(slice payload) impure {
...
+                .store_ref(store_price(ema_price, ema_conf, expo, publish_time))
+            .end_cell();
+
+            tuple price_feed = empty_tuple();
+            price_feed~tpush(price_id);
+            price_feed~tpush(price_feed_cell);
+            price_feeds~tpush(price_feed);
+
+            ;; Mark this price ID as found
+            found_price_ids~udict_set(256, price_id, begin_cell().store_int(true, 1).end_cell().begin_parse());
```
ali-bahjati, line 228: if by any chance we have two updates for the same feed in a message we will add both of them to the result which is not desired.

ali-bahjati (commented):
See my inline comment about the order of the output. I'll approve as soon as it's fixed.

target_chains/ton/contracts/contracts/Pyth.fc
```diff
@@ -198,34 +211,54 @@ tuple parse_price_feeds_from_data(int msg_value, slice data, tuple price_ids, in
...
+                    .store_ref(store_price(ema_price, ema_conf, expo, publish_time))
+                .end_cell();
+
+                if (price_ids_len == 0) {
+                    ordered_feeds~udict_set(8, index, begin_cell()
+                        .store_uint(price_id, 256)
+                        .store_ref(price_feed_cell)
+                    .end_cell().begin_parse());
+                    index += 1;
+                } else {
```
ali-bahjati, line 249: This is probably when having a top level if to return in this case makes the code cleaner.

PR #2100, opened 2024-11-06
fix(target_chains/near): update dependencies and code to make tests work

ali-behjati:
The old `workspaces` package in near that is used as a test suite couldn't download the required `near-sandbox` for it from an external url and it resulted in integration tests failing.

The new workspaces package is called `near-workspaces` and its `near-sandbox` has some rust version requirements to fully work, otherwise it panics with `CompilationError(PrepareError(Deserialization))`. The rust version 1.79.0 that its sandbox uses works but the latest stable rust 1.82.0 still fails. This assumption is not documented anywhere and I guessed it based on [this](https://github.com/near/nearcore/pull/9140) old issue.

In the journey to make it work with @Riateche, we ended up updating the SDKs hoping to fix the problem and it's probably not needed but as the changes are good and will save us time in the future, i'll be keeping them.

### DIFF
--- a/.github/workflows/ci-near-contract.yml
+++ b/.github/workflows/ci-near-contract.yml
@@ -23,7 +23,5 @@ jobs:
         working-directory: target_chains/near/receiver
     steps:
       - uses: actions/checkout@v2
-      - name: Install nextest
-        run: curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
       - name: Test
         run: ./workspace-test.sh


target_chains/near/receiver/workspace-test.sh
```diff
@@ -17,4 +18,4 @@ cp target/wasm32-unknown-unknown/release/pyth.wasm .
     cp target/wasm32-unknown-unknown/release/wormhole_stub.wasm ../receiver
 )
 
-RUST_LOG=info cargo nextest run
+RUST_LOG=info cargo test
```
Riateche, line 21: I guess we can remove nextest from `.github/workflows/ci-near-contract.yml` as well.

Riateche (approved):

PR #2107, opened 2024-11-08
fix dockerfile

guibescos:


cprussin (approved):
ah nice, thanks and sorry I didn't catch this

PR #2109, opened 2024-11-11
Deny clippy warnings in CI properly

Riateche:
Pre-commit config in CI previously used `clippy --fix`, which makes `-D warnings` have no effect. Now warnings should be properly denied. I fixed the simple warnings and added allow annonations for more complex cases.

### DIFF
--- a/apps/fortuna/src/api/revelation.rs
+++ b/apps/fortuna/src/api/revelation.rs
@@ -40,10 +40,6 @@ pub async fn revelation(
         })
         .inc();
 
-    let sequence: u64 = sequence
-        .try_into()
-        .map_err(|_| RestError::InvalidSequenceNumber)?;
-
     let state = state
         .chains
         .get(&chain_id)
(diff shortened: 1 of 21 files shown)


apps/fortuna/src/api/revelation.rs
```diff
@@ -40,10 +40,6 @@ pub async fn revelation(
         })
         .inc();
 
-    let sequence: u64 = sequence
-        .try_into()
-        .map_err(|_| RestError::InvalidSequenceNumber)?;
```
keyvankhademi, line 45: Are you sure the intention here was not to check if `try_into()` has some errors?

target_chains/solana/pyth_solana_receiver_sdk/src/cpi/accounts.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::vec_init_then_push)]
```
ali-bahjati, line 1: couldn't we move this in?

ali-bahjati (approved):

apps/fortuna/src/api/revelation.rs
Riateche: `sequence` is already a `u64` so `try_into()` never fails.

target_chains/solana/pyth_solana_receiver_sdk/src/cpi/accounts.rs
Riateche: The module contains auto generated code so I feel like it's more convenient to keep this outside so it's easier to update the code if needed.

PR #2112, opened 2024-11-12
fix(apps/hermes/server): add crypto redemption rate asset type

ali-bahjati:


cctdaniel (approved):

apps/hermes/server/src/api/types.rs
```diff
@@ -326,17 +326,70 @@ pub struct PriceFeedMetadata {
...
+        match self {
+            AssetType::Crypto => write!(f, "crypto"),
+            AssetType::FX => write!(f, "fx"),
+            AssetType::Equity => write!(f, "equity"),
+            AssetType::Metal => write!(f, "metal"),
+            AssetType::Rates => write!(f, "rates"),
+            AssetType::CryptoRedemptionRate => write!(f, "crypto_redemption_rate"),
+        }
+    }
+}
```
Riateche, line 350: Maybe implement it using serde? It's an extra allocation but that shouldn't matter too much.
```
impl Display for AssetType {
    fn fmt(&self, f: &mut Formatter) -> FmtResult {
        let value = serde_json::to_value(self).unwrap();
        write!(f, "{}", value.as_str().unwrap())
    }
}
```
(You can also use serde_value instead of serde_json).

PR #2116, opened 2024-11-14
fix(staking): use catalog for react-aria & react-aria-components

cprussin:
This fixes an issue where modal dialogs wouldn't trigger due to react-aria being stateful and there being multiple copies in the dependency graph


cctdaniel (approved):
thanks!

PR #2119, opened 2024-11-15
feat: add lazer contracts and sdks

Riateche:


.gitmodules
ali-bahjati, line 1: In other usages of forge in this repo we've avoided git submodule as we thought submodules are not a great way to handle dependencies (and they are not very ergonomic). It is probably fine though, we should just ensure that we'll have a consistent behaviour.

ali-bahjati (approved):

.gitmodules
Riateche: I'd like to avoid gitmodules too but I don't see a good alternative in case of forge.

PR #2126, opened 2024-11-18
feat(hermes): create latest TWAP endpoint

tejasbadadare:
## Purpose
- Create Hermes endpoint at `/v2/updates/twap/:window_seconds/latest/` to get the latest TWAP prices for a list of price IDs with a dynamic time window (default 300s/5m).
- [See design doc here.](https://www.notion.so/dourolabs/TWAP-with-Dynamic-Time-Window-13b3e0276d9380088560e50e6abad4ad?pvs=4)

## Testing
- [WIP] Creating test suite, will request re-review when ready.

## Next up
- Next PR will include an endpoint for fetching historical TWAP given a `publish_time`.
- Future PR will include functionality to fallback to Benchmarks if the required price updates aren't present in the cache (for TWAPs longer than ~15m.)
- Future PR will add these features to the Solana receiver contract.

### DIFF
--- a/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
+++ b/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
@@ -2,7 +2,7 @@ use {
     crate::{
         api::{
             rest::{validate_price_ids, RestError},
-            types::{BinaryUpdate, EncodingType, PriceFeedTwap, PriceIdInput, TwapsResponse},
+            types::{BinaryUpdate, EncodingType, PriceIdInput, RpcPriceFeedTwap, TwapsResponse},
             ApiState,
         },
         state::aggregate::{Aggregates, RequestTime},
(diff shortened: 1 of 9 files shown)


apps/hermes/server/src/api/rest/v2/latest_twaps.rs
```diff
@@ -0,0 +1,162 @@
...
+        .into_iter()
+        .map(|data| match params.encoding {
+            EncodingType::Base64 => base64_standard_engine.encode(data),
+            EncodingType::Hex => hex::encode(data),
+        })
+        .collect();
+    let binary_price_update = BinaryUpdate {
+        encoding: params.encoding,
+        data: encoded_data,
+    };
```
tejasbadadare, line 143: Does it make sense to return binary update data for this endpoint? I'm not sure it does, especially since it's a computed average price we're returning, but not really sure of the binary data use case in general. For now I'm returning the binary update data of the messages involved in calculating the twaps.

apps/hermes/server/src/api/types.rs
```diff
@@ -244,6 +244,31 @@ impl From<PriceFeedUpdate> for ParsedPriceUpdate {
         }
     }
 }
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct PriceFeedTwap {
+    pub id: PriceIdentifier,
+    pub twap_price: Price,
+}
```
ali-bahjati, line 251: we probably need down_slots as well right?

PR #2127, opened 2024-11-19
feat(price_pusher): add price_pusher for ton

cctdaniel:


apps/price_pusher/src/ton/ton.ts
```diff
@@ -0,0 +1,126 @@
...
+    private priceServiceConnection: PriceServiceConnection,
+    private logger: Logger
+  ) {
+    this.contract = this.client
+      .provider(this.contractAddress)
+      .open(PythContract.createFromAddress(this.contractAddress));
+    const keyPair = keyPairFromSeed(Buffer.from(this.privateKey, "hex"));
+    const wallet = WalletContractV4.create({
+      publicKey: keyPair.publicKey,
+      workchain: 0,
```
ali-bahjati, line 82: what's workchain of 0? can you add a comment?

apps/price_pusher/src/ton/ton.ts
```diff
@@ -0,0 +1,126 @@
...
+      this.logger.error(err, "getPriceFeedsUpdateData failed");
+      return;
+    }
+
+    try {
+      for (const updateData of priceFeedUpdateData) {
+        const updateDataBuffer = Buffer.from(updateData, "base64");
+        const updateFee = await this.contract.getUpdateFee(updateDataBuffer);
+        const totalFee =
+          calculateUpdatePriceFeedsFee(BigInt(updateFee)) + BigInt(updateFee);
```
ali-bahjati, line 112: shouldn't we pass numUpdates here :?

ali-bahjati (approved):
Please address my comments before merging. Also bump the version!

apps/price_pusher/src/ton/ton.ts
cctdaniel: technically updateFee is the same as numUpdates but good point and this can be confusing, will make this clearer

PR #2128, opened 2024-11-19
chore: add missing package name and remove lock files

Riateche:


cprussin (approved):

ali-bahjati (approved):
I think you might need to add it to https://github.com/pyth-network/pyth-crosschain/blob/main/pnpm-workspace.yaml as well

Riateche:
It's already added to `pnpm-workspace.yml`.

PR #2131, opened 2024-11-19
feat(hermes): create latest TWAP endpoint

tejasbadadare:
## Purpose
Create Hermes endpoint at `/v2/updates/twap/:window_seconds/latest/` to get the latest TWAP prices for a list of price IDs with a dynamic time window (default 300s/5m).
- Hermes will calculate the requested TWAP, but on-chain consumers won’t have a way to verify this calculated price. Thus, the “real” calculation needs to happen on-chain, using the verified prices from `t0` and `t1`, in the target chain contract.
  - The calculation returned by Hermes (in `parsed`) will match what our SDK will return from on-chain.
- Thus, the endpoint returns the following structure (`TwapsResponse`):
  - `binary` (list of `BinaryUpdate`, where each update contains the start & end `TwapMessages` for a requested feed)
    - The `TwapMessages` contain everything needed to compute the TWAP on-chain (cumul price, expo, slot, num down slots, etc).
    - `parsed` (list of `ParsedPriceFeedTwap`, which contains the calculated TWAP for each requested feed)

[See a sample response here.](https://app.warp.dev/block/bFpSaiTt3OF8mpi9MKlohb)

## Testing
- Added unit tests

## Next up
- Next PR will include an endpoint for fetching historical TWAP given a `publish_time`.
- Future PR will include functionality to fallback to Benchmarks if the required price updates aren't present in the cache (for TWAPs longer than ~15m.)
- Future PR will add these features to the Solana receiver contract.

### DIFF
--- a/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
+++ b/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
@@ -2,10 +2,7 @@ use {
     crate::{
         api::{
             rest::{validate_price_ids, RestError},
-            types::{
-                BinaryUpdate, EncodingType, ParsedPriceTwapWindow, PriceIdInput,
-                RpcCalculatedPriceFeedTwap, TwapsResponse,
-            },
+            types::{BinaryUpdate, EncodingType, ParsedPriceFeedTwap, PriceIdInput, TwapsResponse},
             ApiState,
         },
         state::aggregate::{Aggregates, RequestTime},
@@ -37,6 +34,8 @@ pub struct LatestTwapsPathParams {
 #[into_params(parameter_in=Query)]
 pub struct LatestTwapsQueryParams {
     /// Get the most recent TWAP (time weighted average price) for this set of price feed ids.
+    /// The `binary` data contains the signed start & end cumulative price updates needed to calculate
+    /// the TWAPs on-chain. The `parsed` data contains the calculated TWAPs.
     ///
     /// This parameter can be provided multiple times to retrieve multiple price updates,
     /// for example see the following query string:
@@ -52,13 +51,9 @@ pub struct LatestTwapsQueryParams {
     #[serde(default)]
     encoding: EncodingType,
 
-    /// If true, include the parsed price update in the `parsed` field of each returned feed. Default is `true`.
+    /// If true, include the calculated TWAP in the `parsed` field of each returned feed. Default is `true`.
     #[serde(default = "default_true")]
     parsed: bool,
-
-    /// If true, include the calculated TWAP in the `calculated` field of each returned feed. Default is `true`.
-    #[serde(default = "default_true")]
-    calculated: bool,
 
     /// If true, invalid price IDs in the `ids` parameter are ignored. Only applicable to the v2 APIs. Default is `false`.
     #[serde(default)]
@@ -82,9 +77,9 @@ fn default_true() -> bool {
     true
 }
 
-/// Get the latest TWAP by price feed id.
+/// Get the latest TWAP by price feed id with a custom time window.
 ///
-/// Given a collection of price feed ids, retrieve the latest Pyth price for each price feed.
+/// Given a collection of price feed ids, retrieve the latest Pyth TWAP price for each price feed.
 #[utoipa::path(
     get,
     path = "/v2/updates/twap/{window_seconds}/latest",
@@ -153,10 +148,10 @@ where
         })
         .collect();
 
-    let parsed: Option<Vec<ParsedPriceTwapWindow>> = if params.parsed {
+    let parsed: Option<Vec<ParsedPriceFeedTwap>> = if params.parsed {
         Some(
             twaps_with_update_data
-                .windows
+                .twaps
                 .into_iter()
                 .map(Into::into)
                 .collect(),
@@ -165,23 +160,6 @@ where
         None
     };
 
-    let calculated: Option<Vec<RpcCalculatedPriceFeedTwap>> = if params.calculated {
-        Some(
-            twaps_with_update_data
-                .calculated_twaps
-                .into_iter()
-                .map(Into::into)
-                .collect(),
-        )
-    } else {
-        None
-    };
-
-    let twap_resp = TwapsResponse {
-        binary,
-        parsed,
-        calculated,
-    };
-
+    let twap_resp = TwapsResponse { binary, parsed };
     Ok(Json(twap_resp))
 }
--- a/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
+++ b/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
@@ -2,7 +2,7 @@ use {
     crate::{
         api::{
             rest::{validate_price_ids, RestError},
-            types::{BinaryUpdate, EncodingType, PriceFeedTwap, PriceIdInput, TwapsResponse},
+            types::{BinaryUpdate, EncodingType, PriceIdInput, RpcPriceFeedTwap, TwapsResponse},
             ApiState,
         },
         state::aggregate::{Aggregates, RequestTime},
@@ -82,7 +82,7 @@ fn default_true() -> bool {
     get,
     path = "/v2/updates/twap/{window_seconds}/latest",
     responses(
-        (status = 200, description = "TWAPs retrieved successfully", body = PriceUpdate),
+        (status = 200, description = "TWAPs retrieved successfully", body = TwapsResponse),
         (status = 404, description = "Price ids not found", body = String)
     ),
     params(
@@ -120,7 +120,6 @@ where
     )
     .await
     .map_err(|e| {
-        eprintln!("Application error: {:#?}", e);
         tracing::warn!(
             "Error getting TWAPs for price IDs {:?} with update data: {:?}",
             price_ids,
@@ -142,8 +141,14 @@ where
         encoding: params.encoding,
         data: encoded_data,
     };
-    let parsed_twaps: Option<Vec<PriceFeedTwap>> = if params.parsed {
-        Some(twaps_with_update_data.twaps)
+    let parsed_twaps: Option<Vec<RpcPriceFeedTwap>> = if params.parsed {
+        Some(
+            twaps_with_update_data
+                .twaps
+                .into_iter()
+                .map(Into::into)
+                .collect(),
+        )
     } else {
         None
     };
--- a/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
+++ b/apps/hermes/server/src/api/rest/v2/latest_twaps.rs
@@ -2,7 +2,10 @@ use {
     crate::{
         api::{
             rest::{validate_price_ids, RestError},
-            types::{BinaryUpdate, EncodingType, PriceIdInput, RpcPriceFeedTwap, TwapsResponse},
+            types::{
+                BinaryUpdate, EncodingType, ParsedPriceTwapWindow, PriceIdInput,
+                RpcCalculatedPriceFeedTwap, TwapsResponse,
+            },
             ApiState,
         },
         state::aggregate::{Aggregates, RequestTime},
@@ -52,6 +55,10 @@ pub struct LatestTwapsQueryParams {
     /// If true, include the parsed price update in the `parsed` field of each returned feed. Default is `true`.
     #[serde(default = "default_true")]
     parsed: bool,
+
+    /// If true, include the calculated TWAP in the `calculated` field of each returned feed. Default is `true`.
+    #[serde(default = "default_true")]
+    calculated: bool,
 
     /// If true, invalid price IDs in the `ids` parameter are ignored. Only applicable to the v2 APIs. Default is `false`.
     #[serde(default)]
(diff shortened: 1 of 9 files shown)


apps/hermes/server/src/api/types.rs
```diff
@@ -244,6 +246,50 @@ impl From<PriceFeedUpdate> for ParsedPriceUpdate {
...
+    pub end_timestamp: i64,
+    /// The calculated time weighted average price over the window
+    pub twap: RpcPrice,
+    /// The % of slots where the network was down over the TWAP window.
+    /// A value of zero indicates no slots were missed over the window, and
+    /// a value of one indicates that every slot was missed over the window.
+    /// This is a float value stored as a string to avoid precision loss.
+    #[serde(with = "pyth_sdk::utils::as_string")]
+    #[schema(value_type = String, example="0.00001")]
+    pub down_slots_ratio: f64,
```
ali-bahjati, line 264: Can we consider Decimal from [`rust_decimal`](https://crates.io/crates/rust_decimal) for this field to make it more accurate in calculation as well.

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -169,6 +171,42 @@ pub enum Update {
...
+#[derive(Debug, PartialEq)]
+pub struct PriceFeedTwapUpdate {
+    pub price_feed_id: PriceIdentifier,
+    pub cumulative_price: Price,
+    pub cumulative_num_down_slots: u64,
+    pub slot: Option<Slot>,
+    pub received_at: Option<UnixTimestamp>,
+    pub prev_publish_time: Option<UnixTimestamp>,
+}
+impl From<(&TwapMessage, Slot, UnixTimestamp)> for PriceFeedTwapUpdate {
```
ali-bahjati, line 183: It is typically expected in `From` trait to consume the given objects (i.e. no reference). I also generally think these sort of conversions (passing a tuple) are non-obvious and for non-obvious conversions is better to have an explicit function.

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -568,6 +641,129 @@ where
...
+            }
+        }
+    }
+
+    Ok(TwapsWithUpdateData {
+        twaps: calculated_twaps,
+        update_data: all_update_data,
+    })
+}
+fn calculate_twap(start_message: &TwapMessage, end_message: &TwapMessage) -> Result<Price> {
```
ali-bahjati, line 732: I think there is another check that we need to perform here, and that's making sure that the given messages are the first one in their timestamp, and this can be checked by making sure `prev_publish_time != (or <) publish_time`

ali-bahjati (commented):
Amazing!

apps/hermes/server/src/api/types.rs
tejasbadadare: good shout, updated!

apps/hermes/server/src/state/aggregate.rs
tejasbadadare: Good note, will keep that in mind for next time! Also, turns out this struct isn't even needed (leftover from a previous iteration), removed it

apps/hermes/server/src/state/aggregate.rs
tejasbadadare: Added the validation, but is it sufficient to simply throw an error in this case?

apps/hermes/server/src/api/types.rs
```diff
@@ -261,7 +262,7 @@ pub struct ParsedPriceFeedTwap {
     /// This is a float value stored as a string to avoid precision loss.
     #[serde(with = "pyth_sdk::utils::as_string")]
```
ali-bahjati, line 263: you probably don't need it anymore.

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
```diff
@@ -0,0 +1,165 @@
...
+    serde_qs::axum::QsQuery,
+    utoipa::IntoParams,
+};
+
+#[derive(Debug, Deserialize, IntoParams)]
+#[into_params(parameter_in=Path)]
+pub struct LatestTwapsPathParams {
+    /// The time window in seconds over which to calculate the TWAP, ending at the current time.
+    /// For example, a value of 300 would return the most recent 5 minute TWAP.
+    /// Must be greater than 0 and less than or equal to 600 seconds (10 minutes).
```
guibescos, line 27: is the 10 minute limits because we don't have vaas in hermes?

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// for example see the following query string:
+    ///
+    /// ```
+    /// ?ids[]=a12...&ids[]=b4c...
+    /// ```
+    #[param(rename = "ids[]")]
+    #[param(example = "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43")]
+    ids: Vec<PriceIdInput>,
+
+    /// Optional encoding type. If true, return the price update in the encoding specified by the encoding parameter. Default is `hex`.
```
guibescos, line 50: Does it make sense to call the returned string `price update`?

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
```diff
@@ -0,0 +1,165 @@
...
+where
+    S: Aggregates,
+{
+    let price_id_inputs: Vec<PriceIdentifier> =
+        params.ids.into_iter().map(|id| id.into()).collect();
+    let price_ids: Vec<PriceIdentifier> =
+        validate_price_ids(&state, &price_id_inputs, params.ignore_invalid_price_ids).await?;
+
+    // Collect start and end bounds for the TWAP window
+    let window_seconds = path_params.window_seconds as i64;
```
guibescos, line 109: Maybe you want this to be an i64 in the params to avoid this conversion here?

apps/hermes/server/src/api/types.rs
```diff
@@ -244,6 +247,50 @@ impl From<PriceFeedUpdate> for ParsedPriceUpdate {
...
+    fn from(pft: PriceFeedTwap) -> Self {
+        Self {
+            id: RpcPriceIdentifier::from(pft.id),
+            start_timestamp: pft.start_timestamp,
+            end_timestamp: pft.end_timestamp,
+            twap: RpcPrice {
+                price: pft.twap.price,
+                conf: pft.twap.conf,
+                expo: pft.twap.expo,
+                publish_time: pft.twap.publish_time,
```
guibescos, line 277: this seems a bit redundant

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -568,6 +616,145 @@ where
...
+
+    let conf_diff = end_message
+        .cumulative_conf
+        .checked_sub(start_message.cumulative_conf)
+        .ok_or_else(|| anyhow!("Confidence difference overflow"))?;
+
+    // Perform division before casting to maintain precision
+    // Cast slot_diff to the same type as price / conf diff before division
+    let price = (price_diff / slot_diff as i128) as i64;
+    let conf = (conf_diff / slot_diff as u128) as u64;
```
guibescos, line 748: I think it's better to use the checked casts `u128::from` and `u64::try_from` generally

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -568,6 +616,145 @@ where
...
+where
+    S: Cache,
+{
+    let mut calculated_twaps = Vec::new();
+    let mut all_update_data = Vec::new();
+
+    for price_id in price_ids {
+        let start_messages = state
+            .fetch_message_states(
+                vec![price_id.to_bytes()],
```
guibescos, line 634: why not fetch all price_ids at the same time

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -568,6 +616,145 @@ where
...
+        let end_message = &end_messages[0];
+
+        if let (Message::TwapMessage(start_twap), Message::TwapMessage(end_twap)) =
+            (&start_message.message, &end_message.message)
+        {
+            match calculate_twap(start_twap, end_twap) {
+                Ok(twap_price) => {
+                    // down_slots_ratio describes the % of slots where the network was down
+                    // over the TWAP window. A value closer to zero indicates higher confidence.
+                    let total_slots = end_message.slot - start_message.slot;
```
guibescos, line 662: does this need to be `publish_slot` ?

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
tejasbadadare: 10m limit for now because theres ~15m of updates cached in Hermes. Once we add the ability to hit benchmarks for historical data I'll remove the 10m limit, coming up in the next PR

apps/hermes/server/src/api/types.rs
tejasbadadare: Removed the serde as_string and [added utoipa's `decimal` feature instead which adds support for `rust_decimal.Decimal`.](https://github.com/juhaku/utoipa?tab=readme-ov-file#:~:text=decimal%3A%20Add%20support%20for%20rust_decimal%20Decimal%20type.%20By%20default%20it%20is%20interpreted%20as%20String.%20If%20you%20wish%20to%20change%20the%20format%20you%20need%20to%20override%20the%20type.%20See%20the%20value_type%20in%20component%20derive%20docs.). Ultimately it just treats it as a string anyway, so no functional difference :)

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
tejasbadadare: bit of copypasta, changed to "cumulative price updates"

apps/hermes/server/src/api/rest/v2/latest_twaps.rs
tejasbadadare: i made `window_seconds` of type `DurationInSeconds` from pyth_sdk, which is a u64 since it seemed like a good fit (probably to enforce a nonnegative duration,) but i see your point as well -- any preference on i64 vs u64 here?

apps/hermes/server/src/api/types.rs
tejasbadadare: yeah fair. i reused `RpcPrice` here to avoid creating a slightly different struct just for twap prices, since the other fields fit well (`price`, `conf`, `expo`). thoughts?

apps/hermes/server/src/state/aggregate.rs
tejasbadadare: oo great tip ty

apps/hermes/server/src/state/aggregate.rs
tejasbadadare: Initially my concern was that if `len(all_price_id_start_messages)` != `len(all_price_ids_end_messages)`.
basically if there is an unmatched start/end msg for a given price ID. But thinking about it again, since we aren't supporting partial success, i think it's acceptable to just error out in that case. Fetching all the messages up front does match the `fetch_message_states` API better as well, will make that change

apps/hermes/server/src/state/aggregate.rs
tejasbadadare: not sure of the difference tbh :) updated

apps/hermes/server/src/state/aggregate.rs
```diff
@@ -568,6 +616,145 @@ where
...
+        }
+    }
+
+    Ok(TwapsWithUpdateData {
+        twaps: calculated_twaps,
+        update_data: all_update_data,
+    })
+}
+fn calculate_twap(start_message: &TwapMessage, end_message: &TwapMessage) -> Result<Price> {
+    if end_message.publish_slot <= start_message.publish_slot {
```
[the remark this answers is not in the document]
tejasbadadare: i believe that's happening here. if start_slot == end_slot it will be caught by that check and throw an error.

tejasbadadare:
> Nice! Please bump the version as well.

it is getting bumped to new minor version 0.8.0, want me to bump major version instead?

keyvankhademi:
Nice 👍

apps/hermes/server/src/state/aggregate.rs
ali-bahjati: The `as` keyword for integer conversion is one the biggest mistakes in rust :D very un-rusty

tejasbadadare: is there a rust equivalent to the [zen of python](https://peps.python.org/pep-0020/) out there? :)

apps/hermes/server/src/api/types.rs
```diff
@@ -244,6 +247,48 @@ impl From<PriceFeedUpdate> for ParsedPriceUpdate {
...
+    /// The start unix timestamp of the window
+    pub start_timestamp: i64,
+    /// The end unix timestamp of the window
+    pub end_timestamp: i64,
+    /// The calculated time weighted average price over the window
+    pub twap: RpcPrice,
+    /// The % of slots where the network was down over the TWAP window.
+    /// A value of zero indicates no slots were missed over the window, and
+    /// a value of one indicates that every slot was missed over the window.
+    /// This is a float value stored as a string to avoid precision loss.
```
ali-bahjati, line 262: update this comment?

ali-bahjati (approved):
Very nice!

apps/hermes/server/src/api/types.rs
tejasbadadare: it is actually being returned as a string (utoipa's default behavior for Decimal)

PR #2144, opened 2024-11-25
fix(lazer): fix package build

cprussin:


lazer/sdk/js/package.json
```diff
@@ -6,16 +6,28 @@
     "access": "public"
```
Riateche, line 6: Could you bump the version to 0.1.1?

PR #2146, opened 2024-11-25
feat(lazer): add fees to solana contract

Riateche:
The user will now call our contract so that we can withdraw fees to the treasury account.

Not sure about the migration path for devnet/testnet. We can redeploy the program itself but the storage account also needs to be migrated somehow.


lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
```diff
@@ -116,6 +168,18 @@ pub struct Initialize<'info> {
...
+        payer = payer,
+        space = 0,
+        owner = system_program::ID,
+        seeds = [TREASURY_SEED],
+        bump,
+    )]
+    /// CHECK: this is a system program account but using anchor's `SystemAccount`
+    /// results in invalid output from the Accounts proc macro. No extra checks
+    /// are necessary because all necessary constraints are specified in the attribute.
+    pub treasury: AccountInfo<'info>,
```
ali-bahjati, line 182: Can you tell me why treasury is owned by system_program? If it's owned by system program then we can't withdraw.

keyvankhademi: I think we should just make it an address, not a PDA. Because if it's a PDA then we need a permissioned instruction to withdraw the fees. However with a simple address we can use the DAO treasury or what ever we want.

ali-bahjati: Hmmm not a bad idea. if you make it an address we need to put it in the config then (and people need to look it up) which is a slightly worse ux.

keyvankhademi: We need to save it in a PDA so that we can verify that it's the correct address to transfer funds to. Our SDK can load it by making an RPC call.

Riateche: System program's transfer is only possible if the account is owned by a system program.

I've added the configurable treasury address and a migration instruction and test.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
```diff
@@ -128,5 +257,24 @@ pub struct Update<'info> {
...
+    #[account(mut)]
+    pub payer: Signer<'info>,
+    #[account(
+        seeds = [STORAGE_SEED],
+        bump,
+        has_one = treasury
+    )]
+    pub storage: Account<'info, StorageV2>,
+    /// CHECK: this account doesn't need additional constraints.
+    pub treasury: AccountInfo<'info>,
```
ali-bahjati, line 274: shouldn't you check it matches the treasury inside storage?

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
ali-bahjati: well in this case what you can do is manually += -= lamports from the accounts because both are writable.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
```diff
@@ -34,13 +37,13 @@ impl TrustedSignerInfo {
...
 #[account]
-pub struct Storage {
+pub struct StorageV1 {
     pub top_authority: Pubkey,
     pub num_trusted_signers: u8,
     pub trusted_signers: [TrustedSignerInfo; MAX_NUM_TRUSTED_SIGNERS],
 }
 
-impl Storage {
+impl StorageV1 {
```
keyvankhademi, line 46: Since we haven't deployed on mainnet yet. Can we just make it a breaking change and not have V1 and V2? It feels weird that we have V1 and V2 even before we have launched on mainnet.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
```diff
@@ -49,15 +52,72 @@ impl Storage {
...
+pub struct StorageV2 {
+    pub top_authority: Pubkey,
+    pub treasury: Pubkey,
+    pub single_update_fee_in_lamports: u64,
+    pub num_trusted_signers: u8,
+    pub trusted_signers: [TrustedSignerInfo; MAX_NUM_TRUSTED_SIGNERS],
+}
+
+impl StorageV2 {
+    const SERIALIZED_LEN: usize = PUBKEY_BYTES
```
keyvankhademi, line 65: We can make also have some extra space in this account to keep room for adding more stuff in the future.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
keyvankhademi: it is checked with the `has_one` constraint in `storage` account.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
ali-bahjati: due to the way it's serialized i think extension is reserved for increasing number of signers.

ali-bahjati: if we are going to have upgrades that add fields like this we need to have a way for it so an upgrade doesn't break the config.

ali-bahjati (approved):

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
keyvankhademi: we can add extra space in between too `_extra_space: [u8; 100]` to reserver 100 bytes.
Also if we think we might increase the trusted signers we should allocate space for it too. The array length can be more than `MAX_NUM_TRUSTED_SIGNERS`.

Riateche: Added.

lazer/contracts/solana/programs/pyth-lazer-solana-contract/src/lib.rs
Riateche: I've removed v1 and v2 from the account names and method names, since we cannot easily change them later (that's a breaking change in anchor). I've removed `initialize_v1` method entirely. The migration method uses a temporary `StorageV010` struct to read the previous storage format. We can safely remove that after the migration is done.

ali-bahjati (approved):

keyvankhademi (approved):

PR #2149, opened 2024-11-27
fix(governance/xc_admin): init publishers for price store only when n…

ali-bahjati:
…eeded

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
@@ -290,7 +290,7 @@ const General = ({ proposerServerUrl }: { proposerServerUrl: string }) => {
   const handleSendProposalButtonClick = async () => {
     if (pythProgramClient && dataChanges && !isMultisigLoading && squads) {
       const instructions: TransactionInstruction[] = []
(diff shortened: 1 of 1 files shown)


governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
```diff
@@ -301,9 +301,14 @@ const General = ({ proposerServerUrl }: { proposerServerUrl: string }) => {
           ? mapKey(multisigAuthority)
           : multisigAuthority
 
-        const initPublisher = async (publisherKey: PublicKey) => {
+        const initPublisherInPriceStore = async (publisherKey: PublicKey) => {
+          // Price store is only available in Pythnet and Pythtest-crosschain
+          if (cluster !== 'pythnet' && cluster !== 'pythtest-crosschain') {
```
Riateche, line 306: Maybe a better option would be to check if price store program is deployed on the current chain?

Riateche (approved):

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
ali-bahjati: Hmmm that also works. I wanted to avoid doing it for pythtest-conformance but it doesn't matter probably.

PR #2151, opened 2024-11-28
add comments and ignore bounced messages

cctdaniel:


target_chains/ton/contracts/README.md
```diff
@@ -30,3 +30,21 @@ CHAIN_ID=<CHAIN-ID> npx blueprint run --custom https://testnet.toncenter.com/api
...
+
+**CRITICAL**: Integrators MUST validate the sender address in their receive function to ensure messages are coming from the Pyth Oracle contract. Failure to do so could allow attackers to:
+
+- Send invalid price responses
+- Impersonate users via the sender_address and custom_payload fields
+- Potentially drain the protocol
+
+### Message Bouncing Behavior
+
+- If the target protocol bounces the message (e.g., due to invalid custom payload or other errors), the forwarded TON will remain in the Pyth contract and will not be automatically refunded to the original sender.
```
ali-bahjati, line 48: can't we bounce back?

cctdaniel: to handle bounced messages and return funds to the original sender, we need to:
- store the sender's address when receiving the initial request
- add a bounce handling function that checks for bounced messages and refunds the original sender

its not as straightforward but if you think its worth it then we can consider implementing this

ali-bahjati: oof it seems pretty complicated.

PR #2163, opened 2024-12-03
feat(lazer/contracts/evm): Add fees for verification

darunrs:


lazer/contracts/evm/src/PythLazer.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 
 contract PythLazer is OwnableUpgradeable, UUPSUpgradeable {
     TrustedSignerInfo[2] public trustedSigners;
+    uint256 public verification_fee = 0.00001 ether;
```
darunrs, line 9: Not sure what is an appropriate fee. Changing this value does not impact the tests (Unless the fee is a fractional wei which I don't think is possible?) so we can just change this to whatever we want.

lazer/contracts/evm/test/PythLazer.t.sol
```diff
@@ -20,4 +20,40 @@ contract PythLazerTest is Test {
...
+        bytes memory update = hex"2a22999a577d3cc0202197939d736bc0dcf71b9dde7b9470e4d16fa8e2120c0787a1c0d744d0c39cc372af4d1ecf2d09e84160ca905f3f597d20e2eec144a446a0459ad600001c93c7d3750006240af373971c01010000000201000000000005f5e100";
+
+        uint256 fee = pythLazer.verification_fee();
+
+        address alice = makeAddr("alice");
+        vm.deal(alice, 1 ether);
+        address bob = makeAddr("bob");
+        vm.deal(bob, 1 ether);
+
+        // Alice provides appropriate fee
```
darunrs, line 38: I decided to just bunch these all together. I can also make them separate unit tests too, but I liked the idea of making more than one call on an account and confirming their balance changes as expected.

lazer/contracts/evm/src/PythLazer.sol
keyvankhademi: Is this 1 wei? I thought 1 eth was 10^18 wei

lazer/contracts/evm/src/PythLazer.sol
```diff
@@ -62,7 +63,13 @@ contract PythLazer is OwnableUpgradeable, UUPSUpgradeable {
 
     function verifyUpdate(
         bytes calldata update
-    ) external view returns (bytes calldata payload, address signer) {
+    ) external payable returns (bytes calldata payload, address signer) {
+        // Require fee and refund excess
+        require(msg.value >= verification_fee, "Insufficient fee provided");
+        if (msg.value > verification_fee) {
+            payable(msg.sender).transfer(msg.value - verification_fee);
+        }
```
keyvankhademi, line 71: I'm a little confused, what happens when msg.value == verification_fee?

darunrs: Then we wouldn't make a transfer call, since the user provided exactly the amount of fee required. That's my understanding anyway.

keyvankhademi: oh i saw this is the refund.

ali-bahjati (dismissed):

lazer/contracts/evm/src/PythLazer.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 
 contract PythLazer is OwnableUpgradeable, UUPSUpgradeable {
     TrustedSignerInfo[2] public trustedSigners;
```
ali-bahjati, line 8: The storage is incremental in EVM. I recommend we leave room for more signers here so we can add more signers. A better thing to do is to transition from having a list of signers to a map of addressSigner to expiresAt.

darunrs: Just an FYI for this, by replacing it with a mapping, we seem to incur some costs:
1. We take up semi-random storage slots for hashed items in the mapping instead of a linear sequence of slots. This means collisions become possible, though very unlikely. 
2. Scanning through the items in the mapping are basically impossible, especially considering gas fees. This means migrating the mapping in any way would be pretty difficult since modifying the struct changes the slots taken up by each item, and making a new map would be tricky since it would be hard to transfer the data over.

I would rather increase the size to a comfortable size like 100, rather than change to a map which we would have issues working with. This is based on what I've learned so far, so I could be wrong on some of these things. Please do let me know if something is wrong in my considerations above.

ali-bahjati: I see I see. Yeah makes sense to me.

darunrs:
I was able to successfully upgrade the contract using a local anvil. Will attempt to upgrade the actual contract tomorrow.

lazer/contracts/evm/script/PythLazerDeploy.s.sol
```diff
@@ -168,8 +169,25 @@ contract PythLazerDeployScript is Script {
...
+    }
+
     function run() public {
         address impl = deployImplementation("lazer:impl");
         deployProxy("lazer:proxy", impl);
     }
+
+    function migrate() public {
+        // Deploys new version and updates proxy to use new address
+        address proxyAddress =  getProxyAddress("lazer:proxy");
```
darunrs, line 185: This should deploy to a different address than the previous version since the code itself is used as part of the hash to produce the address.

darunrs:
Performed the following test which succeeded:
1. Reverted to main and deployed contract (took about 2M gas)
2. Added trusted signer (Same one in the unit test)
3. Switched to this branch
4. Updated migrate call to also call the below code after its original upgrade code
```
vm.startBroadcast();
        PythLazer lazer = PythLazer(payable(proxyAddress));
        bytes
            memory update = hex"2a22999a577d3cc0202197939d736bc0dcf71b9dde7b9470e4d16fa8e2120c0787a1c0d744d0c39cc372af4d1ecf2d09e84160ca905f3f597d20e2eec144a446a0459ad600001c93c7d3750006240af373971c01010000000201000000000005f5e100";
        lazer.verifyUpdate{value: 1}(update);
        vm.stopBroadcast();
```
5. Verified that the verifyUpdate function succeeds
6. Called verifyUpdate with no fee attached and received expected revert Error

So, it seems we are able to carry over the previous signature and correctly validate with it still, as well as expect the fee.

lazer/contracts/evm/src/PythLazer.sol
```diff
@@ -5,7 +5,9 @@ import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 
 contract PythLazer is OwnableUpgradeable, UUPSUpgradeable {
-    TrustedSignerInfo[2] public trustedSigners;
+    TrustedSignerInfo[2] private deprecatedTrustedSigners;
```
ali-bahjati, line 8: Can't you just extend this one to become like 100, then add verification_fee after it?

darunrs: Ah that's a good idea. I'll change to doing it that way.

ali-bahjati (approved):
Approving to unblock you from merging it after changing storage layout.

darunrs:
Contract was upgraded successfully. Merging.

PR #2166, opened 2024-12-04
fix: minor fixes

guibescos:
- set minimum publishers in proposal after changing publisher set
- fix proposals time

### DIFF
--- a/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/ProposalRow.tsx
+++ b/governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/ProposalRow.tsx
@@ -55,15 +55,13 @@ export const ProposalRow = ({
 
         // set proposal time
         if (!time) {
-          connection
-            .getConfirmedSignaturesForAddress2(proposal.publicKey)
-            .then((txs) => {
-              if (isCancelled) return
-              const firstBlockTime = txs?.[txs.length - 1]?.blockTime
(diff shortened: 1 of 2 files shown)


governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/ProposalRow.tsx
```diff
@@ -55,15 +55,13 @@ export const ProposalRow = ({
...
-                setTime(new Date(firstBlockTime * 1000))
-              }
-            })
+          connection.getSignaturesForAddress(proposal.publicKey).then((txs) => {
+            if (isCancelled) return
+            const firstBlockTime = txs?.[txs.length - 1]?.blockTime
+            if (firstBlockTime) {
+              setTime(new Date(firstBlockTime * 1000))
+            }
+          })
```
cprussin, line 64: We should probably add a `catch`, even if it's just to log -- generally you should include a `catch` any time you use a promise without async/await, otherwise uncaught async errors can put the application in a bad state but be very hard to debug

cprussin (approved):

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/ProposalRow.tsx
guibescos: done

PR #2169, opened 2024-12-04
feat(xc-admin-frontend): add a filter to show voted proposals

cprussin:
This PR adds a dropdown to filter proposals by only those where you've voted a particular way, if there is a wallet connected.

I've also added a status badge that shows how you voted for each proposal, and I've done some cleanup to the codebase to remove some unnecessary context, extract out a `<Select>` component, and to use [nuqs](https://nuqs.47ng.com/) for managing query params.

screenshot-2024-12-04-125932


cprussin:
Ignore the failure on insights, that's unrelated and is caused by a change to the underlying clickhouse view that insights uses which a separate PR will migrate to

PR #2170, opened 2024-12-05
fix: use higher commitment for blockhash to remove 'Minimum context slot has not been reached' errors

guibescos:


KemarTiti (approved):
lgtm

PR #2174, opened 2024-12-07
feat(solana-receiver-sdk): add `get_twap_no_older_than` to consume TwapUpdates

tejasbadadare:
## Purpose
Enables SDK users to easily deserialize and use TWAP prices stored in `TwapUpdate` accounts using `get_twap_no_older_than` . This function checks that the update is fully verified and not outdated, and returns a `TwapPrice`. To get unchecked prices, `get_twap_unchecked` is also available.

## Testing
Added unit test


ali-bahjati (approved):
Nice! Can you bump the crate version?

keyvankhademi (approved):

PR #2180, opened 2024-12-09
feat(solana-receiver-cli): Add `post-twap-update` cmd to CLI, update TWAP data model & validations

tejasbadadare:
## Purpose
This PR makes changes to the Solana receiver contract, the receiver SDK, and the receiver CLI. 
- Add `post-twap-update` command to the CLI. This is useful for users to test the update procedure, and to see an example of how to verify start & end data from Hermes via Wormhole and post it to the receiver contract. 
   - Usage: `pyth-solana-receiver-cli --wormhole <WORMHOLE> post-twap-update --start-payload <START_PAYLOAD> --end-payload <END_PAYLOAD>`
   - Refactored some existing functions to avoid code duplication

- Add extra validations to `calculate_twap` as brought up in earlier code review.
   - Validate that the start & end messages have the same `feed_id` and `exponent`

- Update `TwapUpdate` data model to remove the `verification_level` & `posted_slot` fields.
   - After a discussion with @guibescos , we decided the `verification_level` field is unnecessary because we don't support an "atomic" version of this instruction. It's required to verify VAAs before using `post_twap_update` and the instruction will error if unverified VAAs are used. Therefore all TwapUpdates produced by the program can be assumed to be fully verified.
   - We also deemed `posted_slot` not very useful since it's primarily used by price pushers to measure recency, which doesn't apply to our use case, unless we want to support some kind of scheduled/constantly pushed TWAP updates.

## Testing
Updated unit tests for `solana_receiver_sdk`. The CLI doesn't have a test suite.

### DIFF
--- a/target_chains/solana/cli/src/main.rs
+++ b/target_chains/solana/cli/src/main.rs
@@ -92,7 +92,32 @@ fn main() -> Result<()> {
                 &merkle_price_updates[0],
             )?;
         }
+        Action::PostTwapUpdate {
+            start_payload,
+            end_payload,
+        } => {
+            let rpc_client = RpcClient::new(url);
+            let payer =
+                read_keypair_file(&*shellexpand::tilde(&keypair)).expect("Keypair not found");
 
+            let start_payload_bytes: Vec<u8> = base64::decode(start_payload)?;
+            let end_payload_bytes: Vec<u8> = base64::decode(end_payload)?;
+
+            let (start_vaa, start_merkle_price_updates) =
+                deserialize_accumulator_update_data(start_payload_bytes)?;
+            let (end_vaa, end_merkle_price_updates) =
+                deserialize_accumulator_update_data(end_payload_bytes)?;
+
+            process_write_encoded_vaa_and_post_twap_update(
+                &rpc_client,
+                &start_vaa,
+                &end_vaa,
+                wormhole,
+                &payer,
+                &start_merkle_price_updates[0],
+                &end_merkle_price_updates[0],
+            )?;
+        }
         Action::InitializeWormholeReceiver {} => {
             let rpc_client = RpcClient::new(url);
             let payer =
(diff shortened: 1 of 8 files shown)


target_chains/solana/cli/src/main.rs
```diff
@@ -367,36 +392,174 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+    )?);
+
+    update_instructions.push(pyth_solana_receiver::instruction::PostUpdate::populate(
+        payer.pubkey(),
+        payer.pubkey(),
+        encoded_vaa_keypair.pubkey(),
+        price_update_keypair.pubkey(),
+        merkle_price_update.clone(),
+        get_random_treasury_id(),
+    ));
```
ali-bahjati, line 427: Does this really fit a single transaction in mainnet? cc: @guibescos (asking because iirc landing a price update tx required 3 txes 🤔 ). (this same question applies to the post twap code below)

target_chains/solana/cli/src/main.rs
```diff
@@ -367,36 +392,174 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+        end_vaa,
+        &end_encoded_vaa_keypair.pubkey(),
+        wormhole,
+    )?);
+    process_transaction(rpc_client, verify_instructions, &vec![payer])?;
+
+    // Transaction 4: Post TWAP update
+    let twap_update_keypair = Keypair::new();
+    let post_instructions = vec![
+        ComputeBudgetInstruction::set_compute_unit_limit(400_000),
```
ali-bahjati, line 504: I see 600_000 above and 400_000 here and I'm wondering whether TWAP construction is computationally cheaper than posting a price update 🤔

target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
```diff
@@ -59,29 +59,30 @@ pub struct PriceUpdateV2 {
...
     pub const LEN: usize = 8 + 32 + 2 + 32 + 8 + 8 + 4 + 8 + 8 + 8 + 8 + 8;
 }
-/// A time weighted average price account. This account is used by the Pyth Receiver program to store a verified TWAP update from a Pyth price feed.
+/// A time weighted average price account.
+/// This account is used by the Pyth Receiver program to store a TWAP update from a Pyth price feed.
+///
+/// ## Warning
+/// This function doesn't check whether the TwapUpdate is constructed from verified price updates.
+///     TwapUpdates should be only created and used after the client has already verified the VAAs via the Wormhole contract.
+///     Check out `target_chains/solana/cli/src/main.rs` for an example of how to do this.
```
ali-bahjati, line 68: What do you mean here? From what i understand the TwapUpdate can be created only from verified VAAs (we get VAA accounts that are owned by wormhole and we check the sources)

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
```diff
@@ -507,6 +505,18 @@ fn post_twap_update_from_vaas<'info>(
...
 fn calculate_twap(start_msg: &TwapMessage, end_msg: &TwapMessage) -> Result<(i64, u64, u32)> {
+    // Validate feed ids match
+    require!(
+        start_msg.feed_id == end_msg.feed_id,
+        ReceiverError::FeedIdMismatch
+    );
+
+    // Validate exponents match
+    require!(
+        start_msg.exponent == end_msg.exponent,
```
guibescos, line 516: almost feel like these checks shouldn't be in `calculate_twap` but in `post_twap_update`.
It should be immediately obvious reading the code that the assignment `twap.feed_id = start_msg.feed_id` is fine because we have that check so I recommend putting them closer.

target_chains/solana/cli/src/main.rs
guibescos: landing a price update is 2 txs

target_chains/solana/cli/src/main.rs
```diff
@@ -367,36 +392,174 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+        &wormhole,
+        &encoded_vaa_keypair,
+    )?;
+    process_transaction(
+        rpc_client,
+        init_instructions,
+        &vec![payer, &encoded_vaa_keypair],
+    )?;
+
+    // Transaction 2: Write remaining VAA data, verify VAA, and post price update
```
guibescos, line 409: This flow is suboptimal, transaction 1 is almost empty. Transaction 2 is probably too big so it won't land.

target_chains/solana/cli/src/main.rs
```diff
@@ -367,36 +392,174 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+        &start_encoded_vaa_keypair,
+    )?;
+    process_transaction(
+        rpc_client,
+        start_init_instructions,
+        &vec![payer, &start_encoded_vaa_keypair],
+    )?;
+
+    // Transaction 2: Create and initialize end VAA
+    let end_init_instructions = create_and_init_encoded_vaa_instructions(
```
guibescos, line 473: confusing that you are writing a part of the vaa here already

target_chains/solana/cli/src/main.rs
```diff
@@ -367,36 +392,174 @@ pub fn process_write_encoded_vaa_and_post_price_update(
...
+        &start_encoded_vaa_keypair.pubkey(),
+        wormhole,
+    )?);
+    verify_instructions.extend(write_and_verify_vaa_instructions(
+        &payer.pubkey(),
+        end_vaa,
+        &end_encoded_vaa_keypair.pubkey(),
+        wormhole,
+    )?);
+    process_transaction(rpc_client, verify_instructions, &vec![payer])?;
```
guibescos, line 499: you can fix later but these two verify_instructions won't fit probably

guibescos: you can spend some time later figuring out what is the most efficient way to pack all these instructions (that's what I did and where VAA_SPLIT_INDEX comes from).
For this case you need to split the first vaa and the second vaa differently.

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
guibescos: or even something like twap.feed_id = check_equal_and_return(start_msg.feed_id, end_msg.feed_id)

target_chains/solana/cli/src/main.rs
tejasbadadare: yeah makes sense. i mostly split up the txns here based on best guess - is there an easy way to measure the size of an instruction before trying to execute it? or is the easiest thing to just try executing it on devnet?

target_chains/solana/programs/pyth-solana-receiver/src/lib.rs
tejasbadadare: agreed yeah, i had all the checks in `calculate_twap` to make it easy to unit test all the validations. i'll think i'll go the route of moving the validations out into `validate_twap_messages(start, end)` or similar so it remains easy to test and obvious in the context of assignments like `twap.feed_id = start_msg.feed_id` that checks have been done.

target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
tejasbadadare: yeah your understanding is correct, the comment is a bit confusing -- will update

target_chains/solana/cli/src/main.rs
tejasbadadare: will tune these once i test and see how much the computation budget actually needs to be!

target_chains/solana/cli/src/main.rs
tejasbadadare: Found this, looks useful 
https://solana.stackexchange.com/questions/5628/is-there-a-way-to-estimate-the-transaction-size

ali-bahjati (approved):
Nice!

PR #2181, opened 2024-12-10
feat(lazer): add solana contract migration script, add message parsing to protocol

Riateche:
This script was used for migration on devnet and testnet.

The consumer contracts will need to use these parsing functions because it's not possible to return data from a CPI.

### DIFF
--- a/lazer/sdk/rust/protocol/src/message.rs
+++ b/lazer/sdk/rust/protocol/src/message.rs
@@ -6,7 +6,7 @@ use {
 };
 
 /// EVM signature enveope.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct EvmMessage {
     pub payload: Vec<u8>,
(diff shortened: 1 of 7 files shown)


lazer/sdk/rust/protocol/src/message.rs
```diff
@@ -0,0 +1,89 @@
...
+        let payload_len: usize = reader.read_u16::<LE>()?.into();
+        let mut payload = vec![0u8; payload_len];
+        reader.read_exact(&mut payload)?;
+        Ok(Self {
+            payload,
+            signature,
+            public_key,
+        })
+    }
+}
```
ali-bahjati, line 89: Maybe these are somewhere else but i think having round-trip ser/de tests are good to have.

ali-bahjati (approved):
Thanks!

lazer/sdk/rust/protocol/src/message.rs
Riateche: Added.

PR #2186, opened 2024-12-11
feat(solana-receiver-js-sdk): verify & post TWAPs

tejasbadadare:
## Purpose
Add `postTwapUpdates` and associated functions to the Solana Receiver JS SDK, which enables users to verify and post TWAP updates from Hermes to the Pyth Receiver contract on Solana. This allows the TWAPs to be consumed on-chain.

## Implementation details
- This SDK uses the [same instruction layout as in the rust CLI](https://app.warp.dev/block/oNDPmPYUU3OwHGZKZIiZr2), which is implemented in `buildPostEncodedVaasForTwapInstructions`.
- Compared with the standard layout from `postPriceUpdates` (calling `buildPostEncodedVaaInstructions` twice for the the start & end VAAs), the new layout uses 1 less transaction and has more efficient packing. 
    - The layout is as follows:
      - Tx 1. Create, init and write initial data for start VAA
         - 1143 bytes (92.78% of max 1232 bytes)
      - Tx 2. Create, init and write initial data for end VAA
         - 1143 bytes (92.78% of max 1232 bytes)
      - Tx 3. Write remaining data for both VAAs and verify both VAAs
        - 730 bytes (59.25% of max 1232 bytes)
      - Tx 4+: Post updates to receiver contract. One tx per price feed.
        - each tx is 1103 bytes (89.53% of max 1232 bytes)
      - Any additional price consumer ixs
      - Optional close ephemeral accounts ixs
      
- The IDL file that describes the receiver contract interface was manually updated to reflect the new `postTwapUpdate` function. We couldn't auto generate the new IDL because the account data models were moved from the receiver contract to the rust SDK at some point after the initial IDL was created. The new IDL wouldn't contain the necessary accounts. 
    - The most straightforward way to solve this would be to upgrade to `anchor-lang 0.30.0`,  which introduced a new "build IDL via compilation" strategy, which allows them to pull in types from external crates (like our SDK) into the IDL. However, we are on anchor `0.28.0`, and  `0.30.0` introduces several breaking changes. Critically, the wormhole SDKs have a hard dependency on `0.28.0`. We could fork it and upgrade it to 0.30.0 ourselves to unblock our own upgrade path, but that would be a time consuming process. 
    - I didn't think that was worth delaying this PR over, so I went with the path of least resistance of manually patching the IDL.
- Misc: update README, tune CLI ix compute budgets, add example. Example relies on HermesClient update, coming soon. 
    
## Testing
Manually tested, no test suite for the JS SDK.


target_chains/solana/sdk/js/pyth_solana_receiver/package.json
```diff
@@ -45,7 +45,7 @@
   "dependencies": {
     "@coral-xyz/anchor": "^0.29.0",
     "@noble/hashes": "^1.4.0",
-    "@pythnetwork/price-service-sdk": ">=1.6.0",
+    "@pythnetwork/price-service-sdk": "workspace:*",
```
tejasbadadare, line 48: Not sure why `@pythnetwork/price-service-sdk` already set to `"workspace:*"`, but we need its `parseTwapMessage` from v1.8.0.

ali-bahjati (commented):
LGTM but since I'm not the expert here I'll wait for @guibescos to review.

target_chains/solana/sdk/js/pyth_solana_receiver/README.md
```diff
@@ -175,6 +175,50 @@ Price updates are relatively large and can take multiple transactions to post on
 You can reduce the size of the transaction payload by using `addPostPartiallyVerifiedPriceUpdates` instead of `addPostPriceUpdates`.
 This method does sacrifice some security however -- please see the method documentation for more details.
 
+### Post a TWAP price update
+
+TWAP prices updates are calculated using a pair of verifiable cumulative price updates per price feed (the "start" and "end" updates for the given time window), and then performing an averaging calculation on-chain to create the time-weighted average price.
+
+The flow of using verifying, posting, and consuming these prices are the same as standard price updates. Get the binary update data from Hermes or Benchmarks, verify the VAAs via the Wormhole contract, and post the VAAs to the Pyth receiver contract. After this, you can consume the calculated TWAP posted to the TwapUpdate account. You can also optionally close these ephemeral accounts after the TWAP has been consumed to save on rent.
```
guibescos, line 182:
```suggestion
The flow of using, verifying, posting, and consuming these prices are the same as standard price updates. Get the binary update data from Hermes or Benchmarks, verify the VAAs via the Wormhole contract, and post the VAAs to the Pyth receiver contract. After this, you can consume the calculated TWAP posted to the TwapUpdate account. You can also optionally close these ephemeral accounts after the TWAP has been consumed to save on rent.
```

target_chains/solana/sdk/js/pyth_solana_receiver/README.md
```diff
@@ -175,6 +175,50 @@ Price updates are relatively large and can take multiple transactions to post on
...
+// Post the updates and calculate the TWAP
+await transactionBuilder.addPostTwapUpdates(binary_data_array);
+
+// You can now use the TWAP prices in subsequent instructions
+await transactionBuilder.addTwapConsumerInstructions(
+  async (
+    getTwapUpdateAccount: (priceFeedId: string) => PublicKey
+  ): Promise<InstructionWithEphemeralSigners[]> => {
+    // Generate instructions here that use the TWAP updates posted above.
+    // getTwaoUpdateAccount(<price feed id>) will give you the account for each TWAP update.
```
guibescos, line 203:
```suggestion
    // getTwapUpdateAccount(<price feed id>) will give you the account for each TWAP update.
```

target_chains/solana/sdk/js/pyth_solana_receiver/examples/post_twap_update.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Wallet } from "@coral-xyz/anchor";
+import fs from "fs";
+import os from "os";
+import { HermesClient } from "@pythnetwork/hermes-client";
+import { sendTransactions } from "@pythnetwork/solana-utils";
+
+// Get price feed ids from https://pyth.network/developers/price-feed-ids#pyth-evm-stable
+const SOL_PRICE_FEED_ID =
+  "0xef0d8b6fda2ceba41da15d4095d1da392a0d2f8ed0c6c7bc0f4cfac8c280b56d";
+const ETH_PRICE_FEED_ID =
```
guibescos, line 12: ETH_PRICE_FEED_ID seems unused here

target_chains/solana/sdk/js/pyth_solana_receiver/examples/post_twap_update.ts
```diff
@@ -0,0 +1,94 @@
+import { Connection, Keypair, PublicKey } from "@solana/web3.js";
+import { InstructionWithEphemeralSigners, PythSolanaReceiver } from "../";
+import { Wallet } from "@coral-xyz/anchor";
+import fs from "fs";
+import os from "os";
```
guibescos, line 5: Where you able to test this example works?

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -192,6 +199,42 @@ export class PythTransactionBuilder extends TransactionBuilder {
...
+   * // Get the price feed ids from https://pyth.network/developers/price-feed-ids#pyth-evm-stable
+   * const twapUpdateData = await hermesClient.getLatestTwaps([
+   *    SOL_PRICE_FEED_ID,
+   *    ETH_PRICE_FEED_ID,
+   * ]);
+   *
+   * const transactionBuilder = pythSolanaReceiver.newTransactionBuilder({});
+   * await transactionBuilder.addPostTwapUpdates(priceUpdateData);
+   * console.log("The SOL/USD price update will get posted to:", transactionBuilder.getPriceUpdateAccount(SOL_PRICE_FEED_ID).toBase58())
+   * await transactionBuilder.addPriceConsumerInstructions(...)
```
guibescos, line 219:
```suggestion
   * await transactionBuilder.addTwapConsumerInstructions(...)
```

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -192,6 +199,42 @@ export class PythTransactionBuilder extends TransactionBuilder {
...
+   * ```typescript
+   * // Get the price feed ids from https://pyth.network/developers/price-feed-ids#pyth-evm-stable
+   * const twapUpdateData = await hermesClient.getLatestTwaps([
+   *    SOL_PRICE_FEED_ID,
+   *    ETH_PRICE_FEED_ID,
+   * ]);
+   *
+   * const transactionBuilder = pythSolanaReceiver.newTransactionBuilder({});
+   * await transactionBuilder.addPostTwapUpdates(priceUpdateData);
+   * console.log("The SOL/USD price update will get posted to:", transactionBuilder.getPriceUpdateAccount(SOL_PRICE_FEED_ID).toBase58())
```
guibescos, line 218:
```suggestion
   * console.log("The SOL/USD price update will get posted to:", transactionBuilder.getTwapUpdateAccount(SOL_PRICE_FEED_ID).toBase58())
```

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -489,13 +587,12 @@ export class PythSolanaReceiver {
...
     });
 
-    postInstructions.push(
-      ...(await buildWriteEncodedVaaWithSplitInstructions(
-        this.wormhole,
-        trimmedVaa,
-        encodedVaaKeypair.publicKey
-      ))
+    const writeInstructions = await buildWriteEncodedVaaWithSplitInstructions(
+      this.wormhole,
```
guibescos, line 591: why this change of syntax here? Aren't they equivalent?

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
```diff
@@ -594,6 +691,98 @@ export class PythSolanaReceiver {
...
+          .instruction(),
+        signers: [twapUpdateKeypair],
+        computeUnits: POST_UPDATE_COMPUTE_BUDGET,
+      });
+
+      priceFeedIdToTwapUpdateAccount[
+        "0x" + parseTwapMessage(startUpdate.message).feedId.toString("hex")
+      ] = twapUpdateKeypair.publicKey;
+      closeInstructions.push(
+        await this.buildClosePriceUpdateInstruction(twapUpdateKeypair.publicKey)
```
guibescos, line 775: note: I think this doesn't work. This instructions only allow closing `PriceUpdateV2` s and not `TwapUpdate`

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -160,3 +166,185 @@ export async function findEncodedVaaAccountsByWriteAuthority(
...
+  };
+}
+
+/**
+ * Helper function to create, init, and write initial data for a VAA
+ * @param wormhole The Wormhole program instance
+ * @param updateData The accumulator update data containing the VAA
+ * @returns The trimmed VAA, generated keypair, and instructions
+ */
+async function buildInitEncodedVaaAndWriteInitialDataInstructions(
```
guibescos, line 257: feels like too much repeated code with `buildPostEncodedVaaInstructions`

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -160,3 +166,185 @@ export async function findEncodedVaaAccountsByWriteAuthority(
...
+  wormhole: Program<WormholeCoreBridgeSolana>,
+  updateData: AccumulatorUpdateData
+): Promise<{
+  trimmedVaa: Buffer;
+  encodedVaaKeypair: Keypair;
+  postInstructions: InstructionWithEphemeralSigners[];
+}> {
+  const trimmedVaa = trimSignatures(updateData.vaa, 13);
```
guibescos, line 265: This feels unnecessary in this context, we don't need to trim the signatures

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -160,3 +166,185 @@ export async function findEncodedVaaAccountsByWriteAuthority(
...
+ */
+export async function buildPostEncodedVaasForTwapInstructions(
+  wormhole: Program<WormholeCoreBridgeSolana>,
+  startUpdateData: AccumulatorUpdateData,
+  endUpdateData: AccumulatorUpdateData
+): Promise<{
+  postInstructions: InstructionWithEphemeralSigners[];
+  closeInstructions: InstructionWithEphemeralSigners[];
+  startVaa: {
+    trimmedVaa: Buffer;
```
guibescos, line 191: I wouldn't used trimmedVaa here since these are fully verified VAAs

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -160,3 +166,185 @@ export async function findEncodedVaaAccountsByWriteAuthority(
...
+
+  closeInstructions.push({
+    instruction: await wormhole.methods
+      .closeEncodedVaa()
+      .accounts({
+        encodedVaa: endVaa.encodedVaaKeypair.publicKey,
+      })
+      .instruction(),
+    signers: [],
+    computeUnits: 0,
```
guibescos, line 240: should this be `CLOSE_ENCODED_VAA_COMPUTE_BUDGET`

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
guibescos: I think you only need to return encodedVaa as a Pubkey like in `buildPostEncodedVaaInstructions`

guibescos (changes requested):
Imo you should try to refactor the functions for the VAA posting instructions so we share as much code between `buildPostEncodedVaaInstructions` and `buildPostEncodedVaasForTwapInstructions`

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
tejasbadadare: yeah just a remnant of debugging

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
tejasbadadare: Ooof good point looks like ill need to update the receiver's reclaim_rent method as well

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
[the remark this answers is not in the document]
tejasbadadare: yeah that would be worthwhile. will do!

target_chains/solana/sdk/js/pyth_solana_receiver/examples/post_twap_update.ts
tejasbadadare: working on it, will update when i've validated it works e2e

tejasbadadare: it works! can check out the txs from the output: https://app.warp.dev/block/dS1PbuFr3e4CgtG4MXMfCq

target_chains/solana/sdk/js/pyth_solana_receiver/src/PythSolanaReceiver.ts
tejasbadadare: added a new `reclaimTwapRent` method to the recv contract

target_chains/solana/sdk/js/pyth_solana_receiver/README.md
```diff
@@ -175,6 +175,50 @@ Price updates are relatively large and can take multiple transactions to post on
 You can reduce the size of the transaction payload by using `addPostPartiallyVerifiedPriceUpdates` instead of `addPostPriceUpdates`.
 This method does sacrifice some security however -- please see the method documentation for more details.
 
+### Post a TWAP price update
+
+TWAP prices updates are calculated using a pair of verifiable cumulative price updates per price feed (the "start" and "end" updates for the given time window), and then performing an averaging calculation on-chain to create the time-weighted average price.
```
guibescos, line 180:
```suggestion
TWAP price updates are calculated using a pair of verifiable cumulative price updates per price feed (the "start" and "end" updates for the given time window), and then performing an averaging calculation on-chain to create the time-weighted average price.
```

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -54,81 +76,250 @@ export function trimSignatures(
...
+    ],
+    closeInstructions: groups.closeInstructions,
+  };
+}
+
+/**
+ * Build instructions to post two VAAs for TWAP (Time-Weighted Average Price) calculations,
+ * optimized for 3 transactions. This is specifically designed for posting start and end
+ * accumulator update VAAs efficiently.
+ * The instructions are packed into 3 transactions:
```
guibescos, line 227: Nitpick: this function doesn't handle the packing.
Maybe you can say that this function splits the vaa writes the `writeEncodedVaa` in multiple instructions with the intention that these instructions can be packed in that way.

guibescos (approved):
tyvm

target_chains/solana/sdk/js/pyth_solana_receiver/src/vaa.ts
```diff
@@ -54,81 +76,250 @@ export function trimSignatures(
...
+ * - initInstructions: Create and initialize the encoded VAA account
+ * - writeFirstPartInstructions: Write the first part of the VAA data (up to VAA_SPLIT_INDEX)
+ * - writeSecondPartAndVerifyInstructions: Write remaining VAA data and verify signatures
+ * - closeInstructions: Close the encoded VAA account to recover rent
  */
-export const VAA_START = 46;
+interface VaaInstructionGroups {
+  initInstructions: InstructionWithEphemeralSigners[];
+  writeFirstPartInstructions: InstructionWithEphemeralSigners[];
+  writeSecondPartAndVerifyInstructions: InstructionWithEphemeralSigners[];
```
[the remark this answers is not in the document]
tejasbadadare: hm yeah they can be consolidated. but i think for newcomers to the code, seeing them split out as (init, writeFirstPart, writeSecondPart) is a little conceptually clearer imo

PR #2203, opened 2024-12-18
feat(pyth-lazer-sdk): add ed25519 ix builder function

keyvankhademi:


lazer/sdk/js/src/ed25519.ts
```diff
@@ -0,0 +1,76 @@
+import * as BufferLayout from "@solana/buffer-layout";
+import { Ed25519Program, TransactionInstruction } from "@solana/web3.js";
+
+const ED25519_INSTRUCTION_LEN = 16;
+const SIGNATURE_LEN = 64;
+const PUBKEY_LEN = 32;
+const MAGIC_LEN = 4;
+const MESSAGE_SIZE_LEN = 2;
+
+const readUint16LE = (data: Uint8Array, offset: number) => {
```
Riateche, line 10: Why not use `Buffer` which has `readUInt16LE` method?

keyvankhademi: good idea, done.

darunrs (approved):

PR #2221, opened 2025-01-02
feat(pyth-lazer-protocol): remove TMP_EXPONENT

keyvankhademi:


lazer/contracts/solana/programs/pyth-lazer-solana-contract/Cargo.toml
```diff
@@ -19,7 +19,7 @@ no-log-ix-name = []
 idl-build = ["anchor-lang/idl-build"]
 
 [dependencies]
-pyth-lazer-protocol = { version = "0.1.2", path = "../../../../sdk/rust/protocol" }
+pyth-lazer-protocol = "0.1.2"
```
darunrs, line 22: Is there a reason we only use 0.1.2 for the contracts?

lazer/sdk/rust/protocol/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-protocol"
-version = "0.1.3"
+version = "0.2.0"
```
darunrs, line 3: When should we be doing a minor version upgrade like here? Not sure why we are going to 2.0 instead of 1.4.

darunrs (approved):

lazer/contracts/solana/programs/pyth-lazer-solana-contract/Cargo.toml
keyvankhademi: i think at the time of the implementation this was the latest.

lazer/sdk/rust/protocol/Cargo.toml
keyvankhademi: Changing the traits is considered a breaking change.

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -77,13 +74,20 @@ impl Price {
...
+        let left_value = i128::from(self.0.get());
+        let right_value = i128::from(rhs.0.get());
 
-    fn try_into(self) -> Result<f64, Self::Error> {
-        Ok(self.0.get() as f64 / 10i64.checked_pow(Self::TMP_EXPONENT).context("overflow")? as f64)
+        let value = left_value * right_value / 10i128.pow(rhs_exponent);
+        let value = value.try_into()?;
+        NonZeroI64::new(value)
+            .context("zero price is unsupported")
+            .map(Self)
```
ali-bahjati, line 90: any reason you removed the traits?

keyvankhademi: With the need to pass an exponent to the `mul` and `f64 cast` functionalities, the traits are no longer suitable. So I it's has been changed to functions.

PR #2222, opened 2025-01-03
fix(pyth-solana-receiver): improve perf and security

tejasbadadare:
## Purpose
Performance and security improvements to the SVM TWAP feature.

## Implementation details
- Perf optimization: Reused `slot_diff` calculation in `calculate_twap` 
- Security optimization: Previously, the SDK's `get_twap_no_older_than` function only checked that the twap's `end_time` was more recent than `maximum_age`. However, the window size of the twap wasn't checked. This could result in a scenario where an attacker could frontrun a transaction that consumes a hardcoded TwapUpdate PDA. The attacker could overwrite the account with a verified twap of an unexpected window (very long/short). This could have the effect of manipulating, for example, the LTV ratio calculation of a lending protocol. 
  - The check has a +/- 1 second tolerance to account for Solana block time variations.

## Testing
Updated SDK test suite

### DIFF
--- a/target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
+++ b/target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
@@ -87,10 +87,11 @@ impl TwapUpdate {
     /// # Warning
     /// This function does not check :
     /// - How recent the price is
+    /// - If the TWAP's window size is expected
     /// - Whether the price update has been verified
     ///
     /// It is therefore unsafe to use this function without any extra checks,
-    /// as it allows for the possibility of using unverified or outdated price updates.
+    /// as it allows for the possibility of using unverified, outdated, or unexpected price updates.
     pub fn get_twap_unchecked(
         &self,
         feed_id: &FeedId,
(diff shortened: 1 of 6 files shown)


target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
```diff
@@ -87,10 +87,11 @@ impl TwapUpdate {
     /// # Warning
     /// This function does not check :
     /// - How recent the price is
+    /// - If the TWAP's window size is expected
     /// - Whether the price update has been verified
     ///
     /// It is therefore unsafe to use this function without any extra checks,
-    /// as it allows for the possibility of using unverified or outdated price updates.
+    /// as it allows for the possibility of using unverified, outdated, or unexpected price updates.
```
guibescos, line 94:
```suggestion
    /// as it allows for the possibility of using unverified, outdated, or arbitrary window length twap updates.
```

target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
```diff
@@ -138,6 +146,16 @@ impl TwapUpdate {
                 >= clock.unix_timestamp,
             GetPriceError::PriceTooOld
         );
+
+        // Ensure the twap window size is as expected
+        // Allow for +/- 1 second tolerance to account for the imprecision introduced by Solana block times
+        const TOLERANCE: i64 = 1;
```
guibescos, line 152: I don't get why you need `TOLERANCE` here. There should be at least one VAA per second with 400 milliseconds blocktimes.

guibescos (changes requested):
I'd remove the +- tolerance, let me know if you disagree

target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
tejasbadadare: My understanding was that there's a VAA roughly every 400ms. That means that if I request a 1 second TWAP (t0=1000 and t1=2000), the start VAA might have timestamp 800 and the end VAA might have timestamp 2000. That would result in a 1200ms window. If i had a strict check for window size, it would fail since 1200 != 1000. This is why i added the tolerance... but now i'm noticing that the start_time and end_time precision is in seconds and not milliseconds, so this shouldn't matter! Will go ahead and remove the tolerance. Does this check out?

ali-bahjati (approved):
Nice!

PR #2225, opened 2025-01-06
feat(lazer-protocol): Record publisher count for aggregated price feeds

darunrs:


darunrs:
Fairly certain this is sufficient. Is there an easy way to use these changes locally in tilt before deploying them as a new version? I tried pointing it to local path but it requires copying the files into the docker build which didn't seem super straightforward?

keyvankhademi (approved):

PR #2229, opened 2025-01-07
feat(lazer-protocol): Add latest price api interface

darunrs:


lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -0,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+use crate::{payload::AggregatedPriceFeedData, router::PriceFeedId};
+
+/// A request sent from the client to the server.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+#[serde(rename_all = "camelCase")]
+pub enum ApiRequest {
```
darunrs, line 9: Not sure if I should name this to something more specific like "RouterApiRequest" but I kinda figured the router would be the only consumer endpoint for Lazer, with history service defining its own req/res structs.

keyvankhademi: yeah

keyvankhademi (approved):

PR #2230, opened 2025-01-08
fix(lazer-protocol): Update Latest Price Interface

darunrs:


lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -15,8 +15,16 @@ pub enum ApiRequest {
...
 pub struct LatestPriceResponse {
-    pub price_feed_ids: Vec<AggregatedPriceFeedData>,
+    pub latest_prices: Vec<LatestPrice>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LatestPrice {
+    id: PriceFeedId,
+    prices: AggregatedPriceFeedData,
```
darunrs, line 29: Previous response did not include IDs. In this case, I could either send over the map or conform it to a struct. I chose to use a struct instead of a map of id -> prices as I felt that made more sense to send over HTTP.

keyvankhademi (approved):

ali-bahjati (approved):

PR #2236, opened 2025-01-10
feat(lazer): Improve JS SDK reliability via redundant parallel websocket cxns

tejasbadadare:
## Purpose
Improve JS SDK reliability via redundant parallel websocket connections (default of 3.) This reduces the chance of dropped messages during reconnection events, since other websockets will be receiving the otherwise missed messages. The parallel streams are merged and deduped to provide a single reliable stream to the SDK user.

Try it out by running `pnpm run example` (examples/index.ts)
## Technical details
- `PythLazerClient` is a thin wrapper around `WebSocketPool` which provides some convenient parsing on the received messages.

- Introduced `WebSocketPool`, which handles resilience at the pool level. 
  - It wraps the collection of redundant `ReliableWebSocket` instances, and provides usage semantics similar to a single standard WebSocket so that users don't have to deal with the details of managing multiple connections.
  - Takes multiple URLs and round robins the sockets between them.
  - Handles broadcasting `send` commands to all sockets in the pool for un/subscribing.
  - Manages the user's subscriptions. Subscription requests are replayed to the underlying ReliableWebSockets to reestablish the stream upon reconnection events.
  - Deduplicates messages from all of the cxns using a TTLCache before forwarding them to the user-provided message handler.
    - A TTLCache was used to avoid unbounded memory usage, and the TTL is set to a short 10 seconds since the deduping only needs to happen between messages received very close to each other in time. No locking needed due to the single threaded event loop.

- Reused and adapted `ReliableWebSocket` to handle resilience at the individual connection level. 
  - This class wraps WebSocket and restarts unexpectedly dropped connections. 
  - Handles restarting the cxn if no message or ping received for 10s, or if the underlying WebSocket closes unexpectedly
  - Takes wsOptions to pass to the WebSocket init, enabling using native WS features like header based auth
  - Known sharp corner here is that the ReliableWebSocket class doesn't wait to receive an `open` message before sending messages. However, this hasn't been an issue in practice when testing manually. 

## Testing
Tested manually successfully with up to 50 parallel connections. Reconnections are handled cleanly and no duplicate messages are delivered.

### DIFF
--- a/lazer/sdk/js/examples/index.ts
+++ b/lazer/sdk/js/examples/index.ts
@@ -6,7 +6,7 @@ console.debug = () => {};
 const client = new PythLazerClient(
   ["wss://pyth-lazer.dourolabs.app/v1/stream"],
   "access_token",
-  3, // Optionally specify number of parallel redundant connections to reduce the chance of dropped messages. Default is 3.
+  3, // Optionally specify number of parallel redundant connections to reduce the chance of dropped messages. The connections will round-robin across the provided URLs. Default is 3.
   console // Optionally log socket operations (to the console in this case.)
 );
 
--- a/lazer/sdk/js/examples/index.ts
+++ b/lazer/sdk/js/examples/index.ts
@@ -1,3 +1,6 @@
+/* eslint-disable no-console */
+/* eslint-disable @typescript-eslint/no-empty-function */
+
 import { PythLazerClient } from "../src/index.js";
 
 // Ignore debug messages
@@ -37,7 +40,7 @@ client.addMessageListener((message) => {
 });
 
 // Create and remove one or more subscriptions on the fly
-client.subscribe({
+await client.subscribe({
   type: "subscribe",
   subscriptionId: 1,
   priceFeedIds: [1, 2],
@@ -48,7 +51,7 @@ client.subscribe({
   parsed: false,
   jsonBinaryEncoding: "base64",
 });
-client.subscribe({
+await client.subscribe({
   type: "subscribe",
   subscriptionId: 2,
   priceFeedIds: [1, 2, 3, 4, 5],
@@ -60,8 +63,8 @@ client.subscribe({
   jsonBinaryEncoding: "hex",
 });
 
-await new Promise((resolve) => setTimeout(resolve, 10000));
+await new Promise((resolve) => setTimeout(resolve, 10_000));
 
-client.unsubscribe(1);
-client.unsubscribe(2);
+await client.unsubscribe(1);
+await client.unsubscribe(2);
 client.shutdown();
--- a/lazer/sdk/js/examples/index.ts
+++ b/lazer/sdk/js/examples/index.ts
@@ -1,13 +1,21 @@
 import { PythLazerClient } from "../src/index.js";
 
-/* eslint-disable no-console */
-const client = new PythLazerClient("ws://127.0.0.1:1234/v1/stream", "ctoken1");
+// Ignore debug messages
+console.debug = () => {};
+
+const client = new PythLazerClient(
+  ["wss://pyth-lazer.dourolabs.app/v1/stream"],
+  "access_token",
+  3, // Optionally specify number of parallel redundant connections to reduce the chance of dropped messages. Default is 3.
+  console // Optionally log socket operations (to the console in this case.)
+);
+
 client.addMessageListener((message) => {
-  console.log("got message:", message);
+  console.info("got message:", message);
   switch (message.type) {
     case "json": {
       if (message.value.type == "streamUpdated") {
-        console.log(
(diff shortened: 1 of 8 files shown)


lazer/sdk/js/examples/index.ts
```diff
@@ -1,13 +1,21 @@
 import { PythLazerClient } from "../src/index.js";
 
-/* eslint-disable no-console */
-const client = new PythLazerClient("ws://127.0.0.1:1234/v1/stream", "ctoken1");
+// Ignore debug messages
+console.debug = () => {};
+
+const client = new PythLazerClient(
+  ["wss://pyth-lazer.dourolabs.app/v1/stream"],
```
tejasbadadare, line 7: ok to include the staging link?

keyvankhademi: yeah

lazer/sdk/js/src/socket/ResillientWebSocket.ts
```diff
@@ -0,0 +1,183 @@
...
+ * This class wraps websocket to provide a resilient web socket client.
+ *
+ * It will reconnect if connection fails with exponential backoff. Also, it will reconnect
+ * if it receives no ping request or regular message from server within a while as indication
+ * of timeout (assuming the server sends either regularly).
+ *
+ * This class also logs events if logger is given and by replacing onError method you can handle
+ * connection errors yourself (e.g: do not retry and close the connection).
+ */
+export class ResilientWebSocket {
```
keyvankhademi, line 18: is this copy pasted from the other ResilientWebSocker? Can we reuse that one?

lazer/sdk/js/src/socket/WebSocketPool.ts
```diff
@@ -0,0 +1,162 @@
...
+   * and storing it for replay on reconnection
+   * @param request The subscription request to send
+   */
+  addSubscription(request: Request) {
+    if (request.type !== "subscribe") {
+      throw new Error("Request must be a subscribe request");
+    }
+    this.subscriptions.set(request.subscriptionId, request);
+    this.sendRequest(request);
+  }
```
keyvankhademi, line 123: I think it's good to listen for subscribed response here and throw an error if it was not successful

lazer/sdk/js/src/socket/WebSocketPool.ts
```diff
@@ -0,0 +1,162 @@
...
+   * and removing it from stored subscriptions
+   * @param subscriptionId The ID of the subscription to remove
+   */
+  removeSubscription(subscriptionId: number) {
+    this.subscriptions.delete(subscriptionId);
+    const request: Request = {
+      type: "unsubscribe",
+      subscriptionId,
+    };
+    this.sendRequest(request);
```
keyvankhademi, line 136: same here

lazer/sdk/js/examples/index.ts
```diff
@@ -1,13 +1,21 @@
 import { PythLazerClient } from "../src/index.js";
 
-/* eslint-disable no-console */
-const client = new PythLazerClient("ws://127.0.0.1:1234/v1/stream", "ctoken1");
+// Ignore debug messages
+console.debug = () => {};
```
keyvankhademi, line 4: I think this can cause issues for the end user. let's avoid overriding global stuff.

keyvankhademi: oh actually this is the example so it's fine I guess

lazer/sdk/js/package.json
```diff
@@ -63,6 +63,8 @@
     "@solana/buffer-layout": "^4.0.1",
     "@solana/web3.js": "^1.98.0",
```
keyvankhademi, line 64: let's increase the version so we can deploy

keyvankhademi (approved):

lazer/sdk/js/src/socket/WebSocketPool.ts
tejasbadadare: Added error handling code to dedupeHandler, which will throw an Error if the server responds with a `subscriptionError` or `error` type message. Btw in my manual testing, looks like the server abruptly closes the connection for issues like bad subscription payload.
image

lazer/sdk/js/src/socket/ResillientWebSocket.ts
tejasbadadare: it's got some changes, like handling bearer header auth and modifications to the heartbeat mechanism to work with lazer

lazer/sdk/js/src/socket/WebSocketPool.ts
ali-bahjati: we should improve the server side error handling for sure

PR #2238, opened 2025-01-10
feat(fortuna): Dynamically increase gas limit for a transaction on failure

jayantk:
This PR adds auto-scaling for the transaction gasLimit parameter when the callback fails. There seems to be an issue on abstract-testnet where some transactions hit this limit for unknown reasons.

I also drive-by fixed an error with the fee calculation.

### DIFF
--- a/apps/fortuna/src/config.rs
+++ b/apps/fortuna/src/config.rs
@@ -134,6 +134,10 @@ pub struct EthereumConfig {
     /// The gas limit to use for entropy callback transactions.
     pub gas_limit: u64,
 
+    /// The percentage multiplier to apply to the gas limit for each backoff.
+    #[serde(default = "default_backoff_gas_multiplier_pct")]
+    pub backoff_gas_multiplier_pct: u64,
+
     /// The minimum percentage profit to earn as a function of the callback cost.
     /// For example, 20 means a profit of 20% over the cost of the callback.
     /// The fee will be raised if the profit is less than this number.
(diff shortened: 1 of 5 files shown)


apps/fortuna/src/keeper.rs
```diff
@@ -303,7 +307,9 @@ pub async fn run_keeper_threads(
             chain_state.provider_address,
             ADJUST_FEE_INTERVAL,
             chain_eth_config.legacy_tx,
-            chain_eth_config.gas_limit,
+            // NOTE: we adjust fees based on the maximum gas that the keeper will submit a callback with.
+            // This number is *larger* than the configured gas limit, as we pad gas on transaction submission for reliability.
+            (chain_eth_config.gas_limit * DEFAULT_GAS_ESTIMATE_MULTIPLIER_PCT) / 100,
```
jayantk, line 312: We were previously multiplying by 5/4 in the function that sends the TX, but we weren't accounting for that when setting the tx fee.

apps/fortuna/src/config.rs
```diff
@@ -172,6 +176,10 @@ pub struct EthereumConfig {
     pub priority_fee_multiplier_pct: u64,
 }
 
+fn default_backoff_gas_multiplier_pct() -> u64 {
+    100
```
jayantk, line 180: we may want to set this to some reasonable % in the future, but let's see how the abstract thing goes first.

m30m (approved):

PR #2242, opened 2025-01-10
fix(price-service-sdk): make price-service-sdk portable

cprussin:
The price-service-sdk package previously used `Buffers` in it's interface in many places.  However, `Buffer` is a node-specific interface which does not exist in browsers.

Many browsers ship [a polyfill](https://github.com/feross/buffer), however I've discovered that the polyfill is not actually fully compatible with the Node module.  In particular, while the Node module [supports aliases such as `readUint8` for `readUInt8`](https://nodejs.org/api/buffer.html#bufreaduint8offset) (note the difference in capitalization), [the polyfill does not](https://github.com/feross/buffer/blob/master/index.d.ts).

As a result, attempting to utilize `price-service-sdk` in the browser was either going to cause errors because the Buffer polyfill isn't present, or it would cause errors because of the use of aliased function names.

There were three options to fix this issue:

1. Switch everything to using unaliased function names.  This would work portably, but only assuming that the Buffer polyfill is present.

2. Switch everything to using [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array), which is a more generic [typed array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) and largely supersedes the need to use the `Buffer` class at all (and which is a superclass of `Buffer`).  This would work, however it would be a breaking change.

3. Switch everything to using a generic interface which accepts anything which extends `Uint8Array` (including `Buffer`), and use `Uint8Array` methods exclusively inside `price-service-sdk`.  This requires a few gross typescript hacks to type check correctly and avoid anything being a breaking change, but it makes the code the most portable without requiring any polyfills or requiring a major version bump.

This commit implements option 3.

### DIFF
--- a/price_service/sdk/js/src/AccumulatorUpdateData.ts
+++ b/price_service/sdk/js/src/AccumulatorUpdateData.ts
@@ -1,4 +1,5 @@
 import BN from "bn.js";
+import type { Buffer } from "buffer";
 
 const ACCUMULATOR_MAGIC = "504e4155";
 const MAJOR_VERSION = 1;
@@ -7,12 +8,12 @@ const KECCAK160_HASH_SIZE = 20;
 const PRICE_FEED_MESSAGE_VARIANT = 0;
 const TWAP_MESSAGE_VARIANT = 1;
 
-export type AccumulatorUpdateData = {
-  vaa: Buffer;
-  updates: { message: Buffer; proof: number[][] }[];
+export type AccumulatorUpdateData<T extends Uint8ArrayLike = Buffer> = {
+  vaa: T;
+  updates: { message: T; proof: number[][] }[];
 };
-export type PriceFeedMessage = {
-  feedId: Buffer;
+export type PriceFeedMessage<T extends Uint8ArrayLike = Buffer> = {
+  feedId: T;
   price: BN;
   confidence: BN;
   exponent: number;
@@ -22,8 +23,8 @@ export type PriceFeedMessage = {
   emaConf: BN;
 };
 
-export type TwapMessage = {
-  feedId: Buffer;
+export type TwapMessage<T extends Uint8ArrayLike = Buffer> = {
+  feedId: T;
   cumulativePrice: BN;
   cumulativeConf: BN;
   numDownSlots: BN;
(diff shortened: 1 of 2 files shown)


price_service/sdk/js/src/AccumulatorUpdateData.ts
```diff
@@ -213,3 +224,62 @@ export function parseAccumulatorUpdateData(
...
+  ...args: Parameters<typeof Uint8Array.from>
+) {
+  return Object.getPrototypeOf(buf.constructor).from(...args) as T;
+}
+
+// Similar to `fromAsTypeOf`, here we want to be able to create either a
+// `Buffer` or a `Uint8Array`, matching the type of a passed in value.  But this
+// is a bit more complex, because for `Uint8Array` we should do that with the
+// `Uint8Array` constructor, where for `Buffer` we should do that using
+// `Buffer.from`.  This is a bit of a weird hack but I don't know a better way
```
tejasbadadare, line 275: do you mean `Buffer.alloc` here? don't see a reference to `from`

tejasbadadare (approved):
lgtm! version bump?

price_service/sdk/js/src/AccumulatorUpdateData.ts
cprussin: ah yes thanks will fix

cprussin:
thanks @tejasbadadare , I updated the version

price_service/sdk/js/src/AccumulatorUpdateData.ts
```diff
@@ -213,3 +224,62 @@ export function parseAccumulatorUpdateData(
...
+  for (const arr of inputs) {
+    out.set(arr, offset);
+    offset += arr.length;
+  }
+  return out;
+}
+
+function toHex(input: Uint8ArrayLike): string {
+  return Array.from(input)
+    .map((value) => value.toString(16).padStart(2, "0"))
```
guibescos, line 245: doesn't this need to be `0x`

guibescos, line 245: why the padding

guibescos (commented):
Seems to pass the tests, but it's a bit complicated imo tbh

cprussin:
Closing this as stale, I'll revisit if this ever comes up again

PR #2246, opened 2025-01-13
feat(xc-admin): add support for deserializing lazer instructions in xc-admin

cctdaniel:
I had to regenerate the lazer idl with anchor `0.29.0` and comment out `idl-build = ["anchor-lang/idl-build"]` for it to work otherwise all the existing idls on xc-admin will break if i used the idl from anchor `0.30.1` since the idl format changed


ali-bahjati (approved):
thanks a lot!

ali-bahjati (approved):

PR #2250, opened 2025-01-13
Lazer solana audit fixes

Riateche:
- Disallow config account migration if it's already migrated. Repeatedly applied migration could overwrite the config account in an invalid way.
- Fetch message data from the instructions sysvar based on the signature program arguments when verifying instead of trusting the caller to pass correct `message_data` and `message_offset`. Generally, it was the caller contract's responsibility to make sure they specify `message_data` and `message_offset` correctly regardless of the untrusted input. However, this looks unsafe in non-CPI scenario (even though that scenario is not intended for use) and opens a possibility of an incorrect use via CPI. The new approach does not require passing `message_offset` and verifies that `message_data` equals the data at the offset for the verified signature.


ali-bahjati (approved):
Thanks!

PR #2251, opened 2025-01-13
feat(fortuna): Add a couple more config values

jayantk:
Two simple changes here
* add a padding multiplier for the gas limit so that the keeper is willing to submit txs that slightly exceed the limit. I noticed a few failing txs in mainnet that are at ~510k gas, which could happen due to differences in the number of hashes. This defaults to 10% and that number should probably be fine without configuration.
* support negative target profit percentages. It turns out that for abstract-testnet, you get a gas refund at the end of the tx which is like 50% of the tx cost. However, you need to set the priority fee *without counting this refund*. Given that, we're systematically overestimating the needed fee.

### DIFF
--- a/apps/fortuna/src/config.rs
+++ b/apps/fortuna/src/config.rs
@@ -199,17 +199,9 @@ pub struct EscalationPolicyConfig {
     #[serde(default = "default_gas_multiplier_cap_pct")]
     pub gas_multiplier_cap_pct: u64,
 
-    /// The initial fee multiplier to apply to the fee estimate.
-    /// Note: consider carefully whether you want to adjust this value or priority_fee_multiplier_pct in the chain configuration.
-    /// The difference between these is that priority_fee_multiplier_pct additionally influences the fees charged by the keeper.
-    /// Setting this value to >100 means that the keeper may lose money on callbacks -- only do this if you know what you are doing.
-    #[serde(default = "default_initial_fee_multiplier_pct")]
-    pub initial_fee_multiplier_pct: u64,
-
     /// The fee multiplier to apply to the fee during backoff retries.
-    /// The fee estimate for the chain (which itself may be padded based on our chain configuration) is multiplied by the fee multiplier.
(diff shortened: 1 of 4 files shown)


apps/fortuna/src/config.rs
```diff
@@ -143,20 +143,23 @@ pub struct EthereumConfig {
...
+    pub target_profit_pct: i64,
 
     /// The maximum percentage profit to earn as a function of the callback cost.
-    /// For example, 100 means a profit of 100% over the cost of the callback.
+    /// For example, 100 means a profit of 100% over the cost of a callback that uses the full gas limit.
     /// The fee will be lowered if it is more profitable than specified here.
     /// Must be larger than min_profit_pct.
-    pub max_profit_pct: u64,
+    /// The minimum value for this is -100. If set to < 0, it means the keeper may lose money on callbacks that use the full gas limit.
+    pub max_profit_pct: i64,
```
jayantk, line 162: (In hindsight, I wish that i had specified these values as multipliers, so 0 would be the natural lower bound and you could just say "75" instead of "-25". That's how I did the other scalars and it would have been better here.

However, I'm choosing to support up to -100 now in order to maintain backward compatibility on the config format)

apps/fortuna/src/keeper.rs
```diff
@@ -338,9 +338,10 @@ pub async fn run_keeper_threads(
             // In the unlikely event that the keeper fees aren't sufficient, the solution to this is to configure the target
             // fee percentage to be higher on that specific chain.
             chain_eth_config.gas_limit,
-            chain_eth_config.min_profit_pct,
-            chain_eth_config.target_profit_pct,
-            chain_eth_config.max_profit_pct,
+            // NOTE: unwrap() here so we panic early if someone configures these values below -100.
+            u64::try_from(100 + chain_eth_config.min_profit_pct).unwrap(),
```
m30m, line 342: maybe just switch to expect to give more context

m30m (approved):

PR #2252, opened 2025-01-14
feat: export lazer program id in js sdk

cctdaniel:
- **export lazer program id**
- **bump**


ali-bahjati (approved):
Please add a Solana prefix to the constants so people don't get confused here.

PR #2261, opened 2025-01-15
chore(lazer): update governance scripts to make them work

ali-bahjati:


governance/xc_admin/packages/xc_admin_common/src/multisig_transaction/idl/lazer.json
ali-bahjati, line 1: this file is manually generated because idl format is different in the version of anchor in this package. unfortunately the ProgramErrors were not defined and as we really don't use them I just removed them.

keyvankhademi (approved):

Riateche (approved):

PR #2262, opened 2025-01-15
feat: allow proposing without connecting a wallet

guibescos:
Initially proposing required a wallet because we sent the transactions directly to the blockchain.
For a long time, we switched to our backend at proposer server submitting the transactions, but we still required a wallet connection. This PR removes that requirement.


governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
```diff
@@ -73,21 +75,31 @@ export const useMultisig = (): MultisigHookData => {
...
+  }, [wallet, connection])
+
+  useEffect(() => {
+    setReadOnlySquads(
+      new SquadsMesh({
+        connection,
+        wallet: new NodeWallet(new Keypair()),
+      })
+    )
+  }, [connection])
```
cprussin, line 94: This shouldn't be done with an effect & state, this should be done instead with `useMemo`

governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
```diff
@@ -73,21 +75,31 @@ export const useMultisig = (): MultisigHookData => {
         })
       )
     } else {
-      setSquads(undefined)
+      setSquads(
+        new SquadsMesh({
+          connection,
+          wallet: new NodeWallet(new Keypair()),
+        })
+      )
```
cprussin, line 83: I'm a bit confused why we even have `readOnlySquads` if `squads` falls back to the same value, can't we just use `squads` everywhere?

cprussin (changes requested):

governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
guibescos: It's because I don't want to reload all the data when a user connects the wallet.

governance/xc_admin/packages/xc_admin_frontend/hooks/useMultisig.ts
guibescos: fixed

cprussin (approved):

PR #2267, opened 2025-01-15
chore: remove node tests from solana lazer contract workflow

cprussin:
I was going to fix this so that it installs pnpm properly to use the correct version, however as it turns out this is actually redundant since we already run tests with https://github.com/pyth-network/pyth-crosschain/blob/main/.github/workflows/ci-turbo-test.yml so there's really no reason to duplicate the tests here as well.


keyvankhademi (approved):

Riateche (approved):
Thanks!

PR #2279, opened 2025-01-20
feat(pulse): add provider

cctdaniel:


target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -171,10 +190,15 @@ abstract contract Pulse is IPulse, PulseState {
...
+        uint256 callbackGasLimit,
+        address provider
     ) public view override returns (uint128 feeAmount) {
+        if (provider == address(0)) {
+            provider = _state.defaultProvider;
+        }
         uint128 baseFee = _state.pythFeeInWei;
-        uint256 gasFee = callbackGasLimit * tx.gasprice;
+        uint128 providerFeeInWei = _state.providers[provider].feeInWei;
+        uint256 gasFee = callbackGasLimit * providerFeeInWei;
```
cctdaniel, line 201: as discussed, changed this to use provider fee to protect from future price jumps/drops

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -105,15 +124,6 @@ abstract contract Pulse is IPulse, PulseState {
...
 
-        // Check if enough gas remains for callback + events/cleanup
-        // We need extra gas beyond callbackGasLimit for:
-        // 1. Emitting success/failure events
-        // 2. Error handling in catch blocks
-        // 3. State cleanup operations
-        if (gasleft() < (req.callbackGasLimit * 3) / 2) {
-            revert InsufficientGas();
-        }
-
```
cctdaniel, line 116: removed this because if insufficient gas error occurs within callback then `PriceUpdateCallbackFailed` will be emitted with `"low-level error (possibly out of gas)"` and otherwise if insufficient gas for the `executeCallback` function itself, it will revert with out-of-gas

ali-bahjati (commented):
I thought what we agreed was that the user would not be involved in selecting provider. we instead give exclusivity to providers to update prices early; if they don't do it after a few seconds anyone can do it.

cctdaniel:
> I thought what we agreed was that the user would not be involved in selecting provider. we instead give exclusivity to providers to update prices early; if they don't do it after a few seconds anyone can do it.

my understanding was that we agreed to add providers first, with regards to the user not selecting provider the logic is not that straightforward as it requires us to figure out what is "reliable" for e.g. we would have to design a `calculateReliabilityScore` function that generates a score based on the provider performance

i can add this in a following PR

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -9,7 +9,7 @@ import "./PulseState.sol";
 interface IPulseConsumer {
     function pulseCallback(
         uint64 sequenceNumber,
-        address updater,
+        address provider,
```
ali-bahjati, line 12: I'm wondering whether we really need to know the provider in the callback. the end user doesn't care, does it?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -174,7 +193,10 @@ abstract contract Pulse is IPulse, PulseState {
         uint256 callbackGasLimit
     ) public view override returns (uint128 feeAmount) {
         uint128 baseFee = _state.pythFeeInWei;
-        uint256 gasFee = callbackGasLimit * tx.gasprice;
+        uint128 providerFeeInWei = _state
+            .providers[_state.defaultProvider]
+            .feeInWei;
+        uint256 gasFee = callbackGasLimit * providerFeeInWei;
```
ali-bahjati, line 199: alright so now provider fee is per gas, and pyth fee is the constant, right?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -13,16 +13,24 @@ abstract contract Pulse is IPulse, PulseState {
         address admin,
```
ali-bahjati, line 13: just to be clear, here the default provider is just set and should be initialized in a separate tx? (maybe make a comment on it)

ali-bahjati (approved):
LGTM!

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
cctdaniel: yeah good point, will remove

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: yup exactly

`baseFee` is the fixed fee to Pyth
`providerFeeInWei` is the provider's per-gas fee

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: yup, will do

PR #2282, opened 2025-01-22
feat(pulse): add provider exclusivity period

cctdaniel:


target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -29,6 +29,7 @@ abstract contract Pulse is IPulse, PulseState {
         _state.pyth = pythAddress;
         _state.currentSequenceNumber = 1;
         _state.defaultProvider = defaultProvider;
+        _state.exclusivityPeriodSeconds = 15; // Default to 15 seconds
```
ali-bahjati, line 32: I recommend making it configurable in deployment as well. I think 15s is good but in some chains we might need faster and in some we need slower (like eth)

ali-bahjati (approved):
I like going with default provider for now. You'd need to update some other functions like getFee. 

I think it might be easier to review all in the previous PR so feel free to merge it and I'll review that one later.

PR #2285, opened 2025-01-22
feat(insights): responsive layouts and components

partyparrotgreg:
Chores around making Insights Hub layout and components responsive.

### DIFF
--- a/apps/insights/src/components/Overview/index.tsx
+++ b/apps/insights/src/components/Overview/index.tsx
@@ -1,8 +1,9 @@
-import styles from "./index.module.scss";
-import { PageTitle } from "../PageTitle/page-title";
+import { Card } from "@pythnetwork/component-library/Card";
+
+import { PageLayout } from "../PageLayout/page-layout";
 
 export const Overview = () => (
-  <div className={styles.overview}>
-    <PageTitle>Overview</PageTitle>
-  </div>
+  <PageLayout title={"Overview"}>
+    <Card title="Overview"></Card>
+  </PageLayout>
 );
--- a/apps/insights/src/components/PageLayout/page-layout.module.scss
+++ b/apps/insights/src/components/PageLayout/page-layout.module.scss
@@ -1,0 +1,24 @@
+@use "@pythnetwork/component-library/theme";
+
+.pageLayout {
+    @include theme.max-width;
+    display: flex;
+    gap: theme.spacing(6);
+    flex-direction: column;
+
+    .pageTitleContainer {
+        display: flex;
+        flex-flow: row nowrap;
+        gap: theme.spacing(3);
+        width: 100%;
+        align-items: center;
+        justify-content: space-between;
+
+        .pageTitle {
+            @include theme.h3;
+            color: theme.color("heading");
+            font-weight: theme.font-weight("semibold");
+            flex-grow: 1;
+        }
+    }
+}
--- a/apps/insights/src/components/PageLayout/page-layout.tsx
+++ b/apps/insights/src/components/PageLayout/page-layout.tsx
@@ -1,0 +1,15 @@
+import type { ReactNode } from "react";
+
+import styles from "./page-layout.module.scss";
+
+export const PageLayout = ({ children, title, actions }: { children: ReactNode; title: ReactNode, actions?: ReactNode }) => {
+    return (
+        <div className={styles.pageLayout}>
+            <div className={styles.pageTitleContainer}>
+                <h1 className={styles.pageTitle}>{title}</h1>
+                {actions && <div className={styles.actions}>{actions}</div>}
+            </div>
+            {children}
+        </div>
+    )
+}
--- a/apps/insights/src/components/PageTitle/page-title.module.scss
+++ b/apps/insights/src/components/PageTitle/page-title.module.scss
@@ -1,7 +1,0 @@
-@use "@pythnetwork/component-library/theme";
-
-.pageTitle {
-    @include theme.h3;
-    color: theme.color("heading");
-    font-weight: theme.font-weight("semibold");
-}
--- a/apps/insights/src/components/PageTitle/page-title.tsx
+++ b/apps/insights/src/components/PageTitle/page-title.tsx
@@ -1,9 +1,0 @@
-import type { ReactNode } from "react"
-
-import styles from "./page-title.module.scss"
-
-export const PageTitle = ({ children }: { children: ReactNode }) => {
-    return (
-        <h1 className={styles.pageTitle}>{children}</h1>
-    )
-}
--- a/apps/insights/src/components/PriceFeeds/index.module.scss
+++ b/apps/insights/src/components/PriceFeeds/index.module.scss
@@ -1,76 +1,53 @@
 @use "@pythnetwork/component-library/theme";
 
-.priceFeeds {
-  @include theme.max-width;
+.toolbarContainer {
+  display: flex;
+  flex-flow: row nowrap;
+  gap: theme.spacing(2);
+}
 
-  .header {
-    @include theme.h3;
+.feedKey {
+  margin: 0 -#{theme.button-padding("xs", true)};
+}
 
-    color: theme.color("heading");
-    font-weight: theme.font-weight("semibold");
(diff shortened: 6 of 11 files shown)


apps/insights/src/components/MobileMenu/mobile-menu.v2.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => {
+    const { theme } = useTheme();
+    const [isOpen, setIsOpen] = useState(false);
+
+    const toggleMenu = () => {
+        setIsOpen(!isOpen);
+    };
+
+    return (
+        <div className={clsx(styles.mobileMenuTrigger, className)} {...props}>
```
cprussin, line 29: Rather than manually coding out this overlay, we should be using the `ModalDialog` component (or one of the components that use it like `Drawer`) from the component library.  The `ModalDialog` will handle a ton of accessibility issues like tab locking, setting proper aria attributes, focus control, and will also apply animation in/out states automatically for you.  It should generally be used any time you are coding up something that is modal and overlays the content of the page.

apps/insights/src/components/MobileMenu/mobile-menu.v2.tsx
```diff
@@ -0,0 +1,61 @@
...
+                        <div className={styles.mobileMenuHandle} />
+                        <Button
+                            href="https://docs.pyth.network"
+                            size="md"
+                            rounded
+                            target="_blank"
+                        >
+                            Dev Docs
+                        </Button>
+                        <SupportDrawer>
```
cprussin, line 43: This is a bit strange, does this mean the support drawer will come in on top of the mobile overlay?  Seems like a weird UX to me, to stack overlays, or am I misunderstanding?

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
+@use "@pythnetwork/component-library/theme";
+
+.mobileMenuTrigger {
+    display: none;
+    @include theme.mobile {
+        display: block;
+    }
```
cprussin, line 7: We should flip these around (mobile-first)

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.mobileMenuOverlay {
+    background: rgba(0,0,0,0.4);
```
cprussin, line 11: `rgba` is effectively deprecated and should be replaced with `rgb` instead: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb

So this would be:

```scss
background: rgb(0 0 0 / 40%);
```

or (I prefer this):

```scss
background: rgb(from black r g b /40%)
```

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
...
+    position: fixed;
+    inset: 0;
+    z-index: 999;
```
cprussin, line 14: hmm I generally recommend against this kind of thing, using super high z-indexing generally shouldn't be necessary and if it is, it means something else is broken with stacking contexts and it will cause hard to debug issues elsewhere.

I suggest avoiding z-indexing if possible, but if not possible just set it to the minimum necessary value.  You can usually avoid z-indexing issues simply by setting stacking contexts correctly and explicitly (sometimes you need to use `isolate` to create a stacking context if you aren't implicitly creating one where one is needed.

If you see issues with stacking context elsewhere when using the minimum necessary z-indexing, it's best to try to resolve those issues early to avoid causing hard to debug problems down the line.

In this particular case, however, I think you can get away without having to use this entirely if you instead use the `ModalDialog` component as I mention below -- that component will render to a portal at the document root that is properly overlaid and it will avoid all issues with z-indexing entirely.

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.mobileMenuContainer {
+    border-top-left-radius: theme.border-radius("2xl");
+    border-top-right-radius: theme.border-radius("2xl");
+    background: theme.color("background", "modal");
+    position: absolute;
+    bottom: 0;
+    left: 0; 
+    right: 0;
+    padding: 1rem;
```
cprussin, line 25: Let's use `theme.spacing(4)` instead

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
...
+    display: flex;
+    flex-flow: column nowrap;
+    gap: theme.spacing(4);
+}
+
+.mobileMenuHandle {
+    background: theme.color("background", "secondary");
+    width: 33%;
+    height: 6px;
```
cprussin, line 34: We should avoid px values entirely and stick to `theme.spacing`

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
```diff
@@ -0,0 +1,53 @@
...
+    justify-content: space-between;
+    align-items: center;
+}
+
+.mobileThemeSwitcherFeedback {
+    display: flex;
+    flex-flow: row nowrap;
+    align-items: center;
+    gap: theme.spacing(3);
+    text-transform: capitalize;
```
cprussin, line 51: We shouldn't use `text-transform`, I know we aren't localizing but if we ever do, text-transforms aren't consistently accurate across locales.  For instance, in some languages, it is incorrect to have some characters capitalized in some contexts even if the rest of the text is all-caps.  It's also a completely unnecessary added amount of processing for the browser.

Instead, we should just write the string as it's meant to be rendered.  If you want all caps, just put all caps in the html.  If we ever go to localize, this will save us a ton of headache, and it will reduce unnecessary processing in the meantime.

apps/insights/src/components/Root/footer.tsx
```diff
@@ -21,10 +22,10 @@ export const Footer = () => (
         <div className={styles.divider} />
         <div className={styles.help}>
           <SupportDrawer>
-            <Link>Help</Link>
+            <Link>Support</Link>
           </SupportDrawer>
           <Link href="https://docs.pyth.network" target="_blank">
-            Documentation
+            Documentation <ArrowSquareOut />
```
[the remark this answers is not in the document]
partyparrotgreg: I think we could make this smarter by recognising if it's an external link and have a default styling with icon for this. This will happen quite often

apps/insights/src/components/MobileMenu/mobile-menu.v2.tsx
partyparrotgreg: I need to implement a way of closing a menu when opening support drawer

apps/insights/src/components/Root/header.tsx
```diff
@@ -25,22 +28,29 @@ export const Header = ({ className, ...props }: ComponentProps<"header">) => (
         <MainNavTabs />
       </div>
       <div className={styles.rightMenu}>
-        <SupportDrawer>
-          <Button beforeIcon={Lifebuoy} variant="ghost" size="sm" rounded>
-            Support
-          </Button>
-        </SupportDrawer>
+        <div className={styles.hideOnMobile}>
```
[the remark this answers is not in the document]
partyparrotgreg: yeah, but since I'm using the same component in two different places where one needs to be shown on desktop (top header), and the other one on mobile (menu drawer) they has to be visible on both breakpoints.

apps/insights/src/components/MobileMenu/mobile-menu.module.scss
partyparrotgreg: This one is to Capitalize the `theme` return (it returns `light, dark, system`), otherwise I wouldn't use it

cprussin: It's still best to map that in the jsx side, not the css

PR #2289, opened 2025-01-22
chore(lazer-protocol-js): Add exponent and pub count to payload

darunrs:


lazer/sdk/js/src/protocol.ts
```diff
@@ -26,6 +26,8 @@ export type ParsedFeedPayload = {
   price?: string | undefined;
   bestBidPrice?: string | undefined;
   bestAskPrice?: string | undefined;
+  publisher_count?: number | undefined;
```
keyvankhademi, line 29: I think the casing should be `publisherCount`

keyvankhademi (approved):

PR #2297, opened 2025-01-24
fix(apps/price_pusher): fix bug causing price_pusher to hand when invalid price feed ids are passed in to hermes ws

cctdaniel:


ali-bahjati (approved):
Thank you!

PR #2302, opened 2025-01-27
chore: downgrade g++ to fix librocksdb-sys build in CI

Riateche:
The rocksdb version we use in message buffer has a bug that prevents it from building on g++ 13. This PR sets up g++ 12 to fix the build. It also adds a missing libssl1.1 library.

### DIFF
--- a/.github/workflows/ci-message-buffer-idl.yml
+++ b/.github/workflows/ci-message-buffer-idl.yml
@@ -25,6 +25,8 @@ jobs:
           components: rustfmt, clippy
       - name: Install Solana
         run: |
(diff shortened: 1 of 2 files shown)


keyvankhademi (approved):

pythnet/message_buffer/Cargo.toml
```diff
@@ -11,3 +11,4 @@ codegen-units = 1
 opt-level = 3
 incremental = false
 codegen-units = 1
+#
```
keyvankhademi, line 14: dont forget to remove this before merging

cprussin (approved):
Nice, thank you!

PR #2304, opened 2025-01-28
feat: Remove Real Time Channel

darunrs:


lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -212,9 +208,7 @@ fn id_supports_all_fixed_rates() {
 }
 
 fn parse_channel(value: &str) -> Option<Channel> {
-    if value == "real_time" {
-        Some(Channel::RealTime)
-    } else if let Some(rest) = value.strip_prefix("fixed_rate@") {
```
keyvankhademi, line 217: isn't it better to map "real_time" to fixed rate 1ms here?

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -184,7 +182,6 @@ impl Serialize for Channel {
 mod channel_ids {
     use super::ChannelId;
 
-    pub const REAL_TIME: ChannelId = ChannelId(1);
     pub const FIXED_RATE_50: ChannelId = ChannelId(2);
     pub const FIXED_RATE_200: ChannelId = ChannelId(3);
     pub const FIXED_RATE_1: ChannelId = ChannelId(4);
```
keyvankhademi, line 187: can we keep the channel id to be 1? it help with db, ...

lazer/sdk/rust/protocol/src/router.rs
darunrs: Yeah we can do that, though it does bring up some questions. Given it is testing parsing of a channel name, where does this channel name come from in actual use cases? I actually don't know anywhere where we are storing the channel in string format.

lazer/sdk/rust/protocol/src/router.rs
darunrs: Yep I thought it might be good but wanted a second opinion on that.

lazer/sdk/rust/protocol/src/router.rs
darunrs: This is used in deserialization actually. Making this change allows us to support the history service APIs and latest price APIs still so this is good. Thanks for the suggestion!

keyvankhademi (approved):

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -173,7 +172,6 @@ impl Serialize for Channel {
         S: serde::Serializer,
     {
         match self {
-            Channel::RealTime => serializer.serialize_str("real_time"),
             Channel::FixedRate(fixed_rate) => {
                 serializer.serialize_str(&format!("fixed_rate@{}ms", fixed_rate.value_ms()))
             }
```
keyvankhademi, line 177: here fixed rate @ 1ms should serialize to "real_time".

keyvankhademi (approved):

PR #2312, opened 2025-01-30
chore(price-puhser) Use Hermes Client

aditya520:


apps/price_pusher/src/fuel/command.ts
```diff
@@ -46,28 +45,33 @@ export default {
...
+    if (invalidPriceItems.length > 0) {
+      logger.error(
+        `Invalid price id submitted for: ${invalidPriceItems
+          .map(({ alias }) => alias)
+          .join(", ")}`
+      );
+    }
 
-    const priceItems = priceConfigs.map(({ id, alias }) => ({ id, alias }));
+    priceItems = existingPriceItems;
```
ali-bahjati, line 71: I recommend you to do it in the PythPriceListener to avoid replicating the logic on each network

apps/price_pusher/src/pyth-price-listener.ts
```diff
@@ -34,108 +34,47 @@ export class PythPriceListener implements IPriceListener {
...
-              publishTime: latestAvailablePrice.publishTime,
-            });
-          });
-        }
-      }
-    }
+    eventSource.onerror = (error: Event) => {
+      console.error("Error receiving updates from Hermes:", error);
+      eventSource.close();
+    };
```
ali-bahjati, line 77: i think here you should either crash the service, or try to reconnect. doing otherwise will result in one dead component while the rest are working.

ali-bahjati (commented):
This is a very good change. Thank you. Can you double check whether pusher keeps working or crashes upon network failures on Hermes? (look at my comment about it). A good way to test is to run it, turn off internet and wait a little while and turn it on again. A good behaviour is that it either crashes when you turn it off, or it doesn't crash but keeps working after you turn the internet on. anything else is a silent dead state.

ali-bahjati (approved):

apps/price_pusher/src/fuel/command.ts
aditya520: Either we filter pricefeeds at `src/index.ts` or at every chain level since we pass price ids to both chain listener and price listener.

apps/price_pusher/src/pyth-price-listener.ts
aditya520: It's better it crashes when the connection goes down. Just tested it.

PR #2314, opened 2025-01-30
feat(lazer-protocol): Update EVM payload

darunrs:


lazer/sdk/js/src/protocol.ts
```diff
@@ -7,7 +7,7 @@ export type PriceFeedProperty =
   | "bestAskPrice"
   | "exponent"
   | "publisherCount";
-export type Channel = "real_time" | "fixed_rate@50ms" | "fixed_rate@200ms";
+export type Channel = "fixed_rate@1ms" | "fixed_rate@50ms" | "fixed_rate@200ms";
```
keyvankhademi, line 10: I think this should remain `real_time` because that will be serialized to a string when passed to WS

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -74,6 +76,24 @@ library PythLazerLib {
...
+    }
+
+    function parseFeedValueInt16(
+        bytes calldata update,
+        uint16 pos
+    ) public pure returns (int16 value, uint16 new_pos) {
+        value = int16(uint16(bytes2(update[pos:pos + 2])));
+        pos += 2;
+        new_pos = pos;
+    }
```
keyvankhademi, line 95: where are these used?

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -8,12 +8,14 @@ library PythLazerLib {
...
-        BestAskPrice
+        BestAskPrice,
+        Exponent,
+        PublisherCount
     }
 
     enum Channel {
         Invalid,
-        RealTime,
+        FixedRate1,
```
keyvankhademi, line 18: is this consumer facing? maybe we should keep it as RealTime no?

lazer/contracts/evm/src/PythLazerLib.sol
keyvankhademi, line 95: This function is confusing, a parseInt should read int by default.

lazer/sdk/js/src/protocol.ts
darunrs: Hmmm yeah I suppose we can leave it.

lazer/contracts/evm/src/PythLazerLib.sol
darunrs: These are used in this PR: https://github.com/pyth-network/pyth-examples/pull/46

I need to be able to pull these changes in from foundry to validate. I guess I can see if there's a way to use them early to make sure they work as expected.

keyvankhademi: can we use them in test?

darunrs: I couldn't find another way to parse a int16 easily. You cannot feed it a 2 byte array apparently. It fails if you do.

keyvankhademi: @Riateche Any idea?

ali-bahjati: what is exactly confusing here :? this `int(uint(...))` is common in solidity as there's no other way to do the parsing.

ali-bahjati (approved):

lazer/contracts/evm/src/PythLazerLib.sol
keyvankhademi: The confusing part is that this function is actually reading `UInt` from the bytes which has a different binary encoding than `Int`. But I don't really know how Solidity's cast works, like what will happen if the number is negative? The Unit will be a large number and casting will correctly change it back to negative?

lazer/sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/pyth-lazer-sdk",
-  "version": "0.3.2",
+  "version": "0.4.0",
```
keyvankhademi, line 3: I think you should revert this back too.

lazer/contracts/evm/src/PythLazerLib.sol
darunrs: They are used here: https://github.com/pyth-network/pyth-examples/pull/46

darunrs: Yes, because the UInt just reads the bytes as is, and converting to Int just reinterprets the number into 2's complement. We don't actually change the underlying data in any way. I'm not sure why I can't use Int16 with a byte array though. I've verified in test that this works.

keyvankhademi (approved):

lazer/contracts/evm/src/PythLazerLib.sol
darunrs: Kept real time

darunrs:
Contract deployed

PR #2315, opened 2025-01-31
fix: use key with sol for simulations

guibescos:
Users that don't have SOL in their wallet can't see their claimable rewards.
I want to fix this by running the simulations with a fixed key that always has SOL, this is possible because claiming rewards for a user is actually permissionsless.

### DIFF
--- a/apps/staking/src/config/server.ts
+++ b/apps/staking/src/config/server.ts
@@ -76,7 +76,7 @@ export const GOVERNANCE_ONLY_REGIONS = transformOr(
 export const PROXYCHECK_API_KEY = demandInProduction("PROXYCHECK_API_KEY");
 // This needs to be a public key that has SOL in it all the time, it will be used as a payer in the transaction simulation to compute the claimable rewards
 // such simulation fails when the payer has no funds.
-export const SIMULATION_PAYER = process.env.SIMULATION_PAYER;
+export const SIMULATION_PAYER = getOr("SIMULATION_PAYER", "E5KR7yfb9UyVB6ZhmhQki1rM1eBcxHvyGKFZakAC5uc");
 class MissingEnvironmentError extends Error {
   constructor(name: string) {
     super(`Missing environment variable: ${name}!`);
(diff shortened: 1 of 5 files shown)


apps/staking/src/hooks/use-api.tsx
```diff
@@ -95,7 +96,7 @@ const State = {
       dashboardDataCacheKey,
 
       loadData: () =>
-        api.loadData(client, pythnetClient, hermesClient, account),
+        api.loadData(client, pythnetClient, hermesClient, account, simulationPayer ? new PublicKey(simulationPayer) : undefined),
```
guibescos, line 99: I'm not sure where to put the string to public key conversion.
Ideally the deployment should fail if SIMULATION_PAYER is an invalid base58 pubkey

apps/staking/src/config/server.ts
```diff
@@ -74,7 +74,9 @@ export const GOVERNANCE_ONLY_REGIONS = transformOr(
   [],
 );
 export const PROXYCHECK_API_KEY = demandInProduction("PROXYCHECK_API_KEY");
-
+// This needs to be a public key that has SOL in it all the time, it will be used as a payer in the transaction simulation to compute the claimable rewards
+// such simulation fails when the payer has no funds.
+export const SIMULATION_PAYER = process.env.SIMULATION_PAYER;
```
guibescos, line 79: I doesn't seem to be working in the vercel preview and I can't quite figure out why

cprussin: What's failing exactly?  Is the expected key not being used?

Also, is this key safe to reveal publicly and hardcode into the codebase?  I assume so since it will be serialized to the browser and it's a public key; if so it's better to do this as `getOr("SIMULATION_PAYER", "<default value>")` and leave it unset in Vercel; ideally we would keep the environment as minimal as possible and only set things when we need a quick override or for secrets

cprussin (approved):

apps/staking/src/config/server.ts
guibescos: sure

PR #2323, opened 2025-02-03
chore(lazer): add crate publish workflow for pyth-lazer-client

tejasbadadare:


bplatak (approved):

darunrs (approved):
Strange that it was deploying solana before when its called rust sdk. Do we need to add another publish workflow for solana still? Any way we can test that this works?

keyvankhademi (approved):

tejasbadadare:
> Strange that it was deploying solana before when its called rust sdk. Do we need to add another publish workflow for solana still? Any way we can test that this works?

There looks to be a new one `publish-rust-lazer-contract.yml` that deploys the solana contract

PR #2324, opened 2025-02-04
refactor(governance/xc_admin): add more logs on relayer

ali-bahjati:
If the relayer fails for whatever reason it is hard to debug what is going on. This change adds more logs to see the details of the proposals which is quite helpful in debugging the issues.

### DIFF
--- a/governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
+++ b/governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
@@ -15,7 +15,6 @@ import {
   Connection,
   Keypair,
   PublicKey,
-  SystemProgram,
   TransactionInstruction,
 } from "@solana/web3.js";
 import * as fs from "fs";
@@ -32,9 +31,6 @@ import {
   REMOTE_EXECUTOR_ADDRESS,
   envOrErr,
   PriceStoreMultisigInstruction,
-  findDetermisticPublisherBufferAddress,
-  PRICE_STORE_BUFFER_SPACE,
-  PRICE_STORE_PROGRAM_ID,
   createDeterministicPublisherBufferAccountInstruction,
 } from "@pythnetwork/xc-admin-common";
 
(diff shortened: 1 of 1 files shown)


governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
```diff
@@ -200,7 +203,7 @@ async function run() {
             .postInstructions([
               ComputeBudgetProgram.setComputeUnitLimit({ units: 1000000 }),
             ])
-            .rpc({ skipPreflight: true });
+            .rpc({ skipPreflight: false });
```
ali-bahjati, line 206: I think we get better logs with simulation. Why was it true before?

governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
```diff
@@ -101,16 +97,19 @@ async function run() {
...
+        console.log(`Found VAA ${lastSequenceNumber}, relaying vaa ...`);
+
         await postVaaSolana(
           provider.connection,
           signTransactionFactory(KEYPAIR),
           WORMHOLE_ADDRESS[CLUSTER]!,
           provider.wallet.publicKey,
           Buffer.from(response.vaaBytes, "base64"),
-          { commitment: COMMITMENT }
+          { commitment: COMMITMENT, maxRetries: 10 }
```
ali-bahjati, line 108: I couldn't see what the default is. It was often failing at this step.

governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
```diff
@@ -15,7 +15,6 @@ import {
   Connection,
   Keypair,
   PublicKey,
-  SystemProgram,
```
ali-bahjati, line 18: Unused imports

governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
guibescos: meeeh, people in Solana recommend 0 retries at this level
`https://github.com/rpcpool/optimized-txs-examples/blob/main/selfSolTransfer.mjs#L116`
instead if you'd like to retry you should put `postVaaSolana` inside

governance/xc_admin/packages/crank_pythnet_relayer/src/index.ts
guibescos: I normally got the logs from explorer, but it might be nicer to get them from Grafana

guibescos (approved):
I don't think `maxRetries` should be used

PR #2327, opened 2025-02-04
fix(pyth-lazer-protocol): deser Channel from String, not &str

tejasbadadare:
The current deserializer for the Channel type fails to work with string values loaded from toml by [config](https://crates.io/crates/config) -- the config crate is trying to deserialize an owned `String` but the deserializer expects a borrowed `&str`, causing this error: 
```
invalid type: string "fixed_rate@50ms", expected a borrowed string for key lazer.subscriptions[0]channel" 
```

Fixed by making the deserializer accept both `String` and `&str` (Serde will coerce a `&str` into a `String`).
Also threw in a minor clippy fix (`dev_dependencies` -> `dev-dependencies`).

### DIFF
--- a/lazer/sdk/rust/client/Cargo.toml
+++ b/lazer/sdk/rust/client/Cargo.toml
@@ -17,7 +17,7 @@ anyhow = "1.0"
 tracing = "0.1"
 url = "2.4"
 
-[dev_dependencies]
+[dev-dependencies]
 bincode = "1.3.3"
 ed25519-dalek = { version = "2.1.1", features = ["rand_core"] }
 hex = "0.4.3"
--- a/lazer/Cargo.lock
+++ b/lazer/Cargo.lock
@@ -3771,7 +3771,7 @@ dependencies = [
  "futures-util",
  "hex",
  "libsecp256k1 0.7.1",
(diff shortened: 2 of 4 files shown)


lazer/sdk/rust/client/Cargo.toml
```diff
@@ -17,7 +17,7 @@ anyhow = "1.0"
 tracing = "0.1"
 url = "2.4"
 
-[dev_dependencies]
```
keyvankhademi, line 20: what is changed here?

keyvankhademi: was it not working at all? or both works?

keyvankhademi (approved):

lazer/sdk/rust/client/Cargo.toml
tejasbadadare: both work, but this format is deprecated:
```
`dev_dependencies` is deprecated in favor of `dev-dependencies` and will not work in the 2024 edition
```

tejasbadadare: it was causing clippy pre-commit step to fail

PR #2333, opened 2025-02-05
feat(pyth-lazer-protocol): add confidence to aggregated data

keyvankhademi:


darunrs (approved):
Could you also add the field to the JS protocol and bump that too?

keyvankhademi:
> Could you also add the field to the JS protocol and bump that too?

Let's deploy this first so we can test the JS stuff. I have to add this to multiple places.

PR #2339, opened 2025-02-06
feat(target_chains/near): add reproducible build

Riateche:


target_chains/near/receiver/Cargo.toml
```diff
@@ -47,3 +49,9 @@ lto                        = "fat"
 debug                      = false
 panic                      = "abort"
 overflow-checks            = true
+
+[package.metadata.near.reproducible_build]
+image                   = "sourcescan/cargo-near:0.13.3-rust-1.84.1"
+image_digest            = "sha256:baa712c5d2b7522d38175e36330d336ad2c4ce32bfaaa41af94ce40407ecd803"
+passed_env              = []
+container_build_command = ["cargo", "near", "build", "non-reproducible-wasm", "--locked"]
```
ali-bahjati, line 57: such a strange thing: the reproducible build runs `non-reproducible-wasm` build command lol

ali-bahjati (approved):
Nice!

Maybe update the readme or build scripts to use this as well?

target_chains/near/receiver/Cargo.toml
Riateche: I guess the difference between them is that reproducible-wasm spins up a docker container, so it makes sense that it's not needed inside the container.

Riateche:
I updated the readme and removed a stale deploy script (now it's just a single command that's in the readme). I'll update the upgrade scripts later.

PR #2340, opened 2025-02-06
fix: better error messages for big instructions

guibescos:


governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
```diff
@@ -287,306 +303,317 @@ const General = ({ proposerServerUrl }: { proposerServerUrl: string }) => {
...
-        const fundingAccount = isRemote
-          ? mapKey(multisigAuthority)
-          : multisigAuthority
-
-        const initPublisherInPriceStore = async (publisherKey: PublicKey) => {
-          // Ignore this step if Price Store is not initialized (or not deployed)
-          if (!connection || !(await isPriceStoreInitialized(connection))) {
-            return
-          }
+    const handleSendProposalButtonClickAsync = async () => {
```
guibescos, line 306: this is just a copy paste refactor so we catch the promise error

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
```diff
@@ -36,6 +36,22 @@ import PermissionDepermissionKey from '../PermissionDepermissionKey'
 import { PriceRawConfig } from '../../hooks/usePyth'
 import { Wallet } from '@coral-xyz/anchor/dist/cjs/provider'
 
+const MAX_SIZE_ADD_PRODUCT_INSTRUCTION_DATA = 369
+const MAX_SIZE_UPD_PRODUCT_INSTRUCTION_DATA = 403 // upd product has one account less
```
cprussin, line 40: Where do these numbers come from?  Are they just randomly selected?  Is there some calculation that leads to them?

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
cprussin: You should remove the `async` marker from the outer function since it's no longer async.

Also, I generally recommend hoisting this kind of thing to the module scope and passing in arguments.  For one, it avoids the perf overhead of creating closures at runtime.  For two, it makes the code much easier to follow and understand.  For three, explicitly passing arguments rather than relying on closure scoped variables makes the information flow much more obvious.

This function is already gigantic and really hard to trace and understand its flow, breaking it up outside the component would help a lot.

cprussin (approved):
Small stylistic nits but up to you if you think they're worth the effort

ali-bahjati (approved):

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
```diff
@@ -36,6 +36,22 @@ import PermissionDepermissionKey from '../PermissionDepermissionKey'
...
+
+const checkSizeOfProductInstruction = (
+  instruction: TransactionInstruction,
+  maxSize: number,
+  symbol: string
+) => {
+  const size = instruction.data.length
+  if (size > maxSize) {
+    throw new Error(
```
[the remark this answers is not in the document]
guibescos: won't spend time refactoring this

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
guibescos: added a comment

governance/xc_admin/packages/xc_admin_frontend/components/tabs/General.tsx
guibescos: removed the async keywork, but won't refactor since it's too much work

PR #2341, opened 2025-02-06
bump: pyth client to 2.22.1

guibescos:


cprussin (approved):
We should really just move this into the monorepo soon

PR #2342, opened 2025-02-07
fix: update corepack on js dockerfile

ali-bahjati:
The existing corepack version in the base node image that we have has some problems (on its keys) for installation. See https://github.com/nodejs/corepack/issues/612 for more details.


cprussin (approved):

cprussin:
Nice find!

PR #2343, opened 2025-02-07
feat(governance/xc_admin): add support to rename symbol name

ali-bahjati:
This change also sets the max latency upon new feed creation.

I tested the changes all manually to make sure they generate correct instructions.


guibescos (commented):
I think a simpler workflow is:
- download the json
- edit the symbol field in the metadata
- upload the json

ali-bahjati:
closing in favor of #2443, taking @guibescos's suggested approach which was waay simpler.

PR #2372, opened 2025-02-12
(chore): Fortuna refactor

jayantk:
Pure move PR that just moves logic around to help isolate reusable ethereum transaction logic. I made a new eth_utils module to hold the reusable stuff. I also split out the contents of the keeper file into multiple files in the new keeper module. 

There is still some refactoring to do around process_event and the escalator policy and such, but i will do that separately. The modules can also be renamed, but again i will address later.


apps/fortuna/src/eth_utils/utils.rs
```diff
@@ -0,0 +1,66 @@
+use {
+    anyhow::{anyhow, Result},
+    ethers::{contract::ContractCall, middleware::Middleware},
+    std::sync::Arc,
+    tracing,
+};
+
+pub async fn send_and_confirm<A: Middleware>(contract_call: ContractCall<A, ()>) -> Result<()> {
```
jayantk, line 8: Note: the function signatures of this function and the one below changed very slightly so that they don't depend on the pyth-specific type declarations in the chain/ module.

m30m (approved):

PR #2375, opened 2025-02-12
add test_only new method into sui contract

reditable:


ali-bahjati (approved):
Thank you for your contribution!

PR #2381, opened 2025-02-14
feat: init Lazer Aptos contract

tejasbadadare:
First pass at the Aptos Lazer contract. Keeps the same API as the Solana and EVM contracts. 
Uses Ed25519 signatures as recommended by Aptos docs.

Deployed to devnet [here](https://explorer.aptoslabs.com/account/0x8731685005cfb169b4da4bbfab0c91c5ba59508bbd6d26990ee2be7225cb34d1).
Example transaction of initializing the contract [here](https://explorer.aptoslabs.com/txn/13808807?network=devnet).


apps/hermes/server/Cargo.lock
keyvankhademi, line 1: this should not be in the PR?

lazer/contracts/aptos/Move.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+license = "UNLICENSED"
+
+[dependencies.AptosFramework]
+git = "https://github.com/aptos-labs/aptos-framework.git"
+rev = "mainnet"
+subdir = "aptos-framework"
+
+[addresses]
+pyth_lazer = "0x8731685005cfb169b4da4bbfab0c91c5ba59508bbd6d26990ee2be7225cb34d1" # Temporary key during development
```
keyvankhademi, line 12: add this to dev addresses maybe?

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Constants
+    const MAX_NUM_TRUSTED_SIGNERS: u8 = 2;
+    const ED25519_PUBLIC_KEY_LENGTH: u64 = 32;
+
+    /// Stores information about a trusted signer including their public key and expiration
+    struct TrustedSignerInfo has store, drop {
+        pubkey: vector<u8>,  // Ed25519 public key (32 bytes)
+        expires_at: u64,     // Unix timestamp
+    }
```
keyvankhademi, line 25: ahh i don't know much about aptos, i thought `drop` was only for object that are not stored in the global storage?

keyvankhademi: is it because it's only used inside another struct? and not being stored separately?

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+        treasury: address,
+    ) {
+        let storage = Storage {
+            top_authority,
+            treasury,
+            single_update_fee: 1, // Nominal fee
+            num_trusted_signers: 0,
+            trusted_signers: vector::empty(),
+        };
+        move_to(account, storage);
```
keyvankhademi, line 55: shouldn't we move this to the `top_authority` address?

keyvankhademi: I think this way any new signer can call initialize and create a new Storage?

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+    /// Error codes
+    const ENO_PERMISSIONS: u64 = 1;
+    const EINVALID_SIGNER: u64 = 2;
+    const ENO_SPACE: u64 = 3;
+    const ENO_SUCH_PUBKEY: u64 = 4;
+    const EINVALID_SIGNATURE: u64 = 5;
+    const EINSUFFICIENT_FEE: u64 = 6;
+
+    /// Constants
+    const MAX_NUM_TRUSTED_SIGNERS: u8 = 2;
```
keyvankhademi, line 18: let's actually remove this MAX. I don't think we should've had it in the solana contract either.

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: drop is needed for any type that can be discarded, we need it here because it's stored in a vector that might remove elements (when removing a trusted signer)

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: the `move_to` operation will fail if Storage already exists at the account's address, so it's effectively a one-time initialization. looks to be a pattern in aptos contracts: https://aptos.dev/en/build/guides/build-e2e-dapp/1-create-smart-contract#create-list-function

lazer/contracts/aptos/Move.toml
tejasbadadare: sure yeah

lazer/contracts/aptos/README.md
```diff
@@ -0,0 +1,45 @@
...
+This package is built using the Move language and Aptos framework.
+
+`PythLazer` is an Aptos contract that allows consumers to easily verify Pyth Lazer updates for use on-chain.
+
+### Build, test, deploy
+
+Install Aptos CLI and set it up:
+
+```shell
+$ brew install
```
m30m, line 12: brew install what?

lazer/contracts/aptos/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+This package is built using the Move language and Aptos framework.
+
+`PythLazer` is an Aptos contract that allows consumers to easily verify Pyth Lazer updates for use on-chain.
+
+### Build, test, deploy
+
+Install Aptos CLI and set it up:
+
+```shell
```
m30m, line 11: I think we can remove these dollar signs from all these scripts, it just makes it harder to copy paste

lazer/contracts/aptos/README.md
```diff
@@ -0,0 +1,45 @@
...
+$ aptos move publish
+```
+
+Invoke deployed contract functions on-chain:
+
+```shell
+aptos move run --function-id 'default::pyth_lazer::update_trusted_signer' --args 'hex:0x8731685005cfb169b4da4bbfab0c91c5ba59508bbd6d26990ee2be7225cb34d1' 'u64:9999999999'
+```
+
+### Error Handling
```
m30m, line 36: I think these belong to docs.pyth.network

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+            storage.num_trusted_signers = storage.num_trusted_signers + 1;
+        };
+    }
+
+    /// Verify a message signature and collect fee
+    public entry fun verify_message(
+        account: &signer,
+        message: vector<u8>,
+        signature: vector<u8>,
+        public_key: vector<u8>,
```
m30m, line 106: maybe we can also call this `trusted_signer`?

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+        pubkey: vector<u8>,  // Ed25519 public key (32 bytes)
+        expires_at: u64,     // Unix timestamp
+    }
+
+    /// Main storage for the Lazer contract
+    struct Storage has key {
+        top_authority: address,
+        treasury: address,
+        single_update_fee: u64,
+        num_trusted_signers: u8,
```
m30m, line 32: why do we need this?
can't we use `vector::length`?

lazer/contracts/aptos/sources/pyth_lazer.move
m30m: I'm also uncertain here. Shouldn't we move this to `@pyth_lazer`?

lazer/contracts/aptos/tests/pyth_lazer_tests.move
```diff
@@ -0,0 +1,148 @@
...
+        timestamp::set_time_has_started_for_testing(&framework);
+
+        // Initialize contract
+        pyth_lazer::initialize(&lazer_contract, TOP_AUTHORITY, TREASURY);
+
+        (top_authority, treasury, user)
+    }
+
+    #[test]
+    fun test_initialize() {
```
m30m, line 66: do we need this?

lazer/contracts/aptos/tests/pyth_lazer_tests.move
```diff
@@ -0,0 +1,148 @@
...
+
+        // This should succeed as we have a valid signer and sufficient fee
+        pyth_lazer::verify_message(&user, TEST_MESSAGE, TEST_SIGNATURE, TEST_PUBKEY);
+    }
+
+    #[test]
+    fun test_update_add_signer() {
+        let (top_authority, _treasury, _) = setup();
+
+        // Add signer
```
m30m, line 92: we are not checking the state after each function call to assert it's actually updated completely

lazer/contracts/aptos/tests/pyth_lazer_tests.move
```diff
@@ -0,0 +1,148 @@
...
+
+    #[test]
+    fun test_verify_message_success() {
+        let (top_authority, _treasury, user) = setup();
+
+        // Add a valid signer
+        let expires_at = timestamp::now_seconds() + 1000;
+        pyth_lazer::update_trusted_signer(&top_authority, TEST_PUBKEY, expires_at);
+
+        // Create a valid ed25519 signature
```
m30m, line 79: I'm not sure why we have this section here. I guess we need to assume `TEST_SIGNATURE`, `TEST_MESSAGE`, and `TEST_PUBKEY` are correct (some sort of fixture) and call the verify_message. No need to assert it here again.

lazer/contracts/aptos/tests/pyth_lazer_tests.move
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    #[test]
+    fun test_initialize() {
+        let (_top_authority, _treasury, _) = setup();
+        // Contract is already initialized in setup
+    }
+
+    #[test]
+    fun test_verify_message_success() {
```
m30m, line 72: we should also have a test for `EINVALID_SIGNATURE`

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,132 @@
...
+        trusted_signers: vector<TrustedSignerInfo>,
+    }
+
+    /// Events
+    struct TrustedSignerUpdateEvent has drop, store {
```
m30m, line 37: I guess this is going to be used later

lazer/contracts/aptos/tests/pyth_lazer_tests.move
tejasbadadare: yeah ran into some difficulties here around ownership of the Storage resource... seems like i might need to create a getter function that borrows Storage and exposes it. do you happen to know any examples that demonstrate this kind of update-then-verify test pattern?

lazer/contracts/aptos/tests/pyth_lazer_tests.move
tejasbadadare: setup() is used in all tests so no, this was a remnant of sanity testing that setup() actually works

lazer/contracts/aptos/README.md
tejasbadadare: is there a repo somewhere where i can update those docs?

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: my assumption here was that only the contract's wallet can invoke this function, making account == @pyth_lazer. i could be wrong though, let me dig into this some more

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: yeah, this was a remnant of the solana implementation, we can remove

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: seemed useful for others to monitor which signers are trusted, but no real plans yet

lazer/contracts/aptos/tests/pyth_lazer_tests.move
tejasbadadare: fair enough yep, removed the extra assertion

lazer/contracts/aptos/tests/pyth_lazer_tests.move
tejasbadadare: alright, looks like i needed to add some getters to expose the private struct. updated the tests to actually check the state after each call

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: okay checked with the Aptos dev group on telegram. moving to `@pyth_lazer` doesn't work since it expects a `&signer`, but adding a check for account == @pyth_lazer serves our purpose here. this makes it so that initialize can only be called by the contract wallet, and can't be called multiple times. docs on `move_to`: https://aptos.dev/en/build/smart-contracts/book/global-storage-structure

lazer/contracts/aptos/README.md
tejasbadadare: just... brew install, man

lazer/contracts/aptos/README.md
m30m: https://github.com/pyth-network/documentation

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Stores information about a trusted signer including their public key and expiration
+    struct TrustedSignerInfo has store, drop, copy {
+        pubkey: vector<u8>, // Ed25519 public key (32 bytes)
+        expires_at: u64 // Unix timestamp
+    }
+
+    /// Main storage for the Lazer contract
+    struct Storage has key {
+        top_authority: address,
```
ali-bahjati, line 27: This is not the Move way of handling ownership. I recommend you create an `AdminCapability` and the holder of that capability can perform admin actions.

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,152 @@
...
+    ) acquires Storage {
+        let storage = borrow_global<Storage>(@pyth_lazer);
+
+        // Verify fee payment
+        assert!(
+            coin::balance<AptosCoin>(signer::address_of(account))
+                >= storage.single_update_fee,
+            EINSUFFICIENT_FEE
+        );
+        coin::transfer<AptosCoin>(account, storage.treasury, storage.single_update_fee);
```
ali-bahjati, line 71: This method has a lot of access to account. The move way here is to pass coin to this function and later deposit it to the contract. Alternatively you can wrap it with but i like to stick to the idiomatic move ways.

lazer/contracts/aptos/sources/pyth_lazer.move
ali-bahjati: you'll get around these once you use deployer contract.

ali-bahjati (commented):
You'd need to use the Wormhole's deployer to make sure you can upgrade the contract later by itself (DAO)

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: Thanks for the pointer, will add

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: mirrored the pattern from the pyth core aptos contract, and wrapped `verify_message` with `verify_message_with_funder`

tejasbadadare:
I will handle governance actions in the next PR to keep this one from getting too unwieldy (initializing using the Wormhole deployer, setting up a governance module that owns the lazer SignerCapability, and handling governance instructions such as contract upgrades)

lazer/contracts/aptos/sources/pyth_lazer.move
```diff
@@ -0,0 +1,206 @@
...
+
+        // Can only be called once. If storage already exists in @pyth_lazer,
+        // this operation will fail (one-time initialization).
+        move_to(account, storage);
+    }
+
+    /// Allows the designated admin to claim their capability
+    public entry fun claim_admin_capability(account: &signer) acquires PendingAdminCapability {
+        let pending = borrow_global<PendingAdminCapability>(@pyth_lazer);
+        assert!(signer::address_of(account) == pending.admin, ENO_PERMISSIONS);
```
ali-bahjati, line 69: we need to invalidate the admin capability right? otherwise it seems we can mint infinite capabilities.

more for my edification: could the @pyth get admincapability itself in `initialize` and then later transfer it to someone else :?

ali-bahjati (approved):
Thank you!

keyvankhademi (approved):

lazer/contracts/aptos/sources/pyth_lazer.move
tejasbadadare: > we need to invalidate the admin capability right? otherwise it seems we can mint infinite capabilities.

multiple claim attempts by the `admin` will fail, since there can only be a single resource of a given type under an account, and since the capabilities don't have `copy` there shouldn't be a way to distribute copies of the capability. i thought about explicitly destroying the PendingAdminCapability after it has been claimed, but thought it might be misleading since move already guarantees it can't be double-claimed. let me add a comment clarifying this 

> more for my edification: could the @pyth get admincapability itself in initialize and then later transfer it to someone else :?

and get rid of PendingAdminCapability? yeah i think that would work, but i was following the wormhole deployer's pattern of a pending capability. i'll dig into this

ali-bahjati: I looked it up and it seems that in Aptos a non-contract cannot pass it's owned objects to the functions and therefore i don't think they can transfer an object. I'm not sure but i saw this pattern a lot that there's an entrypoint for signer everywhere and then objects are collected (like what you do now) which partially defeats the purpose of this language construct 🤷 SUI handles it properly though.

PR #2393, opened 2025-02-18
feat(contract_manager): add support for near contract upgrade

Riateche:
- Add propose upgrade script for near
- Manually track governance source and sequence to allow use of `sync_governance_vaas` script
- Remove near from `RECEIVER_CHAINS` because it's already in `WORMHOLE_CHAINS`

### DIFF
--- a/contract_manager/src/contracts/near.ts
+++ b/contract_manager/src/contracts/near.ts
@@ -88,7 +88,12 @@ export class NearWormholeContract extends WormholeContract {
 export class NearPriceFeedContract extends PriceFeedContract {
   public static type = "NearPriceFeedContract";
 
-  constructor(public chain: NearChain, public address: string) {
+  constructor(
+    public chain: NearChain,
+    public address: string,
+    public governanceDataSource: DataSource,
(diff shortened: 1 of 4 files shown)


contract_manager/src/contracts/near.ts
```diff
@@ -108,21 +113,38 @@ export class NearPriceFeedContract extends PriceFeedContract {
...
 
   static fromJson(
     chain: Chain,
-    parsed: { type: string; address: string }
+    parsed: {
+      type: string;
+      address: string;
+      governanceDataSourceAddress: string;
+      governanceDataSourceChain: number;
+      lastExecutedGovernanceSequence: number;
```
ali-bahjati, line 130: I wonder where you are using these 🤔

ali-bahjati (approved):
Nice.

contract_manager/src/contracts/near.ts
Riateche: They are used in the corresponding methods of `NearPriceFeedContract`.

PR #2405, opened 2025-02-21
refactor(contract_manager): support batch set fee proposal

ali-bahjati:
Update the set fee proposal generation script to support batches of set fee requests in a single proposal.


contract_manager/scripts/generate_governance_set_fee_payload.ts
```diff
@@ -1,41 +1,66 @@
...
 
 async function main() {
-  const { chain, fee, exponent } = await parser.argv;
+  const {
+    config,
+    "ops-key-path": ops_key_path,
+    vault: vault_id,
+  } = await parser.argv;
+
+  const config_obj = parse(readFileSync(config, "utf8"));
```
m30m, line 34: Not sure about the `_obj` suffix. Maybe change `config` to `config_path` and this just becomes `config`?
Same for `chain_obj`.

contract_manager/scripts/generate_governance_set_fee_payload.ts
```diff
@@ -1,41 +1,66 @@
...
+
+  let update_payloads: Buffer[] = [];
+  for (const chain of config_obj) {
+    const chain_obj = DefaultStore.chains[chain.name];
```
m30m, line 38: can we create a function in the Store to get the chain or throw? I see this pattern a lot

m30m (approved):
looks good, please fix the github actions complaint about using const too.

PR #2406, opened 2025-02-21
chore: Update cache saving policy for github actions

jayantk:
At the moment, pre-commit randomly takes the full 25 mins because it can't find the cache entry for the Rust build. I think what's happening is that we are evicting the `main` branch cache sometimes due to the Github cache size limit. (see inline comment for more details)


cprussin (approved):
can we buy more cache @jayantk ?  seems like something we're likely going to run into more if the limit is so close to what we're already using

PR #2413, opened 2025-02-25
fix(target_chains/ton): soft throw instead of revert

cctdaniel:


ali-bahjati (approved):
LGTM. I recall you had some comments somewhere in the code saying we don't handle errors; if such things exists maybe you can update it too?

PR #2414, opened 2025-02-25
feat(lazer): add new formats to protocol

Riateche:
- Reorganize magics in different modules
- Declare all magics in LE and add tests for BE representation
- Add LeEcdsa format for EVM signatures on Solana
- Add LeUnsigned format for off-chain usage
- Add `Message` enum for ser/deserializing any message type.
- Add `BinaryWsUpdate` for ser/deserializing binary WS updates.
- Update WS client example.
- Properly support binary updates in WS client: it no longer converts payloads to base64 and back. 
- Rename "chains" to "formats" in router API because now it contains off-chain formats. Old name is still supported for compatibility.

I wanted to deprecate `parsed` in router API and replace it with `formats: ["json"]` but unfortunately `parsed` is true by default to we can't easily deprecate it. We can remove it when we do a breaking change and go for a `/v2/..` endpoint.

Tested WS client against a prod router to verify no breaking changes. Tested in tilt for new formats.


lazer/sdk/rust/protocol/src/message.rs
```diff
@@ -15,7 +77,7 @@ pub struct EvmMessage {
 
 impl EvmMessage {
     pub fn serialize(&self, mut writer: impl Write) -> anyhow::Result<()> {
-        writer.write_u32::<BE>(EVM_FORMAT_MAGIC)?;
+        writer.write_u32::<LE>(EVM_FORMAT_MAGIC)?;
```
keyvankhademi, line 80: isn't this a breaking change? How does the current contracts in EVM that expect BE work with this?

keyvankhademi (approved):

lazer/sdk/rust/client/Cargo.toml
```diff
@@ -6,7 +6,7 @@ description = "A Rust client for Pyth Lazer"
 license = "Apache-2.0"
 
 [dependencies]
-pyth-lazer-protocol = "0.5.0"
+pyth-lazer-protocol = { path = "../protocol", version = "0.5.0" }
```
keyvankhademi, line 9: doesn't this break the publishing to cargo?

keyvankhademi:
I think you should increase the protocol version too.

lazer/sdk/rust/protocol/src/message.rs
Riateche: It's not a breaking change because I changed the constant value.

lazer/sdk/rust/client/Cargo.toml
Riateche: No, cargo is smart enough to remove paths on publish.

Riateche:
Bumped protocol version.

PR #2417, opened 2025-02-26
feat(apps/argus): init argus

cctdaniel:
## Summary

This PR creates a new Pulse contract interface by adapting the structure from Fortuna (which was written for Entropy), implements new storage query methods, integrates Hermes API for price updates, and tests individual command functionality.

## Rationale

These changes are necessary to:
1. Create a dedicated Pulse contract interface based on the existing Fortuna/Entropy structure but adapted for Pulse's specific requirements
2. Implement efficient storage query methods for retrieving active requests directly from contract storage
3. Integrate with the Hermes API to fetch price updates for fulfilling Pulse requests
4. Ensure all commands work correctly with the new contract interface

## How has this been tested?

- [ ] Added new tests

So far, we have created the Pulse contract interface based on the Fortuna/Entropy structure and made the codebase compilable. We will test individual functions in subsequent PRs, focusing on:

1. Verifying that all commands work with the new Pulse interface
2. Testing the new storage query methods for performance and correctness
3. Validating the Hermes integration for fetching price updates
4. End-to-end testing of the request processing flow


apps/argus/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG RUST_VERSION=1.82.0
```
ali-bahjati, line 1: I recommend adding CI for this (and the code itself)

apps/argus/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+FROM rust:${RUST_VERSION} AS build
+
+# Build
+WORKDIR /src
+COPY apps/argus apps/argus
+COPY pythnet pythnet
+COPY target_chains/ethereum/entropy_sdk/solidity/abis target_chains/ethereum/entropy_sdk/solidity/abis
```
ali-bahjati, line 9: you'll probably need to add fortuna too

apps/argus/README.md
```diff
@@ -0,0 +1,51 @@
...
+3. Executes the callback on the Pulse contract with the fetched price updates
+
+The Hermes client is implemented in the `keeper/hermes.rs` module and handles:
+
+- Converting price IDs to the format expected by Hermes
+- Fetching price updates from the Hermes API
+- Parsing the response and converting it to the format expected by the Pulse contract
+- Error handling and retries
+
+## Local Development
```
ali-bahjati, line 35: We need a deployed contract for this right? If so let's add it here.

For future: it's good if we can create a fully local deployment setup (that either creates a new ETH network or forks ETH and deploys required contracts)

apps/argus/src/keeper/hermes.rs
```diff
@@ -0,0 +1,413 @@
+/// Hermes Integration
```
ali-bahjati, line 1: Since this is a client for another API I think keeper might not be the best module for it. You can have an `adaptor` module for Hermes.

For future: Pyth Native, Lazer, and now Argus are all using Hermes in Rust and each created their own client for it. We should make a Rust client for Hermes!

apps/argus/src/chain/ethereum.rs
```diff
@@ -0,0 +1,317 @@
...
+        traced_client::{RpcMetrics, TracedClient},
+    },
+    std::sync::Arc,
+};
+
+// TODO: Programmatically generate this so we don't have to keep committed ABI in sync with the
+// contract in the same repo.
+abigen!(
+    Pulse,
+    "../../target_chains/ethereum/contracts/out/IPulse.sol/IPulse.abi.json"
```
ali-bahjati, line 32: This seems like an external dependency. either manually move it in, or add it to the build script of this repo; also mention it in README

ali-bahjati: and i'm wondering what built command you used, i needed to run `forge build --extra-output-files abi`

apps/argus/config.sample.yaml
```diff
@@ -0,0 +1,77 @@
...
+  private_key:
+    # For local development, you can hardcode the private key here
+    value: 0xabcd
+    # For production, you can store the private key in a file.
+    # file: keeper-key.txt
+
+# Hermes API configuration
+hermes:
+  # Base URL for the Hermes API
+  # This can be overridden by setting the HERMES_BASE_URL environment variable
```
ali-bahjati, line 76: Why only this one supports env var?

ali-bahjati (commented):
This is the first review on the overall structure. it looks good. I read READMEs, tried to compile the program and faced some issues (see my comments). In the subsequent reviews I'll look at the business logic.


p.s: I also don't know how much of the functionality is similar to Fortuna but the structure looks identical. I'd recommend you reduce the redundancies as much as you can. At this point I'm also wondering whether we could unify these contracts and services entirly. 

p.s 2: the existence of both config file and CLI arguments is a bit confusing to me and it seems that it's from fortuna. I'll try to understand it later.

apps/argus/src/config/request_price_update.rs
```diff
@@ -0,0 +1,41 @@
+use {
+    crate::{api::ChainId, config::ConfigOptions},
+    clap::Args,
+    ethers::types::Address,
+};
+
+#[derive(Args, Clone, Debug)]
+#[command(next_help_heading = "Request Price Update Options")]
+#[group(id = "RequestPriceUpdate")]
+pub struct RequestPriceUpdateOptions {
```
ali-bahjati, line 10: what's the difference between this command and generate?

apps/argus/src/config/withdraw_fees.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    /// Withdraw the fees on this chain, or all chains if not specified.
+    #[arg(long = "chain-id")]
+    pub chain_id: Option<ChainId>,
+
+    /// If provided, run the command using the keeper wallet. By default, the command uses the provider wallet.
+    /// If this option is provided, the keeper wallet must be configured and set as the fee manager for the provider.
+    #[arg(long = "keeper")]
+    #[arg(default_value = "false")]
+    pub keeper: bool,
```
ali-bahjati, line 21: nit: the name of this parameter is confusing, you could name it `withdrawer` and make it an enum of either 'provider' or 'keeper'.

ali-bahjati (commented):
I'll review the main code-path (run + keeper + chain) later.

PR #2423, opened 2025-02-27
feat(pyth-lazer-sdk): add funding rate

keyvankhademi:
## Summary

This PR adds the required stuff for funding rates in pyth lazer sdk.
This PR also fixes serde of publisher count which was done incorrectly.

## Rationale

This changes are required to support funding rates.

## How has this been tested?

- [x] Added new tests

Manually testes the code with funding rate publisher through both WS and Latest Price API.


lazer/sdk/rust/protocol/src/payload.rs
```diff
@@ -33,18 +33,22 @@ pub enum PayloadPropertyValue {
     Price(Option<Price>),
     BestBidPrice(Option<Price>),
     BestAskPrice(Option<Price>),
-    PublisherCount(Option<u16>),
+    PublisherCount(u16),
     Exponent(i16),
     Confidence(Option<Price>),
+    FundingRate(Option<i64>),
+    FundingTimestamp(Option<u64>),
```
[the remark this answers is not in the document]
keyvankhademi: funding timestamp is the timestamp of the funding rate which is different from the timestamp we have in the aggregated update (which is the time we created the aggregate). Also the timestamp needs to be different per feed as each funding rate has different intervals.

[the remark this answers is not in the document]
Riateche: But do the users care about that timestamp? Why can't they use the timestamp of the aggregate?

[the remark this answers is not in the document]
keyvankhademi: I think it's important to have the timestamp. It needs to be verified too, for example if publishers don't see the latest data that just came out we will send an update with a timestamp for the next period but with the data of the previous rate.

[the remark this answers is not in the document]
Riateche: It will be hard to explain to consumers why there are two different timestamps and how they should use them. Consumers shouldn't be required to even be aware of funding periods, and it's our responsibility to make sure that our data does not contain stale values. To protect against stale values, it makes more sense to require publishers to provide the expiration timestamp for their update. Then we can clear expired updates on aggregation. This way consumers wouldn't need to worry about that at all so they wouldn't need to check a separate timestamp. Actually, if we're choosing to add a timestamp field, the rate expiration timestamp would be more useful to users than the start of the period.

Is it possible to fetch the funding rate for a new period ahead of time? Ideally we should be serving the new rate immediately when the new period starts. If it's not possible, I guess the best action would be to return None until we fetch the new rate.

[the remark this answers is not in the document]
keyvankhademi: I think it's better to keep the data similar to the funding rate apis of the exchanges. It will be easier to explain it to users that the data coming is exactly the same as binance api for example. The timestamp is the actual time that the funding happened, it makes sense to have it in the data. If we go with the expiration date instead, it will be more confusing to the consumers as we have to explain what this value exactly represents.

Getting the funding rate for next interval is impossible as it will change when users buy/see futures. I checked to see if there are any apis to get the instantaneous funding rate (i.e the next funding rate if nothing changes until next interval), but i couldn't find any.

lazer/sdk/rust/protocol/src/payload.rs
```diff
@@ -33,18 +33,22 @@ pub enum PayloadPropertyValue {
     Price(Option<Price>),
     BestBidPrice(Option<Price>),
     BestAskPrice(Option<Price>),
-    PublisherCount(Option<u16>),
+    PublisherCount(u16),
     Exponent(i16),
     Confidence(Option<Price>),
+    FundingRate(Option<Rate>),
+    FundingTimestamp(Option<TimestampUs>),
```
darunrs, line 40: How is this different from the timestamp of the update itself?

lazer/sdk/rust/protocol/src/payload.rs
```diff
@@ -196,14 +220,54 @@ fn read_option_price<BO: ByteOrder>(mut reader: impl Read) -> std::io::Result<Op
...
+        }
+        None => {
+            writer.write_u8(0)?;
+            Ok(())
+        }
+    }
+}
+
+fn read_option_rate<BO: ByteOrder>(mut reader: impl Read) -> std::io::Result<Option<Rate>> {
+    let present = reader.read_u8()? != 0;
```
darunrs, line 240: This is what you were discussing about the onchain payload right? We need to implement this function on contract?

lazer/sdk/rust/protocol/src/publisher.rs
```diff
@@ -12,7 +12,33 @@ use {
...
+    pub price: Option<Price>,
+    /// Last known value of the best bid price of this price feed.
+    /// `None` if no value is currently available.
+    pub best_bid_price: Option<Price>,
+    /// Last known value of the best ask price of this price feed.
+    /// `None` if no value is currently available.
+    pub best_ask_price: Option<Price>,
+    /// Last known value of the funding rate of this feed.
+    /// `None` if no value is currently available.
+    pub funding_rate: Option<Rate>,
```
darunrs, line 33: What about funding timestamp?

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -480,9 +522,11 @@ impl ParsedFeedPayload {
             price: data.price,
             best_bid_price: data.best_bid_price,
             best_ask_price: data.best_ask_price,
-            publisher_count: data.publisher_count,
+            publisher_count: Some(data.publisher_count),
```
darunrs, line 525: Why does this need a Some() now if we changed this to no be an Option? I think the parsed payload type needs to also be updated.

darunrs (approved):
Some comments but otherwise, looks good.

lazer/sdk/rust/protocol/src/payload.rs
darunrs: Never mind this is answered above.

keyvankhademi: > unding timestamp is the timestamp of the funding rate which is different from the timestamp we have in the aggregated update (which is the time we created the aggregate). Also the timestamp needs to be different per feed as each funding rate has different intervals.

lazer/sdk/rust/protocol/src/payload.rs
keyvankhademi: Yes, the contract needs to update to support this.

lazer/sdk/rust/protocol/src/publisher.rs
keyvankhademi: I reused the source_timestamp as funding timestamp.

lazer/sdk/rust/protocol/src/publisher.rs
```diff
@@ -12,7 +12,33 @@ use {
 /// from the publisher to the router.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct PriceFeedData {
+pub struct PriceFeedDataV2 {
```
darunrs, line 15: Wait, how will V1/V2 be communicated in the binary? Do we need to add a magic?

keyvankhademi: there are different urls in the relayers for these 2.

PR #2431, opened 2025-02-27
chore(ci): use Org-defined larger runner

morancj:
## Summary

Increate runner size for CI pre-commit

### Currently defined runners

| Runner name | OS | Arch | vCPU | RAM (GiB) | SSD (GB) |
| --- | --- | --- | --- | --- | --- |
| ubuntu-22.04-x64-m | Ubuntu 22.04 | x64_64 | 4 | 16 | 150 |
| ubuntu-24.04-x64-m | Ubuntu 24.04 | x64_64 | 4 | 16 | 150 |
| ubuntu-latest-x64-m | _Currently_ Ubuntu 24.04 | x64_64 | 4 | 16 | 150 |
| ubuntu-22.04-x64-l | Ubuntu 22.04 | x64_64 | 8 | 32 | 300 |
| ubuntu-24.04-x64-l | Ubuntu 22.04 | x64_64 | 8 | 32 | 300 |
| ubuntu-latest-x64-l | _Currently_ Ubuntu 22.04 | x64_64 | 8 | 32 | 300 |

💡 Initially, these runners are only available on this repo.

See also: [GitHub Actions Runner Images](https://github.com/actions/runner-images)

## Rationale

This GHA workflow currently takes 27.5 minutes and ran out of disk space, [example](https://github.com/pyth-network/pyth-crosschain/actions/runs/13568732621)

## How has this been tested?

If the pre-commit workflow fails, we'll know 🙂

PR #2432, opened 2025-02-27
chore(ci): update reviewers for /.github/ and CODEOWNERS

morancj:
## Summary

CODEOWNERS for GH workflows, CODEOWNERS itself

## Rationale

Reduce the chances of PRs to use the [medium or large runners](https://github.com/pyth-network/pyth-crosschain/pull/2431) racking up our spend

## How has this been tested?

Only via the GH verification:
> This CODEOWNERS file is valid.

I don't believe we can fully test until it's merged.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,7 +14,7 @@ pnpm-workspace.yaml @pyth-network/web-team
 turbo.json @pyth-network/web-team
 .npmrc @pyth-network/web-team
 .nvmrc @pyth-network/web-team
-/.github/ @pyth-network/core-team @pyth-network/web-team
+/.github/ @pyth-network/core-team
 /.github/CODEOWNERS @pyth-network/core-team
 .github/workflows/publish-js.yml @pyth-network/web-team
 .github/workflows/ci-turbo-build.yml @pyth-network/web-team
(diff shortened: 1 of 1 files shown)


cprussin (approved):
You may want to change the title of this PR before merging since it's not actually restricting access, just adding automatic reviewers, and someone seeing the message "restrict access" in the commit log may be confused :)

.github/CODEOWNERS
```diff
@@ -14,6 +14,8 @@ pnpm-workspace.yaml @pyth-network/web-team
 turbo.json @pyth-network/web-team
 .npmrc @pyth-network/web-team
 .nvmrc @pyth-network/web-team
+/.github/ @pyth-network/core-team @pyth-network/web-team
```
cprussin, line 17: I don't think the web team needs to be a codeowner for the whole `.github` directory, the web team is already listed as a codeowner for all the js related workflows below.  I can't recall how github deals with ordering for codeowners, but just make sure that the codeowners for the specific js-related actions below take precedence over the codeowners for the whole directory here.

morancj:
> You may want to change the title of this PR before merging since it's not actually restricting access, just adding automatic reviewers, and someone seeing the message "restrict access" in the commit log may be confused :)

doh, sorry, thanks!

.github/CODEOWNERS
morancj: [Yes](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file), thank you
> Order is important; the last matching pattern takes the most precedence

PR #2438, opened 2025-03-03
Chore(contract manager)  add viem

aditya520:
## How has this been tested?

- [ ] Added new tests


ali-bahjati (approved):
The aim to use viem here is to get a better RPC?

aditya520:
Yes. Use viem as the default provider. If that fails, then use the provider in the contract manager.

PR #2444, opened 2025-03-06
feat(pyth-lazer-protocol): add price request

keyvankhademi:
## Summary

Changes needed to support price request (historical prices with signatures)

## Rationale

Our users need this api

## How has this been tested?

- [ ] Added new tests

Manually tested the api

### DIFF
--- a/lazer/sdk/rust/protocol/src/api.rs
+++ b/lazer/sdk/rust/protocol/src/api.rs
@@ -27,8 +27,6 @@ pub struct PriceRequest {
     pub timestamp: TimestampUs,
     pub price_feed_ids: Vec<PriceFeedId>,
     pub properties: Vec<PriceFeedProperty>,
-    // "chains" was renamed to "formats". "chains" is still supported for compatibility.
-    #[serde(alias = "chains")]
     pub formats: Vec<Format>,
     #[serde(default)]
     pub json_binary_encoding: JsonBinaryEncoding,
--- a/lazer/sdk/rust/protocol/src/api.rs
+++ b/lazer/sdk/rust/protocol/src/api.rs
@@ -1,12 +1,30 @@
 use serde::{Deserialize, Serialize};
 
 use crate::router::{
-    Channel, Format, JsonBinaryEncoding, JsonUpdate, PriceFeedId, PriceFeedProperty,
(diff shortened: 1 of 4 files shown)


lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -21,6 +21,24 @@ pub struct LatestPriceRequest {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PriceRequest {
+    pub timestamp: TimestampUs,
+    pub price_feed_ids: Vec<PriceFeedId>,
+    pub properties: Vec<PriceFeedProperty>,
+    // "chains" was renamed to "formats". "chains" is still supported for compatibility.
+    #[serde(alias = "chains")]
```
Riateche, line 31: We don't need a compatibility alias for a new endpoint.

Riateche (approved):

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -236,6 +243,17 @@ mod channel_ids {
     pub const FIXED_RATE_200: ChannelId = ChannelId(3);
 }
 
+impl Display for Channel {
```
darunrs, line 246: You could probably just use serde to serialize it as a string, which gets you the same result without a duplicate implementation.

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -21,6 +21,22 @@ pub struct LatestPriceRequest {
     pub channel: Channel,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PriceRequest {
```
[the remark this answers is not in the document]
darunrs: Ah ok I see. We take in a timestamp to fetch the price for. Theoretically we could unify them but I guess that's not a big deal right now.

lazer/sdk/rust/protocol/src/router.rs
keyvankhademi: I looked into it but couldn't find a good way of handling it. If you want please give it a try and let me know if it worked.

PR #2454, opened 2025-03-08
fix(pulse): Miscellaneous contract fixes

jayantk:
## Summary

I was looking through the Pulse contract and noticed some issues:
* There's a security issue with the IPulseConsumer interface as anyone can mock data and pass it as the legitimate response to the request
* publishTimes are unix timestamps so should be uint64
* The provider fee / backup mechanism doesn't work properly because it doesn't account for pyth fees, and it also doesn't give any incentive to backup providers. 

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
@@ -18,6 +18,7 @@ interface IPulse is PulseEvents {
     /**
      * @notice Requests price updates with a callback
      * @dev The msg.value must be equal to getFee(callbackGasLimit)
+     * @param provider The provider to fulfill the request
      * @param callbackGasLimit The amount of gas allocated for the callback execution
      * @param publishTime The minimum publish time for price updates, it should be less than or equal to block.timestamp + 60
      * @param priceIds The price feed IDs to update. Maximum 10 price feeds per request.
@@ -30,6 +31,7 @@ interface IPulse is PulseEvents {
      *      the fee estimation unreliable.
      */
     function requestPriceUpdatesWithCallback(
+        address provider,
         uint256 publishTime,
         bytes32[] calldata priceIds,
         uint256 callbackGasLimit
@@ -39,11 +41,13 @@ interface IPulse is PulseEvents {
      * @notice Executes the callback for a price update request
      * @dev Requires 1.5x the callback gas limit to account for cross-contract call overhead
      * For example, if callbackGasLimit is 1M, the transaction needs at least 1.5M gas + some gas for some other operations in the function before the callback
+     * @param providerToCredit The provider to credit for fulfilling the request. This may not be the provider that submitted the request (if the exclusivity period has elapsed).
      * @param sequenceNumber The sequence number of the request
      * @param updateData The raw price update data from Pyth
      * @param priceIds The price feed IDs to update, must match the request
      */
     function executeCallback(
+        address providerToCredit,
         uint64 sequenceNumber,
         bytes[] calldata updateData,
         bytes32[] calldata priceIds
@@ -59,12 +63,17 @@ interface IPulse is PulseEvents {
 
     /**
      * @notice Calculates the total fee required for a price update request
+     * FIXME: is this comment right?
      * @dev Total fee = base Pyth protocol fee + gas costs for callback
+     * @param provider The provider to fulfill the request
      * @param callbackGasLimit The amount of gas allocated for callback execution
+     * @param priceIds The price feed IDs to update.
      * @return feeAmount The total fee in wei that must be provided as msg.value
      */
     function getFee(
-        uint256 callbackGasLimit
+        address provider,
+        uint256 callbackGasLimit,
+        bytes32[] calldata priceIds
     ) external view returns (uint128 feeAmount);
 
     function getAccruedFees() external view returns (uint128 accruedFeesInWei);
@@ -85,7 +94,7 @@ interface IPulse is PulseEvents {
 
     function registerProvider(uint128 feeInWei) external;
 
-    function setProviderFee(uint128 newFeeInWei) external;
+    function setProviderFee(address provider, uint128 newFeeInWei) external;
 
     function getProviderInfo(
         address provider
--- a/target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
@@ -6,11 +6,32 @@ import "@pythnetwork/pyth-sdk-solidity/IPyth.sol";
 import "./PulseEvents.sol";
 import "./PulseState.sol";
 
-interface IPulseConsumer {
+abstract contract IPulseConsumer {
+    // This method is called by Pulse to provide the price updates to the consumer.
+    // It asserts that the msg.sender is the Pulse contract. It is not meant to be
+    // overridden by the consumer.
+    function _pulseCallback(
+        uint64 sequenceNumber,
+        PythStructs.PriceFeed[] memory priceFeeds
+    ) external {
+        address pulse = getPulse();
+        require(pulse != address(0), "Pulse address not set");
+        require(msg.sender == pulse, "Only Pulse can call this function");
+
+        pulseCallback(sequenceNumber, priceFeeds);
+    }
+
+    // getPulse returns the Pulse contract address. The method is being used to check that the
+    // callback is indeed from the Pulse contract. The consumer is expected to implement this method.
+    function getPulse() internal view virtual returns (address);
+
+    // This method is expected to be implemented by the consumer to handle the price updates.
+    // It will be called by _pulseCallback after _pulseCallback ensures that the call is
+    // indeed from Pulse contract.
     function pulseCallback(
         uint64 sequenceNumber,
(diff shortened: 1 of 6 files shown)


target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -6,18 +6,40 @@ import "@pythnetwork/pyth-sdk-solidity/IPyth.sol";
...
+        pulseCallback(sequenceNumber, priceFeeds);
+    }
+
+    // getPulse returns the Pulse contract address. The method is being used to check that the
+    // callback is indeed from the Pulse contract. The consumer is expected to implement this method.
+    function getPulse() internal view virtual returns (address);
+
+    // This method is expected to be implemented by the consumer to handle the price updates.
+    // It will be called by _pulseCallback after _pulseCallback ensures that the call is
+    // indeed from Pulse contract.
```
jayantk, line 30: Note that if this is an external method, then *anyone* could have called this function with a sequenceNumber and a list of PriceFeeds (which are already parsed, and thus mockable). 

Fix here ensures that only the Pulse contract can invoke the callback.

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -59,15 +84,22 @@ interface IPulse is PulseEvents {
...
      */
     function getFee(
-        uint256 callbackGasLimit
+        address provider,
+        uint256 callbackGasLimit,
+        bytes32[] calldata priceIds
     ) external view returns (uint128 feeAmount);
 
-    function getAccruedFees() external view returns (uint128 accruedFeesInWei);
+    function getAccruedPythFees()
```
jayantk, line 99: (renamed this for clarity)

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -85,21 +87,20 @@ abstract contract Pulse is IPulse, PulseState {
         req.requester = msg.sender;
         req.numPriceIds = uint8(priceIds.length);
         req.provider = provider;
+        req.fee = SafeCast.toUint128(msg.value - _state.pythFeeInWei);
```
jayantk, line 90: Store the fee in the request so we can credit the provider *who actually fulfilled the request* in `executeCallback`

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -127,19 +128,41 @@ abstract contract Pulse is IPulse, PulseState {
...
 
-        // Parse price feeds first to measure gas usage
-        PythStructs.PriceFeed[] memory priceFeeds = IPyth(_state.pyth)
-            .parsePriceFeedUpdates(
-                updateData,
-                priceIds,
-                SafeCast.toUint64(req.publishTime),
-                SafeCast.toUint64(req.publishTime)
-            );
+        // TODO: should this use parsePriceFeedUpdatesUnique? also, do we need to add 1 to maxPublishTime?
```
jayantk, line 131: Note that the tests mock the parsePriceFeedUpdates call, so we don't actually know if it's being invoked correctly.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -111,7 +112,7 @@ abstract contract Pulse is IPulse, PulseState {
             block.timestamp < req.publishTime + _state.exclusivityPeriodSeconds
         ) {
             require(
-                msg.sender == req.provider,
+                providerToCredit == req.provider,
```
jayantk, line 115: the `provider` key is a cold key that shouldn't be used in hot actions. (This is why we have the feeManager as the hot key that has limited permissions -- if it is compromised, the provider can rotate it without too much damage)

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -336,22 +366,51 @@ abstract contract Pulse is IPulse, PulseState {
...
+            _state.providers[provider].isRegistered,
             "Provider not registered"
         );
-        uint128 oldFee = _state.providers[msg.sender].feeInWei;
-        _state.providers[msg.sender].feeInWei = newFeeInWei;
-        emit ProviderFeeUpdated(msg.sender, oldFee, newFeeInWei);
+        require(
+            msg.sender == provider ||
+                msg.sender == _state.providers[provider].feeManager,
+            "Only provider or fee manager can invoke this method"
```
jayantk, line 396: the fee manager should be allowed to set fees

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
cctdaniel: I think using `parsePriceFeedUpdatesUnique` here is a good idea due to its deterministic behavior

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
```diff
@@ -30,7 +52,8 @@ interface IPulse is PulseEvents {
      *      the fee estimation unreliable.
      */
     function requestPriceUpdatesWithCallback(
-        uint256 publishTime,
+        address provider,
```
ali-bahjati, line 55: I don't like people choosing their providers. It's an extra hop for integrators to think about (what are providers? how they are different? how should i choose? ...)

cctdaniel (approved):

target_chains/ethereum/contracts/contracts/pulse/IPulse.sol
jayantk: Yeah agreed. I think the thing to do is to add a wrapper method that chooses for users. I'm not exactly sure what the right way to choose is yet though.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -85,21 +87,20 @@ abstract contract Pulse is IPulse, PulseState {
...
-
-        _state.providers[provider].accruedFeesInWei += SafeCast.toUint128(
-            msg.value - _state.pythFeeInWei
-        );
         _state.accruedFeesInWei += _state.pythFeeInWei;
 
         emit PriceUpdateRequested(req, priceIds);
     }
 
+    // TODO: does this need to be payable? Any cost paid to Pyth could be taken out of the provider's accrued fees.
```
ali-bahjati, line 101: good call no it's not needed.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -53,17 +53,19 @@ abstract contract Pulse is IPulse, PulseState {
...
         bytes32[] calldata priceIds,
         uint256 callbackGasLimit
     ) external payable override returns (uint64 requestSequenceNumber) {
-        address provider = _state.defaultProvider;
         require(
             _state.providers[provider].isRegistered,
             "Provider not registered"
         );
 
+        // FIXME: this comment is wrong. (we're not using tx.gasprice)
```
ali-bahjati, line 68: remove the wrong comment then?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -182,13 +197,16 @@ abstract contract Pulse is IPulse, PulseState {
...
 
         for (uint i = 0; i < priceFeeds.length; i++) {
             prices[i] = priceFeeds[i].price.price;
             conf[i] = priceFeeds[i].price.conf;
             expos[i] = priceFeeds[i].price.expo;
-            publishTimes[i] = priceFeeds[i].price.publishTime;
+            // Safe cast because this is a unix timestamp in seconds.
+            publishTimes[i] = SafeCast.toUint64(
+                priceFeeds[i].price.publishTime
+            );
```
ali-bahjati, line 209: not sure it helps with costs, afaik event arguments will end up as u256 regardless.

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
ali-bahjati: should we add a check to make sure the provider is registered?

target_chains/ethereum/contracts/contracts/pulse/Pulse.sol
```diff
@@ -127,19 +128,41 @@ abstract contract Pulse is IPulse, PulseState {
...
+        // with time in order to ensure that the callback eventually gets executed.
+        // (There may be exploits with ^ though if the consumer contract is malicious ?)
+        _state.providers[providerToCredit].accruedFeesInWei += SafeCast
+            .toUint128((req.fee + msg.value) - pythFee);
 
         clearRequest(sequenceNumber);
 
+        // TODO: I'm pretty sure this is going to use a lot of gas because it's doing a storage lookup for each sequence number.
+        // a better solution would be a doubly-linked list of active requests.
+        // After successful callback, update firstUnfulfilledSeq if needed
```
ali-bahjati, line 156: yeah true. let's do some gas benchmarking later.

ali-bahjati (commented):
This PR had a lot of small high context changes. I generally recommend separating bug fixes vs other changes and specially the API changes.

PR #2466, opened 2025-03-11
feat(argus): Delete hash chain logic

jayantk:
## Summary

Continuing work on the fortuna -> argus change. This PR deletes all of the hash chain logic from fortuna that's not needed here.


apps/argus/src/command/register_provider.rs
```diff
@@ -45,27 +43,13 @@ pub async fn register_provider_from_config(
...
-        chain_id,
-        &private_key_string.parse::<LocalWallet>()?.address(),
-        &chain_config.contract_addr,
-        &random,
-        commitment_length,
-        provider_config.chain_sample_interval,
-    )?;
-    tracing::info!("Done generating hash chain");
+    // FIXME: delete this
+    let commitment_length = 1000;
```
jayantk, line 48: this stuff will all get deleted later so I think it's fine to mock them out for now.

ali-bahjati (approved):

cctdaniel (approved):

PR #2469, opened 2025-03-11
feat(target_chains/sui): Use Turbo for Commands

darunrs:
## Summary

Updated SUI JS SDK and CLI to use Turbo to help with dependency management. Updated the README to reflect this. Also fixed the relay example to use feed ids as string instead of number. 

## Rationale

There were a litany of issues getting the SUI JS SDK and CLI to work at all. I got help from Connor to figure out what is wrong and how to fix it. Now, the examples should work out of box and without too much trouble. 

## How has this been tested?

- [ ] Added new tests

Manually tested by cleaning up repo locally using `git clean -fdx` and then installing with pnpm in root. Then, running examples and cli commands using turbo successfully.

### DIFF
--- a/target_chains/sui/sdk/js/src/examples/SuiRelay.ts
+++ b/target_chains/sui/sdk/js/src/examples/SuiRelay.ts
@@ -12,6 +12,7 @@ const argvPromise = yargs(hideBin(process.argv))
   .option("feed-id", {
     description:
       "Price feed ids to update without the leading 0x (e.g f9c0172ba10dfa4d19088d94f5bf61d3b54d5bd7483a322a982e1373ee8ea31b). Can be provided multiple times for multiple feed updates",
+    string: true,
     type: "array",
     demandOption: true,
   })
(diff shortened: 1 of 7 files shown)


target_chains/sui/sdk/js/src/examples/SuiRelay.ts
```diff
@@ -49,7 +50,8 @@ async function run() {
 
   // Fetch the latest price feed update data from the Price Service
   const connection = new SuiPriceServiceConnection(argv["hermes"]);
-  const feeds = argv["feed-id"] as string[];
+  const feeds = argv["feed-id"];
+  if (!Array.isArray(feeds)) { throw new Error("Not a valid input!"); }
```
cprussin, line 54: I think this will fail prettier, try running `pnpm turbo fix`

cprussin (approved):

target_chains/sui/sdk/js/src/examples/SuiRelay.ts
darunrs: Ran the command in root and it surprisingly did not complain about or change anything.

cprussin: ah right this package has the legacy pre-commit based prettier config, instead run `pre-commit run --all`

PR #2470, opened 2025-03-11
fix(lazer/aptos-contract): add coin conversion map to tests

tejasbadadare:
## Summary
Fix Aptos contract CI by explicitly creating a coin conversion map in the test setup.

## Rationale
A recent update to the `AptosFramework` dependency made it required to explicitly define a coin conversion map when initializing a coin. We init an APT coin in tests to verify fee withdrawals. Since our dep was pointed to `mainnet`, this caused CI to randomly start failing. This fix adheres to the requirement and also pins the version to the latest revision to avoid other surprises from deps.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/contracts/aptos/Move.toml
+++ b/lazer/contracts/aptos/Move.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 license = "UNLICENSED"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


jayantk (approved):
🙇

keyvankhademi (approved):

PR #2474, opened 2025-03-12
fix(apps/hermes): fix duplicate events for processed slots

cctdaniel:
# Fix Duplicate Events for Processed Slots

## Problem

The application was experiencing an issue where multiple events could be sent for the same slot under certain conditions:

1. When a slot was processed out of order (e.g., slot 100, then 101, then 100 again), the system would send both a `New` event and an `OutOfOrder` event for the same slot.
2. During concurrent processing of updates for the same slot, race conditions could lead to duplicate events being sent.

This resulted in downstream consumers (like SSE clients) receiving redundant events, which could cause unnecessary processing and potential bugs.

## Solution

We implemented a more efficient mechanism to ensure that only one event is ever sent per slot:

1. Modified `store_wormhole_merkle_verified_message` to check if a state already exists for the given slot before storing it, returning a boolean to indicate whether the state was newly stored.

2. Modified `store_update` to check if accumulator messages already exist for a slot before storing them, returning early if they do.

This approach leverages the existing cache mechanisms (which already have size limits) to determine if a slot has been processed before. When either a VAA or accumulator message for a slot is received again, we detect this early in the processing pipeline and return immediately, preventing duplicate processing and event generation.

This solution is more memory-efficient as it eliminates the need for an unbounded tracking set while still ensuring that each slot will only generate a single event, either New or OutOfOrder, but never both.

## Testing

Two tests were added to verify the fix:

1. `test_out_of_order_updates_send_single_event_per_slot`: Verifies that when processing slots out of order (100 > 101 > 100), only one event is sent for each slot.
2. `test_concurrent_updates_same_slot_sends_single_event`: Verifies that when 100 concurrent updates for the same slot are processed, only one event is sent.

Both tests confirm that our solution effectively prevents duplicate events while maintaining the correct application state.

### DIFF
--- a/apps/hermes/server/src/state/aggregate.rs
+++ b/apps/hermes/server/src/state/aggregate.rs
@@ -1543,7 +1543,7 @@ mod calculate_twap_unit_tests {
 
     #[test]
     fn test_invalid_timestamps() {
-        let start = create_basic_twap_message(100, 100, 90, 1000);
+        let start = create_basic_twap_message(100, 100, 110, 1000);
         let end = create_basic_twap_message(300, 200, 180, 1100);
 
         let err = calculate_twap(&start, &end).unwrap_err();
--- a/apps/hermes/server/src/state/aggregate.rs
+++ b/apps/hermes/server/src/state/aggregate.rs
@@ -1538,7 +1538,7 @@ mod calculate_twap_unit_tests {
 
     #[test]
     fn test_invalid_timestamps() {
-        let start = create_basic_twap_message(100, 100, 90, 1000);
+        let start = create_basic_twap_message(100, 100, 110, 1000);
         let end = create_basic_twap_message(300, 200, 180, 1100);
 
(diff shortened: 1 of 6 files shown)


apps/hermes/server/src/state/aggregate.rs
```diff
@@ -105,6 +105,10 @@ pub struct AggregateStateData {
 
     /// Aggregate Specific Metrics
     pub metrics: metrics::Metrics,
+
+    /// Set of slots for which events have already been sent.
+    /// This prevents sending multiple events for the same slot.
+    pub slots_with_sent_events: HashSet<Slot>,
```
ali-bahjati, line 111: this can grow indefinitely

ali-bahjati (commented):
my recommendation is that when you call `store_wormhole_merkle_verified_message` (and the other one) you return something like it was there or not, and if it was there you return immediately. Those states also have a limited size to make sure they don't grow much.

PR #2478, opened 2025-03-12
chore: move all prettier formatting into individual packages

cprussin:
This PR makes a few related changes:

- Stops using `pre-commit` to run prettier anywhere in the monorepo, instead we now run prettier from each package and trigger the runs using turborepo at the root.

- Upgrade the prettier version used in all packages that were previously relying on `pre-commit` to run prettier.

- Reformat files that need it due to the new prettier version.

- Cleans up the turborepo config so that it now runs some tasks in the root package, and also automatically runs `pnpm install` when needed.

- Sets up `pre-commit` to run `TURBO_SCM_BASE="HEAD^1" pnpm turbo fix --affected` which triggers auto-fixes for any package that has changed in this commit, or any package depending on one that's changed in this commit.  Note there are possible downsides of doing this, we'll have to keep an eye on these and maybe update in a follow up:

    - We might want to consider removing TURBO_SCM_BASE which will make turbo compare against main instead of HEAD^1.  In effect this would cause pre-commit to fix anything that has changed since main in the current branch, not just in the current commit.  I didn't do this because it would cause runs to be even slower and I assume the expectation is for pre-commit to fixup things that matter just for your one commit, not the whole branch.

    - This script still might be quite slow -- especially if you are changing a lot of packages or a package that has a lot of dependencies -- as it's not filtering down the sub-tasks under fix to just the files that have changes.  We can probably do something here to filter further, maybe there's some way to pass the filenames in through turbo or something?  It's a bit tricky as you can make changes that cause things to need to be fixed in other files, so just running fixes on affected files isn't really actually reliable....

- Finally, this PR removes pre-commit from CI entirely, as pre-commit is now entirely redundant of other more comprehensive CI checks, and pre-commit should only be used moving forward as a devex optimization.

Note this PR is structured in logically separated commits, so please look through them independently for an easy review.


keyvankhademi (approved):

ali-bahjati (commented):
I think still a couple of us would like to use pre-commit (for personal convenience) and you can just solve it by adding the command to the pre-commit.

On another note, it's difficult to review this PR in a single commit because it contains a lot of mechanical changes in addition to your changes. Would suggest putting the prettier changes in a separate commit.

ali-bahjati (changes requested):
I think still a couple of us would like to use pre-commit (for personal convenience) and you can just solve it by adding the command to the pre-commit.

On another note, it's difficult to review this PR in a single commit because it contains a lot of mechanical changes in addition to your changes. Would suggest putting the prettier changes in a separate commit.

jayantk (approved):

ali-bahjati (approved):
Amazing, thanks!

PR #2483, opened 2025-03-13
fix(staking): fix header on small screens 2: This Time It's Personal

cprussin:
## Summary

More fixes for #2482 .

## Rationale

The fix in #2482 did not account for the header state when the wallet is connected, which additionally shows the current stake account and requires more space.

## How has this been tested?

Before:
screenshot-2025-03-13-101845

After:
screenshot-2025-03-13-101851

### DIFF
--- a/apps/staking/src/components/AccountSummary/index.tsx
+++ b/apps/staking/src/components/AccountSummary/index.tsx
@@ -74,7 +74,7 @@ export const AccountSummary = ({
     <Image
       src={background}
       alt=""
-      className="absolute -right-40 hidden h-full object-cover object-right [mask-image:linear-gradient(to_right,_transparent,_black_50%)] md:block"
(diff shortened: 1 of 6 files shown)


apps/staking/src/components/AccountSummary/index.tsx
```diff
@@ -74,7 +74,7 @@ export const AccountSummary = ({
     <Image
       src={background}
       alt=""
-      className="absolute -right-40 hidden h-full object-cover object-right [mask-image:linear-gradient(to_right,_transparent,_black_50%)] md:block"
+      className="absolute -right-40 hidden h-full object-cover object-right [mask-image:linear-gradient(to_right,_transparent,_black_50%)] sm:block"
```
cprussin, line 77: Unrelated change but just something I noticed looked odd, the account summary background wasn't showing up on small screens where it should have been

guibescos (approved):

PR #2488, opened 2025-03-16
chore: update all app node dependencies

cprussin:
*DEAR REVIEWER*: please note I've painstakingly split this PR into commits that are logically separated.  It should be fairly easy to review if you look through commit-by-commit -- many of the commits are just the same mechanical change applied to the whole monorepo....

## Summary

This PR moves all of our web apps to using exclusively catalog dependencies and upgrades all catalog dependencies to latest verisons.

While doing this, this PR also includes fixes (mostly automated) for newly enforced eslint rules, stricter typechecking introduced by the upgraded typescript version, and improved prettier formatting in some places.  This PR also migrates some code to remove deprecated API usage and take advantage of newer lib features in a few of our libs where appropriate.

## Rationale

There are a number of reasons to get this upgrade done:

- We were starting to lag behind on some of our libs and were missing out on latest & greatest features, for instance we were missing some great new eslint rules, typechecking capabilities, and react-aria components.
- I have upgraded my `@cprussin` tool configs and the newer version resolves a few dependency pinning issues that we'd been working around
- The newer versions of nextjs & react should pave the way for some performance improvement work I plan to get to in the insights hub soon
- We had a mounting pile of hacks for things that have been fixed upstream and we simply hadn't upgraded and removed the hacks, this was growing and getting concerning

Also fixes UI-25

## How has this been tested?

- [x] Current tests cover my changes
- [ ] Added new tests
- [x] Manually tested the code

PR #2489, opened 2025-03-17
chore: rebuild pnpm lockfile

cprussin:
## Summary

This PR rebuilds the `pnpm-lock.yaml` file.  The steps to do this are:

1. `git clean -fdx` (reset the git repo to a fresh state)
2. `rm pnpm-lock.yaml`
3. `pnpm i`
4. `pnpm dedupe`

This PR also includes some fixes for some packages whose dependencies moved without major version bumps, but which in fact included breaking changes.

## Rationale

It's important to rebuild the pnpm lockfile occasionally for a few reasons:

- Doing so ensures we're using latest versions of dependencies, within compatible ranges declared in `package.json` files
- This ensures we're working against the same dependencies that would be installed by a consumer of our libs
- It's important for ensuring that the lockfile represents an optimized set of dependencies based on the requirements of the workspaces

We haven't been able to do this for a while due to a myriad of dependency issues, but with #2488 , those dependency issues were resolved and we can now do this.

## How has this been tested?

- [ ] Added new tests


target_chains/cosmwasm/tools/src/ci/deployer/osmosis.ts
```diff
@@ -16,16 +15,10 @@ export type OsmosisHost = {
...
 export class OsmosisDeployer implements Deployer {
-  private chain: Chain;
   constructor(
     private endpoint: string,
     private mnemonic: string,
-  ) {
-    const c = chains.find(({ chain_name }) => chain_name === "osmosis");
-    if (c === undefined)
-      throw new Error("Could not find Osmosis in chain registry");
-    this.chain = c;
```
cprussin, line 27: This whole block was warning because `chain` is a private variable that's never consumed, meaning none of this should be needed.  Please someone who knows this package better -- please help take a look and make sure there's nothing hacky that's breaking the static analysis here and actually uses this in some unexpected way...

ali-bahjati (approved):

PR #2490, opened 2025-03-17
fix: fix storybook background color

cprussin:
## Summary

Fix the storybook background config.

## Rationale

The api for this changed in some storybook version between when I implement it originally and now, and I just noticed it isn't working since the storybook has been intermittently broken for a while.  This PR fixes it.

## How has this been tested?

- [ ] Added new tests


guiyumin (approved):
LGTM

PR #2494, opened 2025-03-17
fix: fix broken wallet adapter in proposals app

cprussin:
## Summary

This PR upgrades the wallet adapters used in the proposals app to use the catalog dependencies.

It looks like the solana packages revved some patches but released a number of breaking changes.  Unifying on the catalog dependencies both fixes the issues and also makes it easier for me to keep these up to date with the rest of the monorepo moving forward.

## Rationale

The proposals app wallet button is currently broken, presumably as a result of #2489 .

## How has this been tested?

- [ ] Added new tests


governance/xc_admin/packages/xc_admin_frontend/pages/_app.tsx
```diff
@@ -6,12 +6,9 @@ import {
 import { WalletModalProvider } from '@solana/wallet-adapter-react-ui'
 import '@solana/wallet-adapter-react-ui/styles.css'
 import {
-  GlowWalletAdapter,
   LedgerWalletAdapter,
   PhantomWalletAdapter,
   SolflareWalletAdapter,
-  SolletExtensionWalletAdapter,
-  SolletWalletAdapter,
```
cprussin, line 14: fyi @guibescos these wallet adapters have been deprecated and no longer exist in the current version of the solana wallet adapter library.

package.json
```diff
@@ -24,6 +24,9 @@
   "pnpm": {
     "overrides": {
       "@solana/web3.js@1.77.4>rpc-websockets": "7.11.0"
+    },
+    "patchedDependencies": {
+      "eccrypto@1.1.6": "patches/eccrypto@1.1.6.patch"
```
cprussin, line 29: I removed this patch in #2489 and the proposals app works in prod without it, but I didn't test local dev mode, and there  without this patch we get a syntax error in eccrypto, so I need to resurrect it.

guibescos (approved):

PR #2504, opened 2025-03-19
chore(pyth-solana-receiver-sdk): enable solana-program >= 2 (and thus anchor-lang 0.31.0)

tejasbadadare:
## Summary
Remove the solana-program<2 bound in our `pyth-solana-receiver-sdk` dependencies to enable using the SDK with anchor 0.31.0.

## Rationale
- `anchor-lang` v0.31.0 requires `solana-program=^2`. Thus, this change enables using anchor 0.31.0 with our SDK. This change has been requested by a few consumers. 
- `bytemuck_derive` has been restricted to <=1.8.1 since after this version, minimum rust version 1.84 is required, which breaks compatibility with the other packages in the workspace.
- We did not bump the minimum anchor version to 0.31.0 since solana-program ^2 requires a higher rust version, which would also break compatibility with other packages in the workspace.
- Once this lands, I'll publish [this repo](https://github.com/guibescos/anchor-pyth/tree/v0.30.1) to our GH org which shows a example of using the SDK with anchor 0.31.0 (and other versions).

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
+++ b/target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-solana-receiver-sdk"
-version = "0.5.0"
+version = "0.6.0"
 description = "SDK for the Pyth Solana Receiver program"
 authors = ["Pyth Data Association"]
 repository = "https://github.com/pyth-network/pyth-crosschain"
@@ -15,7 +15,8 @@ name = "pyth_solana_receiver_sdk"
 [dependencies]
 anchor-lang = ">=0.28.0"
 hex = ">=0.4.3"
+bytemuck_derive = "<=1.8.1" # After this version, MSRV 1.84 is required.
 pythnet-sdk = { path = "../../../pythnet/pythnet_sdk", version = "2.1.0", features = [
     "solana-program",
 ] }
-solana-program = ">=1.16.0, <2.0.0"
+solana-program = ">=1.16.0"
--- a/target_chains/solana/Cargo.lock
+++ b/target_chains/solana/Cargo.lock
@@ -3120,9 +3120,10 @@ dependencies = [
 
 [[package]]
 name = "pyth-solana-receiver-sdk"
-version = "0.5.0"
+version = "0.6.0"
 dependencies = [
  "anchor-lang",
+ "bytemuck_derive",
  "hex",
  "pythnet-sdk",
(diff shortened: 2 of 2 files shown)


target_chains/solana/pyth_solana_receiver_sdk/Cargo.toml
```diff
@@ -15,7 +15,8 @@ name = "pyth_solana_receiver_sdk"
 [dependencies]
 anchor-lang = ">=0.28.0"
 hex = ">=0.4.3"
+bytemuck_derive = "<=1.8.1" # After this version, MSRV 1.84 is required.
 pythnet-sdk = { path = "../../../pythnet/pythnet_sdk", version = "2.1.0", features = [
     "solana-program",
 ] }
-solana-program = ">=1.16.0, <2.0.0"
+solana-program = ">=1.16.0"
```
guibescos, line 22: actually anchor people recommend against having `solana-program` here and encourage using the reexport in `anchor_lang::solana_program` , can you try that for the sdk

tejasbadadare: Yeah I saw that in the migration notes, (and its in my cursed [WIP PR](https://github.com/pyth-network/pyth-crosschain/pull/2501)) but the issue is that the reexport doesn't exist in anchor-lang=0.28. 

This is tricky because we use the [`pubkey!` macro in lib.rs](https://github.com/pyth-network/pyth-crosschain/blob/59f3f6f37ca9eedc111bee37218a08d416c48e8f/target_chains/solana/pyth_solana_receiver_sdk/src/lib.rs#L20), but this comes from the solana-program crate, not the reexport. See here: https://github.com/coral-xyz/anchor/pull/3021
Their resolution was just to have consumers specify a solana-program in their Cargo.toml.

Later on they added the [macro to the anchor prelude](https://github.com/coral-xyz/anchor/pull/3026), but this isn't available in 0.28.0, which we need to support due to the other packages in the workspace that need 0.28.0 and have a path dependency on the SDK.

We could publish v0.6.0 of the SDK which enforces anchor-lang>=0.31 and uses the reexport, and then have the workspace programs depend on 0.5.0 of the SDK rather than the path dependency. But of course this removes some of the utility of having the workspace. I think its okay in this case to explicitly depend on solana-program>=1.16.0, especially since we don't use anything that broke in v2.

ali-bahjati (approved):

guibescos (approved):

PR #2506, opened 2025-03-20
feat(apps/price_pusher): add support for aptos and sui metrics

cctdaniel:
## Summary

Added support for Aptos and Sui metrics

## Rationale

For use in grafana dashboard

## How has this been tested?

- [ ] Added new tests


apps/price_pusher/src/balance-tracker.ts
ali-bahjati, line 1: hmm actually upon a closer look at this file i don't think this is really necessary. can't you move these to their respective chains?

ali-bahjati (approved):
Nice! Can you address my comment before merging?

PR #2507, opened 2025-03-20
feat(lazer): add mapping to improve efficiency of evm contract

Riateche:
## Summary

Add a mapping to allow quick check of trusted signer.

## Rationale

Asked by a consumer.

## How has this been tested?

- [ ] Added new tests


Riateche:
>Do you even need the trustedSigners array now?

It's not really necessary now but we could need it later. It's nice to have the list of signers available in the contract. Mapping doesn't have ability to iterate over items.

PR #2510, opened 2025-03-20
chore(target_chains/cosmwasm): replace xion_testnet with xion_testnet_2

ali-bahjati:
## Summary

Deploy cosmwasm to xion_testnet_2. Also removes the previous xion deployment as it doesn't exist anymore.

### DIFF
--- a/target_chains/cosmwasm/deploy-scripts/package.json
+++ b/target_chains/cosmwasm/deploy-scripts/package.json
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@types/adm-zip": "^0.5.0",
     "@types/yargs": "^17.0.18",
-    "prettier": "catalog:"
+    "prettier": "catalog:",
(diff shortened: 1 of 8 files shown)


target_chains/cosmwasm/deploy-scripts/package.json
```diff
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@types/adm-zip": "^0.5.0",
     "@types/yargs": "^17.0.18",
-    "prettier": "catalog:"
+    "prettier": "catalog:",
+    "ts-node": "^10.9.2"
```
cprussin, line 32: We should probably move `ts-node` to the catalog too since we should be managing it alongside typescript, and ideally we should upgrade the tooling across the whole monorepo simultaneously.

aditya520 (approved):

PR #2521, opened 2025-03-25
fix(hermes): improve TWAP reliability - non-optional price selection and consistent time windows

tejasbadadare:
## Summary

This PR refactors how Hermes fetches the data for the TWAP calculation to improve reliability. 

## Rationale 

- The internal API for TWAP now accepts `window_seconds: u64` and `end_time: RequestTime`, instead of `start_time: RequestTime` and `end_time: RequestTime`.
  - This allows us to first fetch the end messages, then calculate the `start_time` based on the publish time of the end messages. Previously, the actual time difference between start and end messages could differ from the intended window size since the latest timestamp in the cache could be older than the wall time. This change guarantees that the start and end messages are `window_seconds` apart.
  - This also aligns better with the semantics we want to uphold for the public API.
- Added a new `RequestTime::LatestTimeEarliestSlot` request mode to the cache that selects messages with the latest publish time but earliest slot number. This follows the same semantics as `FirstAfter`, and helps choose non-optional (canonical) prices.
  - Many requests for latest TWAP were returning "Update data not found" because the TWAP calculation will fail if it is given a message where `prev_publish_time == publish_time`. This is done to avoid optionality of prices, and we consider only the message where `prev_publish_time < publish_time` as the canonical price for that timestamp. Previously we were using the `RequestTime::Latest` mode, which just returns the latest message for a feed, NOT the latest canonical message. The new mode provides these semantics.
- Improved error handling and logging for TWAP calculation failures. Previously, they were not getting logged, and errors were masked by a generic "Update data not found" message, which made investigation difficult.

## How has this been tested?

- [x] Added new tests
  - Added `test_get_verified_twaps_with_update_data_uses_non_optional_prices` to verify correct message selection
  - Added `test_latest_time_earliest_slot_request_works` to test the `LatestTimeEarliestSlot` cache request mode
  - Tested with local server and Postman, we don't see the high rate of "Update not found" errors anymore, and the timestamps are always `window_seconds` apart.

### DIFF
--- a/apps/hermes/server/src/state/cache.rs
+++ b/apps/hermes/server/src/state/cache.rs
@@ -292,7 +292,7 @@ async fn retrieve_message_state(
                     let mut latest_entry_with_earliest_slot = last_entry;
 
                     // Walk backwards through the sorted entries rather than use `range` since we will only
-                    // have 1-2 entries that have the same publish_time.
+                    // have a couple entries that have the same publish_time.
                     // We have acquired the RwLock via read() above, so we should be safe to reenter the cache here.
                     for (k, v) in key_cache.iter().rev() {
                         if k.publish_time < latest_publish_time {
--- a/apps/hermes/server/src/state/cache.rs
+++ b/apps/hermes/server/src/state/cache.rs
@@ -285,6 +285,28 @@ async fn retrieve_message_state(
         Some(key_cache) => {
             match request_time {
                 RequestTime::Latest => key_cache.last_key_value().map(|(_, v)| v).cloned(),
+                RequestTime::LatestTimeEarliestSlot => {
+                    // Get the latest publish time from the last entry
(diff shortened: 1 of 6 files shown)


apps/hermes/server/src/state/cache.rs
```diff
@@ -285,6 +285,28 @@ async fn retrieve_message_state(
         Some(key_cache) => {
             match request_time {
                 RequestTime::Latest => key_cache.last_key_value().map(|(_, v)| v).cloned(),
+                RequestTime::LatestTimeEarliestSlot => {
```
tejasbadadare, line 288: Thought i had while working on this... should we use this for all our `latest` endpoints, not just TWAP?

tejasbadadare: Yeah that makes sense. I guess they are on the hook for ensuring the price they use is canonical?

PR #2524, opened 2025-03-26
Only build stuff that requires Vercel env to build on Vercel

cprussin:
## Summary

Do not build packages that rely on the Vercel environment using `turbo build`, instead move them to build with `turbo build:vercel`

## Rationale

For many of our UI applications, in order to build the package, a number of environment variables must be present.

The values for these env vars are stored in Vercel, however not everyone who contributes to this repository will have access to Vercel.

This PR modifies the turbo config to not build those packages on `turbo build`. This is generally probably the right thing since we don't want those packages to build in Github Actions either, given the environment is not going to be present in GHA and so the builds would likely fail there too.

After this PR, to build any packages that rely on Vercel environments, you will run `turbo build:vercel`.  Running `turbo build` will build everything else and will exclude anything that depends on the Vercel env.

## How has this been tested?

CI will be the test

### DIFF
--- a/.github/workflows/ci-turbo-build.yml
+++ b/.github/workflows/ci-turbo-build.yml
@@ -33,4 +33,4 @@ jobs:
       - name: Cache for Turbo
         uses: rharkor/caching-for-turbo@v1.5
       - name: Build
-        run: pnpm build:ci
+        run: pnpm run turbo build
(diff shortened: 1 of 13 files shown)


guibescos (approved):
lgtm

PR #2526, opened 2025-03-27
feat(target_chains/ethereum): add tx fee to evm contract

cctdaniel:
- **feat(target_chains/ethreum): add transaction fee handling to fee calculations and state management in pyth contract**
- **feat(governance): add SetTransactionFee action and implementation for transaction fee management**

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts
+++ b/governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts
@@ -20,6 +20,7 @@ import {
   StarknetSetWormholeAddress,
 } from "./SetWormholeAddress";
 import { EvmExecute } from "./ExecuteAction";
+import { SetTransactionFee } from "./SetTransactionFee";
 
 /** Decode a governance payload */
 export function decodeGovernancePayload(
@@ -73,6 +74,8 @@ export function decodeGovernancePayload(
     }
     case "Execute":
       return EvmExecute.decode(data);
+    case "SetTransactionFee":
+      return SetTransactionFee.decode(data);
     default:
       return undefined;
   }
@@ -86,5 +89,6 @@ export * from "./GovernanceDataSourceTransfer";
 export * from "./SetDataSources";
 export * from "./SetValidPeriod";
 export * from "./SetFee";
+export * from "./SetTransactionFee";
 export * from "./SetWormholeAddress";
 export * from "./ExecuteAction";
--- a/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
@@ -330,7 +330,8 @@ abstract contract Pyth is
     function getTotalFee(
         uint totalNumUpdates
     ) private view returns (uint requiredFee) {
-        return totalNumUpdates * singleUpdateFeeInWei();
+        return
+            (totalNumUpdates * singleUpdateFeeInWei()) + transactionFeeInWei();
     }
 
     function findIndexOfPriceId(
--- a/governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
+++ b/governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts
@@ -16,6 +16,7 @@ export const TargetAction = {
   RequestGovernanceDataSourceTransfer: 5,
   SetWormholeAddress: 6,
   SetFeeInToken: 7,
+  SetTransactionFee: 8,
 } as const;
 
 export const EvmExecutorAction = {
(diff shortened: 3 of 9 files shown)


governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts
ali-bahjati, line 1: You missed to update PythGovernanceHeader to decode the action properly
https://github.com/pyth-network/pyth-crosschain/blob/main/governance/xc_admin/packages/xc_admin_common/src/governance_payload/PythGovernanceAction.ts#L26

target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -330,7 +330,8 @@ abstract contract Pyth is
     function getTotalFee(
```
ali-bahjati, line 330: there's another function that you need to update 
https://github.com/pyth-network/pyth-crosschain/blob/fe75b45e918d9915a1cca4458347c64481fdda9e/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol#L83

target_chains/ethereum/contracts/forge-test/PythGovernance.t.sol
ali-bahjati, line 1: i see that you've replaced the tests here entirely. can you remove that test file entirely?

nit: perhaps you can say a couple of words about it in the description and mention what has changed, whether anything is skipped, ..

target_chains/ethereum/contracts/forge-test/PythGovernance.t.sol
```diff
@@ -0,0 +1,475 @@
...
+        PythGovernance(address(pyth)).executeGovernanceInstruction(vaa);
+    }
+
+    function testInvalidGovernanceDataSource() public {
+        // Test with wrong emitter
+        bytes memory data = abi.encodePacked(
+            MAGIC,
+            uint8(GovernanceModule.Target),
+            uint8(GovernanceAction.SetValidPeriod),
+            bytes32("ethereum"),
```
ali-bahjati, line 168: this is wrong

target_chains/ethereum/contracts/forge-test/PythGovernance.t.sol
```diff
@@ -0,0 +1,475 @@
...
+
+        PythInternalStructs.DataSource[]
+            memory newDataSources = new PythInternalStructs.DataSource[](1);
+        newDataSources[0] = PythInternalStructs.DataSource(
+            1,
+            0x0000000000000000000000000000000000000000000000000000000000001111
+        );
+
+        vm.expectEmit(true, true, true, true);
+        emit DataSourcesSet(oldDataSources, newDataSources);
```
ali-bahjati, line 216: why is this here? i see it in many other places. we expect the call below to emit them not manually emitting them.

ali-bahjati (changes requested):
Nice! and thanks for moving the tests to foundry. Please address my comments. Also can you update contract version so we don't miss it? (you can use a alpha next version because we are going to have more changes there)

governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts
cctdaniel: hmm this is updated [here](https://github.com/pyth-network/pyth-crosschain/pull/2526/files#diff-46370c63540cbcf9a69fffa014fdbd1407a2d137c7a3f83aac2d1dca499e5671R19-R51)

target_chains/ethereum/contracts/forge-test/PythGovernance.t.sol
cctdaniel: this is not manually emitting them, docs [here](https://book.getfoundry.sh/cheatcodes/expect-emit)

but basically whats happening is:
1. vm.expectEmit(true, true, true, true) tells Forge to expect an event where:
- Each true corresponds to checking topic1, topic2, topic3, and data respectively
- All true means "check everything in the event"
2. The subsequent emit defines the EXPECTED event parameters
3. When executeGovernanceInstruction(vaa) runs, Forge compares its emitted event against this expected template

This pattern is used because we can't directly capture and inspect events after they're emitted. Instead, we declare what you expect before the action, then Forge verifies the actual emission matches.

ali-bahjati: ah my bad here. thanks.

governance/xc_admin/packages/xc_admin_common/src/governance_payload/index.ts
ali-bahjati: my bad.

ali-bahjati (approved):

PR #2530, opened 2025-03-28
chore(apps/price_pusher): update jito bundle endpoint

cctdaniel:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/price_pusher/src/solana/solana.ts
+++ b/apps/price_pusher/src/solana/solana.ts
@@ -177,7 +177,7 @@ export class SolanaPricePusherJito implements IPricePusher {
         "https://bundles.jito.wtf/api/v1/bundles/tip_floor",
       );
       if (!response.ok) {
-        this.logger.warn(
+        this.logger.error(
           { status: response.status, statusText: response.statusText },
           "getRecentJitoTips http request failed",
         );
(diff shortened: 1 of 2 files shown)


apps/price_pusher/src/solana/solana.ts
```diff
@@ -174,7 +174,7 @@ export class SolanaPricePusherJito implements IPricePusher {
   async getRecentJitoTipLamports(): Promise<number | undefined> {
     try {
       const response = await fetch(
-        "http://bundles-api-rest.jito.wtf/api/v1/bundles/tip_floor",
+        "https://bundles.jito.wtf/api/v1/bundles/tip_floor",
       );
       if (!response.ok) {
         this.logger.error(
```
ali-bahjati, line 180: shall we make it warning? because if it doesn't work we default to something else (and hopefully other alerts should catch it up if it results into not landing txs)

ali-bahjati (approved):

apps/price_pusher/src/solana/solana.ts
cctdaniel: yup good point will make the change

KemarTiti (approved):

PR #2534, opened 2025-03-31
refactor(argus): remove fortuna-specific keeper code

tejasbadadare:
## Summary

Remove keeper logic and configs specific to Fortuna/Entropy so that the project compiles again, and is a clean(er) slate on which we can implement Argus logic. 

Not exhaustive, more cleanup will follow as we specify the Argus/Pulse architecture. Suggestions for more stuff to chop are welcome!


ali-bahjati (approved):
LGTM.

PR #2535, opened 2025-03-31
feat: add a staking endpoint to return a stake summary per account

cprussin:
## Summary

Adds an endpoint to the staking app to return a stake summary per account.

## Rationale

This is required to calculate Fogo flames

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/governance/pyth_staking_sdk/src/pyth-staking-client.ts
+++ b/governance/pyth_staking_sdk/src/pyth-staking-client.ts
@@ -21,6 +21,8 @@ import {
   Transaction,
   TransactionInstruction,
 } from "@solana/web3.js";
+import { JSONParser } from "@streamparser/json";
+import { z } from "zod";
 
 import {
   GOVERNANCE_ADDRESS,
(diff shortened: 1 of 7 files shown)


governance/pyth_staking_sdk/src/pyth-staking-client.ts
```diff
@@ -1031,4 +1037,114 @@ export class PythStakingClient {
...
+  z.object({
+    value: z.object({
+      account: z.object({
+        data: z
+          .array(z.string())
+          .min(1)
+          .transform((data) =>
+            // Safe because `min(1)` guarantees that `data` is nonempty
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            Buffer.from(data[0]!, "base64"),
```
keyvankhademi, line 1129: nit: can we use an if to avoid having !?

keyvankhademi (approved):

governance/pyth_staking_sdk/src/pyth-staking-client.ts
cprussin: In this case we actually want `!` because the `min(1)` line from Zod already guarantees the length, but can't encode that in the type.  I generally prefer letting Zod verify this kind of thing since it will generate nice consistent error messages, but it requires the `!` to avoid having a second null check at runtime.  In this case since it's so many accounts, it's not huge but it is actually a relevant optimization.

I usually wrap this kind of thing into a `nonEmptyArray` Zod helper which returns `[T, ...T[]]` instead of `T[]` which typescript can then guarantee is nonempty...

PR #2542, opened 2025-04-01
feat(argus): Implement Keeper state and main request processing loop

tejasbadadare:
## Summary

Implement KeeperState which holds the state of open Pulse requests, and the core loop for operating on these open requests, i.e. invoking callback execution.

## Rationale

The motivation here was to decouple the blockchain state implementation details from the core logic of the Keeper. 

The Keeper should only be concerned with storing the current set of open requests (`KeeperState.pending_requests`), and operating on them with some kind of predictable policy (`process_pending_requests()` ). This function spawns fulfillment tasks (and retries) for requests that are ready to be fulfilled. (See the docstring for a more in-depth flow.)

This approach is easier to unit test, versus the Fortuna approach that uses an MPSC channel to pass block ranges from a blockchain watcher to a consumer that optimistically tries to execute with backoffs.

A different process (not implemented yet) will be responsible for keeping the set of requests updated. Since we are keeping requests on-chain, this will be a simple task that polls Pulse's view function `getFirstActiveRequests` every second or so and calls `KeeperState.update()`. We can optimize latency by also subscribing to events and proactively updating the state alongside of the polling.

## How has this been tested?

- [x] Added new tests

## Next up:
- Implement the RequestFulfiller
- Find a nice way to share utilities between Fortuna and Argus, such as EscalationPolicy (likely a common SDK)
- Figure out ethers providers, wire things up to the actual blockchain


apps/argus/src/keeper/state.rs
```diff
@@ -0,0 +1,592 @@
...
+
+    // The amount of time a request can retry until it's considered unfulfillable and is ignored.
+    pub failure_timeout_seconds: u64,
+
+    // Policy that defines the internal retries for landing the callback execution tx.
+    // Increases gas and fees until the tx lands.
+    pub escalation_policy: EscalationPolicy,
+
+    // The Hermes endpoint to fetch price data from
+    pub hermes_url: Url,
```
ali-bahjati, line 82: maybe this can become a HermesClient at some point?

apps/argus/src/keeper/state.rs
```diff
@@ -0,0 +1,592 @@
...
+            // Check if request has been around too long without success
+            let request_age_seconds = current_time - request.publish_time;
+            if !fulfillment_task.success && request_age_seconds > self.failure_timeout_seconds {
+                // Log that this request has been around too long without successful fulfillment
+                error!(
+                    "Request #{} has exceeded timeout of {} minutes without successful fulfillment",
+                    request.sequence_number, self.failure_timeout_seconds
+                );
+
+                // Emit metrics here for monitoring/alerting
```
ali-bahjati, line 244: add TODO

ali-bahjati (approved):
I like it a lot that you are separating different layers and it's great. How do you foresee the overall structure of the code. Is it a shared memory (keeper state) and different tasks interacting with it? If so I'd consider adopting Hermes approach. For instance HermesUrl is not related to the main BD logic. Or you like to have tasks communicating via message passing (like actor model).

apps/argus/src/keeper/state.rs
```diff
@@ -0,0 +1,592 @@
+// Need to be able to unit test whatever we come up with
```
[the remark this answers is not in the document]
tejasbadadare: Ah sorry, forgot to remove this after prototyping!

apps/argus/src/keeper/state.rs
tejasbadadare: Yep definitely, had that thought as well. One doesn't exist for rust (i punted on that when working on Native...), will make one when we implement TaskFulfiller

tejasbadadare:
@ali-bahjati Great! Glad you like the approach :) As far as communication between the layers goes...
> Is it a shared memory (keeper state) and different tasks interacting with it? If so I'd consider adopting Hermes approach. For instance HermesUrl is not related to the main BD logic.  Or you like to have tasks communicating via message passing (like actor model).
Yeah I was debating this myself, had a little think on this, lmk if you agree:

This PR currently assumes that KeeperState would be shared state protected by a RwLock. The request producer obtains a write lock when it wants to update the set of requests, and the consumer mostly obtains non-blocking read locks. I think it's relatively simple this way, and we would gate external access to the state to the update() method to avoid layer violations. We should definitely monitor for lock contention though. 

If we went for an actor model, we would probably create a channel structure like:
```
RequestProducerActor --> KeeperActor <--> FulfillerActor
```
The actor model can be easier to reason about imo, but potentially harder to debug since everything is connected by channels (or similar message passing) rather than function calls. It would be useful though if/when we add other ways to update the state, like via events.

Not super sold either way - for now I think I'll separate concerns better between the Keeper and the Fulfiller (like the hermes_url you mentioned), keep it as shared ownership over the state with locks, and check for perf issues there when its wired up. It should be relatively easy to replace shared state with channels since the interfaces between the layers would be unchanged.

tejasbadadare:
> I like it a lot that you are separating different layers and it's great. How do you foresee the overall structure of the code. Is it a shared memory (keeper state) and different tasks interacting with it? If so I'd consider adopting Hermes approach. For instance HermesUrl is not related to the main BD logic. Or you like to have tasks communicating via message passing (like actor model).

PR #2546, opened 2025-04-02
feat(entropy): Add callback failure states to the contract

jayantk:
## Summary

Currently, the Entropy keeper cannot invoke a callback if the underlying callback fails. This creates confusion for users, who can't distinguish between "the keeper is offline" and "my callback code reverts". 

This change creates a failure flow for callbacks that allows the keeper's transaction to succeed while explicitly reporting a failure of the underlying callback. Requests with failing callbacks remain on the blockchain and can be re-invoked by users. This feature allows users to gracefully recover in cases where the callback fails for transient reasons. (In the future, it will also allow recovery in the case where the callback uses too much gas.)

Note that this PR doesn't cover the case where the callback uses too much gas. I will send a follow-up PR for that. 

## How has this been tested?

- [x] Added new tests


target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -480,26 +490,76 @@ abstract contract Entropy is IEntropy, EntropyState {
...
-        }
-        if (len != 0) {
-            IEntropyConsumer(callAddress)._entropyCallback(
-                sequenceNumber,
-                provider,
+        // Requests that haven't been invoked yet will be invoked safely (catching reverts), and
+        // any reverts will be reported as an event. Any failing requests move to a failure state
+        // at which point they can be recovered. The recovery flow invokes the callback directly
+        // (no catching errors) which allows callers to easily see the revert reason.
+        if (req.status == EntropyConstants.STATUS_CALLBACK_NOT_STARTED) {
```
jayantk, line 497: Note that this is a behavior change at the moment. This means that we won't be able to upgrade the contracts without upgrading the keeper.

However, in the next PR I will additionally condition this on the request having a gas limit, which means that the contract will be 100% backward compatible. (i.e., you don't get the failure flow until the provider explicitly sets a gas limit for callbacks)

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -938,12 +939,13 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents {
         );
     }
 
+    // TODO: restore this test once providers have custom gas limits, which allow us to toggle between
+    // the old and new failure behavior.
```
jayantk, line 943: (restoring this test will ensure backward compatibility of the flow, per the comment above)

target_chains/ethereum/entropy_sdk/solidity/EntropyStructs.sol
```diff
@@ -59,8 +59,8 @@ contract EntropyStructs {
         address requester;
         // If true, incorporate the blockhash of blockNumber into the generated random value.
         bool useBlockhash;
-        // If true, the requester will be called back with the generated random value.
-        bool isRequestWithCallback;
-        // There are 2 remaining bytes of free space in this slot.
+        // Status flag for requests with callbacks. See EntropyConstants for the possible values of this flag.
+        uint8 status;
```
jayantk, line 63: (bools are represented as uint8 so this is fine. I'm reusing this flag because I want the remaining 2 bytes for the gas limit)

target_chains/ethereum/entropy_sdk/solidity/EntropyConstants.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache 2
+
+library EntropyConstants {
+    // Status values for Request.status //
+    // not a request with callback
+    uint8 public constant STATUS_NO_CALLBACK = 0;
```
m30m, line 6: how about naming the library EntropyStatusConstants and remove the `STATUS_` prefix here?

m30m, line 6: the name of this variant is a bit confusing. Maybe something like uninitialized?

target_chains/ethereum/entropy_sdk/solidity/EntropyStructs.sol
m30m, line 63: maybe name this callback_status?

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -480,26 +490,75 @@ abstract contract Entropy is IEntropy, EntropyState {
...
+            req.status = EntropyConstants.STATUS_CALLBACK_IN_PROGRESS;
+            bool success;
+            bytes memory ret;
+            (success, ret) = callAddress.excessivelySafeCall(
+                gasleft(), // TODO: providers need to be able to configure this in the future.
```
m30m, line 502: I think this actually needs to be exactly what the requester asked for and not what provider specifyes

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -480,26 +490,75 @@ abstract contract Entropy is IEntropy, EntropyState {
...
+                    req.requester,
+                    sequenceNumber,
+                    userRandomNumber,
+                    providerRevelation,
+                    randomNumber,
+                    ret
+                );
+                req.status = EntropyConstants.STATUS_CALLBACK_FAILED;
+            } else {
+                // The callback ran out of gas
```
m30m, line 535: I think there are other failure scenarios where ret.length is 0.

m30m (approved):

PR #2551, opened 2025-04-03
refactor(pythnet/pythnet_sdk): add test for update data with fixture

ali-bahjati:
## Summary

add test for update data with fixture

## Rationale

This is a good test but also a good tool to deserialize update data for troubleshooting. If we get many requests to parse the update data we can add it as an API to Hermes or make a CLI for it but I don't think it's necessary for now.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/pythnet/pythnet_sdk/src/wire.rs
+++ b/pythnet/pythnet_sdk/src/wire.rs
@@ -139,10 +139,13 @@ pub mod v1 {
 
 #[cfg(test)]
 mod tests {
-    use crate::wire::{
-        array,
-        v1::{AccumulatorUpdateData, Proof},
-        Deserializer, PrefixedVec, Serializer,
+    use crate::{
+        messages::Message,
+        wire::{
+            array, from_slice,
+            v1::{AccumulatorUpdateData, Proof},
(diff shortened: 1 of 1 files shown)


pythnet/pythnet_sdk/src/wire.rs
```diff
@@ -520,4 +523,20 @@ mod tests {
...
+        let update = hex::decode(update).unwrap();
+        let update = AccumulatorUpdateData::try_from_slice(&update).unwrap();
+        let Proof::WormholeMerkle { vaa: _, updates } = update.proof;
+        println!("Updates: {:?}", updates.len());
```
darunrs, line 534: Did you want to keep these prints in? Should they be assertions instead?

darunrs (approved):

tejasbadadare (approved):

pythnet/pythnet_sdk/src/wire.rs
ali-bahjati: The assertion is mostly the `unwrap` part. I kept the prints for debugging the contents.

PR #2555, opened 2025-04-03
feat: use sendTransaction for OIS

keyvankhademi:
## Summary

Use wallet send transaction for OIS instead of sign and send

## Rationale

For multisig wallets it takes time for everyone to sign and causes block hash to expire.

## How has this been tested?

- [ ] Added new tests
- [ x] Manually tested the code


cprussin (approved):

apps/staking/src/hooks/use-api.tsx
```diff
@@ -195,13 +195,15 @@ const useApiContext = (
               publicKey: wallet.publicKey,
               signAllTransactions: wallet.signAllTransactions,
               signTransaction: wallet.signTransaction,
```
guibescos, line 197: `signTransaction` and `signAllTransactions` are not used anymore?

governance/pyth_staking_sdk/src/utils/transaction.ts
```diff
@@ -20,5 +18,7 @@ export const sendTransaction = async (
     {},
   );
 
-  return sendTransactions(transactions, connection, wallet);
+  for (const transaction of transactions) {
+    await wallet.sendTransaction(transaction.tx, connection);
```
guibescos, line 22: remind me why the interface needs a connection?

guibescos (approved):

governance/pyth_staking_sdk/src/utils/transaction.ts
keyvankhademi: not sure tbh, my guess is that some wallets will use this connection to send the tx instead of their own.

apps/staking/src/hooks/use-api.tsx
keyvankhademi: Yes they are not used afaik, but I thought it's better to keep them.

PR #2557, opened 2025-04-04
feat(pyth-lazer-publisher-sdk): Add SDK for Publisher Transactions

darunrs:
## Summary

Created new publisher_sdk which will store all sdk related code for publishers. Publishers is now more of a gneral term for anyone who publishes transactions to Lazer. Right now that's only Feed Publishers but in the future, there will be Governance, and maybe others too. The new publisher_sdk has proto files for which types are generated. Pyth Lazer will use these for encoding/decoding incoming transactions to the system in the future. 

## Rationale

Lazer needs to support more types of input and be capable of changing its input schema as updates happen. Using a new format for input provides this flexibility. Using protobuf for message passing provides additional flexibility for updating schemas and also improves the message compression size. 

## How has this been tested?

- [ ] Added new tests

Was able to generate types successfully.


lazer/sdk/rust/protocol/build.rs
```diff
@@ -0,0 +1,12 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
+    let proto_files = vec![
+        "../../proto/pyth_lazer.proto",
+        "../../proto/publisher_update.proto",
+    ];
+
+    prost_build::compile_protos(&proto_files, &["../../proto"])?;
```
merolish, line 9: Nice. A previous colleague wanted to use protox to avoid the protoc dependency, but the latter is readily available anywhere that matters.

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,29 @@
...
+  uint64 batch_timestamp_us = 2;
+}
+
+message PublisherUpdate {
+  uint32 price_feed_id = 1;
+  uint64 source_timestamp_us = 2;
+  uint64 publisher_timestamp_us = 3;
+  oneof update {
+    PriceUpdateV1 price_update_v1 = 4;
+    FundingRateUpdateV1 funding_rate_update_v1 = 5;
```
merolish, line 16: Assuming we'll be able to put future message types here compatibly.

darunrs: Yep

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
...
+package pyth_lazer_transaction;
+
+message PublisherUpdate {
+  repeated FeedUpdate updates = 1;
+  uint64 batch_timestamp_us = 2;
+}
+
+message FeedUpdate {
+  oneof update {
+    PriceUpdateV1 price_update_v1 = 3;
```
bplatak, line 12: ❔ why start from 3?

lazer/sdk/js/src/generated/proto.js
tejasbadadare, line 1: Should we gitignore these generated files?

lazer/sdk/rust/protocol/build.rs
tejasbadadare, line 1: Can we add some comments to this file to demystify the build process? Hermes's [build.rs](https://github.com/pyth-network/pyth-crosschain/blob/d3cc7dff00f54bf396f12eec442b615381550bde/apps/hermes/server/build.rs) does a good job of this i think

lazer/sdk/js/src/generated/proto.js
keyvankhademi: i think we should keep them and update them with every change. We probably want to export them as well so everyone can import it in their client codes.

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package pyth_lazer_transaction;
+
+message PublisherUpdate {
+  repeated FeedUpdate updates = 1;
+  uint64 batch_timestamp_us = 2;
```
keyvankhademi, line 7: i think it's better to specify which timestamp it is referring to I assuming it's `publisher_timestamp` as the system time when publisher sends this update.

lazer/sdk/js/src/generated/proto.js
bplatak: IMHO we should keep the generated files committed but we need to be careful to keep them in sync (a CI step could be useful here). We shouldn't force users of this library to depend on protobuf or to have to run code generation as part of their build process.

lazer/sdk/proto/publisher_update.proto
bplatak: Better yet we should add comments to the proto files - they will carry across to the generated sources. We could also use this for API docs

darunrs: Yeah its the publisher timestamp when sending the update. Perhaps something like `publisher_sending_update_timestamp_us`? A little wordy but its clear.

lazer/sdk/rust/protocol/build.rs
darunrs: Good point! I'll add some comments here. Thanks for the link!

lazer/sdk/js/src/generated/proto.js
darunrs: Yeah I think we can. I'll add ignore and update the package.json to ensure the protoc typegen is done before publish.

lazer/sdk/proto/publisher_update.proto
darunrs: Ah good callout. Numbering wasn't reset during some changes. I've reset them starting from 1.

lazer/sdk/js/src/generated/proto.js
darunrs: Yeah I think I'll get Connor's input here on what's the right thing to do regarding protoc and npm library publishing.

lazer/sdk/proto/publisher_update.proto
darunrs: Will add comments to the proto file for sure.

lazer/sdk/js/src/generated/proto.js
cprussin: I'm generally not a fan of having checked in generated files.  It's effectively a cache layer and comes with all the problems of maintaining a cache, and it's very easy to mask problems with the build setup by checking in generated files, which later becomes very painful to resolve.  IMO it's better to make the build reliable and reproducible.

There are, of course, exceptions to that and you'll need to use your judgement on if this is one such exception.  I haven't touched protobufs in a long time so I can't speak intelligently to the tradeoffs here.

I also believe that checking in generated files is generally not the optimization that people think it is, if dev tooling is actually done properly:

- You'll still want your CI running the codegen and checking it to ensure the codegen is done properly and the codegen sources are properly configured, so it doesn't save anything there (UNLESS you just omit the codegen from your CI in which case you are losing any kind of validation that the build ACTUALLY works from source)
- You should expect your developers to have the tools they need to build from source.  I don't buy the idea that it's OK to rely on protoc for the build, but we don't want devs to have protoc on their machine.  What about when someone needs to update things?  If we make it easy to develop without the full suite of dependencies, then we are in effect making it hard to get reproducible development environments.  We then can't be surprised when it's hard to reproduce things when the need arises down the line.
- As an extension to the last point, what about when the underlying tools have upgrades?  Will we even know if things break?  What if things break and far later we have some need to change things, and the tools have changed underneath us in an incompatible way?  Checking in codegen files just makes it easy to not know these problems are happening until it becomes a serious cost.

Again, there are nuances here and your situation may warrant an exception, but I generally recommend caution when checking in generated files because there are real, long term costs

cprussin: (for posterity, clarifying from our Slack conversation) To be clear, my general advice is to not check generated files into the source code, but to include them in the built package that you send to NPM.  In other words, building the package should require generating the files, but _installing_ the package from npm should not.

lazer/sdk/proto/publisher_update.proto
keyvankhademi: `publisher_timestamp_us` should be good enough, we use it a lot when referring to the publisher timestamp when sending the update.

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package pyth_lazer_transaction;
```
Riateche, line 3: Please specify `optional` for all non-repeated fields, as recommended [here](https://protobuf.dev/programming-guides/proto3/#field-labels). Implicit fields are still optional but result in codegen that prevents detection of missing fields when decoding. For every optional field that's really optional, we should also document when it's correct to actually omit it (e.g. for prices where it no known value). And we can add a comment at the top that every field is actually required unless documented otherwise.

lazer/sdk/js/.gitignore
```diff
@@ -20,3 +20,6 @@ node_modules/
 dist/
```
Riateche, line 20: These SDKs are for end-users, so I think it will be confusing to them if we have publisher-related code in the same package. Maybe it's better to create a separate set of packages for publishers under `lazer/publisher_sdk/{js, rust}`?

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
...
+  uint64 publisher_timestamp_us = 3;
+  optional int64 price = 4;
+  optional int64 best_bid_price = 5;
+  optional int64 best_ask_price = 6;
+}
+
+message FundingRateUpdateV1 {
+  uint32 feed_id = 1;
+  uint64 source_timestamp_us = 2;
+  uint64 publisher_timestamp_us = 3;
```
Riateche, line 29: I think it's better to deduplicate the fields here:
```
message FeedUpdate {
  uint32 feed_id = 1;
  uint64 source_timestamp_us = 2;
  uint64 publisher_timestamp_us = 3;
  oneof update_kind {
    PriceFeedUpdate price_feed_update = 1;
    FundingRateUpdate funding_rate_update = 2;
  }
}
```

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
...
+package pyth_lazer_transaction;
+
+message PublisherUpdate {
+  repeated FeedUpdate updates = 1;
+  uint64 batch_timestamp_us = 2;
+}
+
+message FeedUpdate {
+  oneof update {
+    PriceUpdateV1 price_update_v1 = 1;
```
Riateche, line 12: We probably won't need V1 suffix because protobuf has proper schema evolution. If we need it we can rename it later.

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
...
+    FundingRateUpdateV1 funding_rate_update_v1 = 2;
+  }
+}
+
+message PriceUpdateV1 {
+  uint32 feed_id = 1;
+  uint64 source_timestamp_us = 2;
+  uint64 publisher_timestamp_us = 3;
```
Riateche, line 20: If we have publisher timestamp in the top level message we probably don't need it here.

lazer/sdk/proto/pyth_lazer_transaction.proto
```diff
@@ -0,0 +1,21 @@
...
+import "publisher_update.proto";
+
+enum TransactionSignatureType {
+  ed25519 = 0;
+}
+
+message SignedLazerTransaction {
+  TransactionSignatureType signature_type = 1;
+  bytes signature = 2;
+  bytes transaction = 3;
```
Riateche, line 14: I would call this field "payload".

lazer/sdk/proto/publisher_update.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package pyth_lazer_transaction;
+
+message PublisherUpdate {
```
Riateche, line 5: We need `publisher_id` field here.

lazer/sdk/proto/pyth_lazer_transaction.proto
```diff
@@ -0,0 +1,21 @@
...
+}
+
+message LazerTransaction {
+  oneof payload {
+    PublisherUpdate publisher_updates = 1;
```
Riateche, line 19:
```suggestion
    PublisherUpdate publisher_update = 1;
```

lazer/sdk/rust/protocol/Cargo.toml
```diff
@@ -15,6 +16,7 @@ derive_more = { version = "1.0.0", features = ["from"] }
 itertools = "0.13.0"
 rust_decimal = "1.36.0"
 base64 = "0.22.1"
+prost = "0.13.5"
```
Riateche, line 19: Have you considered using `protobuf` and `protobuf-codegen`? I think I'm going to try it for snapshots and see how it works. I think it can be a bit better because it has better typing in the generted code (especially for enums) and doesn't require protoc to be installed. But at this point I'm not sure which one is better.

Riateche (changes requested):

lazer/sdk/proto/publisher_update.proto
ali-bahjati: nit here: use `timestamp` type (see (here)[https://protobuf.dev/best-practices/dos-donts/])

lazer/sdk/proto/publisher_update.proto
ali-bahjati: I recommend seeing how it works internally. I found this one useful: https://martin.kleppmann.com/2012/12/05/schema-evolution-in-avro-protocol-buffers-thrift.html

lazer/sdk/js/.gitignore
darunrs: Ahh good point. I will make a new set of packages for them then. The index.ts example for it can include generating a payload, signing it, etc.

lazer/sdk/proto/publisher_update.proto
darunrs: Yeah I was swinging between the current iteration and having the expected shared fields in the top level message. I thought it looked a bit strange but I think it makes more sense semantically. I'll switch to it.

lazer/sdk/rust/protocol/Cargo.toml
darunrs: Ah that's an interesting suggestion. I'll give that a go and compare the generated code. Thanks!

lazer/sdk/js/.gitignore
darunrs: I'll do it in a separate PR when we are closer to wanting publishers to work with the agent. We have types for Publishers already defined in the rust protocol.

lazer/sdk/js/src/generated/proto.js
darunrs: Added a turbo file and slightly changed the package.json scripts. Please let me know if this is the correct convention for code gen and deployment!

lazer/sdk/proto/publisher_update.proto
darunrs: I was thinking that publisher ID would be something only Relayer knows about. The ID would map to an access token and a public key. The access token used to connect to Relayer determines what public key is accepted, and that key is used to verify the signature. In that scenario, publishers would not know/care what ID they are, and would not include it in any messages. Only the signature.

Riateche: I'd like to use this format in the kafka/nats messages and read them in the aggregator (and add governance instructions as another type of transaction along with PublisherUpdate). Aggregator won't know access tokens so it needs another way to know the publisher id.

darunrs: Well, we do have to attach additional information anyway to the payloads before it goes to NATS. For example, if we reject the update due to price deviation, invalid timestamp, etc. We also attach the information like publisher ID. This is being done in Relayer. Is the consideration that Aggregator will fully handle all of this, and Relayer simplifies to taking data in and verifying it before sending to NATS? I'm good with that approach.

lazer/sdk/rust/protocol/Cargo.toml
darunrs: Yeah I do like the output format for this as well. Having the option for a pure rust compiler is also nice. I'll adopt this.

lazer/sdk/js/.gitignore
darunrs: Hmm actually I'll do it in this PR. I can slowly move any relevant types over if it makes sense.

lazer/sdk/proto/publisher_update.proto
Riateche: With publisher signatures, it will no longer be possible to modify the payload at the relayer. It will only be able to reject the whole message. So for every check we have in relayer today we'll have to decide whether to move it to aggregator or reject at relayer. 

The price deviation is the trickiest to handle. Option one is to remove it (it was always meant as a temporary measure). Option two is implement the pythnet-based publisher and move the whole logic to the aggregator.

Note that your definition of `SignedLazerTransaction` doesn't include public key which is necessary for signature verification. So to verify the signature, we should be able to derive the public key from payload and state. The state will contain public keys of publishers so having publisher_id in the payload should be enough. However, now that I think about it, it's nice to have both. Having public key in `SignedLazerTransaction` will allow anyone to verify a transaction on its own, without state.

cprussin (approved):
Nice!  Overall js packaging stuff looks great, just a few minor points to improve but overall lgtm

lazer/publisher_sdk/js/README.md
```diff
@@ -0,0 +1,17 @@
...
+See [contributing.md](docs/contributing/contributing.md) for information on how to develop or contribute to this project!
+
+## Installation and build
+
+### pnpm
+
+```
+cd to crosschain root
+$ pnpm install
+$ pnpm turbo --filter @pythnetwork/pyth-lazer-publisher-sdk build
```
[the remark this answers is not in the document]
darunrs: Yeah that makes sense. I will add in a followup PR an example showing how to use the SDK to create some of the types, sign them, and then encode them.

lazer/publisher_sdk/js/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./generated/proto.js";
```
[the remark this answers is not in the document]
darunrs: By this, do you mean changing main and types in package.json to point to the proto.js instead of index.ts?

lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
```diff
@@ -0,0 +1,34 @@
...
+  // Type and signature should match
+  optional TransactionSignatureType signature_type = 1;
+
+  // Signature derived by signing payload with private key
+  optional bytes signature = 2;
+
+  // a LazerTransaction message which is already encoded with protobuf as bytes
+  // The encoded bytes are what should be signed
+  optional bytes payload = 3;
+}
```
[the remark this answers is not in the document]
ali-bahjati: I like to have a fixed transaction layout and then have customisability in the payload layer. I see transaction as something for sequencer and content of it as something for the aggregator. It's good to have permissioning over who does what. And then, you can say that the sequencer only streams txs from verified pubkeys.

Riateche (approved):
Very good!

keyvankhademi (approved):

PR #2559, opened 2025-04-04
feat(entropy): Add on-chain gas limits and mark out-of-gas failures on-chain

jayantk:
## Summary

Follow up to the last PR around allowing providers to mark on-chain callbacks as failed. The previous PR covers cases where the contract manually reverts. This PR covers the case where the callback runs out of gas. 

This PR gives each callback a gas limit which is stored in the on-chain struct. Providers can set a default gas limit, or users can manually specify a different gas limit when making a request. We mark callbacks as failed if they revert with an out-of-gas error *and* the transaction had enough gas to forward -- there are some EVM nuances here that we need to be careful about. See inline comments for details.

## Rationale

Explicitly marking on-chain requests as failed will make the state transitions of entropy much clearer to end users.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -521,6 +521,11 @@ abstract contract Entropy is IEntropy, EntropyState {
             bytes memory ret;
             uint256 startingGas = gasleft();
             (success, ret) = callAddress.excessivelySafeCall(
+                // Warning: the provided gas limit below is only an *upper bound* on the gas provided to the call.
+                // If the current context has less gas the desired value, 63/64ths of the current context's gas will
+                // be provided instead. (See CALL opcode docs here https://www.evm.codes/?fork=cancun#f1)
+                // Consequently, out-of-gas reverts need to be handled carefully to ensure that the callback
+                // was truly provided with a sufficient amount of gas.                
                 uint256(req.gasLimit10k) * 10000,
                 256, // copy at most 256 bytes of the return value into ret.
                 abi.encodeWithSelector(
@@ -543,7 +548,9 @@ abstract contract Entropy is IEntropy, EntropyState {
                 );
                 clearRequest(provider, sequenceNumber);
             } else if (ret.length > 0 || startingGas - endingGas >= uint256(req.gasLimit10k) * 10000) {
-                // Callback reverted for some reason. Note ret.length == 0 implies out of gas.
+                // The callback reverted for some reason.
+                // If ret.length == 0, then the callback ran out of gas. In this case, ensure that the
+                // callback was provided with sufficient gas.
                 emit CallbackFailed(
                     provider,
                     req.requester,
@@ -555,7 +562,13 @@ abstract contract Entropy is IEntropy, EntropyState {
                 );
                 req.callbackStatus = EntropyStatusConstants.CALLBACK_FAILED;
             } else {
-                require(false, "provider needs to send more gas");
+                // Callback reverted by running out of gas, but the calling context did not have enough gas
+                // to run the callback. This is a corner case that can happen due to the nuances of gas passing
+                // in calls (see the comment on the call above).
+                // 
+                // (Note that reverting here plays nicely with the estimateGas RPC method, which binary searches for 
+                // the smallest gas value that causes the transaction to *succeed*. See https://github.com/ethereum/go-ethereum/pull/3587 )
+                revert EntropyErrors.InsufficientGas();
             }
         } else {
             // This case uses the checks-effects-interactions pattern to avoid reentry attacks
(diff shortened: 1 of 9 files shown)


target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -961,9 +959,69 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents {
...
+            provider1,
+            assignedSequenceNumber
+        );
+        assertEq(reqAfterReveal.sequenceNumber, 0);
+    }
 
     function testRequestWithRevertingCallback() public {
+        uint32 defaultGasLimit = 100000;
+        vm.prank(provider1);
+        random.setDefaultGasLimit(defaultGasLimit);
```
jayantk, line 1023: these tests require the provider to opt in to the new flow by setting their gas limit

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -703,6 +808,16 @@ abstract contract Entropy is IEntropy, EntropyState {
         );
     }
 
+    // Rounds the provided quantity of gas into units of 10k gas.
+    // If gas is not evenly divisible by 10k, rounds up.
+    function roundGas(uint32 gas) internal pure returns (uint16) {
```
m30m, line 813: naming doesn't imply it's dividing it by 10k. Maybe something like `convertGastoGas10k`?

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -703,6 +808,16 @@ abstract contract Entropy is IEntropy, EntropyState {
...
+        uint32 gas10k = gas / 10000;
+        if (gas10k * 10000 < gas) {
+            gas10k += 1;
+        }
+        return SafeCast.toUint16(gas10k);
```
m30m, line 818: As a dev, I might set the gas limit to a gazillion just for testing and that would cause a revert here.
I don't think the revert error is easily parsable, so I'd rather test manually and revert with a better reason.

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -687,6 +778,20 @@ abstract contract Entropy is IEntropy, EntropyState {
...
+    function setDefaultGasLimit(uint32 gasLimit) external override {
+        EntropyStructs.ProviderInfo storage provider = _state.providers[
+            msg.sender
+        ];
+        if (provider.sequenceNumber == 0) {
+            revert EntropyErrors.NoSuchProvider();
+        }
+
+        uint32 oldGasLimit = provider.defaultGasLimit;
+        provider.defaultGasLimit = gasLimit;
```
m30m, line 791: maybe we should pass this to roundGas and check if it is within the limits and not too high.

target_chains/ethereum/entropy_sdk/solidity/EntropyStructs.sol
```diff
@@ -61,6 +63,9 @@ contract EntropyStructs {
         bool useBlockhash;
         // Status flag for requests with callbacks. See EntropyConstants for the possible values of this flag.
         uint8 callbackStatus;
-        // 2 bytes of space left in this struct.
+        // The gasLimit in units of 10k gas. (i.e., 2 = 20k gas). We're using units of 10k in order to fit this
+        // field into the remaining 2 bytes of this storage slot. The dynamic range here is 10k - ~65M, which should
```
m30m, line 67: it's actually 655M

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -703,6 +808,16 @@ abstract contract Entropy is IEntropy, EntropyState {
         );
     }
 
+    // Rounds the provided quantity of gas into units of 10k gas.
+    // If gas is not evenly divisible by 10k, rounds up.
+    function roundGas(uint32 gas) internal pure returns (uint16) {
+        uint32 gas10k = gas / 10000;
```
m30m, line 814: can we put this 10000 as a constant somewhere in this file? It's being used 4-5 times across multiple functions.

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -590,7 +645,43 @@ abstract contract Entropy is IEntropy, EntropyState {
...
+        // existing feeInWei parameter for the callback failure flow instead of defining new
+        // configuration values.
+        uint32 roundedGasLimit = uint32(roundGas(gasLimit)) * 10000;
+        if (
+            provider.defaultGasLimit > 0 &&
+            roundedGasLimit > provider.defaultGasLimit
+        ) {
+            // This calculation rounds down the fee, which means that users can get some gas in the callback for free.
+            // However, the value of the free gas is < 1 wei, which is insignificant.
+            uint128 additionalFee = ((roundedGasLimit -
```
m30m, line 678: can't this lead to an underflow? if roundedGasLimit < provider.defaultGasLimit?

jayantk: (Note that this behavior is also tested in `testGasLimitsAndFeeRounding`)

PR #2572, opened 2025-04-09
pulse-scheduler: init ethereum contracts, subscriptions, updating prices

tejasbadadare:
## Summary

Initialize the Pulse Scheduler Ethereum contracts with a base layer of functionality. Follows the structure/boilerplate of the existing Pulse contracts.

## Rationale
Add state, functions, and tests for:
- Subscription management
  - Only manager can CRUD subscriptions
- Pushing and consuming price updates
  - An update to a subscription's feeds must have matching timestamps and contain all price IDs for that subscription.
  - Since we plan to make updating permissionless, an update is only allowed if the update criteria is met.
  - Only whitelisted readers can consume the price feeds.
- Viewing active subscriptions (for keepers)

## How has this been tested?

- [x] Added new tests

## Up next
- Balance management (actually using manager funds)
- Keeper/provider management (payment/registration, etc) 
- Gas strategy configuration
- Gas benchmark tests


target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+        // Check if subscription has enough balance
+        if (status.balanceInWei < pythFee) {
+            revert InsufficientBalance();
+        }
+
+        // Parse price feed updates with an expected timestamp range of [-10s, now]
+        // We will validate the trigger conditions and timestamps ourselves
+        // using the returned PriceFeeds.
+        uint64 maxPublishTime = SafeCast.toUint64(block.timestamp);
+        uint64 minPublishTime = maxPublishTime - 10 seconds;
```
jayantk, line 177: i'm not sure 10 seconds is enough. Either way, you probably want this value to be configurable by the admin authority.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+            );
+        for (uint8 i = 0; i < priceIds.length; i++) {
+            // Verify the requested price ID is part of the subscription
+            bool validPriceId = false;
+            for (uint8 j = 0; j < params.priceIds.length; j++) {
+                if (priceIds[i] == params.priceIds[j]) {
+                    validPriceId = true;
+                    break;
+                }
+            }
```
jayantk, line 357: fyi instead of doing this loop, you can also check if requestedFeeds[i] is all zero below. (or check 1 field of it which is guaranteed to be nonzero in a valid update). 

That approach also works better in the rare case where someone makes a subscription then tries to read it before it's ever updated.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+        view
+        override
+        returns (
+            uint256[] memory subscriptionIds,
+            SubscriptionParams[] memory subscriptionParams
+        )
+    {
+        // TODO: This is gonna be expensive because we're iterating through
+        // all subscriptions, including deactivated ones. But because its a view
+        // function maybe it's not bad? We can optimize this.
```
jayantk, line 411: yeah there may be some maximum size limit on the result (?) may need to paginate this or something like that

I also worry that there's a potential vulnerability here where someone goes and creates a billion subscriptions. If there are too many subscriptions, then maybe the keeper can't find the active ones, and that causes the service to go down.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+            revert InvalidGasConfig();
+        }
+
+        // Update subscription parameters
+        _state.subscriptionParams[subscriptionId] = newSubscriptionParams;
+
+        emit SubscriptionUpdated(subscriptionId);
+    }
+
+    function deactivateSubscription(
```
jayantk, line 124: should this be a special case of updateSubscription -- then you can also re-activate deactivated subscriptions?

target_chains/ethereum/contracts/forge-test/PulseScheduler.t.sol
```diff
@@ -0,0 +1,896 @@
...
+
+        // Expect revert because heartbeat condition is not met
+        vm.expectRevert(
+            abi.encodeWithSelector(UpdateConditionsNotMet.selector)
+        );
+        vm.prank(pusher);
+        scheduler.updatePriceFeeds(subscriptionId, updateData2, priceIds);
+    }
+
+    function testUpdatePriceFeedsRevertsOnUpdateConditionsNotMet_Deviation()
```
jayantk, line 454: what's with the _ in the name here and in the method above

jayantk (approved):
looks like a good start to me. Feel free to address my comments in follow up PRs, as this is pretty big and it's generally fine to merge as is.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: Yeah true, will set it to 1 minute to allow for retries/flaky RPCs which might result in the tx landing later than expected.

target_chains/ethereum/contracts/contracts/pulse/scheduler/SchedulerState.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache 2
+
+pragma solidity ^0.8.0;
+
+import "@pythnetwork/pyth-sdk-solidity/PythStructs.sol";
+
+contract SchedulerState {
+    // Maximum number of price feeds per subscription
+    uint8 public constant MAX_PRICE_IDS = 10;
```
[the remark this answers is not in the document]
tejasbadadare: Yeah it's an arbitrary number, will raise it to something higher

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: Hm, when writing this I had considered deactivating a subscription to be a one-way operation. But it is better UX to allow someone to pause their subscription but maintain the balance and settings, so they can resume later. To make it a special case of updateSubscription, we'd need to move `isActive` from `SubscriptionStatus` to `SubscriptionParams`, which maybe feels a little strange from a data model perspective. I considered adding a `updateSubscriptionIsActive(isActive: bool)`, but I think i prefer just one unified `updateSubscription()`. Will make that change

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: Yeah good shout on a potential max size limit on the result, i'll check on that. If we make addSubscription payable and enforce a minimum balance, we should be able to mitigate people flooding the service with subscriptions hopefully.

target_chains/ethereum/contracts/forge-test/PulseScheduler.t.sol
tejasbadadare: Hahah, they're kind of variants of the update criteria unit tests. I suppose a more standard naming convention would be `testUpdatePriceFeedsRevertsOnDeviationUpdateConditionNotMet`

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: Oh yeah, reduces it from O(n^2) to O(n), nice! One sharp corner here is that if the subscribed feeds change, this function would still let them query old price ids (unless we cleared `_state.priceUpdates[subscriptionId]`). But that's pretty minor, and the stale price would quickly fail age validation.

tejasbadadare:
@jayantk addressed your comments in #2585 , @ali-bahjati i'll probably make a new PR to address your feedback since it's already sizable and covers a bunch of disjoint changes

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+        // Parse price feed updates with an expected timestamp range of [-10s, now]
+        // We will validate the trigger conditions and timestamps ourselves
+        // using the returned PriceFeeds.
+        uint64 maxPublishTime = SafeCast.toUint64(block.timestamp);
+        uint64 minPublishTime = maxPublishTime - 10 seconds;
+        PythStructs.PriceFeed[] memory priceFeeds = pyth.parsePriceFeedUpdates{
+            value: pythFee
+        }(updateData, priceIds, minPublishTime, maxPublishTime);
+
+        // Verify all price feeds have the same timestamp
```
ali-bahjati, line 182: a corner case here is when markets are closed for some assets. in these cases using `slot` is better but i doubt we expose it from parse price feeds.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -0,0 +1,482 @@
...
+        // Store the price updates
+        for (uint8 i = 0; i < priceFeeds.length; i++) {
+            _state.priceUpdates[subscriptionId][priceFeeds[i].id] = priceFeeds[
+                i
+            ];
+        }
+        status.priceLastUpdatedAt = priceFeeds[0].price.publishTime;
+        status.balanceInWei -= pythFee;
+        status.totalUpdates += 1;
+        status.totalSpent += pythFee;
```
ali-bahjati, line 221: you'll handle publishers rewards and cost in a separate pr right?

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
ali-bahjati: probably you can clear the state for excluded assets.

just thinking in a fancy way another way is that you can always store a "version" for each subscription and then after each update increment it and set all the included price feeds to be of that version. then here you just check the version :D

target_chains/ethereum/contracts/forge-test/PulseScheduler.t.sol
```diff
@@ -0,0 +1,896 @@
...
+            reader.verifyPriceFeeds(
+                subscriptionId,
+                selectedPriceIds,
+                expectedFeeds
+            ),
+            "Price feeds verification failed"
+        );
+    }
+
+    function testOptionalWhitelist() public {
```
ali-bahjati, line 670: this kind of tests whielist being disabled. the name of test was confusing.

ali-bahjati: where are you testing that whitelisting works actually? can't find it.

ali-bahjati (approved):
Nice work! looking forward to the subsequent prs.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: oh yeah good callout -- lets use slot for this

target_chains/ethereum/contracts/contracts/pulse/scheduler/SchedulerState.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 maxGasLimit;
+    }
+
+    struct UpdateCriteria {
+        bool updateOnHeartbeat;
+        uint32 heartbeatSeconds;
+        bool updateOnDeviation;
+        uint32 deviationThresholdBps;
+
+        // TODO: add updateOnConfidenceRatio?
```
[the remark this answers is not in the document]
tejasbadadare: yeahh sounds good, will leave it out

target_chains/ethereum/contracts/contracts/pulse/scheduler/SchedulerState.sol
```diff
@@ -0,0 +1,64 @@
...
+        // Pyth contract for parsing updates and verifying sigs & timestamps
+        address pyth;
+        // Sub ID -> subscription parameters (which price feeds, when to update, etc)
+        mapping(uint256 => SubscriptionParams) subscriptionParams;
+        // Sub ID -> subscription status (metadata about their sub)
+        mapping(uint256 => SubscriptionStatus) subscriptionStatuses;
+        // Sub ID -> price ID -> latest parsed price update for the subscribed feed
+        mapping(uint256 => mapping(bytes32 => PythStructs.PriceFeed)) priceUpdates;
+        // Sub ID -> manager address
+        mapping(uint256 => address) subscriptionManager;
```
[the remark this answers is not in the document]
tejasbadadare: Yeahh that honestly might work better now that theres more state. I was thinking of saving a hop by maintaining multiple mappings. It's not hard to change this, so i'll table this until we build out the rest of the contract

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: that would be neat :) i guess we only really need to update the price feeds if they were updated, and not if some other attribute was changed (like whitelist). will stick with clearing the old feeds for now

PR #2573, opened 2025-04-10
feat(target_chains/ethereum): add WithdrawFee action and implement related functionality in governance payload

cctdaniel:
## Summary

Added support for the `WithdrawFee` governance action that allows withdrawing collected fees from Pyth contracts to a specified address. The changes include:

- New `WithdrawFee` action in governance payload system
- Added serialization/deserialization support in `xc_admin_common`
- Added Solidity implementation for fee withdrawal
- Added test coverage for the new action

## Rationale

This feature is necessary to:
1. Enable fee collection from Pyth contracts
2. Provide a secure, governance-controlled mechanism to withdraw accumulated fees
3. Allow flexible fee distribution to specified addresses

## How has this been tested?

- [x] Current tests cover my changes
  - Added `WithdrawFee` case to existing governance payload serialization tests
  - Added `WithdrawFee` to property-based testing via `governanceActionArb`
  - Existing governance message verification tests cover the new action

- [x] Added new tests
  - Added serialization/deserialization tests for `WithdrawFee` action
  - Added test coverage for hex string to Buffer conversion in governance payload tests

- [x] Manually tested the code
  - Verified correct encoding/decoding of withdrawal amounts and target addresses
  - Confirmed proper integration with existing governance message flow
  - Validated error handling for insufficient fees

The changes are well-tested through:
1. Unit tests for TypeScript serialization
2. Integration with existing governance payload system
3. Property-based testing for round-trip serialization
4. Solidity contract tests for fee withdrawal functionality

PR #2543, opened 2025-04-02
feat(target_chains/ethereum): add twap

cctdaniel:
## Summary

- Added comprehensive TWAP (Time-Weighted Average Price) functionality to the Pyth contract system including:
  - New data structures: `TwapPriceFeed`, `TwapPriceInfo`, `TwapPriceFeedMessage`
  - New methods for TWAP calculations and updates
  - Optimized fee structure for TWAP updates
  - Added extensive test coverage including negative test cases
- Added new error types for TWAP-related validation

## Rationale

These changes are necessary to:
1. Support TWAP price feed calculations which provide more stable price data by averaging over time
2. Improve UX by optimizing fee structure (charging for single update instead of both updates)
3. Ensure robust error handling for TWAP-related operations
4. Keep dependencies up to date for security and compatibility
5. Maintain code quality through comprehensive test coverage

## How has this been tested?

- [x] Current tests cover my changes
- [x] Added new tests
- [x] Manually tested the code

Testing steps:
1. Added comprehensive positive test case for TWAP calculation verifying:
   - Basic properties (ID, timestamps, exponents)
   - Price calculations
   - Confidence calculations
   - Down slots ratio

2. Added extensive negative test cases covering:
   - Invalid update data length
   - Mismatched price IDs
   - Invalid time ordering
   - Mismatched exponents
   - Invalid previous publish times
   - Insufficient fees

3. Manually verified:
   - TWAP calculation accuracy
   - Fee optimization changes
   - Error handling
   - Event emissions
   - Integration with existing price feed functionality

PR #2574, opened 2025-04-10
feat(target_chains/ethereum): don't store update in contract storage in parsePriceFeedUpdates

tejasbadadare:
## Summary
This PR removes the `updatePriceFeedsIfNecessary` call from the Pyth contract's `parsePriceFeedUpdates` API. Now the function will just parse and return the PriceFeed struct in-memory, and will not store the update in contract storage.

## Rationale

The Pyth ethereum contract's `parsePriceFeedUpdates` function currently has a side effect where if the input update is fresher than the one stored in the contract, it will store the new update via `updatePriceFeedsIfNecessary`.

In our new Pulse Scheduler design, we want to avoid free-riding our push feeds. This side-effect of the parse function conflicts with this, since anyone can fetch the new price via Pyth's `queryPriceFeeds`. We need to call the parse function in Scheduler's `updatePriceFeeds` function, but we want to lock down reading the feeds to the user's whitelisted addresses. Thus, we are removing the `updatePriceFeedsIfNecessary` call from the parse function.

This will be a potentially breaking change in behavior for users, but this behavior was initially intended as an optimization (if someone is bringing new data, let's store it), without regard for free-riding. We don't expect people are calling parse and then queryPriceFeed, since that would be a misuse/incorrect integration of our contract.

## How has this been tested?

  - Removed tests that checked that parsePriceFeedUpdates would update the price feed state.
- [ ] Added new tests


target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -397,6 +396,6 @@ abstract contract Pyth is
     }
 
     function version() public pure returns (string memory) {
-        return "1.4.4-alpha.1";
+        return "1.4.5-alpha.1";
```
tejasbadadare, line 399: Not sure what our versioning semantics are, but there are a couple open PRs that will increment this (#2543 and #2573 ), so I will wait to merge this until those land, and update the version number accordingly.

tejasbadadare: Figured this out (ty daniel), updated the version to `1.4.4-alpha.4` since we will bundle it into the 1.4.4 deployment.

ali-bahjati (approved):

PR #2585, opened 2025-04-15
feat(pulse-scheduler): address initial comments, update sub management, enforce mininum balance

tejasbadadare:
## Summary and Rationale
Various improvements from PR and design doc review.

- Address PR review comments from #2572 
  - Paginate getActiveSubscriptions
  - Make addSubscription payable to allow for atomic addFunds + addSubscription
- Match IPyth signatures for the read prices functions (e.g. `getPricesUnsafe`) to make migration easier for users
- Make deactivateSubscription a special case of updateSubscription
  - Expanding `updateSubscription` to handle activating and deactivating subscriptions makes it possible to reactivate a subscription. This is better UX, since the user may want to pause and resume a subscription without modifying its balance or params.
- Enforce a minimum balance for active subscriptions
  - We now enforce a minimum balance for a subscription when withdrawing funds from an active subscription, and when adding/reactivating a subscription. This helps guard against bad actors flooding the system with subscriptions, which could cause the keepers to become overwhelmed. it also encourages users maintain enough balance to weather high gas environments. The actual balance is a placeholder value right now.
  - An effect of this is that `isActive` has moved from `SubscriptionStatus` to `SubscriptionParams`

- Remove the `onlyPusher` modifier since we will allow anyone to push updates.

- Update GasConfig to expose `maxGasMultiplierCapPct` and `maxFeeMultiplierCapPct` controls to the user. Set defaults to be very high, since most users would rather land txs at a higher cost than not land them due to limits set too low.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/scheduler/IScheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/scheduler/IScheduler.sol
@@ -14,10 +14,11 @@ interface IScheduler is SchedulerEvents {
      * @notice Adds a new subscription
      * @param subscriptionParams The parameters for the subscription
      * @return subscriptionId The ID of the newly created subscription
+     * @dev Requires msg.value to be at least the minimum balance for the subscription
      */
     function addSubscription(
         SchedulerState.SubscriptionParams calldata subscriptionParams
-    ) external returns (uint256 subscriptionId);
+    ) external payable returns (uint256 subscriptionId);
 
     /**
      * @notice Gets a subscription's parameters and status
@@ -45,11 +46,7 @@ interface IScheduler is SchedulerEvents {
         SchedulerState.SubscriptionParams calldata newSubscriptionParams
     ) external;
 
-    /**
-     * @notice Deactivates a subscription
-     * @param subscriptionId The ID of the subscription to deactivate
-     */
-    function deactivateSubscription(uint256 subscriptionId) external;
+    // Deactivation is now handled through updateSubscription by setting isActive to false
 
     /**
      * @notice Updates price feeds for a subscription.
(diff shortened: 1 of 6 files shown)


target_chains/ethereum/contracts/contracts/pulse/scheduler/IScheduler.sol
```diff
@@ -64,16 +59,36 @@ interface IScheduler is SchedulerEvents {
...
-     * @param subscriptionId The ID of the subscription
-     * @param priceIds Optional array of price IDs to retrieve. If empty, returns all price feeds for the subscription.
-     * @return The latest price feeds for the requested price IDs
+    /** @notice Returns the price of a price feed without any sanity checks.
+     * @dev This function returns the most recent price update in this contract without any recency checks.
+     * This function is unsafe as the returned price update may be arbitrarily far in the past.
+     *
+     * Users of this function should check the `publishTime` in the price to ensure that the returned price is
+     * sufficiently recent for their application. If you are considering using this function, it may be
+     * safer / easier to use `getPriceNoOlderThan`.
```
ali-bahjati, line 68: you are going to add getPriceNoOlderThan as well?

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -385,63 +489,124 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+            uint256 totalCount
         )
     {
-        // TODO: This is gonna be expensive because we're iterating through
-        // all subscriptions, including deactivated ones. But because its a view
-        // function maybe it's not bad? We can optimize this.
-
-        // Count active subscriptions first to determine array size
-        uint256 activeCount = 0;
+        // Count active subscriptions first to determine total count
```
ali-bahjati, line 549: i'm worried it gets too expensive to iterate. maybe you could say like startIndex, endIndex and return all active subscriptions within?

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -385,63 +489,124 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
-        _;
+    /**
+     * @notice Returns the minimum balance an active subscription of a given size needs to hold.
+     * @param numPriceFeeds The number of price feeds in the subscription.
+     */
+    function getMinimumBalance(
+        uint8 numPriceFeeds
+    ) external pure override returns (uint256 minimumBalanceInWei) {
+        // Simple implementation - minimum balance is 0.01 ETH per price feed
+        return numPriceFeeds * 0.01 ether;
```
ali-bahjati, line 605: i'm happy you added this method and check. let's make it configurable, it highly depends on the network we are in.

target_chains/ethereum/contracts/contracts/pulse/scheduler/SchedulerState.sol
```diff
@@ -37,28 +42,25 @@ contract SchedulerState {
...
+    /// use a max of 100x (10000%) of the estimated gas as reported by the RPC.
     struct GasConfig {
-        // TODO: Figure out what controls to give users for gas strategy
-
-        // Gas price limit to prevent runaway costs in high-gas environments
-        uint256 maxGasPrice;
-        // Gas limit for update operations
-        uint256 maxGasLimit;
+        /// Gas price multiplier limit percent for update operations
+        uint32 maxGasMultiplierCapPct;
```
ali-bahjati, line 55: this is base fee then i suppose?

ali-bahjati (approved):

target_chains/ethereum/contracts/contracts/pulse/scheduler/IScheduler.sol
tejasbadadare: The [Pyth contract](https://github.com/pyth-network/pyth-crosschain/blob/f32eaf882f46cc92c791afbac7b9e530fc7ee3c1/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol#L154) only specifies the *Unsafe methods, and provides the *NoOlderThan methods [as part of the sdk](https://github.com/pyth-network/pyth-crosschain/blob/59ea3c8f9050dcc911df080efb5af22a3ab2d274/target_chains/ethereum/sdk/solidity/IPyth.sol#L28), as an abstraction over the unsafe methods, so I was following that pattern here

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: Are you worried about the maxResults number being too high? I think startIndex, maxResults is a more friendly API than startIndex, endIndex.. we can probably leave it the users to call the with a reasonable maxResults since its a view function.

Also, I think at this point it's worth just maintaining another list of active subscriptions (or similar,) which would avoid us looping through the full list which includes a bunch of deactivated subs. Will tackle this next

target_chains/ethereum/contracts/contracts/pulse/scheduler/SchedulerState.sol
tejasbadadare: Yep -- can give it a clearer name

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
ali-bahjati: i'm mostly worried of running out of gas during simulation because you iterate through all the subscriptions to see where's the start index (of active subscriptions).  i hear you though, your call.

tejasbadadare: Yeah agreed. I think maintaining a list of active subscriptions fixes this since we can directly index into it. Let me handle this in the next PR.

PR #2587, opened 2025-04-16
fix(pyth-lazer-puiblisher-sdk): Add Deployment Script

darunrs:
## How has this been tested?

- [ ] Added new tests


.github/workflows/publish-rust-lazer-publisher-sdk.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v2
 
-      - run: cargo publish --token ${CARGO_REGISTRY_TOKEN}
+      - run: publish.sh
```
merolish, line 16: What triggers a publish? I assume we only want to do this whenever we have a new version.

Riateche: It triggers on tags (see top of the file).

lazer/publisher_sdk/rust/publish.sh
```diff
@@ -0,0 +1,38 @@
...
+echo "updating lib.rs to export local protobuf files"
+cat > "${PACKAGE_DIR}/src/lib.rs" << EOF
+pub mod transaction {
+    pub use crate::protobuf::pyth_lazer_transaction::*;
+}
+pub mod publisher_update {
+    pub use crate::protobuf::publisher_update::*;
+}
+mod protobuf;
+EOF
```
Riateche, line 26: If we change `lib.rs` we'll likely forget to edit this script. Maybe it's better to use `sed` or python to remove the `include!` and keep the rest of the file as-is.

lazer/publisher_sdk/rust/publish.sh
```diff
@@ -0,0 +1,38 @@
...
+
+echo "removing build spec from Cargo.toml"
+# MacOS sed command is slightly different
+# This script shouldn't be run locally anyway, but I may as well make sure it works
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  sed -i '' '/build.*build\.rs/d' Cargo.toml
+else
+  sed -i '/build.*build\.rs/d' Cargo.toml
+fi
```
Riateche, line 35: You can just remove this from `Cargo.toml`. It will pick up `build.rs` if it exists even without that line.

lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -26,6 +27,20 @@ pub struct ChannelId(pub u8);
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct TimestampUs(pub u64);
 
+impl TryFrom<&MessageField<Timestamp>> for TimestampUs {
```
Riateche, line 30: I think it's better to implement `TryFrom<&Timestamp> for TimestampUs` and let the caller handle `MessageField` separately. This way the implementation will be useful both for required and optional timestamp fields.

Riateche (approved):
Please see the comment. Otherwise it looks good!

lazer/sdk/rust/protocol/src/router.rs
darunrs: Makes sense! Will change!

PR #2590, opened 2025-04-17
fix(pulse-scheduler): validate atomic subscription updates using slots instead of timestamps

tejasbadadare:
## Summary
This PR fixes the "atomic" subscription update guarantee (ensuring all feeds in a subscription are updated together,) by using slots instead of timestamps for the validation.

### Use slots instead of timestamps
The main motivation for this is that timestamps (`publishTime`) can be misleading for validating an "atomic" subscription update, as one or more assets might have closed markets, and will retain timestamps from their last trading period.
Thus, we use Pythnet slots for the atomicity validation, since they are consistent across all assets regardless of market status. 
- We still use timestamps to validate the heartbeat update condition. We now take the max of all passed in timestamps (rather than the first) to ensure that if any of the feeds need updating, we perform the update.

### Get the slots data from the Pyth parse functions
Pythnet slot number is available in the AccumulatorUpdate, but unfortunately `Pyth.parsePriceFeedUpdates` doesn't return them, just the Price[] structs. Thus, we add `parsePriceFeedUpdatesWithSlots` to the Pyth contract to expose slot information, and we consume this from the Scheduler contract.
- All the updates in a single WhMerkle message have the same slot, so technically we could have returned a `slots` array of len(updateData) rather than len(priceFeedIds), but I chose to return parallel arrays of `Price[] feeds` and `uint64[] slots` to keep things coherent for the user, especially if len(updateData) > 1.
- It turns out that `parsePriceUpdatesInternal` was already very close to the stack depth limit, and adding a `slots` array took it (quite far) over the line. Thus, this PR also contains significant refactors to that method to split it up into helper methods to reduce stack frame size. We also use structs to pass arguments instead of local variables for the same reason, since passing many arguments to functions takes up lots of stack space.

### Misc
- Expanded timestamp validity periods to 1 hour in the past and 10 seconds in the future and extracted them out to constants.

## Note to reviewers
I'd start in Pyth.sol first with the changes to parsePriceFeedUpdatesInternal. Then Scheduler.sol since it is downstream. The rest is somewhat mechanical fallout. Some of the motivations behind the changes are nonobvious, see below for the rationale.

## How has this been tested?

  - All existing tests were updated to work with slot-based validation
- [x] Added new tests
  - New tests were added to verify correct slot validation behavior and `Pyth.parsePriceFeedUpdatesWithSlots` behavior

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
@@ -239,20 +239,21 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         // Parse price feed updates with an expected timestamp range of [-10s, now]
         // We will validate the trigger conditions and timestamps ourselves
         // using the returned PriceFeeds.
-        uint64 maxPublishTime = SafeCast.toUint64(block.timestamp) +
-            FUTURE_TIMESTAMP_GRACE_PERIOD;
-        uint64 minPublishTime = maxPublishTime - PAST_TIMESTAMP_GRACE_PERIOD;
+        uint64 curTime = SafeCast.toUint64(block.timestamp);
+        uint64 maxPublishTime = curTime + FUTURE_TIMESTAMP_MAX_VALIDITY_PERIOD;
+        uint64 minPublishTime = curTime - PAST_TIMESTAMP_MAX_VALIDITY_PERIOD;
         PythStructs.PriceFeed[] memory priceFeeds;
         uint64[] memory slots;
         (priceFeeds, slots) = pyth.parsePriceFeedUpdatesWithSlots{
             value: pythFee
         }(updateData, priceIds, minPublishTime, maxPublishTime);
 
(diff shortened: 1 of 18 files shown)


ali-bahjati (approved):
This overall looks good to me. I like to know the code size and gas consumption changes before merging though. Another concern that I have is whether we should audit the Pyth contract now that's changed more.

tejasbadadare:
> This overall looks good to me. I like to know the code size and gas consumption changes before merging though. Another concern that I have is whether we should audit the Pyth contract now that's changed more.

Yeah good callouts. Will test both and get back to you. If we can potentially sneak the Pyth contract into the scope that would be great :) maybe just the patches that changed.

tejasbadadare:
@ali-bahjati 
Gas impact looks minimal (~1%), [see this diff vs main](https://app.warp.dev/block/Fu9BhOmuyFU7717h4I1WRo)

As for the contract size, we're juuust slim enough at 24,125 bytes :) [See here](https://app.warp.dev/block/lEBlfwTpqW4551ettJcqDo)

Will merge, lmk if you have any concerns

PR #2592, opened 2025-04-17
feat(staking): hide empty pools

cprussin:
## Summary

Hide empty pools which you aren't staked to in staking

## Rationale

We should not allow people to stake to empty pools

## How has this been tested?

- [ ] Added new tests


guibescos (approved):
I can't see the preview but lgtm

PR #2593, opened 2025-04-17
feat(pulse-scheduler): add permanent subscriptions, allow permissionless funding, clean up tests

tejasbadadare:
## Summary
1. Add the ability to mark subscriptions as permanent. Permanent subscriptions disallow destructive actions, such as removing price feeds or withdrawing funds.
2. Allow anyone to fund a subscription, not just the manager.
3. Refactor tests with helper methods since they were getting a bit unwieldy.

## Rationale
1. It's valuable for some consumers to lock down a subscription so that it can't be tampered by a manager. They may simply be using a vault manager hot wallet there, not a multisig. That wallet could potentially rug the subscription. But, other users may find it useful to be able to stop a subscription, e.g. someone subscribes to a memecoin that goes to 0 and they want to stop the sub. Thus, we add an optional `isPermanent` flag to the subscription, which will disallow destructive actions when enabled. Destructive actions are defined as:
    1. Removing price feeds from a subscription
    3. Withdrawing funds from a subscription
    4. Disabling the isPermanent flag

2. No reason to lock down addFunds. Perhaps the community wants to sponsor a feed. It's also operationally simpler for a customer to use any wallet to fund.

## How has this been tested?
- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
@@ -100,29 +100,7 @@ abstract contract Scheduler is IScheduler, SchedulerState {
 
         // Check for permanent subscription restrictions
         if (currentParams.isPermanent) {
-            // Cannot disable isPermanent flag once set
-            if (!newParams.isPermanent) {
-                revert IllegalPermanentSubscriptionModification();
-            }
-
-            // Cannot remove price feeds from a permanent subscription
-            if (newParams.priceIds.length < currentParams.priceIds.length) {
-                revert IllegalPermanentSubscriptionModification();
-            }
-
-            // Check that all existing price IDs are preserved
-            for (uint i = 0; i < currentParams.priceIds.length; i++) {
-                bool found = false;
-                for (uint j = 0; j < newParams.priceIds.length; j++) {
-                    if (currentParams.priceIds[i] == newParams.priceIds[j]) {
-                        found = true;
-                        break;
-                    }
-                }
-                if (!found) {
-                    revert IllegalPermanentSubscriptionModification();
-                }
-            }
+            _validatePermanentSubscriptionUpdate(currentParams, newParams);
         }
 
         // If subscription is inactive and will remain inactive, no need to validate parameters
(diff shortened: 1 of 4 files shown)


jayantk (approved):

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -98,6 +98,33 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+                    if (currentParams.priceIds[i] == newParams.priceIds[j]) {
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found) {
+                    revert IllegalPermanentSubscriptionModification();
+                }
+            }
+        }
```
jayantk, line 126: I think you can also effectively disable a subscription by (1) changing the readerWhitelist or (2) making the gas config so limiting that no tx ever lands on chain?

jayantk: hmm I think you could also do it by changing the updateCriteria

jayantk (changes requested):
I approved this but then realized there are some issues here

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -98,6 +98,33 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+
+            // Cannot remove price feeds from a permanent subscription
+            if (newParams.priceIds.length < currentParams.priceIds.length) {
+                revert IllegalPermanentSubscriptionModification();
+            }
+
+            // Check that all existing price IDs are preserved
+            for (uint i = 0; i < currentParams.priceIds.length; i++) {
+                bool found = false;
+                for (uint j = 0; j < newParams.priceIds.length; j++) {
```
merolish, line 116: Heh, we're stuck doing an m x n linear search in Solidity?

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: yeah good points -- will disallow changing the whitelist or the config. truly "permanent."

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: we cant create in-memory hashmaps so yeah, back to the basics 🫠

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
tejasbadadare: just realized they can also just set isActive to false 🤦 updated

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -385,6 +390,103 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+
+        // Cannot change gas config for permanent subscriptions
+        if (
+            newParams.gasConfig.maxBaseFeeMultiplierCapPct !=
+            currentParams.gasConfig.maxBaseFeeMultiplierCapPct ||
+            newParams.gasConfig.maxPriorityFeeMultiplierCapPct !=
+            currentParams.gasConfig.maxPriorityFeeMultiplierCapPct
+        ) {
+            revert IllegalPermanentSubscriptionModification();
+        }
```
jayantk, line 487: I think these gas parameters could cause some problems still. The chain usage will change over time, and these parameters may need to be tweaked. If you can't change them, then they could initially be set too low, and then the chain gets more congested and then you don't get updates.

Maybe we need to set these to infinity for the permanent subscriptions? not sure exactly what to do.

(this is a problem we can resolve later also, so I'm approving this PR)

jayantk (approved):

PR #2598, opened 2025-04-22
fix(apps/hermes): use last ip in the headers

ali-bahjati:
## Rationale

In the current implementation we use the first IP in the specified header to detect the user IP but some load balancers always append the IP to it and the first IP can be provider by the user.


bplatak (approved):

ali-bahjati:
closing because this behaviour might result blocking everyone in some setups.

PR #2599, opened 2025-04-22
feat(hermes-server) Add `ignore_invalid_price_ids` flag to hermes ws endpoint

bplatak:
## Summary
Add a new flag `ignore_invalid_price_ids` to WebSocket subscriptions that enables partial subscriptions when some price IDs are invalid. When enabled, Hermes will establish the WebSocket connection and stream updates only for the valid price IDs while silently ignoring any invalid ones. This enhancement aligns the WebSocket API behaviour with the existing REST API functionality introduced in #2091.

## Rationale
This change simplifies integration for clients who need to subscribe to multiple price feeds but don't require all of them to be valid. Current workarounds are suboptimal:
1. Sending individual subscription requests for each ID (increasing connection overhead)
2. Attempting a bulk subscription, parsing error messages to identify invalid IDs, then retrying with filtered IDs

The new flag provides a more efficient solution by allowing clients to subscribe to all desired price feeds in a single request while gracefully handling any invalid IDs. This reduces implementation complexity for clients and improves connection efficiency by eliminating the need for multiple subscription attempts or error handling logic.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/hermes/server/src/api/ws.rs
+++ b/apps/hermes/server/src/api/ws.rs
@@ -151,6 +151,8 @@ enum ClientMessage {
         binary: bool,
         #[serde(default)]
         allow_out_of_order: bool,
+        #[serde(default)]
+        ignore_invalid_price_ids: bool,
     },
     #[serde(rename = "unsubscribe")]
     Unsubscribe { ids: Vec<PriceIdInput> },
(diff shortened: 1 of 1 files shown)


apps/hermes/server/src/api/ws.rs
```diff
@@ -556,10 +562,22 @@ where
...
+                                    error: "No price feeds available".to_string(),
+                                },
+                            ))?
+                            .into(),
+                        )
+                        .await?;
+                    return Ok(());
                 } else {
-                    for price_id in price_ids {
+                    for price_id in found_price_ids {
```
merolish, line 578: Probably not worth warn logging any not found ids, I assume.

merolish (approved):
Are there any corresponding docs that need to be updated?

apps/hermes/server/src/api/ws.rs
ali-bahjati: probably not a warn for us, but maybe an warn message to the client would be good (as an error response).

ali-bahjati (approved):

tejasbadadare:
> Are there any corresponding docs that need to be updated?

The WS endpoint is undocumented since we want the public to use the SSE endpoint instead. And that one already supports this flag.

tejasbadadare (approved):

PR #2604, opened 2025-04-23
feat(pulse-scheduler): optimize gas when querying active subscriptions

tejasbadadare:
## Summary

Optimize gas usage when querying for all active subscriptions.

## Rationale

This is one the most frequent access patterns in scheduler, since keepers need to poll the set of active subscriptions periodically to keep their state consistent. It's a view function so it will just be simulated, but since we could potentially store thousands of subscriptions, we still want to make it efficient.

This optimization sacrifices a bit of gas when activating/deactivating subscriptions (which should be relatively uncommon), for efficiency gains in the `getActiveSubscriptions` function. Previously we needed to iterate through all subscriptions to collect the active ones. This doesn't scale well as more subscriptions get deactivated. 

Now, we store the active subscriptions in an `activeSubscriptionIds` array for efficient querying and paging, and store a map of `sub id -> index in activeSubscriptionIds` for efficient removal when deactivating subscriptions.

## How has this been tested?

- [ ] Added new tests

### Gas benchmark results

Test case | Original gas | Optimized gas | Delta
|--------|--------|--------|--------|
| Query 5 active subscriptions (out of 10) | 110,317 | 82,547 | -25%
| Query 50 active subscriptions (out of 100) | 1,104,981 | 818,371 | -26%
| Query 500 active subscriptions (out of 1000) | 16,622,552 | 13,747,542 | -17%

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
@@ -784,6 +784,8 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         // Only add if not already in the list
         if (_state.activeSubscriptionIndex[subscriptionId] == 0) {
             _state.activeSubscriptionIds.push(subscriptionId);
+
+            // Store the index as 1-based, 0 means not in the list
             _state.activeSubscriptionIndex[subscriptionId] = _state
                 .activeSubscriptionIds
                 .length;
--- a/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
@@ -56,6 +56,8 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         // Map subscription ID to manager
         _state.subscriptionManager[subscriptionId] = msg.sender;
 
+        _addToActiveSubscriptions(subscriptionId);
+
         emit SubscriptionCreated(subscriptionId, msg.sender);
         return subscriptionId;
     }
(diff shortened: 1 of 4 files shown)


ali-bahjati (approved):
I assume we are sacrificing order to achieve linear removal and pagination; right? I'm fine with it. Just make sure to add documentation to the active subscription query that start index is not the subscription id.

target_chains/ethereum/contracts/contracts/pulse/scheduler/Scheduler.sol
```diff
@@ -790,4 +775,45 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+     * @notice Adds a subscription to the active subscriptions list.
+     * @param subscriptionId The ID of the subscription to add.
+     */
+    function _addToActiveSubscriptions(uint256 subscriptionId) internal {
+        // Only add if not already in the list
+        if (_state.activeSubscriptionIndex[subscriptionId] == 0) {
+            _state.activeSubscriptionIds.push(subscriptionId);
+            _state.activeSubscriptionIndex[subscriptionId] = _state
+                .activeSubscriptionIds
+                .length;
```
jayantk, line 789: maybe leave a comment here noting that this array is 1-indexed. I thought this was a bug until i looked further

that said, is there a test that adds & removes a bunch of subscriptions to validate that this logic works?

tejasbadadare: Yep, the 1-based indexing is documented in other places, but it's definitely not intuitive so i'll mention it every place the map is directly used

There are tests that cover adding and removing subscriptions, and the gas benchmark adds and removes hundreds. I'll augment the existing unit tests to stress test the implementation some more.

PR #2607, opened 2025-04-23
fix(entropy): Use the block number of event instead of request data for revelation api

m30m:
## Summary

Improve revelations api for arbitrum based chains

## Rationale

Current revelations api doesn't work well where `block.number` on-chain is different from what rpc returns. In arbitrum based chains, one points to the L1 block and the other to the L2 block.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/src/api/revelation.rs
+++ b/apps/fortuna/src/api/revelation.rs
@@ -1,3 +1,4 @@
+use crate::chain::reader::BlockNumber;
 use {
     crate::api::{ChainId, RequestLabel, RestError},
     anyhow::Result,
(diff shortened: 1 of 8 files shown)


apps/fortuna/src/api/revelation.rs
```diff
@@ -20,7 +21,7 @@ use {
 /// Callers must pass the appropriate chain_id to ensure they fetch the correct random number.
 #[utoipa::path(
 get,
-path = "/v1/chains/{chain_id}/revelations/{sequence}",
+path = "/v1/chains/{chain_id}/revelations/{block_number}/{sequence}",
```
jayantk, line 24: this is a very awkward API for someone to call -- how would they know to get the block number also?

I think a better solution is to implement the cache of events for the explorer, and then use that to do the lookup from sequence number to block number.

jayantk (commented):
I think this introduces a fairly awkward API -- see inline comment for a suggestion on a better way

jayantk (approved):

PR #2608, opened 2025-04-24
feat(pulse): add governance instructions, admin transfer capability

tejasbadadare:
## Summary

- Add governance instructions to Pulse
- Add ability to transfer admin capability to another address using a 2-step propose & accept flow
  - Admin and Owner can execute governance instructions 
  - Owner can upgrade the contract
- Enforce minimum balance when adding feeds to a subscription
- Minor misc updates to docs/formatting

## Rationale

- Added the following governance instructions:
  - `upgradeContract`
  - `setMinimumBalancePerFeed`
    - Minimum balance per feed a subscription needs to maintain. Enforced when adding feeds or withdrawing funds.
  - `setSingleUpdateKeeperFeeInWei`
    - Fee per single feed update to pay the keeper. Unenforced right now, will implement in the next PR.
- These instructions haven't been added to other parts of the governance stack yet. We can likely do that when we want to actually change one of these values. 
- Overall the governance and admin transfer pattern follows Entropy.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
@@ -73,13 +73,16 @@ abstract contract Scheduler is IScheduler, SchedulerState {
     function updateSubscription(
         uint256 subscriptionId,
         SubscriptionParams memory newParams
-    ) external override onlyManager(subscriptionId) {
+    ) external payable override onlyManager(subscriptionId) {
         SubscriptionStatus storage currentStatus = _state.subscriptionStatuses[
             subscriptionId
         ];
         SubscriptionParams storage currentParams = _state.subscriptionParams[
             subscriptionId
         ];
+
+        // Add incoming funds to balance
+        currentStatus.balanceInWei += msg.value;
 
         // Updates to permanent subscriptions are not allowed
         if (currentParams.isPermanent) {
--- a/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
@@ -12,12 +12,20 @@ import "./SchedulerState.sol";
 import "./SchedulerErrors.sol";
 
 abstract contract Scheduler is IScheduler, SchedulerState {
-    function _initialize(address admin, address pythAddress) internal {
+    function _initialize(
+        address admin,
+        address pythAddress,
+        uint128 minimumBalancePerFeed,
+        uint128 singleUpdateKeeperFeeInWei
+    ) internal {
         require(admin != address(0), "admin is zero address");
         require(pythAddress != address(0), "pyth is zero address");
 
         _state.pyth = pythAddress;
+        _state.admin = admin;
         _state.subscriptionNumber = 1;
+        _state.minimumBalancePerFeed = minimumBalancePerFeed;
+        _state.singleUpdateKeeperFeeInWei = singleUpdateKeeperFeeInWei;
     }
 
     function createSubscription(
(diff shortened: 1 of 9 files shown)


target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
```diff
@@ -91,6 +99,19 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         // Validate the new parameters, including setting default gas config
         _validateAndPrepareSubscriptionParams(newParams);
 
+        // Check minimum balance if number of feeds increases and subscription remains active
+        if (
+            willBeActive &&
+            newParams.priceIds.length > currentParams.priceIds.length
```
ali-bahjati, line 105: I think it's better if you do this check regardless. Otherwise I can increase feeds and deactivate, and then later activate it to bypass it right?

target_chains/ethereum/contracts/forge-test/PulseScheduler.t.sol
```diff
@@ -1092,6 +1100,102 @@ contract SchedulerTest is Test, SchedulerEvents, PulseSchedulerTestUtils {
...
+
+        // Action 2: Add enough funds to meet the new minimum balance
+        uint256 newMinimumBalance = scheduler.getMinimumBalance(newNumFeeds);
+        uint256 requiredFunds = newMinimumBalance - initialMinimumBalance;
+        scheduler.addFunds{value: requiredFunds}(subscriptionId);
+
+        // Verification 2: Update should now succeed
+        vm.expectEmit();
+        emit SubscriptionUpdated(subscriptionId);
+        scheduler.updateSubscription(subscriptionId, newParams);
```
ali-bahjati, line 1144: I recommend you let updateSubscription be payable as well so people can update it in 1 go (now they need to add funds first, and then update it)

ali-bahjati (approved):
Nice! Please address my comments before merging.

Also, one question. What's the difference between owner and admin here? I think we can just have one of them right?

target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
tejasbadadare: yeahh was debating this, but one thing i was thinking was we shouldn't enforce this for people trying to deactivate or reduce their feeds. maybe they are low on balance and trying to reduce their burn rate. ultimately the minimum balance thing serves 2 purposes: disincentivize the attack vector of flooding the system with subscriptions, and help the user maintain enough balance in their account to ensure uninterrupted operation. 

also, we do enforce the check when activating, so that wouldn't be a problem.

tejasbadadare:
> What's the difference between owner and admin here? I think we can just have one of them right?

You know, I had the same thought on this, but I took cues from Entropy's here and assumed that there is a reason :)
As it stands, the owner is the superadmin and can upgrade the contract. The admin can run governance ixs but cannot upgrade the contract. 

The executor will need to be the owner to be able to upgrade the contract, which makes the admin role perhaps redundant? But i did notice in git history for Entropy that folks wanted the admin role and for it to be transferable.

I'll keep both in for now, but it's easy to remove. Let me know what you think.

tejasbadadare:
> > What's the difference between owner and admin here? I think we can just have one of them right?
> 
> @ali-bahjati You know, I had the same thought on this, but I took cues from Entropy here and assumed that there is a reason :) As it stands, the owner is the superadmin - it can run governance ixs including upgrading the contract. The admin can run governance ixs but cannot upgrade the contract.
> 
> The executor will need to be the owner to be able to upgrade the contract, which makes the admin role perhaps redundant? But i did notice in git history for Entropy that folks wanted the admin role and for it to be transferable.
> 
> I'll keep both in for now, but it's easy to remove. Let me know what you think.

We resolved this offline, but for posterity:

turns out we don't really remember why it was done this way in Entropy. the Owner and Admin roles are functionally the same because they are both held by Executor. Perhaps admin was intended to be some kind of "light governance" that can execute smaller administrative instructions, but not have the big hammer authority to upgrade the contract? In any case we'll leave it the same in Pulse for consistency and have Executor assume both roles (edited)

PR #2612, opened 2025-04-24
feat(apps/hermes): add connection timeout for SSE & WebSocket connections

cctdaniel:
## Summary

- Added a 24-hour connection timeout for SSE endpoint
- Added a 24-hour connection timeout for WS endpoint
- Each client connection will automatically disconnect after 24 hours of continuous connection, with a proper error message sent before closing.

## Rationale

- Prevents resource leaks from indefinite connections
- Ensures proper cleanup of long-running connections
- Each client gets their own independent 24h window from their connection time
- Provides clear error messaging to clients about connection termination

## How has this been tested?

- [ ] Added new tests

Testing steps:
1. Verified that each new SSE/WS client connection gets its own independent 24h timeout
2. Confirmed that the timeout message is properly sent before connection close
3. Tested that the stream properly terminates after the timeout period
4. Verified that existing functionality (price updates, error handling) continues to work during the connection lifetime
5. Tested that multiple concurrent clients have independent timeouts based on their individual connection times

### DIFF
--- a/apps/hermes/server/src/api/rest/v2/sse.rs
+++ b/apps/hermes/server/src/api/rest/v2/sse.rs
@@ -26,7 +26,7 @@ use {
 };
 
 // Constants
-const MAX_CONNECTION_DURATION: Duration = Duration::from_secs(10); // 24 hours
+const MAX_CONNECTION_DURATION: Duration = Duration::from_secs(24 * 60 * 60); // 24 hours
 
 #[derive(Debug, Deserialize, IntoParams)]
 #[into_params(parameter_in = Query)]
--- a/apps/hermes/server/src/api/rest/v2/sse.rs
+++ b/apps/hermes/server/src/api/rest/v2/sse.rs
@@ -78,6 +78,9 @@ fn default_true() -> bool {
     params(StreamPriceUpdatesQueryParams)
 )]
 /// SSE route handler for streaming price updates.
+///
+/// The connection will automatically close after 24 hours to prevent resource leaks.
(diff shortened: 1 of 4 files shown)


apps/hermes/server/src/api/rest/v2/sse.rs
```diff
@@ -93,7 +96,14 @@ where
     // Convert the broadcast receiver into a Stream
     let stream = BroadcastStream::new(update_rx);
 
+    // Set connection deadline
+    let connection_deadline = Instant::now() + MAX_CONNECTION_DURATION;
+
     let sse_stream = stream
+        .take_while(move |_| {
+            let now = Instant::now();
+            now < connection_deadline
```
ali-bahjati, line 105: I recommend using `start_time = Instant::now()` and then check `start_time.elapsed() < MAX_....`

ali-bahjati (approved):
Please address my comment before merging and bump the version. Also please update the docs (for SSE at least) to document this behaviour.

tejasbadadare (approved):
nice! one thing i'm wondering about is that there are many clients out there that maintain open connections to hermes. when we deploy this change, they'll all reconnect and start their connection TTLs at the same time. then every 24h there will be a thundering herd of reconnections. we could add a jitter to the 24h to avoid this (give them a variable extra 1-5m after the 24h).

in reality, many of them will drop their connections naturally due to network conditions before 24h elapses, and hermes can probably handle the thundering herd anyway, but just something to be aware of.

PR #2617, opened 2025-04-25
feat(pulse): keeper payment

tejasbadadare:
## Summary

Add keeper payment in `updatePriceFeeds`

## Rationale

Keepers are permissionless and we pay them in-band if they successfully call `updatePriceFeeds`. If the trigger condition isn't satisfied or the subscription doesn't have enough funds, they won't get paid.

This is a temporary payment model, the long term goal is to add a monthly subscription to make it easy for users to consume the service, rather than keeping their balance topped up. However, the basic mechanism of paying the keeper for successful updates will be the same, just the fee calculation will change.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
+++ b/target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
@@ -250,6 +250,8 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         bytes[] calldata updateData,
         bytes32[] calldata priceIds
     ) external override {
+        uint256 startGas = gasleft();
+
         SubscriptionStatus storage status = _state.subscriptionStatuses[
             subscriptionId
         ];
@@ -261,9 +263,12 @@ abstract contract Scheduler is IScheduler, SchedulerState {
             revert InactiveSubscription();
         }
 
-        // Verify price IDs match subscription
+        // Verify price IDs match subscription length
         if (priceIds.length != params.priceIds.length) {
-            revert InvalidPriceIdsLength(priceIds[0], params.priceIds[0]);
+            revert InvalidPriceIdsLength(
+                priceIds.length,
+                params.priceIds.length
+            );
         }
 
         // Keepers must provide priceIds in the exact same order as defined in the subscription
@@ -277,7 +282,7 @@ abstract contract Scheduler is IScheduler, SchedulerState {
         IPyth pyth = IPyth(_state.pyth);
         uint256 pythFee = pyth.getUpdateFee(updateData);
 
-        // Check if subscription has enough balance
+        // If we don't have enough balance, revert
         if (status.balanceInWei < pythFee) {
             revert InsufficientBalance();
         }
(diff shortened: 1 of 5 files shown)


target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
```diff
@@ -277,26 +282,24 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
-            : 0;
         (
             PythStructs.PriceFeed[] memory priceFeeds,
             uint64[] memory slots
         ) = pyth.parsePriceFeedUpdatesWithSlots{value: pythFee}(
                 updateData,
                 priceIds,
-                minPublishTime,
-                maxPublishTime
+                curTime > PAST_TIMESTAMP_MAX_VALIDITY_PERIOD
```
tejasbadadare, line 299: had to inline this (and extract logic into helper functions) to avoid stack depth issues

target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
```diff
@@ -737,4 +725,55 @@ abstract contract Scheduler is IScheduler, SchedulerState {
...
+
+        // Check balance
+        if (status.balanceInWei < totalFee) {
+            revert InsufficientBalance();
+        }
+
+        // Update status and pay keeper
+        status.balanceInWei -= totalFee;
+        status.totalSpent += totalFee;
+        (bool sent, ) = msg.sender.call{value: totalKeeperFee}(""); // Pay only the keeper portion
```
ali-bahjati, line 774: just double check that we have updated the state everywhere (to make sure there's no double entry attack)

target_chains/ethereum/contracts/contracts/pulse/SchedulerState.sol
```diff
@@ -17,6 +17,9 @@ contract SchedulerState {
     /// Maximum time in the future (relative to current block timestamp)
     /// for which a price update timestamp is considered valid
     uint64 public constant FUTURE_TIMESTAMP_MAX_VALIDITY_PERIOD = 10 seconds;
+    /// Fixed gas overhead component used in keeper fee calculation.
+    /// This is a rough estimate of the tx overhead for a keeper to call updatePriceFeeds.
+    uint256 public constant GAS_OVERHEAD = 30000;
```
ali-bahjati, line 22: How did you get to this number?

ali-bahjati (approved):

target_chains/ethereum/contracts/contracts/pulse/SchedulerState.sol
tejasbadadare: It's a rough estimate, we will probably want to set it per blockchain. The idea was to capture the gas used before the first gasleft() call, which would be the flat 21k fee for an external contract call, ~2.6k for the proxy's delegatecall, and some for the calldata. We can measure it more carefully when we do the math to ensure our pricing model works.

target_chains/ethereum/contracts/contracts/pulse/Scheduler.sol
tejasbadadare: I better implemented checks-effects-interactions pattern. Now we apply it independently to the pyth fee and the keeper fee. This keeps accounting accurate if either of the payments fail.

PR #2620, opened 2025-04-25
fix(entropy): Restore old signatures for methods returning structs

jayantk:
## Summary

As the description says -- it turns out that you can't decode the old struct data with the new struct format. This restores the old method signatures and adds new methods for retrieving the new versions of the structs.

I don't love that we have new methods here for retrieving requests / providers, but that's the only way I see to ensure that all users of the contract continue working.

## How has this been tested?

- [ ] Added new tests

I tested getting both a request and a provider from an old contract.

### DIFF
--- a/apps/fortuna/src/command/get_request.rs
+++ b/apps/fortuna/src/command/get_request.rs
@@ -14,10 +14,7 @@ pub async fn get_request(opts: &GetRequestOptions) -> Result<()> {
         &Config::load(&opts.config.config)?.get_chain_config(&opts.chain_id)?,
     )?);
 
-    let p = contract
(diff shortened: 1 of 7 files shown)


apps/fortuna/src/command/get_request.rs
```diff
@@ -14,6 +14,10 @@ pub async fn get_request(opts: &GetRequestOptions) -> Result<()> {
         &Config::load(&opts.config.config)?.get_chain_config(&opts.chain_id)?,
     )?);
```
jayantk, line 16: drive-by. the docs say it gets the provider but it doesn't. It's useful also.

m30m (approved):

PR #2621, opened 2025-04-25
feat(entropy): Event V2 spec

jayantk:
## Summary

Add a new set of EntropyEvents to improve on the previous spec. Specifically, add indexing for the salient attributes, stop depending on struct definitions (which, if changed, break the event signature), and add dummy bytes for future extensibility.

In the process, I have omitted events for code paths that we are going to kill off (like the old pull flow).

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/apps/fortuna/src/chain/ethereum.rs
+++ b/apps/fortuna/src/chain/ethereum.rs
@@ -83,7 +83,7 @@ impl<T: JsonRpcClient + 'static + Clone> SignablePythContractInner<T> {
         {
             // Extract Log from TransactionReceipt.
             let l: RawLog = r.logs[0].clone().into();
-            if let PythRandomEvents::RequestedFilter(r) = PythRandomEvents::decode_log(&l)? {
+            if let PythRandomEvents::Requested1Filter(r) = PythRandomEvents::decode_log(&l)? {
                 Ok(r.request.sequence_number)
             } else {
                 Err(anyhow!("No log with sequence number"))
@@ -147,7 +147,7 @@ impl<T: JsonRpcClient + 'static + Clone> SignablePythContractInner<T> {
             .await?
             .await?
         {
-            if let PythRandomEvents::RevealedFilter(r) =
+            if let PythRandomEvents::Revealed1Filter(r) =
                 PythRandomEvents::decode_log(&r.logs[0].clone().into())?
             {
                 Ok(r.random_number)
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -150,7 +150,9 @@ abstract contract Entropy is IEntropy, EntropyState {
 
         provider.sequenceNumber += 1;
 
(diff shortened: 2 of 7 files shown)


apps/fortuna/src/chain/ethereum.rs
```diff
@@ -83,7 +83,7 @@ impl<T: JsonRpcClient + 'static + Clone> SignablePythContractInner<T> {
         {
             // Extract Log from TransactionReceipt.
             let l: RawLog = r.logs[0].clone().into();
-            if let PythRandomEvents::RequestedFilter(r) = PythRandomEvents::decode_log(&l)? {
```
jayantk, line 86: this naming system seems a bit brittle (I think it's based on the order of the events in the file) but (1) the ABI ordering seems stable and (2) it's temporary anyway so i think it's fine.

target_chains/ethereum/entropy_sdk/solidity/EntropyEventsV2.sol
```diff
@@ -0,0 +1,141 @@
...
+ * @dev This interface is used to emit events that track the lifecycle of random number requests,
+ * provider registrations, and system configurations.
+ */
+interface EntropyEventsV2 {
+    /**
+     * @notice Emitted when a new provider registers with the Entropy system
+     * @param provider The address of the registered provider
+     * @param dummy Additional data field (unused)
+     */
+    event Registered(address indexed provider, bytes dummy);
```
m30m, line 19: Can we change this parameter name to something like `extraArgs`?

Also, I'm pretty sure we will forget to update and remove the `(unused)` comment when we start to use this event 😅

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -349,6 +364,13 @@ abstract contract Entropy is IEntropy, EntropyState {
             userRandomNumber,
             EntropyStructConverter.toV1Request(req)
         );
+        emit EntropyEventsV2.Requested(
+            provider,
+            req.requester,
+            req.sequenceNumber,
+            userRandomNumber,
+            bytes("")
```
m30m, line 372: shouldn't we include more info like gasLimit from the getgo?

target_chains/ethereum/entropy_sdk/solidity/EntropyEventsV2.sol
```diff
@@ -0,0 +1,141 @@
...
+    );
+
+    /**
+     * @notice Emitted when a provider reveals the generated random number
+     * @param provider The address of the provider that generated the random number
+     * @param caller The address of the user who requested the random number (and who receives a callback)
+     * @param sequenceNumber The unique identifier of the request
+     * @param randomNumber The generated random number
+     * @param callbackFailed Whether the callback to the caller failed
+     * @param callbackErrorCode If the callback failed, the error code from the callback. Note
```
m30m, line 44: ErrorCode may not be the best name here since this is the return value of the callback. Also maybe you can specify that this is the first 256 bytes of the return value.

PR #2623, opened 2025-04-27
feat(pyth-lazer/publisher-sdk): Move Transaction Envelope Definitions to SDK

darunrs:
## Summary

- Removes the publisher ID field from the publisher update sent by publishers. Publishers don't need know what their ID is, and its better to not ask them to self report it. 
- Removes update index from feed update context. It's unnecessary, and I will make assertions to make sure the ordering and length of the array match the publisher updates vec exactly. 
- Moved transaction envelope definition to publisher SDK to skip one decode step. Also added comments, adopting Pavel's comment styling in the Pyth Lazer repo.
- Switched the parser to use protoc and then added the argument to generate rust docs for the generated types, as per Pavel's [PR](https://github.com/pyth-network/pyth-lazer/pull/346). 

## Rationale

Reduces some burden on publishers by not requiring them to know their own IDs. The update index field was also confusing. It's better to write thoroughly tested code and make some assumptions instead. The added rust docs should also help with understanding the generated types. 

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/lazer/publisher_sdk/proto/transaction_envelope.proto
+++ b/lazer/publisher_sdk/proto/transaction_envelope.proto
@@ -18,24 +18,24 @@ message TransactionEnvelope {
 // Has different context data depending on the type of transaction.
 // Submitted over Message Queue to be read by rest of Pyth Lazer service.
 message PayloadContext {
-    // [required] ID of publisher based on the access token used to connect
-    optional uint32 publisher_id = 1;
-
     // [required] timestamp wwhen relayer received the signed transaction
-    optional google.protobuf.Timestamp relayer_receive_timestamp = 2;
+    optional google.protobuf.Timestamp relayer_receive_timestamp = 1;
 
     // [required] context set based on type of transaction
     oneof context {
-        PublisherUpdateContext publisher_update_context = 3;
+        PublisherUpdateContext publisher_update_context = 2;
     }
 }
 
 // Context contains status of each feed update found in transaction
 message PublisherUpdateContext {
+    // [required] ID of publisher based on the access token used to connect
+    optional uint32 publisher_id = 1;
+
     // [required] context for each feed update
     // must exactly match length and order of feed updates
     // order of updates are preserved between encoding/decoding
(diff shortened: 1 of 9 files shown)


merolish (approved):
So we determine the publisher id from auth?

darunrs:
> So we determine the publisher id from auth?

Yep, Lazer will determine their ID based on their public key/access key. Also sorry I will do some more things with this PR actually. Will ping when I'm done.

lazer/publisher_sdk/proto/transaction_envelope.proto
```diff
@@ -0,0 +1,67 @@
...
+    // [required] context attached by pyth lazer relayer
+    optional PayloadContext payload_context = 2;
+}
+
+// Context attached by Pyth Lazer Relayer containing information necessary for processing transaction.
+// Has different context data depending on the type of transaction.
+// Submitted over Message Queue to be read by rest of Pyth Lazer service.
+message PayloadContext {
+    // [required] ID of publisher based on the access token used to connect
+    optional uint32 publisher_id = 1;
```
Riateche, line 22: The next option we will likely add to `PayloadContext.context` is a governance update context. That one doesn't have a `publisher_id`, so I think it makes more sense to move this `publisher_id` field to `PublisherUpdateContext` message.

Riateche (approved):
Looks good! See one comment below.

lazer/publisher_sdk/proto/transaction_envelope.proto
darunrs: I don't imagine that we would have a permissionless endpoint on Relayer. If permissioned, we need some way for governance to authn/z with Relayer, at which point we could read their ID. We may not necessarily want to. Maybe the ID takes another form. 

Just curious what you had in mind. For now, I don't mind moving it into publisher context. I had a comment on your PR adding governance on how we determine the governance source.

Riateche: Discussed in Slack. If we add publisher's public key to the payload, I guess we won't need publisher id in the context at all.

PR #2627, opened 2025-04-28
feat(Entropy): Include gas limits and gas usage in events

jayantk:
## Summary

As the PR title says

## Rationale

Showing the gas limit and usage will be useful for users to see if their callback is using too much gas, and also to adjust their usage.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -600,6 +600,7 @@ abstract contract Entropy is IEntropy, EntropyState {
                     randomNumber,
                     false,
                     ret,
+                    SafeCast.toUint32(startingGas - gasleft()),
                     bytes("")
                 );
                 clearRequest(provider, sequenceNumber);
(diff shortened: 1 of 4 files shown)


target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -877,7 +878,7 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents, EntropyEventsV2 {
                 0
             )
         );
-        vm.expectEmit(false, false, false, true, address(random));
+        vm.expectEmit(true, true, true, false, address(random));
```
jayantk, line 881: it turns out that these booleans are which fields of the event to check (4 bools for the 3 topics and data) 🤦 . (This ^ means we check the topics but not the data (which has the callbackGasUsage, which is hard to compute exactly). There is a more explicit test for the data field of this event below.

Also, I'm pretty sure the other calls to expectEmit in this file are doing this wrong, but i'll fix in a separate PR.

target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -599,6 +600,7 @@ abstract contract Entropy is IEntropy, EntropyState {
                     randomNumber,
                     false,
                     ret,
+                    SafeCast.toUint32(startingGas - gasleft()),
```
m30m, line 603: can we move this calculation right after the call return? They may get further away over time and the calculation becomes a bit inaccurate.

m30m (approved):

PR #2629, opened 2025-04-28
fix(staking): use wallet's `sendTransaction` for wallet tester

cprussin:
## Summary

This PR updates the wallet tester to use the wallet's `sendTransaction` method, instead of using `sendTransaction` from `@pythnetwork/solana-utils`.  This should help account for slower wallets like Fireblocks.

## Rationale

We have a fireblocks user who is complaining about not being able to get the tester to work, and we believe the root cause is that tx is being delayed by the wallet -- which would make sense for Fireblocks' UX.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/staking/src/components/WalletTester/index.tsx
+++ b/apps/staking/src/components/WalletTester/index.tsx
@@ -3,14 +3,15 @@
 import type { Idl } from "@coral-xyz/anchor";
 import { Program, AnchorProvider } from "@coral-xyz/anchor";
 import { WalletIcon } from "@heroicons/react/24/outline";
-import {
-  TransactionBuilder,
-  sendTransactions,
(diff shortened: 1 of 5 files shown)


apps/staking/src/components/WalletTester/index.tsx
```diff
@@ -205,28 +206,28 @@ const getHasAlreadyTested = async (
...
-        wallet.publicKey,
-        connection,
-        [
-          {
-            instruction: await testMethod().instruction(),
-            signers: [],
-          },
-        ],
-        {},
-      ),
```
keyvankhademi, line 226: you can still use the transaction builder and use it in `wallet.sendTransaction` instead of manually creating it.

keyvankhademi (approved):

PR #2632, opened 2025-04-28
feat(entropy): More ergonomic v2 interface

jayantk:
## Summary

Add a set of `requestV2` methods that give callers optionality on specifying providers etc. I think we get a set of questions like "what's a provider" or "what's userRandomNumber" which originate from the confusing function signatures in the user interface.

I'm not sure why we didn't do this in the first place -- i think for some reason we thought we couldn't override method names. It works though so I think we go for it.

I think we may want to pull out some of these methods into a separate solidity interface to make it easier for users of entropy to see the methods they want to call as well. If you're onboard with the interface changes, then I can do that next.

## Rationale

This should make it easier for developers to use entropy -- just call `requestV2()` and get a callback.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -295,6 +295,40 @@ abstract contract Entropy is IEntropy, EntropyState {
         }
     }
 
+    function requestV2()
+        external
+        payable
+        override
+        returns (uint64 assignedSequenceNumber)
+    {
+        assignedSequenceNumber = requestV2(
+            getDefaultProvider(),
+            random(),
+            0
+        );
+    }
+
+    function requestV2(
+        uint32 gasLimit
+    ) external payable override returns (uint64 assignedSequenceNumber) {
+        assignedSequenceNumber = requestV2(
+            getDefaultProvider(),
+            random(),
+            gasLimit
+        );
+    }
+
+    function requestV2(
+        address provider,
+        uint32 gasLimit
+    ) external payable override returns (uint64 assignedSequenceNumber) {
+        assignedSequenceNumber = requestV2(
+            provider,
+            random(),
+            gasLimit
+        );
+    }
+
     // As a user, request a random number from `provider`. Prior to calling this method, the user should
     // generate a random number x and keep it secret. The user should then compute hash(x) and pass that
     // as the userCommitment argument. (You may call the constructUserCommitment method to compute the hash.)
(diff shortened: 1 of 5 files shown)


target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -295,6 +295,32 @@ abstract contract Entropy is IEntropy, EntropyState {
         }
     }
 
+    function requestV2()
+        external
+        payable
+        override
+        returns (uint64 assignedSequenceNumber)
+    {
+        assignedSequenceNumber = requestV2(getDefaultProvider(), random(), 0);
```
jayantk, line 304: I think we get a bunch of questions of the form "what's userRandomNumber". I've defaulted it with an in-contract prng here.

m30m: Shouldn't we add some WARNING around to discourage users from doing this?
Also, anything specific to do to make sure we don't receive bounty reports?

m30m: Shouldn't we add some WARNING around to discourage users from doing this?
Also, anything specific to do to make sure we don't receive bounty reports around this?

target_chains/ethereum/contracts/contracts/entropy/EntropyState.sol
```diff
@@ -37,6 +37,7 @@ contract EntropyInternalStructs {
         // proposedAdmin is the new admin's account address proposed by either the owner or the current admin.
         // If there is no pending transfer request, this value will hold `address(0)`.
         address proposedAdmin;
+        bytes32 seed;
```
m30m, line 40: pls add necessary comments

target_chains/ethereum/entropy_sdk/solidity/IEntropy.sol
```diff
@@ -29,6 +29,75 @@ interface IEntropy is EntropyEvents, EntropyEventsV2 {
...
+    /// @return assignedSequenceNumber A unique identifier for this request
+    /// @dev The address calling this function should be a contract that inherits from the IEntropyConsumer interface.
+    /// The `entropyCallback` method on that interface will receive a callback with the returned sequence number and
+    /// the generated random number.
+    ///
+    /// `entropyCallback` will be run with the `gasLimit` provided to this function.
+    /// The `gasLimit` will be rounded up to a multiple of 10k (e.g., 19000 -> 20000), and furthermore is lower bounded
+    /// by the provider's configured default limit.
+    ///
+    /// This method will revert unless the caller provides a sufficient fee (at least `getFeeV2(provider, gasLimit)`) as msg.value.
```
m30m, line 52: `getFeeV2(getDefaultProvider(), gasLimit))`

target_chains/ethereum/entropy_sdk/solidity/IEntropy.sol
```diff
@@ -29,6 +29,75 @@ interface IEntropy is EntropyEvents, EntropyEventsV2 {
...
+    /// Note that provider fees can change over time. Callers of this method should explicitly compute `getFeeV2(provider, gasLimit)`
```
m30m, line 53: `getFeeV2(getDefaultProvider(), gasLimit)`

m30m (approved):

PR #2637, opened 2025-04-29
feat(target_chains/ethereum/pyth): strict minimal updateData parsing

tejasbadadare:
## Summary

Introduces a strict parse mode for `parsePriceUpdatesInternal` named `checkUpdateDataIsMinimal` that reverts if the number of passed in updates in `updateData` is different than `priceIds`. If the update data is minimal and sufficient to satisfy the requested priceIds, parsing succeeds.

## Rationale

- There is a potential attack vector for Pulse where a bad keeper could provide huge updatedata that contains many more price IDs than requested. This doesn't affect correctness, but it could burn more gas than intended during parsing, which the manager pays for. This could be used to  drain manager wallets. Using this new strict parsing mode in Pulse prevents that.

- Gas benchmark delta looks good, <0.5% difference: https://app.warp.dev/block/XCs2XtRCKZ1FhQ7StYoYb7
- Unfortunately this change causes the contract size to exceed the limit by 666 bytes 👿: https://app.warp.dev/block/x6K9sJRtbLRVMdb4aM212C
  - Any hints on things we can chop or refactor to satisfy the limit would be much appreciated! If we can't resolve the size issue, then we need to find a different way to solve this problem.

## How has this been tested?

- [x] Added new tests


target_chains/ethereum/contracts/contracts/pyth/Pyth.sol
```diff
@@ -299,6 +299,24 @@ abstract contract Pyth is
             if (msg.value < requiredFee) revert PythErrors.InsufficientFee();
         }
 
+        // In minimal update data mode, revert if we have more or less updates than price IDs
```
ali-bahjati, line 302: Can you try out doing it on the fly instead of parsing again to save gas (and hopefully contract size)

tejasbadadare: will try 🙏

tejasbadadare: We calculate on the fly now and inlined ParseConfig to chop a struct to save space, but we are still [372 bytes short](https://app.warp.dev/block/xzo3Rd3Km2Zl3dUdC3ESbe) 😭

Any idea if there's something unused that we can remove? Otherwise i'll abandon this approach and just use a heuristic check in `Pulse.updatePriceFeeds` to check that the updatedata isn't too large

ali-bahjati: i recommend you reduce the optimizer runs a bit to match it. the optimizer inlines things to reduce jump costs.

ali-bahjati (approved):

PR #2638, opened 2025-04-29
feat(fortuna): Improve startup flow

m30m:
## Summary

- Each chain is started separately
- Run block timestamp tracking only if chain is registered
- Run hash generation in a blocking thread so that tokio can continue async operations
- Retry chain startup on failure

## Rationale

These changes improve the reliability of running fortuna in production.

## How has this been tested?

- [ ] Added new tests


apps/fortuna/src/api.rs
```diff
@@ -43,7 +43,7 @@ pub struct ApiMetrics {
 
 #[derive(Clone)]
 pub struct ApiState {
-    pub chains: Arc<HashMap<ChainId, BlockchainState>>,
+    pub chains: Arc<RwLock<HashMap<ChainId, BlockchainState>>>,
```
jayantk, line 46: you can use Dashmap for a concurrent hashmap https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html

jayantk: Also I think this should be a `HashMap<ChainId, Option<BlockchainState>>` so "chain is uninitialized" isn't conflated with "unknown chain id".  You could make BlockchainState an enum or something as well to handle this.

apps/fortuna/src/command/run.rs
```diff
@@ -136,41 +101,49 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
-    let mut chains: HashMap<ChainId, BlockchainState> = HashMap::new();
-    for result in states {
-        let (chain_id, state) = result?;
-
-        match state {
-            Ok(state) => {
-                chains.insert(chain_id.clone(), state);
+            match state {
+                Ok(state) => {
+                    keeper_metrics.add_chain(chain_id.clone(), state.provider_address);
```
jayantk, line 113: it's probably better to run this for every chain before it's set up (?) wondering how we will report metrics in the case where the chain never gets initialized.

apps/fortuna/src/state.rs
```diff
@@ -59,11 +60,14 @@ impl PebbleHashChain {
...
         let secret: [u8; 32] = Keccak256::digest(input).into();
-        Ok(Self::new(
-            secret,
-            chain_length.try_into()?,
-            sample_interval.try_into()?,
-        ))
+
+        let chain_length: usize = chain_length.try_into()?;
+        let sample_interval: usize = sample_interval.try_into()?;
+        let hash_chain = spawn_blocking(move || Self::new(secret, chain_length, sample_interval))
```
jayantk, line 66: I don't understand how this helps. This ran synchronously before, so how does the blocking thread help exactly? (I don't totally understand blocking threads)

apps/fortuna/src/keeper.rs
```diff
@@ -215,6 +216,14 @@ pub async fn run_keeper_threads(
...
                     .in_current_span(),
                 );
+                spawn(
+                    track_block_timestamp_lag(
+                        chain_id.clone(),
+                        contract.client(),
+                        keeper_metrics.clone(),
+                    )
+                    .in_current_span(),
+                );
```
jayantk, line 226: how does moving this help us detect that the chain isn't initialized?

apps/fortuna/src/state.rs
```diff
@@ -42,7 +43,7 @@ impl PebbleHashChain {
         }
     }
 
-    pub fn from_config(
+    pub async fn from_config(
```
jayantk, line 46: can i suggest making a separate constructor method that spawns the original from_config in a blocking thread? like from_config_async

apps/fortuna/src/api.rs
```diff
@@ -94,6 +94,12 @@ pub struct BlockchainState {
     pub confirmed_block_status: BlockStatus,
 }
 
+#[derive(Clone)]
+pub enum ApiBlockChainState {
+    Uninitialized,
+    Initialized(BlockchainState),
+}
```
jayantk, line 101: 👍

apps/fortuna/src/command/run.rs
```diff
@@ -185,27 +144,45 @@ pub async fn run(opts: &RunOptions) -> Result<()> {
...
-    if let Some(keeper_private_key) = config.keeper.private_key.load()? {
-        spawn(run_keeper(
-            chains.clone(),
-            config.clone(),
+    run_api(opts.addr, chains.clone(), metrics_registry.clone(), rx_exit).await?;
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn setup_chain(
```
jayantk, line 152: setup_chain_and_run_keeper ?

apps/fortuna/Cargo.lock
```diff
@@ -1554,7 +1554,7 @@ dependencies = [
 
 [[package]]
 name = "fortuna"
-version = "7.5.2"
+version = "7.5.3"
```
jayantk, line 1557: why is this here but not in Cargo.toml

jayantk (approved):
nice

apps/fortuna/Cargo.lock
m30m: main branch didn't updated the Cargo.lock

PR #2643, opened 2025-05-01
refactor(hermes/server): use program accounts instead of mapping acct

ali-bahjati:
## Rationale

We used mapping account to get the product metadata but it involves many calls and many node providers have been facing issues with it. This change removes mapping account and uses `getProgramAccounts` to fetch all accounts in one go.

## How has this been tested?

- [x] Manually tested the code. I tested the readiness probe, made queries in the symbols endpoint, and made sure that other functionalities keep working fine.


tejasbadadare (approved):
lgtm! left a couple comments

PR #2656, opened 2025-05-06
feat(apps/price_pusher): update Aptos balance tracker to use new SDK and improve balance retrieval

cctdaniel:
## Summary

Updated the Aptos client implementation to use the new `@aptos-labs/ts-sdk` instead of the deprecated `aptos` package, and simplified balance tracking by using the dedicated `getAccountAPTAmount` method.

## Rationale

These changes are necessary to:
1. Use the latest Aptos TypeScript SDK which offers improved functionality and better support
2. Replace the complex manual balance tracking with the SDK's built-in method that automatically handles both Fungible Asset and Coin formats
3. Ensure compatibility with upcoming Movement chain updates that follow Aptos SDK standards

## How has this been tested?

- [ ] Added new tests

The changes have been manually tested against both Movement and Aptos chains to ensure the balance tracker correctly fetches and displays token balances. Tests verified that the `getAccountAPTAmount` method correctly tracks balances regardless of which resource format is used by the chain.

### DIFF
--- a/apps/price_pusher/package.json
+++ b/apps/price_pusher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/price-pusher",
-  "version": "9.3.1",
+  "version": "9.3.2",
   "description": "Pyth Price Pusher",
   "homepage": "https://pyth.network",
   "main": "lib/index.js",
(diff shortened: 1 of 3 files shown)


ali-bahjati (approved):
Nice! can you remove the `aptos` dependency package before merging?

cctdaniel:
its still being used in the main price pusher package and im reluctant to replace it since its working for now

PR #2659, opened 2025-05-07
feat(argus): internal interfaces and actor model

tejasbadadare:
## Summary

Skeleton for Argus. Defines internal interfaces and sets up the actor model using [ractor](https://github.com/slawlor/ractor) framework.

Note to reviewer: the interesting stuff is in `actors` and `adapters`, see the `types.rs` in each for the big picture. 

### Actors 
- In a nutshell, actors have independent responsibilities and can communicate via message-passing. Each actor specifies the messages it can receive and reply to.
  - For example, the `PythPriceListener` actor is responsible for tracking off-chain prices. It can handle a `GetLatestPrice(feed_ids)` message, which it serves from its internal store of prices that are being polled/streamed from Hermes in the background. It can also handle a `UpdateFeedIdSet(feed_ids)` message, which can be used to adjust the feeds being tracked. 
  - Similar actors exist for the other parallel tasks that happen in the keeper: keeping active subscriptions up to date, tracking on-chain prices, etc. The orchestration is handled by the Controller actor.
- See `actors/types.rs` for the message definitions, these enums describe the actors' interfaces.
- The actors are currently spawned in `keeper.rs`, but this will eventually be done using ractor's Factory concept, which provides durability and lifecycle mgmt for actors.

### Adapters
- Interactions with external services like Hermes and the blockchain are modeled by the traits in `adapters/types.rs`. These traits are used by the actors, and are implemented by `hermes.rs`, `contract.rs`, etc.
  - For example, the `SubscriptionListener` actor expects an implementor of `ReadChainSubscriptions`. This trait specifies `get_active_subscriptions()` and `subscribe_to_subscription_events()`, which are impl'd by `PythPulse<M>` in `contract.rs`.

## How has this been tested?
Not tested yet. Next up is writing the Controller's main update loop, wiring it up to mock implementations, and ensuring the actor communication works. Then, we will implement individual actor logic, and eventually wire it up to the blockchain instead of mock adapters.


apps/argus/src/actors/types.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum SubscriptionListenerMessage {
+    GetActiveSubscriptions,
+    RefreshSubscriptions,
+}
+
+#[derive(Debug)]
+pub enum PythPriceListenerMessage {
+    GetLatestPrice(PriceId, RpcReplyPort<Option<Price>>),
```
ali-bahjati, line 15: I would advice against using reply between actors because on the sender side it results in an extra async blocker (other than the main message queue) and it defeats the concurrency benefits of the actors (not getting into livelocks and deadlocks). This is mostly useful if an external observer outside the actors wants to get a reply from the system.

ali-bahjati: I kind of expected to have a good construct for a reply-based mechanism using cast but couldn't find it. probably you need to pass a reference to the requesting actor (and track the request/...)

ali-bahjati (commented):
I like the design. I recommend adding module docs to each of your actors now that you’re establishing the structure and design of the service.

apps/argus/src/actors/chain_price_listener.rs
```diff
@@ -0,0 +1,96 @@
...
+        time::Duration,
+    },
+    tokio::{sync::RwLock, time},
+    tracing,
+};
+
+#[allow(dead_code)]
+pub struct ChainPriceListenerState {
+    chain_id: String,
+    contract: Arc<dyn GetChainPrices + Send + Sync>,
```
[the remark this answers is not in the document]
jayantk: (and for ChainId also)

apps/argus/src/actors/types.rs
jayantk: I think we had a bunch of issues with two-way communication between actors in pyth-agent. I don't know if we had a good solution to that though. maybe you simply use shared memory with a lock for these kinds of read requests?

jayantk (commented):
seems generally reasonable to me. I do think that revisiting the pyth-agent experience with actors is worthwhile, as that system got pretty janky over time. Probably some lessons there about how (not) to use Rust actors.

apps/argus/src/actors/types.rs
ali-bahjati: yeah i think we should either do full actor model or full shared memory model (like hermes/agent/..)

apps/argus/src/actors/chain_price_listener.rs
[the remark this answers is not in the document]
tejasbadadare: We've got one actually, let me use that here

apps/argus/src/actors/types.rs
tejasbadadare: Ah interesting.  Do either of you remember why 2-way communication didn't work so well in pyth-agent? 

Re @ali-bahjati about the reply-based mechanisms in ractor using cast:
- `cast` supports fully async replies -- you can pass an ActorRef in the message, the receiver can use it to send a reply message to the sender at a later time. 
- `call` supports awaitable replies -- the receiver gets an RpcReplyPort that it can use to reply while handling the incoming message. the sender can await this reply. the receiver gets messages serially, so it doesn't need to track requests.

I find `call` semantics more ergonomic for certain things like fetching the latest price in the Controller, not sure how we would get it to work with `cast`, since we'd need the latest price in the middle of the Controller's update loop. Lmk if you guys have ideas here.

So the tradeoff of using `call` with GetLatestPrice operations would be that we're slightly compromising the fully async nature of the actors (but prices would still get polled in an unblocked background thread), but we still get the isolation benefits. I'm inclined to stick with `call` here for simplicity, but I hear you guys saying this might have been a regret in the past :) Let me know what you think.

apps/argus/src/actors/chain_price_listener.rs
```diff
@@ -0,0 +1,96 @@
...
+    feed_ids: HashSet<PriceId>,
+    latest_prices: Arc<RwLock<HashMap<PriceId, Price>>>,
```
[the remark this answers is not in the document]
tejasbadadare: Oo very cool, thanks for sharing

apps/argus/src/actors/price_pusher.rs
```diff
@@ -0,0 +1,120 @@
...
+};
+
+pub struct PricePusherState {
+    chain_name: String,
+    contract: Arc<dyn UpdateChainPrices + Send + Sync>,
+    pyth_price_client: Arc<dyn ReadPythPrices + Send + Sync>,
+    backoff_policy: ExponentialBackoff,
+}
+
+pub struct PricePusher;
```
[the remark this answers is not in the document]
tejasbadadare: Yeah good point. We can definitely handle this better in PricePusher. I'm thinking we can `spawn` a task to attempt the update with backoff and track the handles. If a newer push for a subscription comes in while the last one is stuck in backoff, we can cancel the outstanding task and spawn a new one. This also means the actor's handle function doesn't block, and we can multiple push attempts for different subscriptions active at a time.

apps/argus/src/actors/types.rs
tejasbadadare: Did some thinking, and yeah we are indeed in between paradigms here. By introducing synchronous dependencies between actors (using `call`,) we trade away the high concurrency of the actor model for simplicity. From what I gather, this became an issue in pyth-agent as the code evolved and added implicit dependencies between actors resulting in dead/livelocks that were hard to diagnose and fix. In Argus, our actor design is a DAG, so it felt okay to introduce these sync dependencies, but ultimately I settled on going the full shared memory model route here: https://github.com/pyth-network/pyth-crosschain/pull/2682

PR #2660, opened 2025-05-07
refactor(governance/xc_admin_frontend): refactor xc_admin_frontend to support lazer and other programs

cctdaniel:
## Summary

This PR introduces a modular architecture for supporting multiple Pyth programs (Pyth Core and Pyth Lazer) in our frontend application. Key changes include:

1. Added an extensible adapter pattern with `ProgramAdapter` interface and program-specific implementations
2. Created a `ProgramType` enum to differentiate between Pyth Core and Pyth Lazer
3. Implemented UI components for program selection and program-specific views
4. Added context providers to manage program state across the application
5. Refactored existing code to use the new adapter-based architecture
6. Eliminated code duplication by leveraging shared components

## Rationale

These changes were necessary to:

1. Support the upcoming Pyth Lazer program alongside the existing Pyth Core program
2. Isolate program-specific logic into separate modules for better maintainability
3. Allow for easier integration of additional Pyth programs in the future
4. Provide a clean, intuitive UI for users to switch between different programs
5. Reduce code duplication through a common interface for all program interactions
6. Make the codebase more maintainable by ensuring program-specific changes only need to be made in isolated locations

## How has this been tested?

- [ ] Added new tests

Tested by:
- Verifying program switching works correctly in the UI
- Confirming Pyth Core functionality remains intact with the new architecture
- Ensuring Pyth Lazer placeholder components display correctly
- Checking that all UI components adapt appropriately to the selected program
- Validating that the adapter pattern correctly redirects requests to the appropriate program implementation
- Confirming program context state persists across page navigation


governance/xc_admin/packages/xc_admin_frontend/components/programs/PythCore.tsx
ali-bahjati, line 1: i think this component can be shared between the two oracles (the UI aspect is kind of the same, no?)

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/Proposals.tsx
ali-bahjati, line 1: Hmmm why do we need to change proposals?

ali-bahjati (commented):
this is a great start, i think we can change the design a bit to have better boundaries and also better types. if there needs to be any change on the proposals side we can do it as part of another PR (but i don't think we need it)

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_adapter.ts
```diff
@@ -0,0 +1,787 @@
...
+   */
+  private sortObjectByKeys(obj: any) {
+    const sortedObj: any = {};
+    Object.keys(obj)
+      .sort()
+      .forEach((key) => {
+        sortedObj[key] = obj[key];
+      });
+    return sortedObj;
+  }
```
cprussin, line 326: I have 2 thoughts here:

1. [Javascript object property ordering is complex](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582), engine dependent, brittle, and is not intended to be used to drive real code.  If you need guaranteed ordering, you should be using an array of tuples, not an object.

2. If you were to write this code, in getting into the habit of writing immutable, functional code, this kind of thing is better done using `Object.fromEntries` and `entries`, e.g. something like this:

```tsx
const sortObjectByKeys = (obj: Record<string, unknown>) =>
  Object.fromEntries(obj.entries().toSorted((a, b) => a[0].localeCompare(b[0])));
```

governance/xc_admin/packages/xc_admin_common/src/programs/program_adapter.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Parse raw on-chain accounts into a standardized configuration object
+   * for the specific program.
+   *
+   * @param accounts Array of account data from the blockchain
+   * @param cluster The Pyth cluster where the accounts were fetched from
+   * @returns Program-specific configuration object
+   */
+  getConfigFromRawAccounts(accounts: any[], cluster: PythCluster): any;
```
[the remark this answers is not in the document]
cprussin: major +1, `any` is a cancer and it causes a lot of code well beyond what you may intend to silently avoid typechecking.  Use `unknown` if you must but never use `any`.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -0,0 +1,14 @@
...
+  /**
+   * Original Pyth oracle program
+   */
+  PYTH_CORE = "PYTH_CORE",
+
+  /**
+   * Next-generation Pyth oracle program
+   */
+  PYTH_LAZER = "PYTH_LAZER",
+}
```
cprussin, line 14: I generally recommend against string enums, just use regular enums -- string enums don't really have any benefits over regular enums apart from automatic human-readable serialization, but that comes at a greater memory overhead cost and usually isn't actually useful -- enums should intentionally be opaque and if you want to map to clean names, use a record.

cprussin (commented):
Just leaving some initial thoughts, I haven't been able to read through the whole PR yet, but I'll give it a closer look when it's nearer to being ready for release

governance/xc_admin/packages/xc_admin_common/src/programs/program_adapter.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This adapter pattern allows the frontend to use a consistent interface
+ * while supporting multiple program implementations.
+ */
+export interface ProgramAdapter {
+  /**
+   * Program address on Solana
+   * Each program may have different addresses on different clusters
+   */
+  getProgramAddress(cluster: PythCluster): PublicKey;
```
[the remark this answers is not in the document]
cctdaniel: lazer is deployed on solana too right?

governance/xc_admin/packages/xc_admin_frontend/components/tabs/Proposals/Proposals.tsx
cctdaniel: just better segregation and allowing users to view core/lazer proposals, but can also just use the single proposals file and show all proposals

governance/xc_admin/packages/xc_admin_common/src/programs/adapter_factory.ts
```diff
@@ -0,0 +1,23 @@
...
+import { PythLazerAdapter } from "./lazer/lazer_adapter";
+
+/**
+ * Factory function to get the appropriate program adapter based on program type.
+ *
+ * @param type The type of program to get the adapter for
+ * @returns The corresponding program adapter
+ * @throws Error if the adapter for the given program type is not implemented
+ */
+export function getProgramAdapter(type: ProgramType): ProgramAdapter {
```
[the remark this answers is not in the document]
cctdaniel: thanks for the detailed explanation, will change to use functional pattern

governance/xc_admin/packages/xc_admin_common/src/programs/program_registry.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Function to format the configuration for downloading as a JSON file
+ */
+export const getDownloadableConfig: Record<
+  ProgramType,
+  (config: ProgramConfig) => DownloadableConfig
+> = {
+  [ProgramType.PYTH_CORE]: (config: ProgramConfig) => {
+    if ("mappingAccounts" in config) {
+      return pythCore.getDownloadableConfig(config as RawConfig);
```
cprussin, line 64: We should avoid typecasting -- in particular:

1. Typescript narrowing can be used to determine if a value from a union type matches a more specific type; so in particular if you have something like this:

```ts
type A = { kind: "A" as const; someOtherValue: "foo" };
type B = { kind: "B" as const; something: "bar" };
type C = A | B;
```

then typescript can narrow a `C` to an `A` or `B` by checking the `kind` field:

```
const doSomething = (value: C) => {
  if (value.kind === "A") {
    console.log(value.someOtherValue); // typescript knows you have an `A`, so you can access it's fields
  } else {
    console.log(value.something); // typescript knows you have an `B`, so you can access it's fields
  }
}
```

See [here](https://dev.to/darkmavis1980/what-are-typescript-discriminated-unions-5hbb) for some more info on discriminated unions -- note that the discriminator does not need to be a string field but can be any field that clearly disambiguates between members of the union.  I believe that this is a simplified analogy for what you have here, but let me know if not.


2. As a second point, for this function, it seems that specifying the program type is actually redundant, given that the config can be disambiguated so you can determine the program type just given the config.  Does it  make sense for this interface to explicitly map the program type to the `getDownloadableConfig` interface, or in this case should you just have `const getDownloadableConfig = (config: ProgramConfig) => ...` and let the function determine the correct implementation automatically based on the shape of `config`?

governance/xc_admin/packages/xc_admin_common/src/programs/program_registry.ts
```diff
@@ -0,0 +1,120 @@
...
+};
+
+/**
+ * Function to generate the necessary instructions to apply configuration changes
+ */
+export const generateInstructions = {
+  [ProgramType.PYTH_CORE]:
+    pythCore.generateInstructions as GenerateInstructionsFn<ProgramType.PYTH_CORE>,
+  [ProgramType.PYTH_LAZER]:
+    pythLazer.generateInstructions as GenerateInstructionsFn<ProgramType.PYTH_LAZER>,
```
cprussin, line 98: Same comment about avoiding  type casting; we should figure out a way to implement this without type casts if possible; type casts are a really nasty thing that cripples typescript's ability to make guarantees about your code and should only be used in very specific scenarios where truly necessary; it's morally equivalent to an implicit `unsafe {}` block in Rust.

In this case, I think the issue is that your generic types don't actually make a lot of sense for what you're returning so I would rethink the types -- here in particular, you shouldn't have any reason why you need to modify the inferred types of the `generateInstructions` functions; it's not clear to me what this type cast is actually trying to accomplish but depending on the goals I would look at changing the types of the `generateInstructions` functions, perhaps adding return types or more strict typing in the implementations or arguments so that the inferred types are what you want them to be.

governance/xc_admin/packages/xc_admin_common/src/programs/program_registry.ts
```diff
@@ -0,0 +1,120 @@
...
+  return {
+    getProgramAddress: getProgramAddress[type],
+    isAvailableOnCluster: isAvailableOnCluster[type],
+    getConfig: getConfig[type],
+    getDownloadableConfig: getDownloadableConfig[type],
+    validateUploadedConfig: validateUploadedConfig[type],
+    generateInstructions: generateInstructions[
+      type
+    ] as GenerateInstructionsFn<T>,
+  };
```
cprussin, line 119: I would not export this; there's really no good reason for this to exist.  Any consumer of this will just be adding indirection:

```ts
import { ProgramType, getProgramFunctions } from "...";

const theProgramType = ProgramType.PYTH_CORE;
...
const functions = getProgramFunctions(theProgramType);
...
const config = functions.getConfig(...);
```

Now, tracing from `functions.getConfig` to what is actually happening involves backtracing the code / using debuggers / etc, jumping through layers of indirection, etc., rather than just jumping to the function definition.  This can be especially painful if someone decided to put `functions` in some very disconnected context, e.g. if they put it into a React context or something.

Compared to:

```ts
import { ProgramType, getConfig } from "...";

const theProgramType = ProgramType.PYTH_CORE;
...
const config = getConfig[theProgramType](...);
```

Now it's trivial to find what `getConfig` is trivial.  You expect your consumers to keep the `ProgramType` as their context which is trivial to understand since this is a simple enum type and all functionality is trivially traceable from the call site.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -0,0 +1,254 @@
...
+ */
+export type LazerConfig = {
+  programType: ProgramType.PYTH_LAZER;
+  // Make cluster optional since Lazer might not be tied to a specific cluster
+  cluster?: PythCluster;
+  // More generic data source instead of Solana-specific accounts
+  feeds: LazerFeed[];
+  // Additional metadata that might be relevant for Lazer
+  metadata?: Record<string, unknown>;
+};
```
cprussin, line 89: I generally recommend trying to keep types where they are used.  In this case, this is a lazer specific type, so it should be in `lazer_functions.ts`.

I generally avoid `types.ts` files unless you have entities which really don't have a clear home / system-wide entities.  That may be true for some of these types, but I don't think it's true for stuff that's specific to one system or the other.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -0,0 +1,254 @@
...
+export type GetConfigParams =
+  | ({
+      programType: ProgramType.PYTH_CORE;
+    } & CoreConfigParams)
+  | ({ programType: ProgramType.PYTH_LAZER } & LazerConfigParams);
+
+/**
+ * Function to get configuration data
+ */
+export type GetConfigFn = (params: GetConfigParams) => ProgramConfig;
```
cprussin, line 113: Types like this should not be necessary; in general it's best to let typescript infer function types.  If you must hint the function type (e.g. if the body is complex and you want to guarantee you're returning what you expect), the you can do that by adding the return type to the function definition rather than letting typescript infer the return type.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -0,0 +1,254 @@
...
+  address: string;
+  metadata: Omit<Product, "price_account">;
+  priceAccounts: DownloadablePriceAccount[];
+};
+
+/**
+ * Type for downloadable configuration
+ */
+export type DownloadableConfig = {
+  [symbol: string]: DownloadableProduct;
```
cprussin, line 149: Generally we should prefer `Record<string, DownloadableProduct>` (or at least, we should be consistent about how we defined indexed objects, and most of our code uses `Record`), see [here](https://typescript-eslint.io/rules/consistent-indexed-object-style) for details.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -0,0 +1,254 @@
...
+ * Collection of functions for each program type
+ */
+export interface ProgramFunctions<T extends ProgramType = ProgramType> {
+  getProgramAddress: GetProgramAddressFn;
+  isAvailableOnCluster: IsAvailableOnClusterFn;
+  getConfig: GetConfigFn;
+  getDownloadableConfig: GetDownloadableConfigFn;
+  validateUploadedConfig: ValidateUploadedConfigFn;
+  generateInstructions: GenerateInstructionsFn<T>;
+}
```
cprussin, line 254: I think a lot of the types in this file will be removed or become much simpler if you allow typescript to infer types and move types to where they are used as much as possible.  There will still be some common interfaces that you want to enforce, but my advice is to make those as minimal as possible, and to enforce those in places like `program_registry.ts` where you're using them.  As a dumb example, I would approach the architecture here using patterns like this:

```ts
// core.ts
export const doAThing = (arg: { foo: string, bar: string }) => doSomethingWith(arg);

// lazer.ts
export const doAThing = (arg: { foo: number, bar: string }) => doSomethingWith(arg);

// program_registry.ts
export const doAThing = {
  [ProgramType.Core]: core.doAThing,
  [ProgramType.Lazer]: lazer.doAThing
} as const;
```

That should be enough so that typescript knows that when calling `doAThing[someProgramType](config)`, that `config` must be compatible with the arg type depending on what it can statically analyze about `someProgramType`.  It may be OK for the type of `config` to be incompatible between the program types (or even to have wildly different function signatures), as in the example here, which means you are expecting your caller to understand how to disambiguate between them (which in this case is probably correct, since how will they know how to construct a valid `config` if the shape depends on the program type?).  But maybe in other cases, you want to ensure that the maps exported from `program_registry.ts` map to functions which all match the same shape, and you want typescript to tell you if you accidentally mess that up, in which case you do something like this:

```ts
export const doAThing = {
  [ProgramType.Core]: core.doAThing,
  [ProgramType.Lazer]: lazer.doAThing
} as const satisfies Record<ProgramType, (config: { foo: string; bar: string }) => string>
```

In cases where you can't have a consistent interface, I'd question if the abstraction in `program_registry.ts` is even valuable -- in such cases, the caller must know and handle the different program types manually anyways, so you may as well just export the program-specific functions directly and let the caller figure out which one to use.  And if you do that, it's ok to have a mix of things, some of which can operate on either program type and can be generic, and others which only work for one type or the other, and require the caller to statically know which program type they're using.  One of the really good things about simplifying these types is that it helps make it very obvious where you have functions that can be generic, and where you have stuff that really requires the caller to know which program type they're interfacing with.

governance/xc_admin/packages/xc_admin_frontend/components/ProgramSwitch.tsx
```diff
@@ -0,0 +1,101 @@
...
+  </svg>
+)
+
+/**
+ * Format display name from enum name (e.g., "PYTH_CORE" -> "Pyth Core")
+ */
+const formatDisplayName = (value: ProgramType): string => {
+  const enumName = PROGRAM_TYPE_NAMES[value]
+  return enumName
+}
```
cprussin, line 31: Minor nit but this looks like unnecessary abstraction

governance/xc_admin/packages/xc_admin_frontend/components/ProgramSwitch.tsx
```diff
@@ -0,0 +1,101 @@
...
+const ProgramSwitch = ({ light = false }: { light?: boolean }) => {
+  const { programType, setProgramType } = useProgramContext()
+
+  // Convert enum to array of options
+  const programOptions = Object.values(ProgramType)
+    .filter((value): value is ProgramType => typeof value === 'number') // Filter out reverse mappings
+    .map((value) => ({
+      value,
+      label: formatDisplayName(value),
+    }))
```
cprussin, line 46: I think this can just be `Object.entries(PROGRAM_TYPE_NAMES).map(([value, label]) => ({ value, label }))`

cprussin (changes requested):
Definitely looking much much better, thanks a ton for all the effort!  I have some more suggestions on the `xc_admin_common` side; once we nail this code down then I'll look closer at the React stuff

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,827 @@
...
+    .map((account) => {
+      const parsed = parseMappingData(account.account.data);
+      return {
+        next: parsed.nextMappingAccount,
+        address: account.pubkey,
+        products: parsed.productAccountKeys
+          .filter((key) => productRawConfigs[key.toBase58()])
+          .map((key) => {
+            const product = productRawConfigs[key.toBase58()];
+            delete productRawConfigs[key.toBase58()];
```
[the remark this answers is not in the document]
cctdaniel: this deletion prevents duplicate product accounts from being included in multiple mapping accounts, once a product is assigned to a mapping account, it's removed from productRawConfigs to ensure it's not assigned to another mapping account later in the process, will use a `Map` to track the processed key instead to make it more functional and not modify the original `productRawConfig` object

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,827 @@
...
+  }
+
+  const accounts = params.accounts;
+
+  // First pass: Extract price accounts
+  const priceRawConfigs = Object.fromEntries(
+    accounts
+      .filter(
+        (account) =>
+          parseBaseData(account.account.data)?.type === AccountType.Price,
```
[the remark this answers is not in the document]
cctdaniel: yeah good point will refactor

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,827 @@
...
+
+  return sortData(symbolToData);
+}
+
+/**
+ * Validate an uploaded configuration against the current configuration
+ */
+export function validateUploadedConfig(
+  existingConfig: DownloadableConfig,
+  uploadedConfig: unknown,
```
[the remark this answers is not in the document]
cctdaniel: will make it `DownloadableConfig` for now and make changes to the respective program config once Lazer config is confirmed

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,827 @@
...
+    // Check for changes to existing symbols
+    for (const symbol of Object.keys(uploadedConfigTyped)) {
+      // Remove duplicate publishers
+      if (
+        uploadedConfigTyped[symbol]?.priceAccounts?.[0]?.publishers &&
+        Array.isArray(uploadedConfigTyped[symbol].priceAccounts[0].publishers)
+      ) {
+        uploadedConfigTyped[symbol].priceAccounts[0].publishers = [
+          ...new Set(uploadedConfigTyped[symbol].priceAccounts[0].publishers),
+        ];
```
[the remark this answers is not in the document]
cctdaniel: yeah i'll create a deep copy of it

governance/xc_admin/packages/xc_admin_frontend/components/ProgramSwitch.tsx
```diff
@@ -0,0 +1,93 @@
...
+      {({ open }) => (
+        <>
+          <Menu.Button
+            className={`inline-flex w-full items-center justify-between py-3 px-6 text-sm outline-0 ${
+              light ? 'bg-beige2' : 'bg-darkGray2'
+            }`}
+          >
+            <span className="mr-3">
+              {programOptions.find((option) => option.value === programType)
+                ?.label || PROGRAM_TYPE_NAMES[programType]}
```
cprussin, line 51: Minor nit but we should use `??` instead of `||`

governance/xc_admin/packages/xc_admin_frontend/components/programs/PythCore.tsx
```diff
@@ -0,0 +1,505 @@
...
+import PermissionDepermissionKey from '../PermissionDepermissionKey'
+import { Wallet } from '@coral-xyz/anchor/dist/cjs/provider'
+
+interface PythCoreProps {
+  proposerServerUrl: string
+}
+
+const PythCore = ({ proposerServerUrl }: PythCoreProps) => {
+  const [data, setData] = useState<any>({}) // eslint-disable-line @typescript-eslint/no-explicit-any
+  const [dataChanges, setDataChanges] = useState<Record<string, any>>() // eslint-disable-line @typescript-eslint/no-explicit-any
```
cprussin, line 39: Must we use `any` for these?  I think these types are well known right?

governance/xc_admin/packages/xc_admin_frontend/components/programs/PythCore.tsx
```diff
@@ -0,0 +1,505 @@
...
+          setIsSendProposalButtonLoading(false)
+        }
+      }
+    }
+
+    handleSendProposalButtonClickAsync()
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const MetadataChangesRows = ({ changes }: { changes: any }) => {
```
cprussin, line 172: We should try very hard not to create components inside of closures, especially not in the render path of other components -- this will break React's ability to optimize the render tree as React will see this as a new function on each render pass and as such it has no way to know that the part of the subtree is not changing.

I recommend hoisting this component out to the top level of the module.

Same goes for the other components created inside this closure.

Also if we can avoid `any` here we should.

cprussin (changes requested):
Definitely getting a lot closer, but still a few things to fix.  The most important things to me are to avoid using `any` as much as possible (still a few places here that introduce it) and to avoid creating components in the render path of other componets.

I realize that a lot of this stuff was just moved around and isn't actually new code, so maybe it's best to leave the things that aren't net new alone and get this shipped, and to create a follow up PR to fix those patterns?

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,891 @@
...
+} from "../types";
+import { Program } from "@coral-xyz/anchor";
+import { PythOracle } from "@pythnetwork/client/lib/anchor";
+import { MessageBuffer } from "message_buffer/idl/message_buffer";
+
+/**
+ * Maximum sizes for instruction data to fit into transactions
+ */
+const MAX_SIZE_ADD_PRODUCT_INSTRUCTION_DATA = 369;
+const MAX_SIZE_UPD_PRODUCT_INSTRUCTION_DATA = 403; // upd product has one account less
```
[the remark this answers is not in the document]
cctdaniel: yeah lets fix this in a following pr

governance/xc_admin/packages/xc_admin_common/src/programs/core/core_functions.ts
```diff
@@ -0,0 +1,891 @@
...
+
+/**
+ * Sort configuration data for consistent output
+ */
+function sortData(data: DownloadableConfig): DownloadableConfig {
+  return mapValues(data, (productData: DownloadableProduct) => ({
+    address: productData.address,
+    metadata: Object.fromEntries(
+      sortObjectByKeys(productData.metadata),
+    ) as Omit<Product, "price_account">,
```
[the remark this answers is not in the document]
cctdaniel: same for this, will fix this in following pr

cprussin (approved):

PR #2665, opened 2025-05-08
WIP: Chore(evm) add combine pricefeed

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
+++ b/target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
@@ -399,4 +399,66 @@ contract PythUtilsTest is Test, WormholeTestUtils, PythTestUtils, IPythEvents {
         // 0.00001 and we are targetDecimals is 2.
         assertEq(PythUtils.convertToUint(100, -5, 2), 0);
     }
+
+    function testCombinePrices() public {
+        // Test case 1: Basic price combination (ETH/USD / BTC/USD = ETH/BTC)
+        PythStructs.Price memory ethUsd = PythStructs.Price({
+            price: 99875005,  
+            conf: 10,
+            expo: -8,    
+            publishTime: block.timestamp
+        });
+        
+        PythStructs.Price memory btcUsd = PythStructs.Price({
+            price: 206362333702,
+            conf: 100,
+            expo: -8,     
+            publishTime: block.timestamp
+        });
+
+        (int64 price, int32 expo) = PythUtils.combinePrices(ethUsd, btcUsd);
+        assertApproxEqRel(price, 2016367433623696, 9e17); 
+        assertEq(expo, -18);
+
+        // // Test case 2: Different exponents
+        // PythStructs.Price memory smallPrice = PythStructs.Price({
+        //     price: 100,   // $0.01
+        //     conf: 1,
+        //     expo: -4,     // 4 decimals
+        //     publishTime: block.timestamp
+        // });
+
+        // PythStructs.Price memory largePrice = PythStructs.Price({
+        //     price: 1000,  // $10
+        //     conf: 10,
+        //     expo: -2,     // 2 decimals
+        //     publishTime: block.timestamp
+        // });
+
+        // (price, expo) = PythUtils.combinePrices(smallPrice, largePrice, -2);
+        // assertEq(price, 1);    // 0.01
+        // assertEq(expo, -2);
+
+        // // Test case 3: Revert on negative prices
+        // PythStructs.Price memory negativePrice = PythStructs.Price({
+        //     price: -100,
+        //     conf: 10,
+        //     expo: -2,
+        //     publishTime: block.timestamp
+        // });
+
+        // vm.expectRevert();
+        // PythUtils.combinePrices(negativePrice, btcUsd, -2);
+
+        // // Test case 4: Revert on positive exponents
+        // PythStructs.Price memory invalidExpo = PythStructs.Price({
+        //     price: 100,
+        //     conf: 10,
+        //     expo: 2,
+        //     publishTime: block.timestamp
+        // });
+
+        // vm.expectRevert();
+        // PythUtils.combinePrices(ethUsd, invalidExpo, -2);
+    }
 }
--- a/target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
+++ b/target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
@@ -403,22 +403,22 @@ contract PythUtilsTest is Test, WormholeTestUtils, PythTestUtils, IPythEvents {
     function testCombinePrices() public {
         // Test case 1: Basic price combination (ETH/USD / BTC/USD = ETH/BTC)
         PythStructs.Price memory ethUsd = PythStructs.Price({
-            price: 99875005,  
+            price: 206487956502,  
             conf: 10,
             expo: -8,    
             publishTime: block.timestamp
         });
         
         PythStructs.Price memory btcUsd = PythStructs.Price({
-            price: 206362333702,
+            price: 206741615681,
             conf: 100,
             expo: -8,     
             publishTime: block.timestamp
         });
 
         (int64 price, int32 expo) = PythUtils.combinePrices(ethUsd, btcUsd);
-        assertApproxEqRel(price, 2016367433623696, 9e17); 
-        assertEq(expo, -18);
+        assertApproxEqRel(price, 99875203, 9e17); 
+        assertEq(expo, -8);
 
         // // Test case 2: Different exponents
         // PythStructs.Price memory smallPrice = PythStructs.Price({
--- a/target_chains/ethereum/sdk/solidity/PythUtils.sol
+++ b/target_chains/ethereum/sdk/solidity/PythUtils.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "./PythStructs.sol";
-
+import "forge-std/console.sol";
 library PythUtils {
     /// @notice Converts a Pyth price to a uint256 with a target number of decimals
     /// @param price The Pyth price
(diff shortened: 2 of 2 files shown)


target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -33,4 +33,37 @@ library PythUtils {
...
         }
     }
+
+    /// @notice Combines two prices to get a cross-rate
+    /// @param price1 The first price (a/b)
+    /// @param price2 The second price (c/b)
+    /// @return combinedPrice The combined price (a/c)
+    /// @return expo The exponent of the combined price
+    /// @dev This function will revert if either price is negative or if the exponents are invalid
+    function combinePrices(
```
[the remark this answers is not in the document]
aditya520: how about `calculateCrossRate` or `deriveCrossRate` ?

target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
```diff
@@ -399,4 +410,44 @@ contract PythUtilsTest is Test, WormholeTestUtils, PythTestUtils, IPythEvents {
...
+
+        // Different Exponent Tests
+        successTest(10_000, -2, 100, -4, 100_000_000, -4);
+        successTest(10_000, -2, 10_000, 0, 1, -2);
+        successTest(10_000, 0, 10_000, 0, 1, 0);
+
+        // End Range Tests
+        successTest(int64(type(int64).max), 0, int64(type(int64).max), 0, 1, 0);
+        successTest(int64(type(int64).max), 0, 1, 0, int64(type(int64).max), 0);
+        successTest(1, 0, int64(type(int64).max), 0, 1 / int64(type(int64).max), 0);
```
jayantk, line 439: this case is the kind of thing you have to worry about with these conversion methods -- the `1 / int64(...)` here is actually `0`, which could be disastrous depending on the use case.

The critical thing to do here is to make sure that users understand *when* they can hit this case. I think what your method does is truncate (ie. round down) to the nearest representable quantity in the smaller exponent. Please document that explicitly. You may also want to let callers pass in the exponent for the result -- it's very easy to be safe if the caller explicitly specifies `-8` or something similar as the exponent.

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -33,4 +33,38 @@ library PythUtils {
...
+        }
+
+        // Convert both prices to the same decimal places (using the larger of the two)
+        uint8 maxDecimals = uint8(uint32(-1 * (expo1 < expo2 ? expo1 : expo2)));
+        uint256 p1 = convertToUint(price1, expo1, maxDecimals);
+        uint256 p2 = convertToUint(price2, expo2, maxDecimals);
+
+        // Calculate the combined price with precision
+        uint256 combined = (p1 * 10**18) / p2;
+        combined = combined / 10 ** (18 - maxDecimals);
```
jayantk, line 62: the other case to be very careful about is when this method can fail due to numeric overflow. This is another gotcha that could trigger if the price of asset goes up a lot.

(I think you're pretty unlikely to overflow given the real-world range of the exponents, but good to be explicit about that. You may even want to restrict the exponent range in the assert further to guarantee that there's no overlfow)

jayantk: sorry this is in reference to the multiplication on line 61

target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
aditya520: I planned to ask user to send an exponent for more precision, but what should be the limit for that? Should we cap it to 18?

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -33,4 +33,38 @@ library PythUtils {
                 10 ** uint32(priceDecimals - targetDecimals);
         }
     }
+
+    /// @notice Combines two prices to get a cross-rate
```
ali-bahjati, line 37: document all the assumptions here that are not captured within the types like prices being positive and the range that you can truly support without reverting. please also give a note that our price feeds might not look like that, we can have a feed with negative value (at some point)

ali-bahjati: It seems you have an assumption that expo's are <= 18

target_chains/ethereum/contracts/forge-test/utils/PythTestUtils.t.sol
```diff
@@ -370,6 +370,17 @@ abstract contract PythTestUtils is Test, WormholeTestUtils, RandTestUtils {
 }
 
 contract PythUtilsTest is Test, WormholeTestUtils, PythTestUtils, IPythEvents {
+    function successTest(int64 price1, int32 expo1, int64 price2, int32 expo2, int64 expectedPrice, int32 expectedExpo) internal {
```
ali-bahjati, line 373: rename it to assertCombinePriceEquals or something.

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -33,4 +33,38 @@ library PythUtils {
...
+        uint8 maxDecimals = uint8(uint32(-1 * (expo1 < expo2 ? expo1 : expo2)));
+        uint256 p1 = convertToUint(price1, expo1, maxDecimals);
+        uint256 p2 = convertToUint(price2, expo2, maxDecimals);
+
+        // Calculate the combined price with precision
+        uint256 combined = (p1 * 10**18) / p2;
+        combined = combined / 10 ** (18 - maxDecimals);
+        // Check if the combined price fits in int64
+        if (combined > uint256(uint64(type(int64).max))) {
+            revert();
```
ali-bahjati, line 65: add reasons to your reverts or use error structs.

target_chains/ethereum/sdk/solidity/Math.sol
ali-behjati, line 1: why didn't you import it?

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -12,25 +15,102 @@ library PythUtils {
...
+
+        // Bounds check: prevent overflow/underflow with base 10 exponentiation
+        // Calculation: 10 ** n <= (2 ** 256) - 1
+        //              n <= log10((2 ** 256) - 1)
+        //              n <= 77.2
+        if (combinedExpo > 77 || combinedExpo < -77) revert PythErrors.ExponentOverflow();
 
-        uint8 priceDecimals = uint8(uint32(-1 * expo));
+        // price is int64, always >= 0 here
+        uint256 unsignedPrice = uint256(uint64(price));
```
jayantk, line 44: the comment here is a bit confusing -- this is safe because you have a revert ^. I would say that explicitly

target_chains/ethereum/sdk/solidity/Math.sol
jayantk: yeah pls import as a dependency

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -12,25 +15,102 @@ library PythUtils {
...
+            revert PythErrors.InvalidInputExpo();
+        }
+
+        // We can safely cast the prices to uint64 because we know they are positive
+        uint256 fixedPointPrice = Math.mulDiv(uint64(price1), 10 ** PRECISION, uint64(price2));
+        int32 combinedExpo = expo1 - expo2 - int32(int8(PRECISION));
+
+        int32 factoredExpo = combinedExpo - targetExpo;
+
+        if (factoredExpo > 77 || factoredExpo < -77) revert PythErrors.ExponentOverflow();
```
jayantk: can't you put all of this exponent stuff into the Math.mulDiv call ^ and be done? 

like you're trying to compute `z` in: 

`z * 10^c = (x * 10^a) / (y * 10^b) `

which you can rewrite to:

`z = (x * 10^(a - (b + c))) / y`

so can't you do like:
```
uint256 result = Math.mulDiv(uint64(price1), 10 ** (expo1 - (expo2 + targetExpo)), uint64(price2));
```

and then you're done?

target_chains/ethereum/sdk/solidity/PythUtils.sol
jayantk: The thing i would document in particular is: under what range of exponents is this function *guaranteed not to overflow* for any set of prices. That's the safe range to use this function in.

target_chains/ethereum/sdk/solidity/PythUtils.sol
jayantk: (Math.mulDiv is a good find -- that's a very useful primitive for this kind of arithmetic)

target_chains/ethereum/sdk/solidity/PythUtils.sol
jayantk: I would also document that you get the floor of the result in the target exponent, so small numbers can round to 0.

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -12,25 +15,102 @@ library PythUtils {
...
     ) public pure returns (uint256) {
-        if (price < 0 || expo > 0 || expo < -255) {
-            revert();
+        if (price < 0) {
+            revert PythErrors.NegativeInputPrice();
         }
+        if (expo < -255) {
+            revert PythErrors.InvalidInputExpo();
+        }
+        int32 combinedExpo = int32(uint32(targetDecimals)) + expo;
```
jayantk: the name of this variable is odd -- this is an offset or multiplier for the result of some sort right? 

I would be careful with names like "expo" and "decimal" because one is the negative of the other. When I see this line, I go "wait you can't add a decimal and an expo and get an expo".

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -12,25 +15,102 @@ library PythUtils {
...
+
+        // Bounds check: prevent overflow/underflow with base 10 exponentiation
+        // Calculation: 10 ** n <= (2 ** 256) - 1
+        //              n <= log10((2 ** 256) - 1)
+        //              n <= 77.2
```
jayantk: this isn't actually the bound you want. You want to use 2**(256 - 64) to guarantee that there's enough space for the full `int64 price` to fit in the result

jayantk: (the case that you want to worry about is when there is a price-dependent overflow. The exponent and decimal will be set once at contract deployment time, but the price will change over time.)

jayantk (commented):
this is much better than last time. couple comments here. I suggest we sit down in person and work through them -- should be pretty quick

target_chains/ethereum/sdk/solidity/Math.sol
aditya520: It’s better to copy specific methods from another library when writing your own, especially if you’re only using a limited subset.
This helps reduce contract code size and lowers deployment costs.
In my opinion, focused Solidity libraries are always the better choice.

target_chains/ethereum/sdk/solidity/PythUtils.sol
```diff
@@ -12,25 +15,97 @@ library PythUtils {
...
+    /// @param expo1 The exponent of the first price
+    /// @param price2 The second price (c/b)
+    /// @param expo2 The exponent of the second price
+    /// @param targetExpo The target exponent for the cross-rate
+    /// @return crossRate The cross-rate (a/c)
+    /// @dev This function will revert if either price is negative or if the exponents are less than -255.
+    /// @notice This function doesn't return the combined confidence interval.
+    /// This function will revert with PythErrors.ExponentOverflow if the combined exponent (targetExpo + expo1 - expo2) is greater than 58 or less than -58.
+    /// Assuming the combined exponent is within bounds, this function will work for full range of int64 prices.
+    /// The result of the computation is rounded down. In particular, if the result is < 1 in the delta exponent, it will be rounded to 0
```
jayantk, line 76: in the targetExponent

jayantk (approved):
nicely done.

target_chains/ethereum/sdk/solidity/Math.sol
ali-behjati: Isn't forge compiler smart enough to not include unused code? If it is, it's much cleaner to have a dependency than copy paste.

PR #2666, opened 2025-05-08
feat(entropy): Add gas tracking for the callback recovery flow

jayantk:
## Summary

As the title says. Track gas usage and emit it in the event.

## Rationale

I previously omitted gas tracking in this recovery flow (which is identical to the old callback flow). However, I realized tracking gas will be useful here -- if a user's callback fails and they recover it, they probably want to know how much gas it used so they can set the gas limit better next time.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
+++ b/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
@@ -669,15 +669,18 @@ abstract contract Entropy is IEntropy, EntropyState {
             }
         } else {
             // This case uses the checks-effects-interactions pattern to avoid reentry attacks
+            address callAddress = req.requester;
+            EntropyStructs.Request memory reqV1 = EntropyStructConverter
+                .toV1Request(req);
             clearRequest(provider, sequenceNumber);
+            // WARNING: DO NOT USE req BELOW HERE AS ITS CONTENTS HAS BEEN CLEARED
 
             // Check if the requester is a contract account.
             uint len;
-            address callAddress = req.requester;
             assembly {
                 len := extcodesize(callAddress)
             }
-            uint256 startingGas = gasleft();            
+            uint256 startingGas = gasleft();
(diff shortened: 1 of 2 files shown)


target_chains/ethereum/contracts/contracts/entropy/Entropy.sol
```diff
@@ -669,38 +669,43 @@ abstract contract Entropy is IEntropy, EntropyState {
...
-                bytes(""),
-                0, // gas usage not tracked in the old no-gas-limit flow.
-                bytes("")
-            );
-
+            address callAddress = req.requester;
+            EntropyStructs.Request memory reqV1 = EntropyStructConverter
+                .toV1Request(req);
             clearRequest(provider, sequenceNumber);
+            // WARNING: DO NOT USE req BELOW HERE AS ITS CONTENTS HAS BEEN CLEARED
```
jayantk, line 676: referencing req.requester in the prior version of this code wasn't a good idea. (that field doesn't get cleared, but not good to make assumptions about that)

target_chains/ethereum/contracts/forge-test/Entropy.t.sol
```diff
@@ -1864,8 +1863,7 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents, EntropyEventsV2 {
             assertEq(callbackFailed, true);
             assertEq(callbackErrorCode, bytes(""));
 
-            // callback gas usage is approximate and only triggered when the provider has set a gas limit.
-            // Note: this condition is somewhat janky, but we hit the stack limit so can't put in any more local variables :(
+            // callback gas usage is approximate
```
jayantk, line 1866: Note that other tests of the `Revealed` event don't check the gas field because it's too hard to predict up front. Thus, we're using `vm.expectEmit` to only check the topics not the data.

m30m (approved):

PR #2672, opened 2025-05-09
Add Developer Hub to monorepo

aaronbassett:
## Summary

Minimal Developer Hub boilerplate with a basic content structure so we can begin Vercel deployments.


apps/developer-hub/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "test:format": "prettier --check .",
+    "test:lint:eslint": "eslint . --max-warnings 0",
+    "test:lint:stylelint": "stylelint 'src/**/*.scss' --max-warnings 0",
+    "test:types": "tsc"
+  },
+  "dependencies": {
+    "@phosphor-icons/react": "catalog:",
+    "@pythnetwork/component-library": "workspace:*",
+    "@react-hookz/web": "catalog:",
+    "@rive-app/react-canvas": "^4.19.1",
```
cprussin, line 26: We should add this to `pnpm-workspace.yamlr` and use the [pnpm catalog protocol](https://pnpm.io/catalogs) instead; we should really only use specific versions if we have some reason why some app needs to move it's dependency version separately from the rest of the repo -- in this case, if we start using rive in more places, we'll want to try to use the same version across the repo as much as possible

apps/developer-hub/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "nuqs": "catalog:",
+    "react": "catalog:",
+    "react-aria": "catalog:",
+    "react-dom": "catalog:",
+    "recharts": "catalog:",
+    "shiki": "catalog:",
+    "superjson": "catalog:",
+    "swr": "catalog:",
+    "zod": "catalog:",
+    "zod-validation-error": "catalog:"
```
cprussin, line 45: You should go through these dependencies and strip out stuff you don't need; I'm pretty sure you won't need the vast majority of stuff listed here...

apps/developer-hub/src/app/opengraph-image.png
cprussin, line 1: This isn't the right opengraph image for dev hub, let's just delete this for now until we have a more fitting og image

apps/developer-hub/src/app/page.tsx
```diff
@@ -0,0 +1,15 @@
...
+const Page = () => {
+  return (
+    <Homepage>
+      <HomepageHero />
+      
+      {/* Other sections will be added here */}
+    </Homepage>
+  );
+};
+
```
cprussin, line 14: Definitely not a requirement and feel free to disregard, but I've found that I prefer keeping everything in the nextjs `app` directory as just re-exports of components from the `components/` directory, along with metadata and other nextjs page config.  That gives me a very clean separation of concerns -- the `app` directory is concerned with configuring the urls, nextjs caching, and metadata, but all UI lives in `components`.

Up to you if you choose to follow that as well but I've found I like it more than having some UI in `app` and others in other places in the repo, it's a bit more predictable and simpler.

apps/developer-hub/src/components/Error/index.module.scss
```diff
@@ -0,0 +1,63 @@
+@use "@pythnetwork/component-library/theme";
```
cprussin, line 1: There's an [ErrorPage component in the component library](https://github.com/pyth-network/pyth-crosschain/tree/main/packages/component-library/src/ErrorPage) now which I recently extracted out of insights hub; I'd suggest using that instead of having a bespoke one for this app (e.g. see how I'm doing it in [the Entropy explorer app](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/src/app/error.ts)

apps/developer-hub/src/components/Homepage/index.module.scss
```diff
@@ -0,0 +1,23 @@
+@use "@pythnetwork/component-library/theme";
+
+.homepage {
+  @include theme.max-width;
+
+  align-items: center;
+
+  > section {
```
cprussin, line 8: I tried doing this same approach with css for the Fogo website where I avoided introducing class names when possible and just styled based on the dom structure instead, but ended up reverting it entirely as I found that with css modules, using rules like this are a pain at scale and it's much easier to maintain when you use explicit class names.  Being explicit and just styling class names means your jsx is clearly and intentionally applying styles where appropriate, rather than having styles implicitly apply and then needing to deal with overrides in cases where the implicit styles don't make sense -- or worse, dealing with things breaking if you change the structure of the page.  For instance, what if you added a section where you _don't_ want the width to be 100%?  Or what if one section needs a different style applied to it?  What if you reorder sections?  All of this is trivial when using explicit classnames, but it's a huge pain when your css makes assumption about the dom structure.

apps/developer-hub/src/components/HomepageHero/HeroContent/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    grid-column: 1 / 2;
+  }
+
+  @include theme.breakpoint(lg) {
+    grid-column: 1 / 3;
+  }
+}
+
+.title {
+  font-size: theme.spacing(6);
```
cprussin, line 18: It would be better to stick to font sizes from the [font-size scale](https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/src/theme.scss#L33-L48) -- I have a [text mixin](https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/src/theme.scss#L874-L880) that I usually apply e.g. `@include theme.text("sm", "medium")`.

TBH I'm not in love with the way the font configuration stuff is done, but at the very least I think we should try to stick to font sizes out of the font-size scale one way or another.

apps/developer-hub/src/components/HomepageHero/HeroContent/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  @include theme.breakpoint(sm) {
+    flex-direction: row;
+    align-items: center;
+  }
+}
+
+.secondaryLink {
+  color: theme.color("foreground");
+  text-decoration: none;
```
cprussin, line 53: In general, [wcag requirements state that links must be discernable](https://dequeuniversity.com/rules/axe/4.10/link-name) in a very obvious way using features other than color.  I see you adjusted the font weight which may be sufficient, but underlining text is the most effective and accessible mechanism for satisfying this requirement, so I generally recommend against suppressing the underline.

If you do choose to move forward with this, you can pass [the invert prop](https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/src/Link/index.tsx#L9) to the Link component so you don't have to override the style manually.

apps/developer-hub/src/components/HomepageHero/HeroGraphic/index.module.scss
```diff
@@ -0,0 +1,30 @@
+@use "@pythnetwork/component-library/theme";
+
+.heroGraphic {
+  position: relative;
+  width: 100%;
+  height: 400px;
```
cprussin, line 6: I recommend avoiding using px values; we should ideally always stick to using sizes via theme.spacing but at the very least, unless we're styling something that must be sized in px, we should be using rem

apps/developer-hub/src/components/HomepageHero/HeroGraphic/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import styles from "./index.module.scss";
+
+export const HeroGraphic = () => {
+  return (
+    <div className={styles.heroGraphic}>
+      <div className={styles.graphicContainer}>
+        <Rive
+          src="/price_feeds_stack.riv"
+          style={{ width: "500px", height: "700px" }}
```
cprussin, line 13: Can this be done in css instead?

apps/developer-hub/src/components/NotFound/index.module.scss
```diff
@@ -0,0 +1,57 @@
+@use "@pythnetwork/component-library/theme";
```
cprussin, line 1: There's also a [NotFoundPage component in the component library](https://github.com/pyth-network/pyth-crosschain/tree/main/packages/component-library/src/NotFoundPage) now which I suggest you move to -- sorry I didn't have these extracted before you started working on this!

apps/developer-hub/src/config/server.ts
```diff
@@ -0,0 +1,29 @@
+// Disable the following rule because this file is the intended place to declare
```
cprussin, line 1: I'll wrap up some utilities to make these configs a bit nicer to write....

pnpm-workspace.yaml
```diff
@@ -84,6 +84,8 @@ catalog:
   "@types/node": ^22.14.0
   "@types/react": ^19.1.0
   "@types/react-dom": ^19.1.1
+  "@types/react-window": ^1.8.8
+  "@types/swagger-ui-react": ^5.18.0
```
cprussin, line 88: It doesn't look like you ended up using any of the dependencies you added here unless I'm missing something -- should we revert these changes?

cprussin (changes requested):
Overall looking great!  Some minor changes but nothing big and then I think this will be good to merge

apps/developer-hub/next.config.js
```diff
@@ -0,0 +1,77 @@
+import createBundleAnalyzer from "@next/bundle-analyzer";
```
cprussin, line 1: Nice!  Note that I'm working on moving other apps to using turbopack; I'm not sure how well this will work with turbopack, but in the meantime I think it would be great to roll it out across all our apps

apps/developer-hub/next.config.js
```diff
@@ -0,0 +1,77 @@
...
+        loaders: ["@svgr/webpack"],
+        as: "*.js",
+      },
+    },
+  },
+
+  pageExtensions: ["ts", "tsx", "mdx"],
+
+  experimental: {
+    useCache: true,
```
cprussin, line 23: I don't think you need this experiment if you aren't using the `"use cache"` directive

apps/developer-hub/next.config.js
```diff
@@ -0,0 +1,77 @@
...
+  reactStrictMode: true,
+  turbopack: {
+    resolveExtensions: [".mdx", ".tsx", ".ts", ".jsx", ".js", ".mjs", ".json"],
+    rules: {
+      "*.svg": {
+        loaders: ["@svgr/webpack"],
+        as: "*.js",
+      },
+    },
+  },
```
cprussin, line 18: It doesn't look like you're actually using turbopack (I don't see `--turbo` in your `next dev` cli args) so I think you should either:

1. Remove this (it will become stale and incorrect if no one is using it)
2. Switch to using turbopack in dev

IMO 2 is preferable over 1

apps/developer-hub/package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "scripts": {
+    "build:vercel": "next build",
+    "build:analyze": "ANALYZE=true next build",
+    "fix:format": "prettier --write .",
+    "fix:lint:eslint": "eslint --fix .",
+    "fix:lint:stylelint": "stylelint --fix 'src/**/*.scss'",
+    "pull:env": "[ $CI ] || VERCEL_ORG_ID=team_BKQrg3JJFLxZyTqpuYtIY0rj VERCEL_PROJECT_ID=prj_TBkf9EyQjQF37gs4Vk0sQKJj97kE vercel env pull",
+    "start:dev": "next dev --port 3003",
+    "start:prod": "next start --port 3003",
```
cprussin, line 17: Can you change these ports to ports that are not in use by any other app?  You can easily list all the ports in use by doing something like `rg '"start:dev":'` or whatever is your preferred code search tool.

The reason I suggest this is that I will often run multiple apps simultaneously if I'm working on something like the component library which is shared between apps, and if you have multiple apps that try to bind to the same port, it will fail to start.

apps/developer-hub/src/app/(docs)/[section]/[...slug]/page.tsx
```diff
@@ -0,0 +1,25 @@
+import { DocumentationPage } from "@/src/components/Pages/DocumentationPage";
```
Comment (cprussin) [inline apps/developer-hub/src/app/(docs)/[section]/[...slug]/page.tsx:1]: I strongly recommend against import aliases (i.e. use relative paths instead of `@`).  Import aliases feel fun to write, but because they are not actually part of the Node module resolution algorithm AND are not trivially parseable relative paths, they have a long history of breaking dev tools and making it impossible to use tools that haven't explicitly rolled out support for resolving import aliases.

With relative paths, support is trivial, anything that resolves standard relative pathnames or follows node's resolver algorithm will work natively.

It's a bit clunky with the app dir, but the other benefit is that, outside the app dir, it makes it really obvious where you have deep coupling and should reorganize the codebase.  If you're walking a ton of paths often, it's a strong indication that the code isn't organized well.

apps/developer-hub/src/app/(homepage)/page.tsx
```diff
@@ -0,0 +1,3 @@
+import { Homepage } from "@/src/components/Pages/Homepage";
```
cprussin, line 1: In cases like this I recommend a simple re-export as webpack will optimize away the module:

```ts
export { Homepage as default } from "..."
```

apps/developer-hub/src/components/Pages/BasePage/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  DocsBody,
+  DocsDescription,
+  DocsPage,
+  DocsTitle,
+} from "fumadocs-ui/page";
+import { notFound } from "next/navigation";
+
+export function BasePage(props: { params: { slug: string[] } }) {
+  const page = source.getPage(props.params.slug);
+  if (!page) return notFound();
```
cprussin, line 13: Very minor nit but I generally prefer to avoid early returns, the reason is very similar to why people avoid gotos -- as code grows it becomes hard to know that control flow has exit paths if those exit paths aren't clearly captured by the control flow structure.  I've seen many cases where functions start simple, eventually grow to be very large, and then debugging them is a nightmare because you can't easily look in a line of code and know whether or not execution will actually reach that line.

So in particular in this case, I recommend something like this:

```ts
if (page) {
  const MDX = page.data.body;
  return (<DocsPage>{...}</DocsPage>;
} else {
  notFound();
}
```

That way it's very trivial to follow simple code structure to understand how execution will ever reach a particular place

apps/developer-hub/src/components/Pages/DocumentationPage/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { BasePage } from "../BasePage";
+
+export async function DocumentationPage(props: {
+  params: Promise<{ section: string; slug: string[] }>;
+}) {
+  const params = await props.params;
+  params.slug.unshift(params.section);
+  return BasePage({ params });
+}
```
cprussin, line 9: 2 things:

1. We shouldn't be mutating params ever
2. We need to use jsx to invoke `BasePage` -- there are a ton of things both in nextjs and in react that will only work properly if using jsx (or `createElement` etc) to invoke components

So this should be written as:

```tsx
const params = await props.params;
return <BasePage params={{ ...params, slug: [ params.section, ...params.slug ] }} />
```

The same comment applies to the other pages too

apps/developer-hub/src/components/Pages/LandingPage/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { BasePage } from "../BasePage";
+
+export async function LandingPage(props: {
+  params: Promise<{ section: string }>;
+}) {
+  const params = await props.params;
+  return BasePage({ params: { slug: [params.section] } });
```
cprussin, line 7: Same comment as above about properly using jsx to invoke `BasePage`

apps/developer-hub/src/components/Root/theme.css
```diff
@@ -0,0 +1,336 @@
...
+    var(--color-steel-900),
+    var(--color-steel-50)
+  );
+  --color-fd-popover: light-dark(var(--color-white), var(--color-steel-950));
+  --color-fd-primary-foreground: light-dark(
+    var(--color-steel-900),
+    var(--color-steel-50)
+  );
+  --color-fd-ring: light-dark(var(--color-violet-600), var(--color-violet-400));
+}
```
cprussin, line 336: Very nice!  I suggest trying to use `theme.scss` to drive this file if possible -- in particular I think the whole color gradient can just use `theme.pallette-color`, and I think all the color tokens can be defined in terms of `theme.color` values right?

e.g. I would imagine this file could look something like this:

```scss
@use "@pythnetwork/component-library/theme";

:root {
  --color-slate-50: theme.pallette-color("slate", 50);
  ...

  --color-fd-background: theme.color("background", "primary");
}
```

One thing to note is that once the new website gets rolled out, we will do a pass at adjusting the design system color tokens a bit to match with the new branding.  If you implement in terms of all these tokes, it will mean less that we have to worry about manually syncing.

apps/developer-hub/vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "ignoreCommand": "../../vercel-ignore.sh",
```
cprussin, line 3: We recently got rid of the ignore command and are using a new vercel feature for this instead, see #2701 for details

cprussin (changes requested):
Looking very good!  A few minor points and notes and then I think we'll be good to ship it

apps/developer-hub/next.config.js
aaronbassett: This has just been copied across from Insights. I didn't really change anything I wasn't 100% sure about. It looks like there's a few other comments on the next config that are probably for similar reasons. I'll compile all the requested config changes into a single commit rather than address them individually.

cprussin: no problem -- in this case we are in fact using the experimental caching in insights hub, so this is needed there, but I don't recommend using it unless you have a really good reason to and I don't think you will for dev hub

apps/developer-hub/package.json
aaronbassett: I counted all the port numbers I could find in the monorepo out of interest. I'll switch it to a free one.

`rg -Po '(?:--port|(?:(?:25[0-5]|(?:2[0-4]|1\d|[1-9]|)\d)\.?\b){4}:|localhost)\W*(\d{4,5})' . | grep -o '\d\{4,5\}' | sort | uniq -c | sort -rn`

<details><summary>Count of ports mentioned in monorepo</summary>
<pre>
  24 9000
   9 9002
   9 39187
   9 39101
   9 36853
   9 36219
   6 46335
   6 45851
   6 45007
   6 44423
   6 44135
   6 44061
   6 43451
   6 41551
   6 40869
   6 39889
   6 39573
   6 38815
   6 38133
   6 38025
   6 37813
   6 37183
   6 37179
   6 36503
   6 35625
   6 34043
   6 33765
   6 33519
   6 32965
   6 32833
   6 3003
   5 8547
   5 26657
   3 38381
   3 38187
   3 35915
   3 35253
   3 26656
   2 8899
   2 4000
   2 34000
   2 3006
   2 3005
   2 3004
   2 3002
   2 3001
   2 1317
   1 9091
   1 9090
   1 8080
   1 7777
   1 7575
   1 6060
   1 5050
   1 33999
   1 33888
   1 26658
</pre>
</details>

cprussin: To be clear all that matters here is that you don't use another port in use by another nextjs app, as these are what we'll run in parallel.  The vast majority of those other ports are ports consumed by clients (i.e. we don't bind to them), or are ports used for services which are unlikely to ever be run in parallel with a nextjs app, so you don't need to worry about considering them.  Just look for anything that has a `start:dev` script as I'd mentioned above since that script is the hook for running a dev server.

apps/developer-hub/src/components/Pages/BasePage/index.tsx
aaronbassett: Totally fair point — I agree that in large, tangled functions, multiple exit points can make reasoning harder. But in small and focused functions, early returns increase readability by avoiding unnecessary nesting, and they don't increase the cyclomatic complexity, a return and a conditional are both decision points. But yeah, once a function grows beyond a certain size, it’s usually a signal to refactor anyway — not necessarily to avoid early returns, but to keep the logic manageable overall.

I try to use them for clear guard clauses, much the same as how we use `require` in Solidity, or `?` error propagation in Rust.

apps/developer-hub/src/components/Root/theme.css
aaronbassett: I didn't want to spend any time on this part as I'm hoping @partyparrotgreg will replace it with the correct color mapping.

apps/developer-hub/src/app/(docs)/[section]/[...slug]/page.tsx
Comment (aaronbassett) [reply apps/developer-hub/src/app/(docs)/[section]/[...slug]/page.tsx:1]: What about using `baseUrl` in `tsconfig.json`? Is that likely to cause any issues?

I find it less brittle to go down from root directory, then traverse up through sub-directories. I find it much easier to read paths like `/src/config` than `../../../config`

apps/developer-hub/package.json
aaronbassett: Yeah a lot of them are used when running local validators for different networks, which probably don't have a major chance of collision. But others are used in config files for the likes of Hermes, which I could definitely see us running locally while spinning up one of the nextjs apps. It's not critical or anything, but if I'm going to pick a specific port for the app I might as well try to make it unique.

.gitignore
```diff
@@ -25,3 +25,4 @@ __pycache__
 .turbo/
 .cursorrules
 .corepack
+justfile
```
cprussin, line 28: Hmm what is this?  Is it intentional?

cprussin (approved):

.gitignore
aaronbassett: Yeah, justfile is like a Makefile (but with a simplified syntax) it's a local command runner I use in terminal which probably isn't helpful to commit

cprussin: totally reasonable, feel free to commit files like this if you think they may be useful for others, for instance I've committed my nix files (flake.nix etc) -- maybe in a separate commit though

PR #2682, opened 2025-05-12
feat(argus): internal interfaces and shared memory model

tejasbadadare:
## Summary

Skeleton for Argus. Defines code organization, internal interfaces, and sets up a services + shared memory model.
Previously I had explored implementing Argus with an actor model (#2659 ), but after useful feedback, it feels more straightforward to go with a more standard shared memory approach. 

We still keep the isolation and separation of concerns from the actor model, but without the message-box concept for IPC.
I also considered implementing the trait system from Hermes to gate access to certain State objects behind traits, but found it didn't add much value here.

### Services
Services are long-running processes (tokio tasks) responsible for executing business logic and keeping relevant States up-to-date (e.g. via polling or subscribing to data sources via Adapters.)
- For example, the ControllerService reads from SubscriptionState, PythPriceState, and ChainPriceState to determine whether or not to trigger an update. The SubscriptionService writes to SubscriptionState as it tracks changes to on-chain subscription metadata.

### State
The State objects hold the state for Argus for a specific blockchain and expose them to Services via read/write APIs, wrapping the underlying data structure with synchronization primitives like RwLock and DashMap.

### Adapters
Interactions with external services like Hermes and the blockchain are modeled by the traits in adapters/types.rs. These traits are used by the actors, and are implemented by hermes.rs, contract.rs, etc.
- For example, the SubscriptionService expects an implementor of ReadChainSubscriptions. This trait specifies get_active_subscriptions() and subscribe_to_subscription_events(), which are impl'd by PythPulse<M> in contract.rs.

## How has this been tested?

- [ ] Added new tests

Not tested yet. Next up is wiring up the Controller's main update loop to mock implementations, and ensuring the actor communication works. Then, we will implement individual actor logic, and eventually wire it up to the blockchain instead of mock adapters.


apps/argus/src/services/controller_service.rs
```diff
@@ -0,0 +1,126 @@
...
+    chain_price_state: Arc<ChainPriceState>,
+}
+
+impl ControllerService {
+    pub fn new(
+        chain_name: ChainName,
+        update_interval: Duration,
+        subscription_state: Arc<SubscriptionState>,
+        pyth_price_state: Arc<PythPriceState>,
+        chain_price_state: Arc<ChainPriceState>,
```
ali-bahjati, line 36: ah so instead of relying on the magic of type inferences you decided to literally pass the states around. good idea!

apps/argus/src/adapters/types.rs
ali-bahjati, line 1: i recommend moving some of these in to the files to reduce indirection if they are only used in one place.

ali-bahjati (approved):
Nice!

apps/argus/src/command/run.rs
```diff
@@ -1,252 +1,202 @@
...
+    private_key: String,
+    chain_eth_config: EthereumConfig,
+    chain_name: String,
+    _metrics: Arc<KeeperMetrics>, // TODO: add metrics
+    rpc_metrics: Arc<RpcMetrics>,
+    exit_rx: watch::Receiver<bool>,
+) -> Result<()> {
+    tracing::info!("Starting keeper for chain {}", chain_name);
+
+    // TODO: create a contract with a WS provider if geth_rpc_wss was provided
```
jayantk, line 90: you probably don't have to worry about the wss feature -- we don't even use this in fortuna. you can probably delete this feature entirely

apps/argus/src/command/run.rs
```diff
@@ -1,252 +1,202 @@
...
-            }
-        },
-        Err(e) => {
-            tracing::error!("Failed to get block - {:?}", e);
-            INF_LAG
-        }
+    let state = Arc::new(ArgusState::new());
+
+    let hermes_client = Arc::new(HermesClient);
+    let subscription_poll_interval = Duration::from_secs(60);
```
jayantk, line 114: in the future: please make these configurable from the config file

apps/argus/src/command/run.rs
```diff
@@ -1,252 +1,202 @@
...
+    let subscription_service = SubscriptionService::new(
+        chain_name.clone(),
+        contract.clone(),
+        subscription_poll_interval,
+        state.subscription_state.clone(),
+        state.pyth_price_state.clone(),
+        state.chain_price_state.clone(),
+    );
+
+    let pyth_price_service = PythPriceService::new(
```
jayantk, line 135: do you really want to instantiate a separate one of these per chain? I think you could probably run a single thread that listens to the hermes websocket and streams the prices to check across all chains.

jayantk (approved):
yeah this is better than the actors for sure

apps/argus/src/adapters/types.rs
```diff
@@ -0,0 +1,44 @@
...
+    async fn subscribe_to_subscription_events(&self) -> Result<()>;
+}
+
+#[async_trait]
+pub trait ReadPythPrices {
+    async fn get_latest_prices(&self, feed_ids: &[PriceId]) -> Result<Vec<Vec<u8>>>;
+    async fn subscribe_to_price_updates(&self, feed_ids: &[PriceId]) -> Result<()>; // TODO: return a stream
+}
+
+// TODO: find a different home for these (public SDK)
```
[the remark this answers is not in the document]
tejasbadadare: yep! will use the types in there

apps/argus/src/services/price_pusher_service.rs
```diff
@@ -0,0 +1,136 @@
...
+use crate::state::ChainName;
+
+pub struct PricePusherService {
+    chain_name: ChainName,
+    name: String,
+    contract: Arc<dyn UpdateChainPrices + Send + Sync>,
+    pyth_price_client: Arc<dyn ReadPythPrices + Send + Sync>,
+    backoff_policy: ExponentialBackoff,
+    request_rx: Mutex<Option<mpsc::Receiver<PushRequest>>>,
+    request_tx: mpsc::Sender<PushRequest>,
```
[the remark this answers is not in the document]
tejasbadadare: It will be shared with other services' new() functions. Only the Controller uses it right now, so we could have just returned a reference to this service directly, but the channel is helpful to decouple the two imo.

apps/argus/src/command/run.rs
tejasbadadare: Ah okay cool. Figured it would be nice to primarily stream in price updates and only poll as a backup, but I can imagine wss support/reliability is spotty with RPCs.

apps/argus/src/command/run.rs
tejasbadadare: Yep definitely

apps/argus/src/command/run.rs
tejasbadadare: Yeahh i had the same thought, this was the path of least resistance since we can just replace the tracked feed ID set per chain. Later I'll extract this out to be a single thread with redundant connections and the like.

PR #2691, opened 2025-05-15
feat(fortuna): Support new Entropy v2 contract

jayantk:
## Summary

Update fortuna to set the default_gas_limit field on-chain, and then work with the new Entropy v2 interface. The main changes there are (1) replace the types of ProviderInfo and Request with their V2 variants, and (2) remove the gas_limit parameters from callback submission.

## Rationale

we want to support the new version of the contract

## How has this been tested?

- [ ] Added new tests

I've tested this locally against the Entropy contract on blast-testnet (using the EntropyTester to simulate all of the various callback success / failure / out of gas conditions) and it works. I still want to test against a zk chain like Abstract, but that's pending the upgrade right now.


apps/fortuna/src/config.rs
```diff
@@ -133,7 +133,7 @@ pub struct EthereumConfig {
     pub legacy_tx: bool,
 
     /// The gas limit to use for entropy callback transactions.
-    pub gas_limit: u64,
+    pub gas_limit: u32,
```
jayantk, line 136: u32 is the size of the field in the contract

apps/fortuna/src/chain/ethereum.rs
```diff
@@ -233,29 +233,21 @@ impl InstrumentedPythContract {
...
-        let r = self
-            .get_request(provider_address, sequence_number)
-            // TODO: This doesn't work for lighlink right now. Figure out how to do this in lightlink
-            // .block(ethers::core::types::BlockNumber::Finalized)
+        let request = self
+            .get_request_v2(provider_address, sequence_number)
             .call()
             .await?;
-
-        // sequence_number == 0 means the request does not exist.
```
m30m, line 248: why is this removed? I know that this check exists in some other places, but if we want to be safe and avoid unintended changes, we can keep it as is?

m30m (commented):
I think in this new flow we should always make sure that the gas we set for the transaction is more than what the user requested. Now that the users set the gas limit, this is much easier and we can solve the cases where the gas used can vary a lot (from simulation time to landing time) based on other factors.

apps/fortuna/src/eth_utils/utils.rs
```diff
@@ -217,34 +178,9 @@ pub async fn submit_tx_with_backoff<T: Middleware + NonceManaged + 'static>(
 pub async fn submit_tx<T: Middleware + NonceManaged + 'static>(
     client: Arc<T>,
     call: &ContractCall<T, ()>,
-    gas_limit: U256,
-    // A value of 100 submits the tx with the same gas/fee as the estimate.
-    gas_estimate_multiplier_pct: u64,
+    // A value of 100 submits the tx with the same fee as the estimate.
     fee_estimate_multiplier_pct: u64,
 ) -> Result<TransactionReceipt, backoff::Error<anyhow::Error>> {
-    let gas_estimate_res = call.estimate_gas().await;
```
jayantk, line 225: keep this gas estimation logic

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -105,7 +104,7 @@ pub async fn process_event_with_backoff(
         sender: event.requestor,
         user_random_number: event.user_random_number,
         state: RequestEntryState::Pending,
-        gas_limit,
+        gas_limit: U256::from(0), // FIXME(Tejas): set this properly
```
tejasbadadare, line 116: Earmarked this to fix when we do the work to expose the needed types to the API for the frontend to consume

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -57,9 +56,11 @@ pub async fn process_event_with_backoff(
 
             // Check if the request is still open after the delay.
             // If it is, we will process it as a backup replica.
+
+            // FIXME(Tejas): check callback status
             match chain_state
                 .contract
-                .get_request(event.provider_address, event.sequence_number)
+                .get_request_v2(event.provider_address, event.sequence_number)
                 .await
```
tejasbadadare, line 64: Just checking that the request isn't on chain is no longer suffcient, need to check the callbackStatus. WIll handle this in another PR.

apps/fortuna/src/history.rs
```diff
@@ -259,20 +259,19 @@ impl History {
...
-                    new_status.created_at,
-                    new_status.last_updated_at,
-                    "Pending",
-                    block_number,
-                    request_tx_hash,
-                    user_random_number,
-                    sender,
-                    gas_limit
-            )
+                sqlx::query("INSERT INTO request(chain_id, network_id, provider, sequence, created_at, last_updated_at, state, request_block_number, request_tx_hash, user_random_number, sender, gas_limit) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
```
tejasbadadare, line 262: Use `query()` instead of the `query!` macro since the compile-time checking of the queries doesn't reliably work with the local sqlite db, preventing building. Besides, since most of the queries are dynamically constructed, the compile time checking doesn't give us a guarantee they're correct anyway.

apps/fortuna/src/keeper/process_event.rs
jayantk: this value is in the requestV2 struct you retrieved ^ to check the status

tejasbadadare: Oh perfect ty

PR #2693, opened 2025-05-15
feat(argus): ControllerService update loop, ChainPriceService poll loop

tejasbadadare:
## Summary

- Create main Controller update loop. This orchestrates the other services and makes the decision on whether or not to push prices for a subscription. 
- Create ChainPriceService polling loop. This runs on an interval to track the on-chain prices.
  - Update ChainPriceState data structures to be subscription oriented, since we need to maintain the on-chain price for each feed for each subscription.

## How has this been tested?

- [x] Added new tests
  - Added integration tests for the update loop that validate the orchestration of other services (e.g. calling PricePusherService when feeds need to be updated)
  - Added unit tests for `needs_update`, which tests scenarios around update criteria. Given a pyth price, chain price, and update criteria, the function returns a boolean indicating whether the subscription should be updated or not.
  - Considered adding integration tests for ChainPriceService, but since it's a light wrapper around an RPC API, figured there's not much value in writing tests for this.

### DIFF
--- a/apps/argus/src/adapters/contract.rs
+++ b/apps/argus/src/adapters/contract.rs
@@ -10,21 +10,39 @@ use std::collections::HashMap;
 
 #[async_trait]
 pub trait GetChainPrices {
-    async fn get_price_unsafe(
+    async fn get_all_prices_for_subscription(
         &self,
         subscription_id: SubscriptionId,
-        feed_id: &PriceId,
-    ) -> Result<Option<Price>>;
+    ) -> Result<Vec<Price>>;
+
+    async fn get_prices_for_subscription(
+        &self,
+        subscription_id: SubscriptionId,
+        price_ids: &Vec<PriceId>,
+    ) -> Result<Vec<Price>>;
 }
 
 #[async_trait]
 impl<M: Middleware + 'static> GetChainPrices for PythPulse<M> {
-    async fn get_price_unsafe(
+    async fn get_all_prices_for_subscription(
         &self,
-        _subscription_id: SubscriptionId,
-        _feed_id: &PriceId,
-    ) -> Result<Option<Price>> {
-        todo!()
+        subscription_id: SubscriptionId,
+    ) -> Result<Vec<Price>> {
+        let price_ids = self.get_prices_unsafe(subscription_id, vec![]).await?;
+        Ok(price_ids.into_iter().map(From::from).collect())
+    }
+    async fn get_prices_for_subscription(
+        &self,
+        subscription_id: SubscriptionId,
+        price_ids: &Vec<PriceId>,
+    ) -> Result<Vec<Price>> {
+        let price_ids = self
+            .get_prices_unsafe(
+                subscription_id,
+                price_ids.into_iter().map(|id| id.to_bytes()).collect(),
+            )
+            .await?;
+        Ok(price_ids.into_iter().map(From::from).collect())
     }
 }
 #[async_trait]
--- a/apps/argus/src/services/chain_price_service.rs
+++ b/apps/argus/src/services/chain_price_service.rs
@@ -7,6 +7,8 @@
 
 use anyhow::Result;
 use async_trait::async_trait;
+use pyth_sdk::Price;
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::watch;
@@ -14,9 +16,11 @@ use tokio::time;
 use tracing;
 
 use crate::adapters::contract::GetChainPrices;
+use crate::adapters::types::PriceId;
 use crate::services::Service;
 use crate::state::ChainName;
 use crate::state::ChainPriceState;
+use crate::state::SubscriptionState;
 
 pub struct ChainPriceService {
     chain_name: ChainName,
@@ -24,6 +28,7 @@ pub struct ChainPriceService {
     contract: Arc<dyn GetChainPrices + Send + Sync>,
     poll_interval: Duration,
     chain_price_state: Arc<ChainPriceState>,
+    subscription_state: Arc<SubscriptionState>,
 }
 
 impl ChainPriceService {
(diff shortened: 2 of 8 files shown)


apps/argus/src/services/chain_price_service.rs
```diff
@@ -32,24 +37,70 @@ impl ChainPriceService {
...
+                        subscription_id = %subscription_id,
+                        "Got prices for subscription"
+                    );
+
+                    // Store the latest price feeds for the subscription
+                    self.chain_price_state
+                        .update_prices(subscription_id, prices_map);
+                }
+                Err(e) => {
+                    // If we failed to get prices for a subscription, we'll retry on the next poll interval.
```
ali-bahjati, line 92: this implicitly enforces having a high poll interval

apps/argus/src/services/chain_price_service.rs
```diff
@@ -32,24 +37,70 @@ impl ChainPriceService {
...
-            service = self.name,
-            feed_count = feed_ids.len(),
-            "Polled for on-chain price updates"
-        );
+        // For each subscription, query the chain for the price of each feed
+        for item in subscriptions.iter() {
+            let subscription_id = item.key().clone();
+            let subscription_params = item.value().clone();
+
+            // TODO: do this in parallel using tokio tasks?
```
ali-bahjati, line 62: yeah let's do it.

apps/argus/src/adapters/contract.rs
```diff
@@ -10,21 +10,39 @@ use std::collections::HashMap;
...
-    async fn get_price_unsafe(
+    async fn get_all_prices_for_subscription(
         &self,
-        _subscription_id: SubscriptionId,
-        _feed_id: &PriceId,
-    ) -> Result<Option<Price>> {
-        todo!()
+        subscription_id: SubscriptionId,
+    ) -> Result<Vec<Price>> {
+        let price_ids = self.get_prices_unsafe(subscription_id, vec![]).await?;
```
ali-bahjati, line 31: `vec![]` seems wrong.

apps/argus/src/services/chain_price_service.rs
```diff
@@ -64,7 +115,13 @@ impl Service for ChainPriceService {
         loop {
             tokio::select! {
                 _ = interval.tick() => {
-                    self.poll_prices(self.chain_price_state.clone()).await;
+                    if let Err(e) = self.poll_prices().await {
```
ali-bahjati, line 118: poll_prices doesn't seem to ever fail (maybe change its signature?)

apps/argus/src/services/controller_service.rs
```diff
@@ -108,7 +138,13 @@ impl Service for ControllerService {
         loop {
             tokio::select! {
                 _ = interval.tick() => {
-                    self.perform_update().await;
+                    if let Err(e) = self.perform_update().await {
+                        tracing::error!(
+                            service = self.name,
+                            error = %e,
+                            "Failed to perform price update"
+                        );
```
ali-bahjati, line 146: no retry here?

apps/argus/src/services/controller_service.rs
```diff
@@ -125,3 +161,595 @@ impl Service for ControllerService {
...
+#[tracing::instrument()]
+fn needs_update(
+    pyth_price: &Price,
+    chain_price_opt: Option<&Price>,
+    update_criteria: &UpdateCriteria,
+) -> bool {
+    // If there's no price currently on the chain for this feed, an update is always needed.
+    let chain_price = match chain_price_opt {
+        None => {
+            tracing::debug!("Update criteria met: No chain price available.");
```
ali-bahjati, line 176: in price pusher sometimes i found these logs helpful, might not be bad to have them at info level, or maybe expose as metrics or trace

apps/argus/src/services/controller_service.rs
```diff
@@ -125,3 +161,595 @@ impl Service for ControllerService {
...
+        }
+        Some(cp) => cp,
+    };
+
+    // 1. Heartbeat Check:
+    // Updates if `update_on_heartbeat` is enabled and the Pyth price is newer than or equal to
+    // the chain price plus `heartbeat_seconds`.
+    if update_criteria.update_on_heartbeat {
+        if pyth_price.publish_time
+            >= chain_price.publish_time + (update_criteria.heartbeat_seconds as i64)
```
ali-bahjati, line 187: let's never use + , instead use `saturated_add` to make sure we never overflow (otherwise it might result in some security issues)

apps/argus/src/services/controller_service.rs
```diff
@@ -125,3 +161,595 @@ impl Service for ControllerService {
...
+                );
+                return true;
+            }
+        } else {
+            let price_diff = pyth_price.price.abs_diff(chain_price.price);
+            let threshold_val = (chain_price.price.abs() as u64
+                * update_criteria.deviation_threshold_bps as u64)
+                / 10000;
+
+            if price_diff > threshold_val {
```
ali-bahjati, line 218: let's double check this matches the onchain logic 100% (it looked the same when i looked)

apps/argus/src/services/controller_service.rs
```diff
@@ -125,3 +161,595 @@ impl Service for ControllerService {
...
+    // This means a price difference of more than 0.5 would trigger an update
+    if update_criteria.update_on_deviation {
+        // Critical assumption: The `expo` fields of `pyth_price` and `chain_price` are identical,
+        // since we directly compare the `price` fields.
+        if chain_price.price == 0 {
+            if pyth_price.price != 0 {
+                tracing::debug!(
+                    "Deviation criteria met: Chain price is 0, Pyth price is non-zero."
+                );
+                return true;
```
ali-bahjati, line 210: this is not in the on-chain logic, however the on-chain logic accepts any update if there's no update, maybe it is this one?

ali-bahjati (approved):
Looks very good! I left some comments, please address them before merging the code.

PR #2705, opened 2025-05-19
chore(target_chains/ethereum): cut another release for pyth

ali-bahjati:
## Summary

We got a new change in that changed the version again. I'm bumping it so we don't go back.


bplatak (approved):
godspeed

PR #2707, opened 2025-05-20
chore(target_chains/ethereum): add new networks

ali-bahjati:
## Summary

Add evm new networks and update the compiler to use `paris` evm version (= the evm instructions at the merge) as some newer instructions in newer versions are not supported in all chains.


contract_manager/store/contracts/EvmWormholeContracts.yaml
```diff
@@ -490,3 +487,12 @@
 - chain: worldchain_testnet
   address: "0xb27e5ca259702f209a29225d0eDdC131039C9933"
   type: EvmWormholeContract
+- chain: mezo_testnet
```
ali-bahjati, line 490: i'm midway with mezo_testnet deployment.

KemarTiti (approved):
lgtm

PR #2708, opened 2025-05-20
feat(fortuna): better logging + avoid unnecessary fee syncs

m30m:
## Summary

- Add a new config that disables fee syncs if the chain is already registered.
- Improve logging:
  - Change retry log levels from ERROR to WARN
  - Add more info on some other logs

## Rationale

Whenever we deploy a new version or change some configuration, the setup script is executed and fees are synced to the default value. This is unnecessary and can cause problems for some of the requests made within the duration that the fee is reset (if the users are paying for the fee and the value sent to our contract is the tx value signed by user).

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/config.sample.yaml
+++ b/apps/fortuna/config.sample.yaml
@@ -32,6 +32,9 @@ chains:
     # How much to charge in fees
     fee: 1500000000000000
 
+    # Set this temporarily to false if you have changed the fees and want to apply a new baseline fee.
+    sync_fee_only_on_register: true
+
     # Configuration for dynamic fees under high gas prices. The keeper will set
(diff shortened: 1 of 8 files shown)


apps/fortuna/config.sample.yaml
```diff
@@ -32,6 +32,9 @@ chains:
     # How much to charge in fees
     fee: 1500000000000000
 
+    # Set this temporarily to false if you have changed the fees and want to apply a new baseline fee.
+    sync_fee_only_on_register: true
```
jayantk, line 36: I think this is a brittle config flag -- people will forget to unset it. I think it is better to auto-detect whether or not you need to set the fees differently.

E.g., can we run the same code we run in the keeper to set the fees based on the gas price?

PR #2717, opened 2025-05-21
feat(fortuna): Improve explorer

m30m:
## Summary

- Revert the unnecessary RETRY_PREVIOUS_BLOCKS PR
- Add more fields to fortuna:
  - gas
  - combined_random_value
  - limit and offset
  - network_id

## Rationale

Network id is a better way to connect the ui, contract manager and other components to this backend since the names are not fully standardised (camelCase/snake_case and sometimes a testnet/mainnet is missing)

## How has this been tested?

- [x] Added new tests

Here is an example json payload:

```json
{
    "chain_id": "monad-testnet",
    "network_id": 10143,
    "provider": "0xde817704b3da538580f665ad248c947acb06976d",
    "sequence": 16,
    "created_at": "2025-05-21T23:39:38.168051Z",
    "last_updated_at": "2025-05-21T23:39:38.168056Z",
    "request_block_number": 17944962,
    "request_tx_hash": "0x7378963ecf2a6a9fc5b8b5b3bc08a002401a5ee28530ee6baba46b67b702fcc0",
    "gas_limit": "500000",
    "user_random_number": "129e247e17d0bd17ae8087eb024b381f2d9832ebe741f120f3e8490664b871f3",
    "sender": "0x78357316239040e19fc823372cc179ca75e64b81",
    "state": {
      "state": "completed",
      "reveal_block_number": 17944966,
      "reveal_tx_hash": "0x3ab2e2c6d1b2857aa5b4be9fa4e214c0b10740bc1127d1c706e25b97f498a2f9",
      "provider_random_number": "5e56b5976b77e8d457ce01daf9d949ffa7851ac95d3db0f54171f8f2b71da921",
      "gas_used": "81880",
      "combined_random_number": "59605786c32b552f41fceecdf8745c9896de00990e6f689dbb2111a25c12b452"
    }
  }
  ```

### CODE REVIEW

File: apps/fortuna/src/serde.rs [hunk 1]
```diff
@@ -0,0 +1,21 @@
+pub mod u256 {
```
jayantk, line 1: maybe we should have a types/ directory where this file could be types/u256 and then we can move out other type definitions there?

m30m: hmmm, this is not our type but just overriding the way we want to (de)serialize it. I scanned the repo and we have managed to get away with `serde_with` so far with other types.

PR #2718, opened 2025-05-22
feat: Lazer publisher agent

merolish:
## Summary

Based on the Lazer relayer, this binary receives incoming update streams from a publisher and sends signed transactions to the Lazer relayer itself.

## Rationale

Managing transaction signatures on the publisher side.

## How has this been tested?

- [ ] Added new tests


darunrs:
At a high level, is there a reason we can't put this in the Pyth Publisher SDK?

merolish:
The sdk is a library crate and this is a binary which will have some docker/devops stuff as well. No strong feelings though.

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub async fn run_relayer_connection(&mut self) -> Result<()> {
+        // Establish relayer connections
+        // Relayer will drop the connection if no data received in 5s
+        let (mut relayer_sender, relayer_receivers) = connect_to_relayers(&self.config).await?;
+        let mut stream_map = StreamMap::new();
+        for (i, receiver) in relayer_receivers.into_iter().enumerate() {
+            stream_map.insert(self.config.relayer_urls[i].clone(), receiver);
+        }
+
+        // Read the keypair from the file using Solana SDK because it's the same key used by the Pythnet publisher
```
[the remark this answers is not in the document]
darunrs: We're also burning time as the longer we take before sending updates, the greater the chance of Relayer dropping the connection. We don't know what the window is for agent running and publishers sending data. We might even need to avoid connecting *until* agent receives data. That way, we aren't forcing our publishers to manage a race condition.

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,220 @@
...
+use tracing::{error, instrument, warn};
+
+use crate::{
+    http_server,
+    lazer_publisher::LazerPublisher,
+    websocket_utils::{handle_websocket_error, send_ping, send_text},
+};
+
+//pub const MAX_EXPIRY_TIME_US: u64 = 5_000_000; // 5s
+//pub const MAX_TIMESTAMP_ERROR_US: u64 = 50_000; // 5ms
```
darunrs, line 31: Remove? Seems some leftover comments.

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,220 @@
...
+            pin!(receive);
+            #[allow(clippy::never_loop)] // false positive
+            loop {
+                select! {
+                    result = &mut receive => {
+                        receive_type = Some(result?);
+                        break
+                    }
+                    _ = ping_interval.tick() => {
+                        send_ping(&mut ws_sender).await;
```
darunrs, line 97: Pinging them is not necessary as Agent will not be able to collect and send this info anywhere. Maybe you can put it behind a flag to log the ping results, but for now it's not necessary.

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,220 @@
...
+                        //last_ping_timestamp = Some(TimestampUs::now());
+                    }
+                    _ = &mut publisher_timeout => {
+                        bail!("no updates received from publisher after {:?}", PUBLISHER_TIMEOUT);
+                    }
```
darunrs, line 102: What's the value in d/c'ing them from Agent? These are both being ran on their end.

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,220 @@
...
+                }
+            }
+        }
+        publisher_timeout
+            .as_mut()
+            .reset(Instant::now() + PUBLISHER_TIMEOUT);
+
+        if let Some(Incoming::Pong(_)) = receive_type {
+            continue;
+        }
```
darunrs, line 112: Let's remove the ping code if its not going to do anything.

pyth-lazer-agent/src/websocket_utils.rs
```diff
@@ -0,0 +1,58 @@
...
+            _ => {
+                warn!("error while handling connection: {:?}", err.to_string())
+            }
+        }
+    } else {
+        warn!("error while handling connection: {:?}", err.to_string());
+    }
+}
+
+pub async fn send_ping<T: AsyncRead + AsyncWrite + Unpin>(sender: &mut Sender<T>) {
```
darunrs, line 34: Same deal here. Don't need it if we don't use it.

darunrs (commented):
Gave it a preliminary review. Broadly speaking, it seems this PR is running along the idea of getting them to run it first, then have them send the new type to a new endpoint after? 

Retaining the V1 and V2 endpoints maintains the confusion behind all the types. I don't think providing them a new reasonably easy type to work with is that large a lift after getting them to run the agent, which is probably harder? But yeah curious to hear what you had in mind in terms of the onboarding process.

pyth-lazer-agent/config/config.toml
```diff
@@ -0,0 +1,5 @@
+relayer_urls = ["ws://localhost:1235/v1/transaction", "ws://localhost:1335/v1/transaction"]
+authorization_token = "token1"
+publish_keypair_path = "/path/to/solana/id.json"
```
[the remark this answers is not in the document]
merolish: Yep, good call.

pyth-lazer-agent/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "pyth-lazer-agent"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+pyth-lazer-publisher-sdk = "0.1.3"
+pyth-lazer-protocol = "0.7.2"
```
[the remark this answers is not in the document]
merolish: As per discussion with Keyvan, publishers publish legacy json to this agent (/v1/publisher and /v2/publisher), and the agent handles batching and signing and publishes to the relayers' /v1/transaction. I have yet to address https://linear.app/douro-labs/issue/PF-133/design-publisher-facing-data-type which is what I think you're referring to by new endpoint/type; we should sync up if I'm misunderstanding.

pyth-lazer-agent/src/config.rs
```diff
@@ -0,0 +1,37 @@
...
+use url::Url;
+
+#[derive(Deserialize, Derivative, Clone, PartialEq)]
+#[derivative(Debug)]
+pub struct Config {
+    pub listen_address: SocketAddr,
+    pub publisher_id: PublisherId,
+    pub relayer_urls: Vec<Url>,
+    pub authorization_token: String,
+    pub publish_keypair_path: PathBuf,
```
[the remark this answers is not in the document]
merolish: Ah yes. Hmm, I assume we'll be copying in the the file to the Docker container anyway? So there wouldn't be any point debug logging it anyway.

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,252 @@
...
+)> {
+    tracing::info!("connecting to the relayer at {}", url);
+    url.set_path("/v1/transaction");
+    let mut req = url.clone().into_client_request()?;
+    let headers = req.headers_mut();
+    headers.insert(
+        "Authorization",
+        HeaderValue::from_str(&format!("Bearer {}", token))?,
+    );
+    let (ws_stream, _) = connect_async_with_config(req, None, true).await?;
```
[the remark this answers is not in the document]
merolish: It's the http response from the websocket connect, which we ignore presuming we successfully connect to the relayer.

pyth-lazer-agent/src/lazer_publisher.rs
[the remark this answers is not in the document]
merolish: Good point, will fix.

pyth-lazer-agent/src/publisher_handle.rs
merolish: Holdover code from relayer.

pyth-lazer-agent/config/config.toml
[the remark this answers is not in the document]
ali-behjati: as a principle I like us to be as similar to agent wrt to config. so let's leave it here

pyth-lazer-agent/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+
+mod config;
+mod http_server;
+mod lazer_publisher;
+mod publisher_handle;
+mod websocket_utils;
+
+#[derive(Parser)]
+#[command(version)]
+struct Cli {
```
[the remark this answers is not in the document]
keyvankhademi: we should support config file + env. We don't need to add the configs in cli args.

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,252 @@
...
+                    "Reading publish keypair returned an error. ",
+                );
+                bail!("Reading publish keypair returned an error. ");
+            }
+        };
+
+        let signing_key = SigningKey::from_keypair_bytes(&publish_keypair.to_bytes())
+            .context("Failed to create signing key from keypair")?;
+
+        let mut publish_interval = interval(self.config.publish_interval_duration);
```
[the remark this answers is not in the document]
keyvankhademi: if the compiler is not complaining it probably means we dont

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,173 @@
...
+                            &UpdateDeserializationErrorResponse {
+                                error: format!("failed to parse binary update: {err}"),
+                            }
+                            .into(),
+                        )?;
+                        send_text(&mut ws_sender, error_json).await?;
+                        continue;
+                    }
+                }
+            } //_ => bail!("Publisher API request set with invalid context"),
```
keyvankhademi: what is this?

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,153 @@
...
+        };
+        for relayer_sender in self.relayer_senders.iter() {
+            if let Err(e) = relayer_sender
+                .sender
+                .send(signed_lazer_transaction.clone())
+                .await
+            {
+                error!("Error sending transaction to Lazer relayer session: {e:?}");
+            }
+        }
```
keyvankhademi: is this sync? does it wait for 1 relayer send to finish before running the next one? In that case we should fix it.
Ideally concurrent, at least async.

pyth-lazer-agent/src/relayer_session.rs
```diff
@@ -0,0 +1,147 @@
...
+    // connection state
+    url: Url,
+    token: String,
+    receiver: Receiver<SignedLazerTransaction>,
+}
+
+impl RelayerSessionTask {
+    pub async fn run(&mut self) {
+        let mut failure_count = 0;
+        let retry_duration = Duration::from_secs(1);
```
keyvankhademi: might wanna have expo backoff. Having 1 sec down time for a random disconnect is not good.
Also make it configurable.

pyth-lazer-agent/src/lazer_publisher.rs
merolish: This is just sending to the channel for each relayer thread to pick up, but I guess I shouldn't call await for that.

pyth-lazer-agent/src/publisher_handle.rs
merolish: Oh, that's left over from the relayer code, will remove.

pyth-lazer-agent/src/relayer_session.rs
darunrs: Pavel used the backoff crate for configuring expo retries. Perhaps you can use that?

pyth-lazer-agent/src/config.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub authorization_token: String,
+    #[derivative(Debug = "ignore")]
+    pub publish_keypair_path: PathBuf,
+    #[serde(with = "humantime_serde", default = "default_publish_interval")]
+    pub publish_interval_duration: Duration,
+}
+
+fn default_publish_interval() -> Duration {
+    Duration::from_millis(50)
+}
```
darunrs: Again, this is not the interval we want to publish at. We want to publish at 0.5ms interval. On tick, we send any new feed updates over. The interval is overall, not per feed. 

If no new data has arrived, we can skip the tick, up until perhaps 25ms where we instead send all current feed data over at once. 

Publisher update rate will now have to be each unique source timestamp dated update. We'll probably need to update some metrics for this. 

IMO, the above makes sense as an approach.

pyth-lazer-agent/src/main.rs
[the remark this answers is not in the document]
darunrs: Yeah I can agree with that. Ali actually mentioned wanting to keep the interface the same as Pyth Agent and that's a good point too.

pyth-lazer-agent/src/lazer_publisher.rs
darunrs: It is sync, though the [docs](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#impl-Sender%3CT%3E) do mention "This method will never block the current thread.". Regardless, it'd be nice to do them concurrently.

pyth-lazer-agent/src/publisher_handle.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use crate::{
+    http_server,
+    lazer_publisher::LazerPublisher,
+    websocket_utils::{handle_websocket_error, send_text},
+};
+
+pub struct PublisherConnectionContext {
+    pub request_type: http_server::Request,
+    pub _remote_addr: SocketAddr,
```
darunrs, line 26: Remove the leading underscore.

darunrs (commented):
Nice, I think this is pretty good. Some small comments. After that, I think we'll be good to go.

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,252 @@
...
+    Vec<SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>>,
+)> {
+    let mut relayer_senders = Vec::new();
+    let mut relayer_receivers = Vec::new();
+    for url in config.relayer_urls.clone() {
+        let (relayer_sender, relayer_receiver) =
+            connect_to_relayer(url, &config.authorization_token).await?;
+        relayer_senders.push(relayer_sender);
+        relayer_receivers.push(relayer_receiver);
+    }
```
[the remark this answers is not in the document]
merolish: relayer sessions have since been broken out into separate threads.

pyth-lazer-agent/src/lazer_publisher.rs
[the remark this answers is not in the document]
merolish: This is now just done once at startup.

pyth-lazer-agent/src/lazer_publisher.rs
merolish: All right, then I'm going to use `futures::future::join_all` whenever possible for fan-out cases like this.

pyth-lazer-agent/src/relayer_session.rs
merolish: Ah, I see it used in a couple places in lazer, will do that here.

pyth-lazer-agent/src/config.rs
merolish: Yeah, I'd updated my own un-checked in config and not the default here 😄 .

So we'll maintain most recent update per feed and then send an entire snapshot if we hit this expiry period?

I don't quite understand the wording of the last statement.

pyth-lazer-agent/src/publisher_handle.rs
merolish: Just did this temporarily for clippy because we don't log or metric this anywhere yet.

pyth-lazer-agent/src/relayer_session.rs
merolish: Done.

pyth-lazer-agent/src/lazer_publisher.rs
darunrs: Ah that's the wrong mpsc sender anyway. We're using tokio's. That one is async. But yeah use join_all. Be careful about failure cases though. Honestly sending into an MPSC should not fail but if it does, we want to still be able to send to the rest. I guess its possible for the channel to close due to filling up and so on.

pyth-lazer-agent/src/config.rs
darunrs: Hmm no I guess on each tick you check if something hasn't been updated/sent for 25ms, and you send whatever you have. Otherwise, just keep sending only the new data that has come in. 

The logic on tick is:
1. Collect all new updates
2. Collect latest data of feeds not updated for 25ms
3. Update all feeds whose data is being sent that they're now being sent in whatever cache we are using
4. Send the data

Something like that. Basically, we make sure we always send data on 0.5ms ticks if new data is coming in, and always send data every 25ms for sure for each feed.

pyth-lazer-agent/src/publisher_handle.rs
darunrs: But we use it don't we? On line 100. 
https://github.com/pyth-network/pyth-crosschain/pull/2718/files/3486c61498ee06a667750284a7c1ee7bd05df23a#diff-bb96dcad51ea57119d35403ef9b490c5ddcfc2af009eb125bb4db2a9f59e9c22R100

pyth-lazer-agent/src/config.rs
merolish: So, if a feed hasn't been published for 25ms we retransmit its latest update? Makes sense, although I admit I don't see what this sort of heartbeating buys us at this step in the data flow.

pyth-lazer-agent/src/publisher_handle.rs
merolish: The field is initialized in the struct but is never read.

pyth-lazer-agent/src/config.rs
darunrs: Its hard as a server to know for sure if a client is behaving appropriately. If the client consistently sends something over, we can expect it is functioning properly. The connection being open is not sufficient to know this. We also still want to trigger aggregations and such. If the data itself (source timestamp) gets old enough, it will be expired anyway.

pyth-lazer-agent/src/publisher_handle.rs
darunrs: Ah ok. If we don't intend to use the field, and we don't have a follow up PR already planned that enables the use of that field, let's remove it.

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -137,15 +137,12 @@ impl LazerPublisherTask {
...
-                .sender
-                .send(signed_lazer_transaction.clone())
-                .await
-            {
-                error!("Error sending transaction to Lazer relayer session: {e:?}");
-            }
-        }
+        futures::future::join_all(
+            self.relayer_senders
+                .iter_mut()
```
darunrs, line 142: Is iter_mut() needed? send should only need a regular reference.

pyth-lazer-agent/src/relayer_session.rs
```diff
@@ -85,8 +87,15 @@ struct RelayerSessionTask {
...
+        let mut backoff = ExponentialBackoffBuilder::new()
+            .with_initial_interval(initial_interval)
+            .with_max_interval(max_interval)
+            .with_max_elapsed_time(None)
+            .build();
+
         let mut failure_count = 0;
-        let retry_duration = Duration::from_secs(1);
 
         loop {
```
darunrs, line 100: Instead of having a manual loop, use backoff's [retry](https://docs.rs/backoff/latest/backoff/future/fn.retry.html). Something like this:
`
retry(expo_backoff_config, || {
match connection().await {
Ok(()) => 'shutdown',
Err(err) => handle_err
}
})
`

pyth-lazer-agent/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /pyth-lazer-agent
+
+RUN cargo build --release
+
+FROM debian:12-slim
+
+RUN apt update && apt install -y libssl-dev && apt clean all
+
+COPY --from=builder /pyth-lazer-agent/target/release/pyth-lazer-agent /pyth-lazer-agent/
+COPY --from=builder /pyth-lazer-agent/config/* /pyth-lazer-agent/config/
```
ali-behjati, line 15: should we add this?

pyth-lazer-agent/src/config.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[derivative(Debug = "ignore")]
+    pub authorization_token: String,
+    #[derivative(Debug = "ignore")]
+    pub publish_keypair_path: PathBuf,
+    #[serde(with = "humantime_serde", default = "default_publish_interval")]
+    pub publish_interval_duration: Duration,
+}
+
+fn default_publish_interval() -> Duration {
+    Duration::from_micros(500)
```
ali-behjati, line 24: isn't this too aggressive?

pyth-lazer-agent/src/lazer_publisher.rs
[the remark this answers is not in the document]
ali-behjati: interval is pinned on its own internally (and implements Unpin).

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,150 @@
...
+            special_fields: Default::default(),
+        };
+        futures::future::join_all(
+            self.relayer_senders
+                .iter_mut()
+                .map(|relayer_sender| relayer_sender.sender.send(signed_lazer_transaction.clone())),
+        )
+        .await;
+
+        self.pending_updates.clear();
```
ali-behjati, line 147: since you release the task a line before, here you might clear some pending updates that you haven't sent yet.

ali-bahjati (approved):

pyth-lazer-agent/src/lazer_publisher.rs
merolish: Nope. Was experimenting with AI suggestions and left this in. Removing.

pyth-lazer-agent/src/config.rs
merolish: As per Darun's comments above, it sounded like half a millisecond was the target frequency. I don't have background on our requirements here.

pyth-lazer-agent/Dockerfile
merolish: I followed what pyth-agent does, although I'm not sure the best way to manage config here. Also we'll need to add the key file.

pyth-lazer-agent/src/lazer_publisher.rs
merolish: Really? pending_updates only gets pushes in the other arm of the select!, I'd have thought that and batch_transaction is atomic with respect to that.

pyth-lazer-agent/src/publisher_handle.rs
merolish: Perhaps at some point perhaps we log it or it goes in a metric?

pyth-lazer-agent/src/relayer_session.rs
merolish: Ah, nice.

pyth-lazer-agent/src/config.rs
darunrs: The background on this is that the frequency needs to support 1ms feeds. Granted we don't have many currently. But the update frequency needs to be low enough to provide granularity here. The relayer side batching does 0.5ms batches itself currently.

merolish:
@darunrs I've done ad hoc integration testing using example publisher -> lazer agent -> relayer in tilt, but yes, it's time for me to get into the habit of `mod tests` in any new code, will try to get some in this weekend.

pyth-lazer-agent/src/config.rs
keyvankhademi: @darunrs 0.5ms doesn't sound right. That means 2000 updates per second per relayer. 50ms is also too much, I think we should fine-tune this number. I'm thinking somewhere between 2-10 ms.

pyth-lazer-agent/Dockerfile
keyvankhademi: this doesn't make sense to me. The config should be passed when running the binary, no?

pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -0,0 +1,150 @@
...
+    async fn batch_transaction(&mut self, signing_key: &SigningKey) -> Result<()> {
+        if self.pending_updates.is_empty() {
+            return Ok(());
+        }
+
+        let publisher_update = PublisherUpdate {
+            updates: self.pending_updates.clone(),
+            publisher_timestamp: MessageField::some(Timestamp::now()),
+            special_fields: Default::default(),
+        };
```
keyvankhademi, line 114: regarding Ali's comment. It's true that we won't add to `self.publisher_updates` in this function because of `tokio::select!`. But as a best practice, you can use `updates = self.pending_updates.drain(..)` to get and remove all the updates in 1 function.

pyth-lazer-agent/src/lazer_publisher.rs
keyvankhademi: in the case that one channel gets full, we will stop sending to all channels. `join_all` doesn't solve this because it waits for all of them to finish. Anyway I don't think this is something we should fix right now, but keep that in mind in case somethings goes wrong.

pyth-lazer-agent/src/config.rs
darunrs: Going slower isn't outright going to be a problem. But as we grow real time feeds, we'll intentionally be bottlenecking them. I don't mind going slower since we have few of them. But we definitely want a low ticker and potentially feed specific ticker evaluations (e.g. real time evaluates every tick, 50ms every 25ms ticks, etc.). 2ms is probably fine I suppose.

pyth-lazer-agent/Dockerfile
merolish: In pyth-agent, the config is copied into the container like this, and the file can be read as a CLA but it defaults to config/config.toml [here](https://github.com/pyth-network/pyth-agent/blob/main/src/bin/agent.rs#L28) which is what happens in practice. I need to fix this in the lazer agent because I didn't set a default [update: set the wrong default, actually], but we can do whatever people think best.

pyth-lazer-agent/src/lazer_publisher.rs
merolish: I like it, will change in the next PR.

pyth-lazer-agent/Dockerfile
keyvankhademi: Do you mean in practice the publisher volume their own config and uses the default path? Even in that case we don't need to copy the config. I don't think there is any use of the default config in the docker.
@ali-bahjati am I missing something?

PR #2719, opened 2025-05-22
feat(lazer/publisher_sdk): add wormhole governance source proto

ali-bahjati:
## Summary

Add wormhole governance source for the governance instructions.

## Rationale

Governance actions are going to come from the Pyth DAO councils that live on Solana.


lazer/publisher_sdk/proto/governance_instruction.proto
```diff
@@ -29,14 +28,15 @@ message GovernanceInstruction {
...
     optional google.protobuf.Timestamp max_execution_timestamp = 4;
-    // [required] Sequence number of this instruction. It must be greater than 0.
-    // It must always be increasing, but not required to be
-    // strictly sequential (i.e. gaps are allowed). Each shard separately keeps track of the last executed
-    // governance instruction and will reject instructions with the same or smaller sequence no.
-    // Note that if instructions are received out of order, some of them may become permanently
-    // rejected (e.g. if instruction #3 has been successfully processed before instruction #2 was observed,
-    // #2 will always be rejected).
-    // Sequence numbers are assigned and tracked separately for each governance source.
+    // [optional] Sequence number of this instruction. Required for SingleEd25519 governance source
```
darunrs, line 31: Why is it optional now?

lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
```diff
@@ -15,7 +15,11 @@ message SignedLazerTransaction {
     // [required] signature with public key
     optional SignatureData signature_data = 1;
 
-    // [required] lazer transaction encoded as bytes through protobuf
+    // [required] lazer transaction payload encoded as bytes.
+    // If the signature data is a Ed25519SignatureData, the payload is the encoded
+    // LazerTransaction protobuf message.
+    // If the signature data is a WormholeMultiSigData, the payload is the encoded
+    // Wormhole VAA body.
```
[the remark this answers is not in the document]
darunrs: Would these be two different payload types? The comment says the payload would be wormhole VAA body. So, it wouldn't be an existing payload enum body but rather an entirely new different byte payload that is decoded in a different way?

lazer/publisher_sdk/proto/governance_instruction.proto
ali-bahjati: wasn't it clear in the sentence after :?

lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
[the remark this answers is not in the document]
ali-bahjati: yeah i just added it.

Well they are 2 different types but all represented as bytes :D i added some docs.

lazer/publisher_sdk/proto/governance_instruction.proto
Riateche: We will validate sequence-no on our side exactly like you described. It's optional here because if we use wormhole it's not possible to specify sequence-no in the payload, since it's generated by wormhole itself after we send the payload. The wormhole's sequence-no will be present in the transaction data, just not here.

PR #2722, opened 2025-05-23
feat: add initial Lazer integration

cctdaniel:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/governance/xc_admin/packages/xc_admin_common/src/programs/lazer/lazer_functions.ts
+++ b/governance/xc_admin/packages/xc_admin_common/src/programs/lazer/lazer_functions.ts
@@ -2,31 +2,10 @@ import { PublicKey, TransactionInstruction } from "@solana/web3.js";
 import { PythCluster } from "@pythnetwork/client";
 import {
   ValidationResult,
-  DownloadableProduct,
-  DownloadableConfig,
   ProgramType,
+  LazerConfig,
+  LazerState,
 } from "../types";
-
-/**
- * Program ID for the Pyth Lazer program
- */
-export const LAZER_PROGRAM_ID = new PublicKey(
-  "pytd2yyk641x7ak7mkaasSJVXh6YYZnC7wTmtgAyxPt",
-);
-
-/**
- * Lazer-specific configuration type
- * TODO: Change to actual Lazer config type
- */
-export type LazerConfig = {
-  programType: ProgramType.PYTH_LAZER;
-  // Make cluster optional since Lazer might not be tied to a specific cluster
-  cluster?: PythCluster;
-  // More generic data source instead of Solana-specific accounts
-  feeds: LazerFeed[];
-  // Additional metadata that might be relevant for Lazer
-  metadata?: Record<string, unknown>;
-};
 
 /**
  * Parameters for getting Lazer configuration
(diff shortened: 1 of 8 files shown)


governance/xc_admin/packages/xc_admin_common/src/programs/lazer/lazer_functions.ts
```diff
@@ -148,24 +154,143 @@ export function getDownloadableConfig(config: LazerConfig): DownloadableConfig {
...
+    // Check for required fields
+    if (
+      !("shardId" in uploadedLazerState) ||
+      !("feeds" in uploadedLazerState) ||
+      !("publishers" in uploadedLazerState)
+    ) {
+      return {
+        isValid: false,
+        error: "Missing required fields in Lazer configuration",
+      };
```
cprussin, line 182: Is this not guaranteed by the type?

In general I recommend using something like [zod](https://zod.dev/) to describe and validate schemas in typescript; if this can't be guaranteed by the type system I'd use Zod to parse instead.

governance/xc_admin/packages/xc_admin_common/src/programs/lazer/lazer_functions.ts
```diff
@@ -148,24 +154,143 @@ export function getDownloadableConfig(config: LazerConfig): DownloadableConfig {
...
+    }
+
+    // Calculate changes between existing and uploaded config
+    const changes: Record<string, { prev?: any; new?: any }> = {};
```
cprussin, line 186: Please do not use `any` -- you can restructure this type slightly by using well known keys to be easier to type (and probably easier to use) with something along these lines:

```ts
type ChangeField<T> = { prev: T, new: T };

type Changes = {
  shard: ChangeField<Shard>;
  feeds: Record<string, ChangeField<Feed>>;
  publishers: Record<string, ChangeField<Publisher>>;
}
```

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -116,3 +119,64 @@ export interface ValidationResult {
...
+};
+
+/**
+ * Lazer-specific configuration type
+ */
+export type LazerConfig = {
+  programType: ProgramType.PYTH_LAZER;
+  // The Lazer state data
+  state: LazerState;
+};
```
cprussin, line 182: These types are all specific to lazer so they should live alongside the lazer code

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -85,8 +85,11 @@ export type DownloadableProduct = {
 
 /**
  * Type for downloadable configuration
+ * Can be either a mapping of symbols to products (Core) or a LazerState (Lazer)
  */
-export type DownloadableConfig = Record<string, DownloadableProduct>;
+export type DownloadableConfig =
+  | Record<string, DownloadableProduct>
+  | LazerState;
```
cprussin, line 92: To me this now looks like two separate types and it's not really clear why the union is ever useful, it doesn't seem like it would ever be usable as a union without the user having to know whether they're working with lazer or core, and if the consumer has to know that, what's the value in having a union or a function that returns the union?  If the user knew whether they were working with lazer or core, there's no point in using a function abstracted over both, they should just use the function that's correct for the system they're on.

Even when I look below in this PR where this and the `getDownloadableConfig` function are being used, they're in `PythCore.tsx` -- a file explicitly for pyth core -- so then there's no real point in using the abstracted function in `program_registry.ts` and we should just use the core function from `core_functions.ts` instead.

Frankly, the more I see stuff like this, the more I question the value of `program_registry.ts` entirely -- the common interface is useful if the interfaces are common, but it really is starting to look like they aren't common, and the abstraction is not even usable without knowing which program you're working with.  If that's the case, it's just added indirection and causes type information to become less precise.

If you can come up with a generic type which Core & Lazer both implement and define the abstraction in terms of that generic type, and you plan to use them in situations where you know you have _a_ pyth program but you don't know _which_, I could see it being useful.  But if you don't have common arguments AND you don't have common return types, then you should just export the functions from core and lazer separately, and expect the consumers to import the correct one directly.

governance/xc_admin/packages/xc_admin_common/src/programs/types.ts
```diff
@@ -116,3 +119,64 @@ export interface ValidationResult {
...
+  publishers: LazerPublisher[];
+};
+
+/**
+ * Lazer-specific configuration type
+ */
+export type LazerConfig = {
+  programType: ProgramType.PYTH_LAZER;
+  // The Lazer state data
+  state: LazerState;
```
cprussin, line 181: This type looks redundant, if the type is explicitly a lazer type, why would you also need a `programType: ProgramType.PYTH_LAZER` ?  This is again making me question the `program_registry.ts` pattern.

governance/xc_admin/packages/xc_admin_frontend/components/programs/PythCore.tsx
```diff
@@ -400,7 +399,10 @@ const PythCore: React.FC<PythCoreProps> = ({ proposerServerUrl }) => {
 
   useEffect(() => {
     if (!dataIsLoading && rawConfig) {
-      const downloadableConfig = getDownloadableConfig(rawConfig)
+      const downloadableConfig = getDownloadableConfig(rawConfig) as Record<
+        string,
+        DownloadableProduct
+      >
```
cprussin, line 405: Why do we need a typecast here?  If the return types of `getDownloadableConfig` are accurate, we shouldn't need the typecast.  If they aren't accurate, then we shouldn't be typecasting, we should be fixing the types to be accurate.

governance/xc_admin/packages/xc_admin_frontend/components/programs/PythLazer.tsx
```diff
@@ -1,41 +1,503 @@
...
+const ShardChangesRows: React.FC<ShardChangesRowsProps> = ({ changes }) => {
+  const isNewShard = !changes.prev && changes.new
+
+  return (
+    <>
+      {Object.entries(changes.new).map(
+        ([key, newValue]) =>
+          (isNewShard ||
+            (changes.prev &&
+              changes.prev[key as keyof typeof changes.prev] !== newValue)) && (
```
cprussin, line 81: This is kind of awkward, if this is how you plan to use this then I'd rethink the type a bit.  The type system can't actually guaranteed that the keys are shared between `prev` and `new`, which is why you have the typecast.  So instead I think it would be better to have the type be something where you have `prev` and `new` underneath each key, i.e. something along the lines of (referencing my prior comment):

```ts
type Changes = {
  shard: Record<keyof Shard, ChangeField<Shard[keyof Shard]>>;
  feeds: Record<string, ChangeField<Feed>>;
  publishers: Record<string, ChangeField<Publisher>>;
}
```

In general, typecasting should be avoided, and if you have to use it for stuff like this, it generally indicates that you should rethink the types you're working with.

governance/xc_admin/packages/xc_admin_frontend/contexts/PythContext.tsx
```diff
@@ -46,15 +51,32 @@ export const PythContextProvider: React.FC<PythContextProviderProps> = ({
...
+        const lazerConfig = getConfig[ProgramType.PYTH_LAZER]({})
+        // Since we know this is a LazerState object, we can cast it
+        setLazerState(lazerConfig.state)
+      } catch (error) {
+        console.error('Error fetching Lazer state:', error)
+      }
+    }
+
+    fetchLazerState()
+  }, [])
```
cprussin, line 76: I'm a bit confused why this is async, it looks like everything it does is sync?

Also, you might like this pattern which I always use for client-side data fetching: https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/src/useData/index.ts.  It has a few nice things:

- This uses `swr` so you can easily configure it to automatically revalidate data if you so desire
- It returns a tagged union of states so handling loading / error / before loading states is not only trivial, but typescript will actually enforce that you don't forget to do it

Here's an example of recent code I wrote that uses it: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/src/components/Home/results.tsx#L31-L62

cprussin (changes requested):
Biggest change I think that needs to happen here is to rethink some of the types, especially the type for changes.  Also I'm calling into question the value of the `program_registry.ts` abstraction.

PR #2727, opened 2025-05-23
refactor: Optional storage flag when parsing price feed updates

ayushboss:
## Summary

I updated the parsePriceFeedUpdatesInternal function to be more flexible and take an additional flag indicating whether or not parsed prices need to be stored. 

## Rationale

Recent changes added the parsePriceFeedUpdatesInternal function to prevent invocations of that function from updating storage. An additional flag allows us to be flexible with use cases that may need to update storage while performing parsing.

## How has this been tested?

- [x] Added new tests

Tested using existing tests. Additionally, I added tests to verify the new parse function both when storage is enabled and disabled, checking that the parsed price signatures are found in the former case, and making sure that the price signatures are not updated in the latter. Also added gas benchmark tests.


target_chains/ethereum/sdk/solidity/MockPyth.sol
```diff
@@ -86,15 +86,21 @@ contract MockPyth is AbstractPyth {
...
+        bool storeUpdatesIfFresh
     )
-        internal
-        returns (PythStructs.PriceFeed[] memory feeds, uint64[] memory slots)
+        public
+        payable
+        returns (
+            PythStructs.PriceFeed[] memory feeds,
+            uint64[] memory slots
+        )
```
tejasbadadare, line 103: IMO this is a little confusing as a public API since it exposes lots of options (prone to misconfiguration) and also returns data most people won't need (slots). Might be better to keep this as a private method and expose public APIs targeted at specific use cases. The only thing using `storeUpdatesIfFresh` would be Pulse via  `parsePriceFeedUpdatesWithSlotsStrict`, so we could expose the `storeUpdatesIfFresh` param there. Curious on your thoughts on this @ali-bahjati .

tejasbadadare (commented):
Thanks! Couple notes
- Can we add a test case that asserts that when the param is enabled it actually updates the underlying public feed, and the negative case as well?
- Can we add a test to the gas benchmark to check how expensive parsing + storing is?
- Do the updated contracts fit within size limits? (`forge build --sizes`)

ayushboss:
> Thanks! Couple notes
> 
> * Can we add a test case that asserts that when the param is enabled it actually updates the underlying public feed, and the negative case as well?
> * Can we add a test to the gas benchmark to check how expensive parsing + storing is?
> * Do the updated contracts fit within size limits? (`forge build --sizes`)

Added both cases to check the updates, as well as an additional test to the gas benchmark. I checked the updated contracts, it seems that the "MockPythUpgrade" contract exceeds the size limit slightly (by 303 B) @tejasbadadare

target_chains/ethereum/sdk/solidity/AbstractPyth.sol
```diff
@@ -136,21 +136,6 @@ abstract contract AbstractPyth is IPyth {
         override
         returns (PythStructs.PriceFeed[] memory priceFeeds);
 
-    function parsePriceFeedUpdatesWithSlotsStrict(
```
tejasbadadare, line 139: did we squeeze within the size limit with this? :)

tejasbadadare (approved):
nice! couple small things, please address before merging. also looks like you may need to run the formatter ([pre-commit](https://pre-commit.com/) should help take care of this automatically going forward, see `.pre-commit-config.yaml`)

PR #2730, opened 2025-05-24
fix(apps/price_pusher) handle price feed removal more gracefully

ali-bahjati:
## Summary

The price pusher in many networks would throw an error if a feed doesn't exist and the behaviour was there to prevent config errors but it falls short when we remove a feed. This change adds a warning for non existent feeds but lets the pusher keep functioning (for the remaining feeds).

## How has this been tested?

- [x] Manually tested the code: tested missing feeds (another code path errors it) and tested hermes reconnections.

p.s: evm pusher already had the flag to ignore invalid feeds.

### DIFF
--- a/apps/price_pusher/src/pyth-price-listener.ts
+++ b/apps/price_pusher/src/pyth-price-listener.ts
@@ -5,6 +5,7 @@ import {
 } from "@pythnetwork/hermes-client";
 import { PriceInfo, IPriceListener, PriceItem } from "./interface";
 import { Logger } from "pino";
+import { sleep } from "./utils";
 
 type TimestampInMs = number & { readonly _: unique symbol };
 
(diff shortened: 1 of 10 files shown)


apps/price_pusher/src/pyth-price-listener.ts
```diff
@@ -71,22 +90,15 @@ export class PythPriceListener implements IPriceListener {
       });
     };
 
-    eventSource.onerror = (error: Event) => {
+    eventSource.onerror = async (error: Event) => {
       console.error("Error receiving updates from Hermes:", error);
       eventSource.close();
+      await sleep(5000); // Wait a bit before trying to reconnect
+      this.startListening(); // Attempt to restart the listener
```
ali-bahjati, line 97: please note that we have another loop checking for latest price update that crashes if there's no update within last 30 seconds.

cctdaniel (approved):

PR #2733, opened 2025-05-26
fix: adjusted @types/node to pull from global setting through catalog setting

ayushboss:
## Summary

Modified package.json of three internal packages to pull type for @types/node from the global setting, as defined in pnpm-workspace.yaml.

## Rationale

I was running into versioning errors when attempting to initially clone and run turbo build on the target chains Ethereum contract. We noticed that the build error was coming from an issue with the version of node that was being run, and these modifications helped line it up.

## How has this been tested?

- [ ] Added new tests

Builds proceed as normal.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1399,6 +1399,9 @@ importers:
         specifier: ^2.1.1
         version: 2.7.1
     devDependencies:
+      '@types/node':
+        specifier: 'catalog:'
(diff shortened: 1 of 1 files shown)


cprussin:
@ayushboss looks like you forgot to check in the update to `pnpm-lock.yaml`, please add that to resolve the CI failures

cprussin (approved):

PR #2743, opened 2025-05-28
feat(lazer) Add response messages for lazer transactions

bplatak:
## Summary
Add new message types to represent responses from the relayer to lazer transactions. Initially only include a very simple structure that we will expand upon later. Also add a synthetic transaction_id derived as sha256 hash of the message payload. Note that this is not a stable ID across proto versions and should only be used by the clients to identify responses to transactions over the same connection they were sent to.

## Rationale
The relayers receive two types of transactions: governance instructions, and publisher updates. Both deserve having an acknowledgment or otherwise a response explain what the error is to allow clients to fix it.  

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
+++ b/lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
@@ -88,3 +88,26 @@ message LazerTransaction {
         GovernanceInstruction governance_instruction = 2;
     }
 }
+
+// Response to SignedLazerTransactionResponse from the relayer
+message SignedLazerTransactionResponse {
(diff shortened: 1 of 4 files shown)


lazer/publisher_sdk/proto/pyth_lazer_transaction.proto
bplatak, line 1: Why is this payload so generic anyway? Couldn't we constrain it at all? https://github.com/pyth-network/pyth-crosschain/blob/518aa8dea6a561ba4ee2aca9432d9df50c5afd32/lazer/publisher_sdk/proto/pyth_lazer_transaction.proto#L20-L30

PR #2754, opened 2025-06-03
feat(fortuna): refactor explorer endpoint & history query methods

cprussin:
## Summary

This commit implements a few fixes for the fortuna explorer endpoint, and in doing so I refactored the query methods.  We are now using a query builder approach which enables the filters, limits, offsets, etc to all work correctly regardless of the query supplied.

This commit also implements a new query param to filter on state.

## Rationale

The prior implementation of the history query methods did not properly support limit, offset, filters, etc on queries containing a search, which cause weird UX (I'd see results for a chain different than the one I selected, or searches based on sender would return too many results).  Implementing this correctly led to a ton of duplication, so it seemed to make sense to move to a query builder pattern here.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/migrations/20250605165549_re-add_tx_hash_indices.down.sql
+++ b/apps/fortuna/migrations/20250605165549_re-add_tx_hash_indices.down.sql
@@ -1,0 +1,4 @@
+-- Add down migration script here
+
+DROP INDEX request__request_tx_hash;
+DROP INDEX request__reveal_tx_hash;
--- a/apps/fortuna/migrations/20250605165549_re-add_tx_hash_indices.up.sql
+++ b/apps/fortuna/migrations/20250605165549_re-add_tx_hash_indices.up.sql
@@ -1,0 +1,4 @@
+-- Add up migration script here
+
+CREATE INDEX request__request_tx_hash ON request (request_tx_hash) WHERE request_tx_hash IS NOT NULL;
+CREATE INDEX request__reveal_tx_hash ON request (reveal_tx_hash) WHERE reveal_tx_hash IS NOT NULL;


apps/fortuna/.gitignore
```diff
@@ -2,3 +2,5 @@
 *config.yaml
 *secret*
 *private-key*
+.envrc
+fortuna.db
```
cprussin, line 6: @m30m are these safe to gitignore?  I noticed they weren't here before; I assume it's just an oversight but was there a good reason?

apps/fortuna/migrations/20250605004757_add_indices_for_advanced_search.up.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE INDEX request__sender__network_id__state__created_at ON request(sender, network_id, state, created_at);
+CREATE INDEX request__sender__network_id__created_at ON request(sender, network_id, created_at);
+CREATE INDEX request__sender__state__created_at ON request(sender, state, created_at);
+CREATE INDEX request__sender__created_at ON request(sender, created_at);
+CREATE INDEX request__sequence__network_id__state__created_at ON request(sequence, network_id, state, created_at);
+CREATE INDEX request__sequence__network_id__created_at ON request(sequence, network_id, created_at);
+CREATE INDEX request__sequence__state__created_at ON request(sequence, state, created_at);
+CREATE INDEX request__sequence__created_at ON request(sequence, created_at);
+CREATE INDEX request__state__created_at ON request(state, created_at);
+CREATE INDEX request__created_at ON request(created_at);
```
cprussin, line 22: I got these indexes by using [sqlite's cli expert](https://sqlite.org/cli.html#index_recommendations_sqlite_expert_).  I did some testing to try to collapse the indices down to a smaller set, but it seems that reordering the columns does not result in indices that are properly reused where you might expect (i.e. I thought that if I swapped `state` and `created_at` in the index on line 11, that it would result in not needing the index on line 12, but testing this and checking query plan explain messages indicates that both are needed and reordering just fails to index properly the query that needs the first case).

apps/fortuna/src/history.rs
```diff
@@ -486,6 +517,10 @@ mod test {
         }
     }
 
+    fn to_hex_string<T: ToHex>(val: T) -> String {
+        format!("0x{}", val.encode_hex::<String>())
```
m30m: do we need this function?
Also, can you add a test or two to make sure search handles all cases of a tx hash or address (without leading `0x` or upper case)

apps/fortuna/flake.nix
```diff
@@ -0,0 +1,63 @@
+{
```
m30m, line 1: I've never worked with nix. If this is an optional dev maybe add some documentation on README?

apps/fortuna/migrations/20250605004757_add_indices_for_advanced_search.down.sql
```diff
@@ -0,0 +1,22 @@
...
+DROP INDEX request__sequence__state__created_at;
+DROP INDEX request__sequence__created_at;
+DROP INDEX request__state__created_at;
+DROP INDEX request__created_at;
+
+
+CREATE INDEX idx_request_sequence ON request (sequence);
+CREATE INDEX idx_request_network_id_created_at ON request (network_id, created_at);
+CREATE INDEX idx_request_created_at ON request (created_at);
+CREATE INDEX idx_request_request_tx_hash ON request (request_tx_hash) WHERE request_tx_hash IS NOT NULL;
```
m30m, line 20: Are you sure this index was useless?

apps/fortuna/src/history.rs
cprussin: Ah no we don't any more, will clean this up & add the tests

apps/fortuna/flake.nix
cprussin: Yep, good call, will do.  Yes, optional for automatically configuring the dev env.

apps/fortuna/migrations/20250605004757_add_indices_for_advanced_search.down.sql
cprussin: What I did was to start by removing all indices and then work through every query combination with sqlite expert to generate the required list of indices to optimize all possible query combinations.  I believe that it's not that these indices were useless, but more that the ones that are useful are superseded by more useful / efficient ones.

PR #2759, opened 2025-06-05
Add pyth quorom server

danimhr:
This PR aims to add pyth quorom server


apps/quorom/src/server.rs
```diff
@@ -0,0 +1,133 @@
...
+#[allow(clippy::large_enum_variant)]
+pub struct RunOptions {
+    #[command(flatten)]
+    pub server: ServerOptions,
+
+    /// The URL of the Pythnet node to connect to.
+    #[arg(long = "pythnet-url")]
+    #[arg(env = "PYTHNET_URL")]
+    #[arg(default_value = "https://api2.pythnet.pyth.network")]
+    pub pythnet_url: String,
```
ali-behjati: let's use `Url` here

apps/quorom/src/server.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// The Wormhole pid on the Pythnet chain.
+    #[arg(long = "wormhole-pid")]
+    #[arg(env = "WORMHOLE_PID")]
+    pub wormhole_pid: String,
```
ali-behjati: you can use a Pubkey here as well.

apps/quorom/src/api.rs
```diff
@@ -0,0 +1,145 @@
...
+        .clone();
+
+    if signatures.len() >= (state.guardian_set.addresses.len() * 2) / 3 + 1 {
+        let vaa: Vaa<Payload> = (Header {
+            version: 1,
+            guardian_set_index: state.guardian_set_index,
+            signatures,
+        }, params.body.clone()).into();
+        if let Err(e) = state.ws.broadcast_sender.send(UpdateEvent::NewVaa(vaa)) {
+            tracing::error!(error = ?e, "Failed to broadcast new VAA");
```
ali-behjati: i think you only get error here when there is no receiver and we should ignore it.

apps/quorom/src/api.rs
```diff
@@ -0,0 +1,145 @@
...
+use axum::{routing::{get, post}, Json, Router};
+use secp256k1::{ecdsa::{RecoverableSignature, RecoveryId}, Message, Secp256k1};
+use serde::{Deserialize};
+use sha3::{Digest, Keccak256};
+use ::time::OffsetDateTime;
+use wormhole_sdk::{vaa::{Body, Header, Signature}, GuardianAddress, GuardianSetInfo, Vaa};
+
+use crate::{server::State, ws::{ws_route_handler, UpdateEvent}};
+
+const OBSERVATION_LIFETIME: u32 = 10; // In seconds
```
ali-behjati: Can we make this configurable? In any case there's a delay in the network i want to be able to tune it a bit.

apps/quorom/src/api.rs
```diff
@@ -0,0 +1,145 @@
...
+    if signatures.len() >= (state.guardian_set.addresses.len() * 2) / 3 + 1 {
+        let vaa: Vaa<Payload> = (Header {
+            version: 1,
+            guardian_set_index: state.guardian_set_index,
+            signatures,
+        }, params.body.clone()).into();
+        if let Err(e) = state.ws.broadcast_sender.send(UpdateEvent::NewVaa(vaa)) {
+            tracing::error!(error = ?e, "Failed to broadcast new VAA");
+        }
+        verification_writer.remove(&params.body);
```
ali-behjati: something that can happen here is that we wait for all these to finish when we shut down. i am somewhat surprised that axum graceful shutdown is not doing it.

ali-bahjati (approved):
LGTM. My comments are mostly minor but please address them before merging.

apps/quorom/src/api.rs
danimhr: Isnt it good to shout when there is no receiver?

apps/quorom/src/api.rs
danimhr: I used task tracker to make sure that we handle all observation before shutting down. Do you think it's not necessary to handle all observations before shutting down?

danimhr: This means we are not going to accept new requests, but we will wait until we handle all prv requests that we received

apps/quorom/src/api.rs
ali-behjati: hmmm i don't know whether it's an error. maybe as info?

apps/quorom/src/api.rs
ali-behjati: got it.

danimhr: When we receive a Ctrl-c signal, the WS connection is being dropped immediately, so I just removed the task tracker for simplicity.

I also tested graceful shutdown and it's working fine.

apps/quorom/src/api.rs
ali-behjati: @danimhr told me we always hold a receiver in wsstate so it's impossible for this to fail. resolving this.

PR #2760, opened 2025-06-05
feat(fortuna): add result count to results

cprussin:
## Summary

Add a total result count field to the fortuna logs response

## Rationale

This is necessary so we can implement a paginator on the UI since we need to know how many pages there are (or at the very least, we need to know if there's a next page).

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/apps/fortuna/src/history.rs
+++ b/apps/fortuna/src/history.rs
@@ -427,8 +427,34 @@ impl<'a> RequestQueryBuilder<'a> {
     }
 
     pub async fn execute(&self) -> Result<Vec<RequestStatus>> {
-        let mut query_builder =
-            QueryBuilder::new("SELECT * FROM request WHERE created_at BETWEEN ");
+        let mut query_builder = self.build_query("*");
+        query_builder.push(" LIMIT ");
+        query_builder.push_bind(self.limit);
+        query_builder.push(" OFFSET ");
+        query_builder.push_bind(self.offset);
+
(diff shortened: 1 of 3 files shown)


apps/fortuna/src/history.rs
```diff
@@ -928,4 +941,38 @@ mod test {
...
+            provider_random_number: None,
+        };
+        History::update_request_status(&history.pool, failed_status.clone()).await;
+
+        let results = history.query().count_results().await.unwrap();
+        assert_eq!(results, 3);
+
+        let results = history
+            .query()
+            .limit(1)
```
m30m, line 963: nice test

apps/fortuna/src/history.rs
```diff
@@ -427,8 +427,34 @@ impl<'a> RequestQueryBuilder<'a> {
...
+        query_builder.push(" LIMIT ");
+        query_builder.push_bind(self.limit);
+        query_builder.push(" OFFSET ");
+        query_builder.push_bind(self.offset);
+
+        let result: sqlx::Result<Vec<RequestRow>> =
+            query_builder.build_query_as().fetch_all(self.pool).await;
+
+        if let Err(e) = &result {
+            tracing::error!("Failed to fetch request by time: {}", e);
```
m30m: I think `by time` here remained from the old functions. Let's remove it

m30m (approved):

PR #2761, opened 2025-06-06
feat(fortuna): add prometheus metrics for history query latency

cprussin:
## Summary

Add prometheus histograms to measure query latency for explorer queries.

## Rationale

So that we can keep an eye on query performance in prod.

## How has this been tested?

- [ ] Added new tests


apps/fortuna/Cargo.toml
```diff
@@ -19,7 +19,7 @@ ethabi = "18.0.0"
 ethers = { version = "2.0.14", features = ["ws"] }
 futures = { version = "0.3.28" }
 hex = "0.4.3"
-prometheus-client = { version = "0.21.2" }
+prometheus-client = { version = "0.23.1" }
```
cprussin, line 22: This upgrade is required because in 0.21.2 there was no implementation of the `EncodeLabelSet` trait for `bool`, which I need

apps/fortuna/src/api/explorer.rs
```diff
@@ -1,17 +1,100 @@
...
+    results_latency: Family<QueryTags, Histogram>,
+    count_latency: Family<QueryTags, Histogram>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, EncodeLabelSet)]
+pub struct QueryTags {
+    search_type: Option<SearchType>,
+    has_network_id_filter: bool,
+    has_state_filter: bool,
+    limit: i64,
```
cprussin: Technically this is a bad idea because we don't want high-cardinality tags in prometheus BUT in this instance I think it's OK since we will only ever call this API with limits mapping to a very limited set of options in the page size selector in the frontend.

If people are concerned about this, however, I'm happy to remove it or map it to a type that has no way to increase in cardinality.

m30m: I prefer removing it for now. This is not an important parameter to identify slow running queries anyway and it has a max of 1000 which is small.

apps/fortuna/src/api/explorer.rs
```diff
@@ -1,17 +1,100 @@
...
+        });
+        sub_registry.register(
+            "results_latency",
+            "The latency of requests to the database to collect the limited results.",
+            results_latency.clone(),
+        );
+
+        let count_latency = Family::<QueryTags, Histogram>::new_with_constructor(|| {
+            Histogram::new(
+                [
```
m30m: I noticed you are using the `into_iter` here but removed it from the `keeper_metrics.rs` file because of `vec!` usage I guess. Let's use the same format and keep them consistent.

apps/fortuna/src/api/explorer.rs
```diff
@@ -1,17 +1,100 @@
...
+                    0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1.0, 2.0, 5.0, 10.0, 20.0,
```
m30m: let's put this in a constant or variable and use it for both metrics.

m30m (approved):

apps/fortuna/src/api/explorer.rs
cprussin: yeah I suppose that's fair, I was mostly thinking it would be useful if we wanted to compare relative query performance between different classes of queries but I suppose you're right that with 1000 as the max, the difference caused by adjusting the limit will be negligible.  Will remove.

apps/fortuna/src/api/explorer.rs
cprussin: Ah yes I removed it from `keeper_metrics.rs` because clippy caught it when I upgraded `prometheus-client`, but I modeled this code after `traced_client.rs` which also uses `into_iter` on an array.  I'll move both this and `traced_client.rs` to using `vec!` instead

apps/fortuna/src/api/explorer.rs
cprussin: Yep, good call, will do.  This also matches the bucket config in `traced_client.rs` so I'll reuse the same constant there too.

PR #2762, opened 2025-06-06
Entropy tester

m30m:
## Summary

Utility app to regularly check entropy contracts by making a request and waiting for callback.

Things left to do:
- [ ] Accept custom RPC url
- [ ] Add alerts

## Rationale

Additional reliability of fortuna

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/entropy-tester/src/index.ts
+++ b/apps/entropy-tester/src/index.ts
@@ -6,7 +6,6 @@ import { EvmEntropyContract } from "@pythnetwork/contract-manager/core/contracts
 import { DefaultStore } from "@pythnetwork/contract-manager/node/store";
 import type { Logger } from "pino";
 import { pino } from "pino";
-import type { ArgumentsCamelCase, InferredOptionTypes } from "yargs";
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
 import { z } from "zod";
@@ -158,13 +157,11 @@ export const main = function () {
     .parserConfiguration({
       "parse-numbers": false,
     })
-    .command({
-      command: "run",
-      describe: "run the tester until manually stopped",
-      builder: RUN_OPTIONS,
-      handler: async (
-        argv: ArgumentsCamelCase<InferredOptionTypes<typeof RUN_OPTIONS>>,
-      ) => {
+    .command(
+      "run",
+      "run the tester until manually stopped",
+      RUN_OPTIONS,
+      async (argv) => {
         const logger = pino();
         const configs = await loadConfig(argv.config);
         if (argv.validate) {
@@ -195,7 +192,7 @@ export const main = function () {
         });
         await Promise.all(promises);
       },
-    })
+    )
     .demandCommand()
     .help();
 };


apps/entropy-tester/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  rules: {
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
+  },
+};
```
cprussin: A few notes:

1. `.eslintrc.js` is [a deprecated config format](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated)
2. This is a pretty loose config that won't catch a ton of issues, including many issues that make the type system actually reliable and trustworthy.

The easiest fix is to replace this file with a file `eslint.config.js` that exports my canned config, e.g. [something like this](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/eslint.config.js).  Note you'll need to set `"type": "module"` in your `package.json` to use this as I expect esm for all packages I support.

apps/entropy-tester/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@pythnetwork/entropy-tester",
+  "version": "1.0.0",
+  "description": "Utility to test entropy provider callbacks",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
cprussin: Note that in modern js packages, it is recommended to use [subpath exports](https://nodejs.org/api/packages.html#subpath-exports) in addition to `main` and `types` at the root, as eventually subpath exports will be the only supported mechanism.

apps/entropy-tester/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "fix:format": "prettier --write \"src/**/*.ts\"",
+    "fix:lint": "eslint src/ --fix --max-warnings 0",
+    "test:format": "prettier --check \"src/**/*.ts\"",
+    "test:lint": "eslint src/ --max-warnings 0",
+    "start": "node lib/index.js",
+    "dev": "ts-node src/index.ts",
+    "prepublishOnly": "pnpm run build && pnpm run test:lint",
+    "preversion": "pnpm run test:lint",
+    "version": "pnpm run test:format && pnpm run test:lint && git add -A src"
+  },
```
cprussin: I suggest modeling your scripts after another app, e.g. https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/package.json#L9-L21 -- many of the scripts here don't make any sense for a package that isn't published like this and you're missing things like having prettier format all valid file types

apps/entropy-tester/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "devDependencies": {
+    "@types/ethereum-protocol": "^1.0.2",
+    "@types/express": "^4.17.21",
+    "@types/jest": "^27.4.1",
+    "@types/yargs": "^17.0.10",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
+    "@typescript-eslint/parser": "^6.0.0",
+    "eslint": "^8.13.0",
+    "jest": "^29.7.0",
```
cprussin: In general you should use `catalog:` dependencies for anything that's listed in [the workspace catalog](https://github.com/pyth-network/pyth-crosschain/blob/main/pnpm-workspace.yaml#L48-L152).  Doing so will enable me to maintain the versions as I upgrade shared monorepo dependencies.

At a minimum I would ask that you do this for code checking tools like eslint, jest, prettier, etc.

apps/entropy-tester/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "pino": "^9.2.0",
+    "prom-client": "^15.1.0",
+    "viem": "^2.19.4",
+    "yaml": "^2.1.1",
+    "yargs": "^17.5.1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "Apache-2.0",
+  "packageManager": "pnpm@10.7.0"
```
cprussin: This field is not necessary on a workspace package, it should only be specified at the root.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env node
```
cprussin: General advice but I usually separate scripts from implementations, similar to how you'd do in rust too.  here's a good example package where I've done this: https://github.com/dourolabs/fogo-web/tree/main/packages/snapshot.  You'll notice there's a simple javascript (not typescript) bin entry point that just imports the cli from a lib file, a `cli.ts` which configures yargs, and everything else is defined in standard library files.

Structuring things like this enables a few nice benefits:

1. By having the entry point be written as javascript, you avoid some issues where e.g. package managers will expect the bin entry point to exist at install time, not at build time
2. You get a nice lib for the implementations that you can reuse if you ever need it
3. The code is generally easier to structure, read, and predict

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+import fs from "fs";
+import pino, { Logger } from "pino";
+import { DefaultStore } from "@pythnetwork/contract-manager/node/store";
+import { EvmEntropyContract } from "@pythnetwork/contract-manager/core/contracts/evm";
+import {
+  PrivateKey,
+  toPrivateKey,
+} from "@pythnetwork/contract-manager/core/base";
+
+type DurationSeconds = number;
```
cprussin: Newtypes don't do anything in typescript and in fact add cognitive overhead.  For instance, this typechecks:

```ts
type DurationSeconds = number;
type SheepCountedDuringSleep = number;

const addOneSecond = (value: DurationSeconds) => value + 1;

const main = () => {
  const sheep: SheepCountedDuringSleep = 20;

  // ....

  const oneSecondLater = addOneSecond(sheep);
};
```

Because typescript does not actually treat these as newtypes and instead treat these as interchangeable aliases, they tend to lead to confusion and cognitive overhead, and situations where type aliases for compatible types are mixed and the code is hard to understand or predict.

For this reason, I generally advise against this pattern in typescript.  Better to make the code obvious than to use aliases that can get abused and just add mental leaps.

Note there are [tricks to making newtypes work](https://kubyshkin.name/posts/newtype-in-typescript/) the way you might expect in Typescript, however they aren't first-class native language features, and I'm dubious that they're worth the effort.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+      return value * 60;
+    case "s":
+      return value;
+    default:
+      throw new Error("Unsupported time unit.");
+  }
+}
+
+function loadConfig(configPath: string): LoadedConfig[] {
+  const configs = YAML.parse(fs.readFileSync(configPath, "utf-8"));
```
cprussin: If you can use json, I recommend it -- I know that yaml is nicer, but json enables using built in loading mechanisms like `async import` etc and you will generally find life easier in typescript if you just stick to json.

Even better would be to use something like https://github.com/cosmiconfig/cosmiconfig, which will support json OR yaml or even js files for config, and generally makes configuration work the way that people expect of most node tools.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+      throw new Error(
+        `Can not find the contract for chain ${config["chain-id"]}, check contract manager store.`,
+      );
+    }
+    if (contracts.length > 1) {
+      throw new Error(
+        `Multiple contracts found for chain ${config["chain-id"]}, check contract manager store.`,
+      );
+    }
+    loadedConfigs.push({ contract: contracts[0], interval });
```
cprussin: Doing:

```ts
const foo = [];
for (const bar of bars) {
  foo.push(getSomeValueFrom(bar));
}
```

is just a procedural equivalent to:

```ts
const foo = bars.map(bar => getSomeValueFrom(bar))
```

I generally recommend getting in the habit of the declarative map over the procedural loop

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+  logger.info(
+    { sequenceNumber, txHash: requestResponse.transactionHash },
+    `Request submitted`,
+  );
+
+  const startTime = Date.now();
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
cprussin, line 112: Don't do this, but for your entertainment, if you want some very esoteric node facts, you can actually do this:

```ts
import { promisify } from "util";

await promisify(setTimeout)(2000);
```

You wouldn't expect it to work due to the argument order of `setTimeout`, but this works because of some absolutely wild Node internals where the `promisify` function actually can return a completely separate implementation which is defined in the Node lib internals.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+    handler: async (argv: any) => {
+      const logger = pino();
+      const configs = loadConfig(argv.config);
+      if (argv.validate) {
+        logger.info("Config validated");
+        return;
+      }
+      const privateKey = toPrivateKey(
+        fs
+          .readFileSync(argv["private-key"], "utf-8")
```
cprussin: I suggest using `fs/promises` and making this async/await.  `readFileSync` is largely unnecessary now that async/await works, and using it at all is kind of a bad habit because it really only serves to mask the async nature of what you're doing, and in some cases it can unintentionally cause major issues because code gets deferred that doesn't need to be.

The only case I recommend using `readFileSync` or other such sync functions is if you absolutely must add code like this in a place where the code is not designed for async, and converting it would be a huge pain.  But here, that's not the case, you're already in an async function.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+        required: true,
+      },
+      "private-key": {
+        type: "string",
+        required: true,
+        description:
+          "Path to the private key to sign the transactions with. Should be hex encoded",
+      },
+    },
+    handler: async (argv: any) => {
```
cprussin: We should never use `any`; yargs will properly typecheck the arguments so the annotation shouldn't even be needed and adding it just cripples the type system.  `any` is a cancer and it will spread and cause the type checker to silently stop verifying a lot more of your code than you probably intend.

If you must, use `unknown`, but never use `any`.

apps/entropy-tester/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "declaration": true,
+    "rootDir": "src/",
+    "outDir": "./lib",
+    "strict": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true
+  },
+  "include": ["src"],
+  "exclude": ["node_modules", "**/__tests__/*"]
+}
```
cprussin, line 5: I suggest using [my canned tsconfig](https://github.com/cprussin/cprussin-monorepo/blob/main/packages/tsconfig/src/base.json) as a base here, as that will enable a lot of much stricter configs which make the type system much more reliable / trustworthy.  You can use it [like this](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/tsconfig.json) if you decide to do so

apps/entropy-tester/src/index.ts
m30m: this is going be read in runtime. Probably k8s will mount this file as a volume into the docker container. Does the argument still hold?

cprussin: Yeah, `async import` is specifically made for runtime imports, and cosmiconfig is the most common way to configure node stuff, so I'd still do it.  I don't think it's a big deal, but it's a good habit to get into and a better default for future code -- sticking to json or cosmiconfig (there's a few other similar config libs out there) will generally make your life easier when dealing with configuring node stuff.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+type ParsedConfig = {
+  "chain-id": string;
+  interval: string;
+};
+
+async function loadConfig(configPath: string): Promise<LoadedConfig[]> {
+  const configs = (await import(configPath, {
+    with: { type: "json" },
+  })) as { default: ParsedConfig[] };
```
cprussin: For all the same reasons you use serde to deserialize in Rust and you don't just unsafe cast external data, we should also avoid typecasting in typescript and should instead validate the schema using zod: https://v3.zod.dev/ (note we still use v3 everywhere in the monorepo but you're welcome to try v4 if you want).  This would look basically like this:

```ts
const configSchema = z.array(z.strictObject({
  "chain-id": z.string(),
  interval: z.string(),
}));

// ...

const configs = configSchema.parse(
  (await import(configPath, { with: { type: 'json' } })).default
);
```

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+      throw new Error(
+        `Can not find the contract for chain ${config["chain-id"]}, check contract manager store.`,
+      );
+    }
+    if (contracts.length > 1) {
+      throw new Error(
+        `Multiple contracts found for chain ${config["chain-id"]}, check contract manager store.`,
+      );
+    }
+    return { contract: contracts[0], interval };
```
cprussin: Marginally better here is to store `contracts[0]` in a variable and reuse that:

```ts
const firstContract = contracts[0];

if (contracts.length === 0 || !firstContract) {
  /* ... */
} else if (contracts.length > 1) {
  /* ... */
} else {
  return { contract: firstContract, interval }
}
```

The reason this is preferable is that typescript can guarantee statically with this code that `firstContract` is not `undefined` in the return branch, but it cannot guarantee the same with `contracts[0]` due to how typescript narrowing is designed.

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  )) as {
+    transactionHash: string;
+    events: {
+      RequestedWithCallback: {
+        returnValues: {
+          sequenceNumber: string;
+        };
+      };
+    };
+  };
```
cprussin: Same comment -- don't typecast, instead parse with zod.  Typecasting is unsafe and it breaks all the type safety the type checker attempts to provide.

cprussin (approved):
Main piece of feedback is to switch to using zod to validate rather than using typecasting for external data, but other than that this LGTM

apps/entropy-tester/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  yargs(hideBin(process.argv))
+    .parserConfiguration({
+      "parse-numbers": false,
+    })
+    .command({
+      command: "run",
+      describe: "run the tester until manually stopped",
+      builder: RUN_OPTIONS,
+      handler: async (
+        argv: ArgumentsCamelCase<InferredOptionTypes<typeof RUN_OPTIONS>>,
```
cprussin: Hmm yargs should be able to infer this type automatically, are you not finding that to be the case?

apps/entropy-tester/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "@cprussin/tsconfig/nextjs.json",
+  "include": ["svg.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
```
cprussin: You should remove the first and last option here give you aren't in a next app so there's no `.next` folder at all, and you have no svgs so the first one doesn't make sense

apps/entropy-tester/src/index.ts
m30m: correct

PR #2766, opened 2025-06-09
feat(entropy-explorer): bind to live apis

cprussin:
## Summary

Set up entropy explorer to use live apis

## How has this been tested?

- [ ] Added new tests


apps/entropy-explorer/src/errors.ts
```diff
@@ -2,52 +2,79 @@ export const ERROR_DETAILS = {
...
+  ],
+  "0xdf51c431": [
+    "NoSuchProvider",
+    "The requested provider does not exist.",
+    "https://docs.pyth.network/entropy/contract-addresses",
+  ],
+  "0xc4237352": [
+    "NoSuchRequest",
+    "The specified request does not exist.",
+    "https://docs.pyth.network/entropy/best-practices#handling-callback-failures",
```
tejasbadadare, line 20: This will only happen in rare instances where keeper A tries to fulfill a request that keeper B has already successfully fulfilled. So the fulfillment actually succeeded but it looks like a failure from the perspective of one keeper, which updates the status. This shouldn't happen anymore due to an extra check we do to ensure the request still exists, but in case it does due to RPC weirdness or a race condition, we may want to update the help text to say something like "The specified request does not exist, likely because the request was already fulfilled."

tejasbadadare: If it proves to still be a problem I can just not update the history when we encounter a NoSuchRequest error.

cprussin:
> Looks good! (But i'm no TS expert.) Any way we can see the explorer plugged into staging data?

@tejasbadadare yes there will be once #2932 is merged & released!  Once that lands I'll rebuild this PR and share out the preview link

apps/entropy-explorer/src/errors.ts
cprussin: Totally your call -- you tell me what makes more sense.  I can just get rid of it from here and we can fall back on the default message too.

PR #2772, opened 2025-06-10
feat(entropy-tester): add docker image workflow

m30m:
## Summary

To be used in deployments

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/.github/workflows/docker-entropy-tester.yml
+++ b/.github/workflows/docker-entropy-tester.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
     paths:
       - "apps/entropy-tester/**"
-      - ".github/workflows/entropy-tester/docker-entropy-tester.yml"
+      - ".github/workflows/docker-entropy-tester.yml"
       - "Dockerfile.node"
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-entropy-tester.yml
```diff
@@ -0,0 +1,50 @@
...
+    paths:
+      - "apps/entropy-tester/**"
+      - ".github/workflows/docker-entropy-tester.yml"
+      - "Dockerfile.node"
+  workflow_dispatch:
+    inputs:
+      dispatch_description:
+        description: "Dispatch description"
+        required: true
+        type: string
```
cprussin, line 16: Why is this argument here?

m30m: I copy pasted it from other configs. I believe it is for manual dispatch of the job which can be useful 🤔

cprussin: I don't think you need the inputs, I think you just need `workflow_dispatch` with no children.  The `inputs` creates a UI where you can fill in values when doing the manual dispatch, but you aren't actually using the input anywhere so I think it should be removed

PR #2773, opened 2025-06-11
fix(entropy-tester): adjust cli execution and tsconfigs

m30m:
## Summary
tsconfig was not working as I hoped for.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/entropy-tester/package.json
+++ b/apps/entropy-tester/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/entropy-tester",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Utility to test entropy provider callbacks",
   "type": "module",
   "main": "dist/index.js",
@@ -22,7 +22,8 @@
     "test:format": "prettier --check .",
     "test:lint": "eslint . --max-warnings 0",
     "test:types": "tsc",
-    "start": "tsc && node cli/run.js"
+    "dev:start": "tsc && node cli/run.js",
+    "start": "node cli/run.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


apps/entropy-tester/package.json
```diff
@@ -22,7 +22,8 @@
     "test:format": "prettier --check .",
     "test:lint": "eslint . --max-warnings 0",
     "test:types": "tsc",
-    "start": "tsc && node cli/run.js"
+    "dev:start": "tsc && node cli/run.js",
```
cprussin: This is named `start:dev` (and the other one is named `start:prod`) in other services

apps/entropy-tester/tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
-  "extends": "@cprussin/tsconfig/nextjs.json",
-  "include": ["**/*.ts", "**/*.tsx"],
+  "extends": "@cprussin/tsconfig/base.json",
+  "compilerOptions": {
```
cprussin: Generally, I will separate out things so that the base tsconfig is just used to check types (i.e. `noEmit: true`), and I have an extended tsconfig for actually building.  See https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/tsconfig.build.json for an example, and https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/package.json#L22 for the corresponding build script (this one is a bit weird because it's running react compiler but in your case you can just omit the `babel` command).

I prefer doing things this way because it allows you to run just a type check as part of CI without actually building.  It's a bit pedantic but I do find it nice.

cprussin (approved):

PR #2774, opened 2025-06-11
feat: wormhole receiver contract for stylus

ayushboss:
## Summary

Implementing a Wormhole receiver contract for the stylus framework. The receiver currently handles parsing + verifying VAAs and will also be able to handle guardian set management.

## Rationale

A Wormhole receiver is necessary for ultimately implementing a Pyth receiver in Stylus. 

## How has this been tested?

- [x] Added new tests

I'm building a test suite for the stylus contract through normal cargo tests. At the moment those tests are focused on VAA verification, more tests will follow for guardian set management soon. I'll be expanding those tests and also running more complete stylus tests once I have everything loaded.


target_chains/stylus/contracts/pyth/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "pyth-contract"
```
tejasbadadare: Can we use a more specific name here like "pyth-stylus-receiver"?

target_chains/stylus/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[workspace]
+members = [
+    "contracts/wormhole",
+    "contracts/pyth",
+    "contracts/sample",
+]
+resolver = "2"
+
+[workspace.package]
+edition = "2021"
```
tejasbadadare: If we're specifying `channel = "1.87.0"` in rust-toolchain.toml, we can bump `edition` to "2024".

target_chains/stylus/contracts/wormhole/rust-toolchain.toml
tejasbadadare, line 1: move this to the workspace level?

target_chains/stylus/src/main.rs
tejasbadadare, line 1: Let's remove the placeholder files and dirs until we need them (`src/`, `sdk/`, etc)

target_chains/stylus/contracts/wormhole/licenses/Apache-2.0
tejasbadadare, line 1: are these licenses necessary to include here? we have a blanket LICENSE (apache 2.0) at the repo root

target_chains/stylus/contracts/wormhole/.cargo/config.toml
tejasbadadare, line 1: could you include a brief note to explain why this is necessary? if it's specific to the stylus env, a link to the docs would be sufficient

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct GuardianSignature {
+    pub guardian_index: u8,
+    pub signature: FixedBytes<65>,
+}
+
+#[derive(Clone, Debug)]
+pub struct VerifiedVM {
```
tejasbadadare, line 38: this data structure looks like a VAA header + VAA body, what is VerifiedVM?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    AlreadyInitialized,
+    NotInitialized,
+    InvalidInput,
+    InsufficientSignatures,
+    InvalidGuardianIndex,
+    InvalidAddressLength,
+    VerfiyVMError,
+}
+
+impl From<WormholeError> for Vec<u8> {
```
tejasbadadare, line 75: why return error strings as Vec<u8>?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    consumed_governance_actions: StorageMap<Vec<u8>, StorageBool>,
+    initialized: StorageBool,
+    guardian_set_sizes: StorageMap<U256, StorageUint<256, 4>>,
+    guardian_set_expiry: StorageMap<U256, StorageUint<256, 4>>,
+    guardian_keys: StorageMap<U256, StorageAddress>,
+}
+
+#[public]
+impl WormholeContract {
+    pub fn initialize(
```
tejasbadadare, line 123: can anyone call initialize?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    ) -> Result<(), Vec<u8>> {
+        if self.initialized.get() {
+            return Err(WormholeError::AlreadyInitialized.into());
+        }
+
+        if initial_guardians.is_empty() {
+            return Err(WormholeError::InvalidInput.into());
+        }
+
+        self.current_guardian_set_index.set(U256::from(0u32));
```
tejasbadadare: why not use 0 directly here?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    fn governance_contract(&self) -> Address;
+    fn submit_new_guardian_set(&mut self, encoded_vm: Vec<u8>) -> Result<(), WormholeError>;
+}
+
+#[entrypoint]
+#[storage]
+pub struct WormholeContract {
+    current_guardian_set_index: StorageUint<256, 4>,
+    chain_id: StorageUint<256, 4>,
+    governance_chain_id: StorageUint<256, 4>,
```
tejasbadadare, line 112: what does StorageUint<256, 4> mean? (i assume 256 is the data size in bytes)

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+            return Err(WormholeError::NotInitialized.into());
+        }
+
+        match self.get_guardian_set_internal(index) {
+            Some(guardian_set) => Ok(guardian_set.keys.abi_encode()),
+            None => Err(WormholeError::InvalidGuardianSetIndex.into()),
+        }
+    }
+
+    pub fn parse_and_verify_vm(&self, encoded_vm: Vec<u8>) -> Result<Vec<u8>, Vec<u8>> {
```
tejasbadadare: why `parse_and_verify_vm`? i see that this is called `parseAndVerifyVAA` in other EVM wormhole impls

tejasbadadare: please add documentation to the public methods (probably in `IWormhole`)

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    governance_chain_id: StorageUint<256, 4>,
+    governance_contract: StorageAddress,
+    consumed_governance_actions: StorageMap<Vec<u8>, StorageBool>,
+    initialized: StorageBool,
+    guardian_set_sizes: StorageMap<U256, StorageUint<256, 4>>,
+    guardian_set_expiry: StorageMap<U256, StorageUint<256, 4>>,
+    guardian_keys: StorageMap<U256, StorageAddress>,
+}
+
+#[public]
```
tejasbadadare, line 121: nit: this is a preference thing, but i think having one `impl` block for the contract and individually marking functions as `#[public]` is more clear

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+    fn quorum(num_guardians: u32) -> u32 {
+        (num_guardians * 2) / 3 + 1
+    }
+}
+
+impl WormholeContract {
+    fn parse_vm(&self, encoded_vm: &[u8]) -> Result<VerifiedVM, WormholeError> {
+        Self::parse_vm_static(encoded_vm)
+    }
```
tejasbadadare: why do we need a non-static version of this?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+    fn parse_vm_static(encoded_vm: &[u8]) -> Result<VerifiedVM, WormholeError> {
```
tejasbadadare: please add docs describing the data layout (can just a be link to the appropriate VAA docs) since we are hand rolling the data

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+            encoded_vm[cursor + 2],
+            encoded_vm[cursor + 3],
+        ]);
+        cursor += 4;
+
+        let len_signatures = encoded_vm[cursor];
+        cursor += 1;
+
+        let mut signatures = Vec::new();
+        for _ in 0..len_signatures {
```
tejasbadadare: let's validate len_signatures doesn't exceed the maximum before we start iterating through them

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        let sequence = u64::from_be_bytes([
+            encoded_vm[cursor],
+            encoded_vm[cursor + 1],
+            encoded_vm[cursor + 2],
+            encoded_vm[cursor + 3],
+            encoded_vm[cursor + 4],
+            encoded_vm[cursor + 5],
+            encoded_vm[cursor + 6],
+            encoded_vm[cursor + 7],
+        ]);
```
tejasbadadare: why not `let sequence = u64::from_be_bytes(encoded_vm[cursor..cursor + 8]`?

you can also use the `byteorder` crate (this is what the [pythnet_sdk](https://github.com/pyth-network/pyth-crosschain/blob/955a805e85dc093b9fff0d3f6af1268052ca358c/pythnet/pythnet_sdk/src/wire/de.rs) uses)

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    fn verify_vm(&self, vm: &VerifiedVM) -> Result<(), WormholeError> {
+        let guardian_set = self.get_guardian_set_internal(vm.guardian_set_index)
+            .ok_or(WormholeError::InvalidGuardianSetIndex)?;
+
+        if vm.guardian_set_index != self.current_guardian_set_index.get().try_into().unwrap_or(0u32)
+            && guardian_set.expiration_time > 0 {
+                return Err(WormholeError::GuardianSetExpired)
+        }
+
+        let required_signatures = Self::quorum(guardian_set.keys.len() as u32);
```
tejasbadadare: let's avoid use of `as` (see [Rust code guidelines](https://www.notion.so/pyth-network/Rust-code-guidelines-20a2eecaaac98060b6b2d9a52c9f9141?source=copy_link)) (applies to other places in this PR too)

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+        if vm.signatures.len() < required_signatures as usize {
+            return Err(WormholeError::InsufficientSignatures);
+        }
+
+        let mut last_guardian_index: Option<u8> = None;
+
+        for signature in &vm.signatures {
+            if let Some(last_index) = last_guardian_index {
```
tejasbadadare, line 343: nit: rename to prev_guardian_index to avoid confusion

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+            }
+
+            let guardian_address = guardian_set.keys[signature.guardian_index as usize];
+            let hashed_vm_hash: FixedBytes<32> = FixedBytes::from(keccak256(vm.hash));
+
+            match self.verify_signature(&hashed_vm_hash, &signature.signature, guardian_address) {
+                Ok(true) => {},
+                Ok(false) => return Err(WormholeError::InvalidSignature.into()),
+                Err(e) => return Err(e),
+            }
```
tejasbadadare: does it need to return `Result<bool, WormholeError>`? The `Ok(false)` case is just another error case, you could simplify the signature to `Result<(), WormholeError>`

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+            let key = self.compute_guardian_key(set_index, i as u8);
+            self.guardian_keys.setter(key).set(*guardian);
+        }
+
+        Ok(())
+    }
+
+    fn verify_signature(
+        &self,
+        hash: &FixedBytes<32>,
```
tejasbadadare, line 397: would be helpful to have a more specific name for future readers, like `vaa_body_hash`

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        if signature.len() != 65 {
+            return Err(WormholeError::InvalidSignature);
+        }
+
+        let secp = Secp256k1::new();
+
+        let recid = RecoveryId::try_from(signature[64] as i32)
+            .map_err(|_| WormholeError::InvalidSignature)?;
+        let recoverable_sig = RecoverableSignature::from_compact(&signature[..64], recid)
+            .map_err(|_| WormholeError::InvalidSignature)?;
```
tejasbadadare: nice use of `map_err` :)

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: also, there are subtleties about what exactly comprises the digest (once-hashed or twice-hashed body, it would be good to be clear about it: https://wormhole.com/docs/protocol/infrastructure/vaas/#signatures)

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    use secp256k1::{ecdsa::{RecoverableSignature, RecoveryId}, Secp256k1, Message, SecretKey, PublicKey};
+    use stylus_sdk::alloy_primitives::keccak256;
+    use base64::engine::general_purpose;
+    use base64::Engine;
+
+    const CHAIN_ID: u16 = 60051;
+    const GOVERNANCE_CHAIN_ID: u16 = 1;
+    const GOVERNANCE_CONTRACT: U256 = U256::from_limbs([4, 0, 0, 0]);
+
+    fn test_real_vaa() -> Vec<u8> {
```
tejasbadadare: i guess these are fixtures of sorts? if so, i'd recommend renaming them to `mock_` since `test_` typically indicates a test case

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        ]).expect("32-byte secret key is valid")
+    }
+
+    fn test_guardian_address1() -> Address {
+        let secret = test_guardian_secret1();
+        let secp = Secp256k1::new();
+        let pubkey = PublicKey::from_secret_key(&secp, &secret);
+        let pubkey_uncompressed = pubkey.serialize_uncompressed();
+        let hash = Keccak256::digest(&pubkey_uncompressed[1..]); // Skip 0x04
+        let address_bytes: [u8; 20] = hash[12..].try_into().unwrap(); // Last 20 bytes
```
tejasbadadare: why skip and slice these specific bytes?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+    fn guardian_set4() -> Vec<Address> {
+        vec![
+            Address::from_str("0x5893B5A76c3f739645648885bDCcC06cd70a3Cd3").unwrap(),
+            Address::from_str("0xfF6CB952589BDE862c25Ef4392132fb9D4A42157").unwrap(),
+            Address::from_str("0x114De8460193bdf3A2fCf81f86a09765F4762fD1").unwrap(),
+        ]
+    }
+
+    fn current_guardians() -> Vec<Address> {
```
tejasbadadare: these will change over time, and shouldn't be functionally different than the other guardian_set fixtures you've got here -- would recommend renaming it `mock_guardian_set13`. 

also in general i'd recommend against calling certain tests "real" (are those tests considered better/more realistic than the non-real ones?) all fixtures should ideally be representative of real world inputs.

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    }
+
+    #[motsu::test]
+    fn test_real_wormhole_vaa_parsing() {
+        let vaa_vec = test_real_vaa();
+        let result = WormholeContract::parse_vm_static(&vaa_vec).unwrap();
+        assert_eq!(result.signatures.len(), 13)
+    }
+
+    #[motsu::test]
```
tejasbadadare, line 809: what does motsu give us? have just seen `cargo test` used in our other projects

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    fn test_verify_vm_invalid_guardian_index() {
+        let contract = deploy_with_test_guardian();
+        let signatures = vec![
+            create_guardian_signature(5),
+        ];
+        let vm = create_test_vm(0, signatures);
+
+        let result = contract.verify_vm(&vm);
+        assert!(matches!(result, Err(WormholeError::InvalidGuardianIndex)));
+    }
```
tejasbadadare: can we also test this with multiple guardian sets loaded into storage?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        let expected = U256::from_be_bytes(keccak256(&key_data).0);
+
+        let key_data2 = [&set_index.to_be_bytes()[..], &[guardian_index]].concat();
+        let result2 = U256::from_be_bytes(keccak256(&key_data2).0);
+
+        assert_eq!(expected, result2);
+    }
+
+    #[motsu::test]
+    fn test_multiple_guardian_sets() {
```
tejasbadadare, line 1078: great test coverage overall!!

tejasbadadare (commented):
Great work! Lots of comments, but most are minor and organizational. Do all the tests pass?

target_chains/stylus/contracts/pyth/Cargo.toml
tejasbadadare: on second thought, we can also just remove this crate until we're ready to implement it to keep this PR atomic

target_chains/stylus/contracts/wormhole/src/main.rs
ali-behjati, line 1: is this needed?

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+#[cfg(not(feature = "std"))]
+#[global_allocator]
+static ALLOC: mini_alloc::MiniAlloc = mini_alloc::MiniAlloc::INIT;
+
+// #[cfg(all(not(feature = "std"), not(test)))]
+// #[panic_handler]
+// fn panic(_info: &core::panic::PanicInfo) -> ! {
+//     loop {}
+// }
```
ali-behjati: what's the context behind this :? shall we remove it?

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati, line 1: as an overal comment i recommend breaking this down to multiple modules.

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: maybe vec<u8> is what stylus accepts

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+    pub nonce: u32,
+    pub emitter_chain_id: u16,
+    pub emitter_address: FixedBytes<32>,
+    pub sequence: u64,
+    pub consistency_level: u8,
+    pub payload: Vec<u8>,
+    pub hash: FixedBytes<32>,
+}
+
+#[derive(Debug, Clone)]
```
ali-behjati: not sure whether it's supported in stylus or not, but thiserror can help to give you Display for the error.

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: damn. this struct stores u64[4] inside instead of u8[256] to probably be more optimized. the 4 is a constant and is checked at compile time to make sure it's the right number for 256 bits.

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: this is the name in EVM implementations and I recommend sticking to the same name. ideally we want anyone from EVM to be able to interact with this contract just as it was the Pyth/Wormhole EVM contract.

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        self.store_guardian_set(0, initial_guardians, 0)?;
+
+        self.initialized.set(true);
+        Ok(())
+    }
+
+    pub fn get_current_guardian_set_index(&self) -> Result<u32, Vec<u8>> {
+        if !self.initialized.get() {
+            return Err(WormholeError::NotInitialized.into());
+        }
```
ali-behjati: it's kind of strange that we need to check it everywhere

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        if cursor + 51 > encoded_vm.len() {
+            return Err(WormholeError::InvalidVMFormat);
+        }
+
+        let timestamp = u32::from_be_bytes([
+            encoded_vm[cursor],
+            encoded_vm[cursor + 1],
+            encoded_vm[cursor + 2],
+            encoded_vm[cursor + 3],
+        ]);
```
ali-behjati: can't this be like `let timestamp = u32::from_be_bytes(encoded_vm.get(cursor..cursor+4)).ok_or(WormholeError::InvalidVMFormat)`

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+        for _ in 0..len_signatures {
+            if cursor + 66 > encoded_vm.len() {
+                return Err(WormholeError::InvalidVMFormat);
+            }
+
+            let guardian_index = encoded_vm[cursor];
+            cursor += 1;
+
+            let mut signature_bytes = [0u8; 65];
+            signature_bytes.copy_from_slice(&encoded_vm[cursor..cursor + 65]);
```
ali-behjati: generally instead of direct memory access you can use `.get()` that returns an Option and you can convert it to error and return if it's empty. this will be safer than manually checking size above and having a direct access.

target_chains/stylus/contracts/wormhole/rust-toolchain.toml
ayushboss: We need this file here for stylus to be able to build

target_chains/stylus/contracts/wormhole/.cargo/config.toml
ayushboss: This came as a part of the initialization process when I started the stylus project. I'll look around to see if there's any formal reference of this file in the docs

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Yeah this is a holdover from porting over from starknet, where the VAAs were seemingly referred to as VM (verified message ig?). I'll go ahead and refactor

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Stylus doesn't support the string type, so if there are any public facing functions that will be called in the context of a stylus contract, I read that it's supported to return bytes.

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: At the moment yes, although it can only be initialized once. Is there a good way to specify a restricted group of people that can initialize it?

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: I updated this to the preferred method for handling stylus errors, so this is no longer a concern

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Yeah, I set this up pretty early on before I was sure about how much space I needed. Gonna try to scale this down though, I figure we probably don't need 256 bytes

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: I had a static function in there just to make testing easier and less state dependent. Do you think I should get rid of that?

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Still getting used to Rust, didn't realize i could use comprehension like that. Will update

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: I tried this, but I couldn't get this to work since I think `#[public]` is a stylus modifier that is only possible on `impl` blocks. I'll keep looking into this

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Is 19 in general the maximum? Or do we need to store a custom maximum for each?

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: motsu is a package that mocks up the requirements of the stylus environment while doing unit tests

target_chains/stylus/contracts/wormhole/src/main.rs
ayushboss: I will need it for deployment, I am still figuring out exactly the best way to handle it though.

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Have separated the tests out into a different module on a different PR, will also think about the best way to branch these out more

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Is there a better way to do this? My other thought was to initialize the contract if we run into it when it's not initialized.

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Yeah, didn't know about this until Tejas mentioned it to me, should make things cleaner.

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: We can follow the same convention as EVM as Ali mentioned

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: Typically the deployer is set as the initial owner for EVM deployments that use the [upgradeable proxy pattern](https://docs.openzeppelin.com/upgrades-plugins/proxies). Although i'm not sure what the standard is on stylus. Happen to know @ali-behjati ?

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: Got it, no worries if that's the standard in stylus

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: I agree, wasn't sure why we needed the variant with the `&self` param :) A free function like you already have would work well for this

target_chains/stylus/contracts/wormhole/src/lib.rs
tejasbadadare: Yeah 19 guardians looks like [an invariant](https://wormhole.com/docs/protocol/infrastructure/guardians/#:~:text=set%20of%2019%20distributed%20nodes), but i guess they could theoretically change this down the line? maybe ignore my initial suggestion here to be safe.

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: There are good type aliases that you can use though. like StorageUint256

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: this is how EVM address is calculated from a public key (last 20 bytes of the keccak hash)

ali-behjati (commented):
Overall very nice work here. I recommend you look into wormhole crates that offer a lot of these logic for you. Have you seen [this](https://docs.rs/wormhole-vaas/latest/wormhole_vaas/)?

target_chains/stylus/contracts/wormhole/src/lib.rs
ayushboss: Yeah, the main idea is that the "real" tests were based on the actual guardian set and actual VAAs I pulled from wormhole, versus the other messages being ones I mocked up

target_chains/stylus/Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
 panic = "abort"
 
 [profile.release]
-panic = "abort"
+opt-level = "z"           # Optimize for size instead of speed
+lto = "fat"              # More aggressive link-time optimization
+strip = "symbols"        # Strip all symbols including debug
+panic = "abort"          # Smaller panic handling
+codegen-units = 1        # Better optimization (slower build, smaller binary)
+overflow-checks = false  # Disable overflow checks for size
```
tejasbadadare, line 30: Do we need to disable overflow-checks? Otherwise, nice work on trimming the fat

target_chains/stylus/contracts/wormhole/contract_tests/test_parse_and_verify_vm.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python3
+"""
+Python script to test parse_and_verify_vm function on deployed Stylus Wormhole contract
+Requires: pip install web3 base64
+"""
```
tejasbadadare, line 5: Just a tip - you can use [`uv`](https://docs.astral.sh/uv/guides/scripts/) for these one-off scripts. It follows a declarative approach to dependencies and creates ephemeral virtual envs for deterministic execution. You can define the dependencies in-line and run with `uv run`.

target_chains/stylus/contracts/wormhole/contract_tests/initialize_contract.sh
tejasbadadare, line 1: Not sure how much value this script provides over just listing the init and verify commands in a README. especially since the commands may take multiple retries (RPCs can be flaky)

target_chains/stylus/contracts/wormhole/contract_tests/test_parse_and_verify_vm.py
tejasbadadare, line 1: i think this belongs in pyth_crosschain/contract_manager (you can see the existing scripts there for our EVM contracts)

ali-behjati: hmmm it kind of seems like a real test so i'm not sure. 

@ayushboss this is like an integration test? Is there anyway to test it locally rather than using arb sepolia?

target_chains/stylus/contracts/wormhole/src/lib.rs
ali-behjati: I don't know either. It's not in the scope of this PR but @ayushboss we should see how upgradability works here.

target_chains/stylus/contracts/wormhole/src/tests.rs
ali-behjati, line 1: the best practice in rust is that the unit tests be in the same file as where the logic is (and i think these should go there). the integration tests however go outside src directory (in tests directory). see [this](https://doc.rust-lang.org/book/ch11-03-test-organization.html)

target_chains/stylus/contracts/wormhole/src/tests.rs
```diff
@@ -0,0 +1,624 @@
...
+            .unwrap();
+        let set4 = contract.get_gs_internal(4)
+            .unwrap();
+
+        assert_eq!(set0.keys, guardian_set0());
+        assert_eq!(set4.keys, guardian_set4());
+    }
+
+    #[motsu::test]
+    fn test_verify_vm_with_valid_signatures() {
```
ali-behjati: can you test using a duplicate signature and make sure it fails?

target_chains/stylus/contracts/wormhole/src/types.rs
ali-behjati, line 1: I know I've asked you to break these into multiple files. but seeing that only `lib.rs` uses them i take back my words and ask to put them all in a single place. Ideally if we want to break it down it should be that we break down the functionalities into multiple modules.

target_chains/stylus/contracts/wormhole/contract_tests/test_parse_and_verify_vm.py
ayushboss: yeah this is an integration test. i was mostly using it as a way to confirm the contract was hooked up for myself, but i found that stylus has a testing framework of its own. ill pivot to using that for more detailed integration tests.

target_chains/stylus/contracts/wormhole/src/lib.rs
```diff
@@ -230,12 +296,14 @@ impl WormholeContract {
 
     fn verify_vm(&self, vaa: &VerifiedVM) -> Result<(), WormholeError> {
 
+        // println!("checkpoint 1: {:?}", vaa);
```
tejasbadadare: likely forgotten comment

tejasbadadare (approved):
Looks good to me, great work! I guess we will have to put WH governance into its own contract to stay within binary size limits. Please get @ali-behjati 's approval before merging, thanks!

target_chains/stylus/contracts/wormhole/src/tests.rs
ayushboss: Just a note on this, due to some of the size restrictions I had issues with trying to separate the integration tests -- essentially when I separated them, under the structure of stylus 0.6.0, I could not run them with a normal cargo test. Upgrading to stylus 0.9.0 would allow me to do more separated integration tests, but they pushed the size of the build above the 24 kb limit, and I couldn't get it cut down after a few days of trying. I have that work stored on a side branch, and I'll revisit it when I come back to wormhole for governance.

PR #2775, opened 2025-06-11
Add pythnet quorum listere to hermes

danimhr:


apps/hermes/server/src/network/pythnet.rs
```diff
@@ -432,10 +436,93 @@ where
...
+    let mut responded_to_ping = true; // Start with a true to not close the connection immediately
+    let request = pythnet_quorum_ws_endpoint.into_client_request()?;
+    let (mut ws_stream, _) = connect_async(request).await?;
+
+    loop {
+        tokio::select! {
+            message = ws_stream.next() => {
+                let vaa_bytes = match message.ok_or_else(|| anyhow!("PythNet quorum stream terminated."))?? {
+                    Message::Frame(_) => continue,
+                    Message::Text(message) => message.try_to_vec().unwrap(),
```
ali-behjati: let's handle it gracefully.

apps/hermes/server/src/network/pythnet.rs
```diff
@@ -432,10 +436,93 @@ where
...
         task_listener,
         task_guardian_watcher,
-        task_price_feeds_metadata_updater
+        task_price_feeds_metadata_updater,
+        task_quorum_listener,
     );
     Ok(())
 }
+
+const QUORUM_PING_INTERVAL: Duration = Duration::from_secs(10);
```
ali-behjati, line 476: since the server is doing the ping I'm not sure that it's necessary.

apps/hermes/server/src/network/pythnet.rs
```diff
@@ -432,10 +436,93 @@ where
...
+
+    loop {
+        tokio::select! {
+            message = ws_stream.next() => {
+                let vaa_bytes = match message.ok_or_else(|| anyhow!("PythNet quorum stream terminated."))?? {
+                    Message::Frame(_) => continue,
+                    Message::Text(message) => message.try_to_vec().unwrap(),
+                    Message::Binary(bytes) => bytes.to_vec(),
+                    Message::Ping(_) => {
+                        let _ = ws_stream.send(Message::Pong(vec![].into())).await;
```
ali-behjati: Tungstenite does this automatically itself.

apps/quorum/src/api.rs
ali-behjati, line 1: bump version here too?

danimhr:
@ali-bahjati I ran Hermes, Pythnet-watcher, and quorum locally with a single guardian inside the guardian set, and override the guardian set list on my local with that specific guardian. Then connect these three services and created a log to see if Hermes could process VAA completely till the end or not.

apps/hermes/server/src/network/pythnet.rs
danimhr: What will happen if Hermes server gets disconnected from network?

ali-behjati: hmm yeah that's a fair point.

PR #2777, opened 2025-06-11
doc, refactor(apps/hermes/server): add code guidelines, enable extra clippy lints

Riateche:
## Summary

We want to add better linting throughout our codebase. I also put together [this guide](https://www.notion.so/pyth-network/Rust-code-guidelines-20a2eecaaac98060b6b2d9a52c9f9141) to help deal with these new lints.

## Rationale

To improve code quality: avoid panics, overflows and other suspicious code patterns.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/hermes/server/Cargo.toml
+++ b/apps/hermes/server/Cargo.toml
@@ -74,3 +74,46 @@ panic                  = 'abort'
 
 [profile.dev]
 panic                  = 'abort'
+
+
+[lints.rust]
+unsafe_code = "deny"
+
+[lints.clippy]
+# See https://www.notion.so/pyth-network/Rust-code-guidelines-20a2eecaaac98060b6b2d9a52c9f9141
+
+wildcard_dependencies = "deny"
+
+collapsible_if = "allow"
+collapsible_else_if = "allow"
+
+allow_attributes_without_reason = "warn"
+
+# Panics
+expect_used = "warn"
+fallible_impl_from = "warn"
+indexing_slicing = "warn"
+panic = "warn"
+panic_in_result_fn = "warn"
+string_slice = "warn"
+todo = "warn"
+unchecked_duration_subtraction = "warn"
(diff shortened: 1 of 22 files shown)


apps/hermes/server/src/state/aggregate.rs
```diff
@@ -663,7 +674,12 @@ where
         ));
     } else {
         // Use the publish time from the first end message
-        end_messages[0].message.publish_time() - window_seconds as i64
+        end_messages
+            .first()
+            .context("no messages found")?
+            .message
+            .publish_time()
+            - i64::try_from(window_seconds).context("window size overflow")?
```
ali-behjati, line 682: can we ban unchecked math operations?

apps/hermes/server/Cargo.toml
```diff
@@ -74,3 +74,46 @@ panic                  = 'abort'
 
 [profile.dev]
 panic                  = 'abort'
+
+
+[lints.rust]
+unsafe_code = "deny"
+
+[lints.clippy]
+# See https://www.notion.so/pyth-network/Rust-code-guidelines-20a2eecaaac98060b6b2d9a52c9f9141
```
tejasbadadare: should we have a link to our private notion in this public repo? maybe you can publish the notion doc as a webpage and link that if you want to include it?

tejasbadadare (approved):
thank you!

apps/hermes/server/src/state/aggregate.rs
Riateche: We should do it but I think it's better to do it a bit later because it requires more work to do it properly.

apps/hermes/server/Cargo.toml
Riateche: Added guides directly to the crosschain repo for better visibility.

PR #2780, opened 2025-06-12
feat(fortuna): better retry mechanism

m30m:
## Summary

If we deploy multiple instances of fortuna, keepers will compete to fulfil requests and only one of them will succeed making the callback. The other instances will continue retrying for 5 minutes and by default this can take up to 13 retries. Knowing that this will happen for every request, the RPC usage will increase substantially which is not acceptable.

This PR fixes this by:
1. Expose the underlying errors in an inspectable manner instead of putting everything inside anyhow
2. Expose an error mapper which can customise the error returned before retrying. Using this mechanism, we customise the `retry_interval` and the number of retries.

Another nice side-effect here is that we get better error msgs for explorer.

## Rationale

To avoid excessive RPC usage

## How has this been tested?

- [ ] Added new tests

Ran two instances of Fortuna locally, created 10 requests on monad-testnet and verified the behavior.

### DIFF
--- a/apps/fortuna/src/eth_utils/utils.rs
+++ b/apps/fortuna/src/eth_utils/utils.rs
@@ -4,11 +4,15 @@ use {
     backoff::ExponentialBackoff,
     ethabi::ethereum_types::U64,
     ethers::{
-        contract::ContractCall,
+        contract::{ContractCall, ContractError},
         middleware::Middleware,
-        types::{TransactionReceipt, U256},
+        providers::ProviderError,
+        types::{transaction::eip2718::TypedTransaction, TransactionReceipt, U256},
     },
-    std::sync::{atomic::AtomicU64, Arc},
+    std::{
+        fmt::Display,
+        sync::{atomic::AtomicU64, Arc},
+    },
     tokio::time::{timeout, Duration},
     tracing,
 };
@@ -156,7 +160,9 @@ pub async fn submit_tx_with_backoff<T: Middleware + NonceManaged + 'static>(
     call: ContractCall<T, ()>,
     gas_limit: U256,
     escalation_policy: EscalationPolicy,
-) -> Result<SubmitTxResult> {
+    error_mapper: impl Fn(u64, backoff::Error<SubmitTxError<T>>) -> backoff::Error<SubmitTxError<T>>
+        + Copy,
+) -> Result<SubmitTxResult, SubmitTxError<T>> {
     let start_time = std::time::Instant::now();
 
     tracing::info!("Started processing event");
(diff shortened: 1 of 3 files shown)


apps/fortuna/src/keeper/process_event.rs
```diff
@@ -160,16 +190,45 @@ pub async fn process_event_with_backoff(
                 .get_request(event.provider_address, event.sequence_number)
                 .await;
 
-            tracing::error!("Failed to process event: {:?}. Request: {:?}", e, req);
```
m30m, line 163: I moved this inside because it was creating a bunch of false alarms.

apps/fortuna/src/eth_utils/utils.rs
```diff
@@ -176,14 +182,15 @@ pub async fn submit_tx_with_backoff<T: Middleware + NonceManaged + 'static>(
...
+            let result = submit_tx(
                 middleware.clone(),
                 &call,
                 padded_gas_limit,
                 gas_multiplier_pct,
                 fee_multiplier_pct,
             )
-            .await
+            .await;
+            result.map_err(|e| error_mapper(num_retries, e))
```
jayantk: (and then you don't need to pass `error_mapper`)

apps/fortuna/src/eth_utils/utils.rs
```diff
@@ -156,7 +160,9 @@ pub async fn submit_tx_with_backoff<T: Middleware + NonceManaged + 'static>(
     call: ContractCall<T, ()>,
     gas_limit: U256,
     escalation_policy: EscalationPolicy,
-) -> Result<SubmitTxResult> {
+    error_mapper: impl Fn(u64, backoff::Error<SubmitTxError<T>>) -> backoff::Error<SubmitTxError<T>>
```
jayantk: can you leave a comment that this lets you customize the backoff behavior based on the error type? It's not obvious what you get from this at the moment

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -74,12 +76,40 @@ pub async fn process_event_with_backoff(
...
+                // Slow down the retries if the request is not found.
+                // This probably means that the request is already fulfilled via another process.
+                // After 5 retries, we return the error permanently.
+                if num_retries >= 5 {
+                    return backoff::Error::Permanent(err);
+                }
+                if 1 < num_retries && num_retries < 5 {
+                    return backoff::Error::Transient {
+                        err,
+                        retry_after: Some(Duration::from_secs(60)),
```
jayantk: i think the spacing here needs to be a bit more granular. retry the first time after 5 seconds, then 10 seconds, then 60 seconds

these errors happen pretty frequently in the first 1-2 seconds because of RPC async issues. The current logic will significantly degrade the UX whenever this happens, because the callback will take 60 seconds now.

m30m: this already kicks in on the 3rd attempt, but I will also increase the delay on the first 2 attempts.

apps/fortuna/src/eth_utils/utils.rs
m30m: I didn't understand this comment

PR #2782, opened 2025-06-13
fix(contract-manager): remove fs from base

cprussin:
## How has this been tested?

- [ ] Added new tests


cprussin:
Closing this as I did some testing and there's still a lot of node dependent code that's pulled in within the `core` part of contract manager.  We will need to do some more significant code restructuring.

PR #2784, opened 2025-06-14
Add verifier to pyth evm contract

danimhr:
## Summary

This PR introduces support for a backup verifier in the Pyth EVM contract to enhance reliability when verifying VAAs.

## Rationale

Pyth currently verifies VAAs exclusively using Wormhole. This creates a single point of failure: if Wormhole becomes unavailable or its guardians are unable to sign messages, Pyth will be unable to verify updates, potentially causing disruptions.

To address this, we’ve integrated a backup verifier into the contract. If Wormhole fails, the new verifier—with its own guardian set—can independently approve and verify VAAs. As a result, Pyth can continue to function even if Wormhole is down. Only if both the Wormhole and backup verifier guardian sets are unavailable will VAA verification become impossible.

This change improves the resilience and fault tolerance of the system without affecting current functionality.

## How has this been tested?

- [X] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
@@ -37,7 +37,15 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
         {
             bool valid;
             (vm, valid, ) = wormhole().parseAndVerifyVM(encodedVm);
-            if (!valid) revert PythErrors.InvalidWormholeVaa();
+            if (!valid) {
+                if (address(verifier()) == address(0)) {
+                    revert PythErrors.InvalidUpdateData();
+                }
+                (vm, valid, ) = verifier().parseAndVerifyVM(encodedVm);
+                if (!valid) {
+                    revert PythErrors.InvalidUpdateData();
+                }
+            }
         }
 
         if (!isValidDataSource(vm.emitterChainId, vm.emitterAddress))
--- a/target_chains/ethereum/contracts/contracts/pyth/PythGovernanceInstructions.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/PythGovernanceInstructions.sol
@@ -37,7 +37,8 @@ contract PythGovernanceInstructions {
         SetWormholeAddress, // 6
         SetFeeInToken, // 7 - No-op for EVM chains
         SetTransactionFee, // 8
-        WithdrawFee // 9
+        WithdrawFee, // 9
+        SetVerifierAddress // 10
     }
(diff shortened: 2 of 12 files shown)


target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -90,13 +110,14 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
...
                 offset += 1;
 
-                // We use another offset for the trailing header and in the end add the
-                // offset by trailingHeaderSize to skip the future headers.
-                //
-                // An example would be like this:
-                // uint trailingHeaderOffset = offset
-                // uint x = UnsafeBytesLib.ToUint8(accumulatorUpdate, trailingHeaderOffset)
-                // trailingHeaderOffset += 1
+                uint trailingHeaderOffset = offset;
```
ali-behjati, line 113: shall we change `MINIMUM_ALLOWED_MINOR_VERSION`?

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
```diff
@@ -90,13 +110,14 @@ abstract contract PythAccumulator is PythGetters, PythSetters, AbstractPyth {
...
+                signer = Signer(
```
ali-behjati, line 114: i've lost my solidity knowledge here, what would happen if this byte does not fit in Signer (which accepts only 0 and 1). let's check it if it's an undefined behaviour.

target_chains/ethereum/contracts/contracts/pyth/PythGovernanceInstructions.sol
```diff
@@ -37,7 +37,8 @@ contract PythGovernanceInstructions {
         SetWormholeAddress, // 6
         SetFeeInToken, // 7 - No-op for EVM chains
         SetTransactionFee, // 8
-        WithdrawFee // 9
+        WithdrawFee, // 9
+        SetVerifierAddress // 10
```
ali-behjati, line 41: note for the future: we need to support it in governance sdk in TS

target_chains/ethereum/contracts/contracts/pyth/PythAccumulator.sol
danimhr: Why do we need to change that?

ali-behjati: it is a metadata to tell us on which version we are right now and reject messages with older minor version. if we don't check it and this gets upgraded before Hermes there might be problems.

PR #2785, opened 2025-06-14
chore(api-reference) Add Contract Manager Support

aditya520:
## How has this been tested?

- [ ] Added new tests


apps/api-reference/src/components/EvmProvider/index.tsx
```diff
@@ -19,9 +19,11 @@ const CHAINS = NETWORK_IDS.map((id) =>
...
 const TRANSPORTS = Object.fromEntries(
   CHAINS.map((chain) => {
-    const url = getRpcUrl(chain.id);
-    if (url) {
-      return [chain.id, http(url)];
+    const rpcUrl = getRpcUrl(chain.id);
+    if (rpcUrl) {
+      return [chain.id, http(rpcUrl)];
+    } else if (chain.rpcUrls.default.http[0]) {
+      return [chain.id, http(chain.rpcUrls.default.http[0])];
```
cprussin: This logic should live in `getRpcUrl` -- it doesn't make sense that we have a function that should get the rpc URL which only sometimes returns the RPC url and other times there is still a valid URL but the consumer needs to know about different logic to retrieve it.

apps/api-reference/src/evm-networks.ts
```diff
@@ -1,11 +1,47 @@
+import * as EvmChains from "@pythnetwork/contract-manager/data/chains/EvmChains.json";
+import * as EvmPriceFeedContracts from "@pythnetwork/contract-manager/data/contracts/EvmPriceFeedContracts.json";
+
+type EvmChains = {
+  id: string;
+  mainnet: boolean;
+  rpcUrl: string;
+  networkId: number;
+  type: string;
+  nativeToken?: string;
```
cprussin: I don't think you need to worry about adding types for fields that aren't in use with this app.  Typescript objects are not closed (meaning they allow undeclared fields to be present), and only including fields you actually need allows typescript to be more helpful if we wanted to e.g. restructure the shape of `EvmChains` (typescript would be able to tell us if we've restructured things in a way that's breaking anything in use)

apps/api-reference/src/evm-networks.ts
```diff
@@ -1,11 +1,47 @@
...
+};
+
+const getPriceFeedContractAddress = (chain: EvmChains) => {
+  const contractAddress = Object.values(EvmPriceFeedContracts).find(
+    (contract) => contract.chain === chain.id,
+  );
+  return contractAddress?.address as `0x${string}`;
```
cprussin: We should not be typecasting.  For one, this is going to incorrectly mask the `undefined` case, and for two, if an invalid string were entered into `EvmPriceFeedContracts`, the typesystem would be incorrect and runtime would not operate in a way that matches what the type checker is guaranteeing.

Instead, it would be better to explicitly verify and throw exceptions if the data fails invariants you expect, e.g. something along these lines:

```ts
if (contractAddress === undefined) {
  throw new Error(`No contract found for ${chain.id}`);
} else if (isZeroXString(contractAddress.address)) {
  return contractAddress.address;
} else {
  throw new Error(`Invalid contract address for ${chain.id}: ${contractAddress.address}`);
}

...

const isZeroXString = (val: string): val is `0x${string}` => val.startsWith("0x");
```

apps/api-reference/src/evm-networks.ts
```diff
@@ -1,11 +1,47 @@
...
+};
+
+const mapEvmChainsToNetworkInfo = (chains: EvmChains[]) => {
+  const networkInfo: Record<number, NetworkInfo> = {};
+
+  for (const chain of chains) {
```
cprussin: A few notes:

1. This doesn't seem to make a lot of sense to me.  Why not just get rid of `NETWORK_INFO` entirely and just use `EvmChains` directly in the consuming files?  This looks like it's just added indirection that just attempts to keep the `NETWORK_INFO` structure the same as it was before, but the indirection doesn't add any other value and it shouldn't be hard to refactor things to get rid of this file entirely (the type checker should be very helpful here)
2. If you're going to do this, a more declarative / functional way is to use `Object.fromEntries`:

```ts
const mapEvmChainsToNetworkInfo = (chains: EvmChains[]) =>
  Object.fromEntries(chains.map(chain => [
    Number.parseInt(chain.networkId, 10),
    {
      name: chain.id,
      rpcUrl: chain.rpcUrl,
      isMainnet: chain.mainnet,
      contractAddress: getPriceFeedContractAddress(chain)
    }
  ]));
```

3. as a minor nit, you should use `Number.parseInt` to convert string to int (I think that's what's happening here), not `Number()`, as the `Number` constructor is ambiguous and unclear (e.g. it's not clear to me from reading the code what conversion is actually happening)

apps/api-reference/src/evm-networks.ts
```diff
@@ -1,11 +1,47 @@
...
+      isMainnet: chain.mainnet,
+      contractAddress: getPriceFeedContractAddress(chain),
+    };
+  }
+
+  return networkInfo;
+};
+
+// Convert EvmChains to array format
+const evmChainsArray = Object.values(EvmChains) as EvmChains[];
```
cprussin: Please don't typecast -- if this does not automatically unify then it means something needs to be verified in order for the typesystem to be reliable.  But I think if you follow my advice of doing away with `NETWORK_INFO` entirely, this will go away.

cprussin (changes requested):
I have a few minor suggestions, but to me the most important thing is that I really don't think `NETWORK_INFO` makes sense when using contract manager, so I don't think it's wise to approach this by trying to rebuild it to match the old shape.  Instead, my suggestion is to get rid of `NETWORK_INFO` entirely, and to update the call sites to use contract manager directly.

apps/api-reference/src/components/EvmProvider/index.tsx
```diff
@@ -1,33 +1,40 @@
...
+// const chainEntries = await Promise.all(
+//   CHAINS.map(async (chain) => {
+//     const url = await getEvmChainRpcUrl(chain.id);
+//     return url ? [chain.id, http(url)] : undefined;
+//   }),
+// );
+
+// const TRANSPORTS = Object.fromEntries(
+//   chainEntries.filter((entry): entry is [number, ReturnType<typeof http>] => entry !== undefined)
+// ) as Record<number, ReturnType<typeof http>>;
```
cprussin: Is this comment relevant?

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
+import Web3 from 'web3';
+import evmChainsData from '../../store/chains/EvmChains.json';
+import evmPriceFeedContractsData from '../../store/contracts/EvmPriceFeedContracts.json';
+import evmWormholeContractsData from '../../store/contracts/EvmWormholeContracts.json';
+import * as chains from "viem/chains";
+
+export const getEvmContractAddress = (chainId: number, contractType: string): `0x${string}` => {
```
cprussin: The interface of this is awkward; `contractType` being a freeform string is both not discoverable (what contract types are valid?) and error prone (what if I accidentally type in `pricefeed` instead of `priceFeed`?  Typescript won't catch this).

You can make this better by using a union (i.e. `contractType: 'wormhole' | 'priceFeed'`), but IMO that's still an awkward interface.

I would suggest you just make two separate functions:

```ts
export const getEvmPriceFeedContractAddress = (chainId: number): `0x${string}` | undefined => { ... }
export const getEvmWormholeContractAddress = (chainId: number): `0x${string}` | undefined => { ... }
```

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  const chain = evmChainsData.find((c) => c.networkId === chainId);
+  if (!chain) {
+    throw new Error(`Chain with network ID ${chainId} not found`);
```
cprussin: We shouldn't be throwing all these exceptions here.  Exceptions should be reserved for exceptional situations -- e.g. out of memory, or some other invariant that failed.  This may be an invariant for the API reference app (i.e. you know in that app that all chain IDs you pass in are valid) but it may not be a valid invariant for all usage of this util (e.g. what if we built an app where the chain ID was a user input?)

There are better ways to signal that the network ID is not valid which do not require raising exceptions and can be validated by the type system.  In particular, change the return type of this function to `` `0x${string}` | undefined`` and return `undefined` here.

Move the exception enforcing the invariant to the app where it is an invariant, not the util where it may not be invariant.

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  
+  if (contractType === 'priceFeed') {
+    const contract = evmPriceFeedContractsData.find((c) => c.chain === chain.id);
+    if (!contract) {
+      throw new Error(`Price feed contract not found for chain ${chainId}`);
```
cprussin: Is this a true invariant?  i.e. are we guaranteed that there will be a contract for all valid chain ids?  I believe this is invariant, but is there ever a case where we'd add a chain to contract manager without a corresponding contract?

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!contract) {
+      throw new Error(`Wormhole contract not found for chain ${chainId}`);
+    }
+    return contract.address as `0x${string}`;
+  }
+  
+  throw new Error(`Unknown contract type: ${contractType}`);
+}
+
+export const getAllEvmChainsIds = () => evmChainsData.map((c) => c.networkId);
```
cprussin: There's no reason to make this a closure, export this as a const.  This takes no arguments and is constant, the closure just means extra cycles every time you use it.

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return contract.address as `0x${string}`;
+  }
+  
+  if (contractType === 'wormhole') {
+    const contract = evmWormholeContractsData.find((c) => c.chain === chain.id);
+    if (!contract) {
+      throw new Error(`Wormhole contract not found for chain ${chainId}`);
+    }
+    return contract.address as `0x${string}`;
```
cprussin: Don't typecast -- typescript should be able to infer this automatically from the json, if it can't infer it automatically then you either need to validate it (e.g. with `startsWith("0x")` and a type guard) or you need to widen the return type of the function

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+//   const viemChain = Object.values(chains).find((c: any) => c.id === chain.id);
+//   if (viemChain && viemChain.rpcUrls && viemChain.rpcUrls.default) {
+//     const viemRpcUrl = viemChain.rpcUrls.default.http[0];
+//     if (await checkRpcUrl(viemRpcUrl)) {
+//       return viemRpcUrl;
+//     }
+//   }
+  
+//   throw new Error(`No working RPC URL found for chain ${chainId}`);
+// }
```
cprussin: Is this comment still relevant?

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+
+export const getEvmChainRpcUrl = (chainId: number) => {
+  const chain = evmChainsData.find((c) => c.networkId === chainId);
+  if (!chain) {
+    throw new Error(`Chain with network ID ${chainId} not found`);
```
cprussin: Same comment -- this should not throw.  It may be invariant in some cases, but in others it may not be.  Don't throw except for things that are truly invariants or exceptional scenarios.

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+    const viemRpcUrl = viemChain.rpcUrls.default.http[0];
+    return viemRpcUrl;
+  } 
+
+  // Now let's try to use the json rpc url
+  if (chain.rpcUrl) {
+    return chain.rpcUrl;
+  }
+  
+  throw new Error(`No working RPC URL found for chain ${chainId}`);
```
cprussin: Same thing -- we should not throw here.  Signal this by returning `undefined`.

cprussin (changes requested):
Looks much better but I think we need to change some of the interfaces a bit if we're going to make it a shared util in a common lib.

contract_manager/src/utils/utils.ts
aditya520: Yes, it is possible that we have a chain with no relevant corresponding contract. For example, If we only deploy on lazer or entropy, we will have that chain id in this list, but the price feed contract address.

cprussin: In that case, this is not invariant and it's not appropriate to raise an exception.  Instead, you should return `undefined`.

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    return undefined;
+  }
+  const contract = evmWormholeContractsData.find((c) => c.chain === chain.id);
+  if (!contract?.address || !contract.address.startsWith("0x")) {
+    return undefined;
+  }
+  return contract.address as `0x${string}`;
+};
+
+export const getAllEvmChainsIds: number[] = evmChainsData.map(
```
cprussin: Minor nit but this should be named `allEvmChainIds`, not `getAllEvmChainsIds`

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+  const chain = evmChainsData.find((c) => c.networkId === chainId);
+  if (!chain) {
+    return undefined;
+  }
+  const contract = evmWormholeContractsData.find((c) => c.chain === chain.id);
+  if (!contract?.address || !contract.address.startsWith("0x")) {
+    return undefined;
+  }
+  return contract.address as `0x${string}`;
+};
```
cprussin: This can be simplified:

```suggestion
export const getEvmPriceFeedContractAddress = (chainId: number) =>
  getContractAddress(chainId, evmPriceFeedContractsData);

export const getEvmWormholeContractAddress = (chainId: number) =>
  getContractAddress(chainId, evmWormholeContractsData);

const getContractAddress = (chainId: number, contractsData: { chain: string, address: string }[]) => {
  const chain = evmChainsData.find(({ networkId }) => networkId === chainId);
  if (chain === undefined) {
    return undefined;
  } else {
    const contract = contractsData.find(contract => contract.chain === chain.id);
    if (contract?.address === undefined) {
      return undefined;
    } else if (isZeroXString(contract.address)) {
      return contract.address;
    } else {
      throw new Error(`Invariant failed: invalid contract address ${contract.address} for chain ${contract.chain}`);
    }
  }
}

const isZeroXString = (str: string): str is `0x${string}` => str.startsWith("0x");
```

Note in particular a few features:

1. Removed code duplication
2. Use of [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) for enforcing the dependent `0x${string}` type, rather than using a typecast
3. Because we said that all contract addresses should always start with `0x`, anything in contract manager that does not is invalid config data.  Thus, this is an invariant failure and we should know about it as soon as possible (and have systems in place that make it impossible).  Therefore, it is appropriate to throw so that it does not go undetected.
4. Removed early returns (I personally prefer to avoid early returns for the exact same reasons why I avoid goto statements, they break control flow and mean the code cannot be analyzed simply based on the control structure -- I've seen a lot of code that grows over time and early returns make grokking the code hard because it masks the actual control structure of the code).
5. Using explicit `value === undefined` rather than `!value`; the latter may not behave how you want in all cases since [falsiness in javascript](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) is more broad and sometimes leads to cases that are broken in ways that are very nonobvious, e.g. if a `0` is ever passed.  It doesn't affect anything in this specific code, but it's good habit to get into (I often catch myself using `!value` too so it takes discipline)
6. Switched `if` clauses to be positives (i.e. `thing === value` as opposed to `thing !== value`) which generally reads more naturally IMO

contract_manager/src/utils/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    return viemChain.rpcUrls.default.http[0];
+  }
+
+  // Now let's try to use the json rpc url
+  if (chain.rpcUrl) {
+    return chain.rpcUrl;
+  }
+
+  return undefined;
+};
```
cprussin: Minor nits but for similar reasons as above, I'd write this as:

```suggestion
export const getEvmChainRpcUrl = (chainId: number) => {
  const chain = evmChainsData.find(({ networkId }) => networkId === chainId);
  if (chain === undefined) {
    return undefined;
  } else {
    const viemChain = Object.values(chains).find(({ id }) =>
      id === Number.parseInt(chain.id, 10)
    );
    return viemChain?.rpcUrls?.default?.http?.[0] ?? chain.rpcUrl;
  }
};
```

Things to note here:

1. Using `Number.parseInt` instead of `Number()`.  Generally, I recommend `parseInt` over `Number` because it's more explicit (e.g. what does `Number(something)` actually mean?  With `parseInt(something)`, it's very obvious that I'm parsing a string into an int)
2. Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) rather than explicit checks, which makes the code far more terse
3. As above, removes early returns

cprussin (approved):
Main things look good, but I have suggestions for how to improve the implementation

PR #2792, opened 2025-06-19
feat(stylus) - Set up pyth receiver contract for stylus

ayushboss:
## Summary

Built on the wormhole stylus contract, this PR establishes the framework for the Pyth contract and implements the get prices functions.

## Rationale

These are the first steps to building out the Pyth stylus receiver contract.

## How has this been tested?

- [ ] Added new tests

Formal tests of the getter function will depend on further implementation. This will be added on in future PRs. the existing program builds at the moment.


target_chains/stylus/contracts/pyth-receiver/src/structs.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub publish_time: StorageU64,
+    pub expo: StorageI32,
+    pub price: StorageI64,
+    pub conf: StorageU64,
+    pub ema_price: StorageI64,
+    pub ema_conf: StorageU64,
+}
+
+// PriceInfo struct storing price information
+pub type PriceInfoReturn = (U64, I32, I64, U64, I64, U64);
```
ali-behjati, line 17: I forgot why we couldn't make this a type. it makes it very difficult to understand the response.

ali-behjati (commented):
sorry i'm confused. what's the difference between this and #2793? i think that one might have been on top of this?

PR #2794, opened 2025-06-20
feat(stylus) - get prices pyth contract functions

ayushboss:
## Summary

Implemented the get prices functions of the pyth stylus contract.

## Rationale

Important for getting most recent prices, and has an important function definition for making sure that an update is no older than a certain specified age (is_no_older_than). 

## How has this been tested?

- [ ] Added new tests

Get prices can only be evaluated when other parts of the contract are completed (i.e. initialization). Until then, cargo build is the main system by which I'm verifying things work.

### DIFF
--- a/target_chains/stylus/contracts/pyth-receiver/src/lib.rs
+++ b/target_chains/stylus/contracts/pyth-receiver/src/lib.rs
@@ -36,15 +36,19 @@ impl PythReceiver {
     pub fn get_price_unsafe(&self, _id: [u8; 32]) -> Result<PriceInfoReturn, PythReceiverError> {
         let id_fb = FixedBytes::<32>::from(_id);
         
-        let price_info = self.latest_price_info.get(id_fb).unwrap_or(PythReceiverError::PriceUnavailable)?;
+        let price_info = self.latest_price_info.get(id_fb);
+        
+        if price_info.publish_time.get() == U64::ZERO {
+            return Err(PythReceiverError::PriceUnavailable);
+        }
 
         Ok((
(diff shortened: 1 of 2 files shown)


ali-behjati (approved):

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
```diff
@@ -31,12 +33,31 @@ pub struct PythReceiver {
...
+            price_info.price.get(),
+            price_info.conf.get(),
+            price_info.ema_price.get(),
+            price_info.ema_conf.get(),
+        ))
     }
 
-    pub fn get_price_no_older_than(&self, _id: [u8; 32], _age: u64) -> PriceInfoReturn {
-        (U64::ZERO, I32::ZERO, I64::ZERO, U64::ZERO, I64::ZERO, U64::ZERO)
+    pub fn get_price_no_older_than(&self, _id: [u8; 32], _age: u64) -> Result<PriceInfoReturn, PythReceiverError> {
```
tejasbadadare, line 55: Not necessary right now (you can do this in a future PR,) but let's move the public components that users will consume like structs and errors into an SDK. That way they can import just the public SDK to consume prices and use the contracts. The private impls will stay in the contract, but will implement the public SDK interfaces. [Example here.](https://github.com/pyth-network/pyth-crosschain/blob/7642a395b965dfbfbf40ea0e1fac462d7016526f/target_chains/ethereum/sdk/solidity/README.md)

tejasbadadare (approved):

PR #2804, opened 2025-06-24
Add metrics over http urls

danimhr:


apps/quorum/src/server.rs
```diff
@@ -66,7 +79,17 @@ lazy_static! {
...
-    pub static ref EXIT: watch::Sender<bool> = watch::channel(false).0;
+
+    // Make sure at least one receiver is created
+    static ref EXIT_CHANNEL: (watch::Sender<bool>, watch::Receiver<bool>) = watch::channel(false);
+    pub static ref EXIT: watch::Sender<bool> = EXIT_CHANNEL.0.clone();
+}
+
+pub async fn wait_for_exit() {
+    let mut rx = EXIT.subscribe();
+    if !*rx.borrow() {
```
ali-behjati: can you add a bit of comment here? or maybe change it to `if *rx.borrow == false`?

apps/quorum/src/server.rs
```diff
@@ -66,7 +79,17 @@ lazy_static! {
     /// - The `Receiver` side of a watch channel performs the detection based on if the change
     ///   happened after the subscribe, so it means all listeners should always be notified
     ///   correctly.
-    pub static ref EXIT: watch::Sender<bool> = watch::channel(false).0;
+
+    // Make sure at least one receiver is created
```
ali-behjati: this seems a bit hacky. can't we call `send_modify` here instead to make sure value changes even without receivers?

ali-behjati (commented):
Looks good!

apps/quorum/src/server.rs
danimhr: using `*rx.borrow == false` would cause some lint issues. Added some comments

ali-behjati (approved):

PR #2806, opened 2025-06-24
feat(fortuna): json format on prod

m30m:
## Summary

Uses another log format when not running in shell

## Rationale

Some alerts are not working properly because the error log is multiline and we can't capture the chain_id of the error log

## How has this been tested?

- [ ] Added new tests

This is exactly how hermes sets up the logging.


tejasbadadare (approved):
thx! beats having to regexp it in grafana haha

PR #2809, opened 2025-06-25
Quorum: Add ws and other metrics

danimhr:


apps/quorum/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "quorum"
-version = "0.2.0"
+version = "0.3.0"
```
ali-behjati: this is not a breaking change is it?

apps/quorum/src/server.rs
```diff
@@ -181,6 +181,15 @@ pub async fn run(run_options: RunOptions) -> anyhow::Result<()> {
             run_options.clone(),
             state.clone()
         )),
+        metric_collector("state".to_string(), || {
+            let state = state.clone();
+            async move {
+                let verification = state.verification.read().await;
+                metrics::gauge!("state_verifications_total").set(verification.len() as f64);
```
ali-behjati: this is not total, it's the size (which might go up and down) so please rename it.

p.s: it took me a while to make sure what `verification` is (it's nor the type were not descriptive). i recommend either renaming it or adding some documentation.

ali-behjati (commented):
I'm generally a bit on the fence with the `metrics` library because it's very flexible and if a metric is used in two places, there's chances of error (typo/...) but i can tolerate it.

Apart from these metrics, i like to see these two metrics too:
- number of active subscribers (so we can see if there's a decrease)
- number of observations sent per signer (so we can detect if a signer goes down)

apps/quorum/Cargo.toml
danimhr: It's not a breaking change

danimhr:
> I'm generally a bit on the fence with the `metrics` library because it's very flexible and if a metric is used in two places, there's chances of error (typo/...) but i can tolerate it.
> 
> Apart from these metrics, i like to see these two metrics too:
> 
> * number of active subscribers (so we can see if there's a decrease)
> * number of observations sent per signer (so we can detect if a signer goes down)
 
I agree with your points. I chose to use the `metrics` crate because it's compatible with the Axum Prometheus. Also, the Quorum project is small, I think the current setup is fine for now.
We can define metric names as enum as well, if you think having it is necessary. Let me know if we have to add it before merge.

We can caculate the number of active subscribers by looking at the total number of ping updates sent by the server, which is why I didn’t add any other metric for active subscribers.

I’ll add the "observations per signer" metric in the upcoming commits.

apps/quorum/Cargo.toml
ali-behjati: well in this case (0.x.y) the convention is to use x as breaking and y as minor or patch.

PR #2810, opened 2025-06-25
Hermes: Add error log when receiving invalid quorum VAA

danimhr:
This PR aims to add error logs if the VAA received by PythNet quorum service was invalid.

### DIFF
--- a/apps/hermes/server/src/network/pythnet.rs
+++ b/apps/hermes/server/src/network/pythnet.rs
@@ -512,7 +512,7 @@ where
                     let state = state.clone();
                     async move {
                         if let Err(e) = state.process_message(vaa_bytes).await {
-                            tracing::debug!(error = ?e, "Skipped VAA.");
+                            tracing::error!(error = ?e, "Received an invalid VAA from PythNet quorum.");
                         }
                     }
                 });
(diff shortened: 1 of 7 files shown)


apps/hermes/server/src/network/pythnet.rs
```diff
@@ -512,7 +512,7 @@ where
                     let state = state.clone();
                     async move {
                         if let Err(e) = state.process_message(vaa_bytes).await {
-                            tracing::debug!(error = ?e, "Skipped VAA.");
+                            tracing::error!(error = ?e, "Received an invalid VAA from PythNet quorum.");
```
ali-behjati, line 517: I didn't notice quorum is in Pythnet module. Would it be possible to move it out? (not necessary as this code is not gonna change but it's nicer).

apps/hermes/server/src/state/wormhole.rs
```diff
@@ -80,27 +83,35 @@ where
...
         }
 
         // Clear old cached VAA sequences.
         while observed_vaa_seqs.len() > OBSERVED_CACHE_SIZE {
             observed_vaa_seqs.pop_first();
         }
 
+        observed_vaa_seqs.insert(sequence);
+        // Drop the lock to allow other threads to access the state.
+        drop(observed_vaa_seqs);
```
ali-behjati, line 104: damn. thanks for catching this.

apps/hermes/server/src/network/pythnet.rs
danimhr: Let me create a new PR for this

Where should I put it?

PR #2812, opened 2025-06-25
feat(fortuna): Multiple replica support

tejasbadadare:
## Summary

Implemented multiple Fortuna replica support with modulo-based request assignment and configurable backup delays.

## Rationale

We want to improve Fortuna's reliability through redundancy, but naively running multiple Fortuna instances caused race conditions and gas waste. Now we distribute requests to specific replicas via `sequence_number % total_replicas`. All replicas still see all requests, so if a replica doesn't fulfill its assigned request, the others will try to fulfill it after the set delay.

The replica assignment & delay is implemented in `process_event_with_backoff`, which plays well with the block delay and retry mechanisms. It kicks in after the block delay has already elapsed, so the effective latency for requests missed by their assigned replica is block delay + replica delay.

## How has this been tested?

- [ ] Added new tests

Tested manually, works well on monad testnet. Stress tested it a bit with the load tester script and both instances were happy. Tested out killing one of them and the other one picks up the requests after the delay as expected.


apps/fortuna/README.md
```diff
@@ -40,6 +40,60 @@ Please add the changed files in the `.sqlx` folder to your git commit.
...
+```
+
+**Three Replica Setup:**
+```yaml
+# Replica 0 - handles sequence numbers 0, 3, 6, 9, ...
+keeper:
+  replica_config:
+    replica_id: 0
+    total_replicas: 3
+    backup_delay_seconds: 45
```
m30m: I would use the same value as the two above here just to avoid confusion.

apps/fortuna/config.sample.yaml
```diff
@@ -47,6 +47,8 @@ chains:
     # blocks after 5 blocks, then again after 10 blocks, and finally after 20 blocks.
     block_delays: [5, 10, 20]
 
+
+
```
m30m: do we need this?

apps/fortuna/src/keeper.rs
```diff
@@ -85,6 +85,13 @@ pub async fn run_keeper_threads(
         contract: contract.clone(),
         gas_limit,
         escalation_policy: chain_eth_config.escalation_policy.to_policy(),
+        keeper_config: crate::config::KeeperConfig {
```
m30m: I guess this should be some actual values?

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,66 @@ pub async fn process_event_with_backoff(
...
+                );
+                return Ok(());
+            }
+            Err(e) => {
+                tracing::warn!(
+                    sequence_number = event.sequence_number,
+                    error = ?e,
+                    "Error checking request status after delay, skipping"
+                );
+                return Ok(());
```
m30m: I would attempt fulfilling the request again here as well. A shaky RPC can cause the backup service to ignore the request.

apps/fortuna/src/keeper.rs
tejasbadadare: looks like devin didn't thread through the actual configs here. will fix

apps/fortuna/src/config.rs
```diff
@@ -94,6 +94,18 @@ impl Config {
             }
         }
 
+        if let Some(replica_config) = &config.keeper.replica_config {
+            if replica_config.total_replicas == 0 {
```
m30m, line 98: nit: can you add a check that if this is not None, `config.keeper.private_key.value/file` is also not None?

apps/fortuna/src/keeper.rs
```diff
@@ -67,6 +67,11 @@ pub async fn run_keeper_threads(
     let latest_safe_block = get_latest_safe_block(&chain_state).in_current_span().await;
     tracing::info!("Latest safe block: {}", &latest_safe_block);
 
+    let private_key = keeper_config
```
m30m: It seems a bit awkward that we are loading the private key here per chain again. I'd suggest having a separate struct here that contains the loaded private key or pass private_key and the ReplicaConfig as 2 separate arguments.

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,69 @@ pub async fn process_event_with_backoff(
         return Ok(());
     }
 
+    // Check if we are the primary replica for this request.
+    // If not, we will wait for a delay before processing the request.
+    let is_primary_replica =
```
m30m: I would put all of this logic nested inside this block:
```rs
if let Some(replica_config) = &process_param.keeper_config.replica_config {
...
}
```

This way, I can easily reason that if replica_config is None, I can skip the whole thing.

Also, if `is_primary_replica` is `false`, we expect `replica_config` to exist, but the code doesn't enforce that right now and it's a bit awkward.

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,69 @@ pub async fn process_event_with_backoff(
...
+        if let Some(replica_config) = &process_param.keeper_config.replica_config {
+            let assigned_replica = event.sequence_number % replica_config.total_replicas;
+            if assigned_replica != replica_config.replica_id {
+                tracing::debug!(
+                    sequence_number = event.sequence_number,
```
m30m: no need to add sequence_number, it's already added via instrumentation.

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,69 @@ pub async fn process_event_with_backoff(
...
+                true
+            }
+        } else {
+            true // No replica config, process all requests
+        };
+
+    if !is_primary_replica {
+        if let Some(replica_config) = &process_param.keeper_config.replica_config {
+            tracing::info!(
+                sequence_number = event.sequence_number,
```
m30m: ditto

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,69 @@ pub async fn process_event_with_backoff(
...
+        // Check if the request is still open after the delay.
+        // If it is, we will process it as a backup replica.
+        match chain_state
+            .contract
+            .get_request(event.provider_address, event.sequence_number)
+            .await
+        {
+            Ok(Some(_)) => {
+                tracing::info!(
+                    sequence_number = event.sequence_number,
```
m30m: ditto

apps/fortuna/src/keeper.rs
tejasbadadare: Yeah thats fair -- passing in both felt awkward, but i agree that it's less awkward than loading twice.

apps/fortuna/src/keeper/process_event.rs
tejasbadadare: good idea thanks!

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,56 @@ pub async fn process_event_with_backoff(
...
+                Err(e) => {
+                    tracing::warn!(
+                        error = ?e,
+                        "Error checking request status after delay, processing as backup replica"
+                    );
+                }
+            }
+        }
+    }
+    // If no replica config we are running standalone, process all requests directly
```
m30m: I would put this comment at the top of this block. I generally expect comments to describe the next block or line but this seems to be a comment for a non-existent else block

apps/fortuna/src/keeper/process_event.rs
```diff
@@ -35,6 +35,56 @@ pub async fn process_event_with_backoff(
...
+    // then check if the request is still open. If it is, we process it as a failover.
+    if let Some(replica_config) = &process_param.replica_config {
+        let assigned_replica = event.sequence_number % replica_config.total_replicas;
+        let is_primary_replica = assigned_replica == replica_config.replica_id;
+
+        if is_primary_replica {
+            tracing::debug!("Processing request as primary replica");
+        } else {
+            tracing::debug!("Processing request as backup replica");
+
```
m30m, line 51: I think we can merge these 2 logs

PR #2814, opened 2025-06-26
feat(entropy-tester): accept custom endpoint

m30m:
## Summary

Use a custom endpoint for making requests

## Rationale

I was hoping that the load is small enough such that public rpcs would be sustainable. But that was not the case.

## How has this been tested?

- [ ] Added new tests


tejasbadadare (approved):
suggest calling it rpcEndpoint or rpcUrl for consistency

PR #2815, opened 2025-06-26
feat(stylus): added crate for types shared across Wormhole and Pyth

ayushboss:
## Summary

Added a new crate that defines types for structs shared between Wormhole and Pyth stylus contracts

## Rationale

Allows for more flexible refactoring of structs in the future.

## How has this been tested?

- [ ] Added new tests

The crate currently builds. I will test it further when integrating it into their respective libraries.


target_chains/stylus/contracts/pyth-types/src/lib.rs
tejasbadadare, line 1: Ik this is a draft PR but wondering why we want to extract out the wormhole types here? they aren't needed outside of the wormhole crate right?

Issue #2796, opened 2025-06-23
Need a Rust SDK for SUI blockchain


0xbarchitect:
Hi team,
I am building a Defi applications on SUI blockchain and need to use Pyth price feed.
My applications is built in Rust, so I need to use a Pyth Rust SDK. Unfortunately, it seems that the official SDK for SUI is only available in JS. 
I have checked [this tutorial](https://docs.pyth.network/price-feeds/use-real-time-data/sui).
Are there an available Rust SDK for SUI?

Thanks!

Ifechukwudaniel:
I can do this

Ifechukwudaniel:
@madninja  just started working on this

PR #2829, opened 2025-07-02
sui: rust sui sdk

Ifechukwudaniel:
## Rationale

Currently, the only official way to interact with Pyth on Sui is through JavaScript, which doesn’t make sense for teams building entirely in Rust. This SDK enables you to remain within the Rust ecosystem while accessing up-to-date price data and pushing it on-chain. This is just the foundation, but it clears the path for full-on-chain integration using Rust.


Ifechukwudaniel:
This is my attempt at #2796

PR #2830, opened 2025-07-02
chore: Rust toolchain/CI update and workspace setup

Riateche:
## Summary

- Added top level Cargo workspace and rust-toolchain.toml file
- Updated all Rust CI actions to use rust-toolchain.toml file instead of hardcoding version in the action
- Updated all Rust CI actions to use `actions-rust-lang/setup-rust-toolchain`
- Removed redundant `Swatinem/rust-cache` as the above action calls it internally
- Removed rust-toolchain files so that the top level version is used where possible
- Enabled new lints for a few projects

Smart contracts remain in their separate workspaces. There are a couple of reasons why they cannot be included in the main workspace (pinned dependencies, lack of chain tooling support, different release profile configs).

## Rationale

- It's easier to upgrade toolchain for all projects.
- It's easier to maintain projects if they're built/checked using a recent Rust toolchain.
- Now it's easier to add new projects to the shared workspace.
- Better Rust tooling support (e.g. rust-analyzer).

## How has this been tested?

- [ ] Added new tests


lazer/contracts/solana/clippy.toml
```diff
@@ -0,0 +1,3 @@
+allow-unwrap-in-tests = true
+allow-expect-in-tests = true
+#allow-indexing-slicing-in-tests = true
```
ali-behjati: ?

Riateche: Fixed.

PR #2831, opened 2025-07-02
Lazer: Add inactive symbol state

danimhr:


lazer/publisher_sdk/proto/state.proto
```diff
@@ -44,6 +44,12 @@ message Publisher {
     optional bool is_active = 4;
 }
 
+enum FeedMetadataState {
```
Riateche: How about just `FeedState`?

lazer/sdk/rust/protocol/src/symbol_state.rs
```diff
@@ -8,13 +8,15 @@ use {
 pub enum SymbolState {
     Stable,
     ComingSoon,
+    Deprecated,
```
Riateche:
```suggestion
    ComingSoon,
    Stable,
    Deprecated,
```
Just to match the order in proto.

PR #2832, opened 2025-07-04
feat(evm/sdk/js): add pyth filler

ali-behjati:
## Summary

This change revamps the deprecated EVM SDK to add a utilty for EVM to fill a call with the Pyth data it needs automatically. Under the hood it performs trace calls to detect Pyth contract usage and extracts the IDs from there, then fetches the update data from Hermes and creates a pyth update call.

This change is the alpha version and is subject to change. In the subsequent PRs more improvements will come. Also I'm going to improve the packaging (like dual export, file structure) later.

Please review the change commit-by-commit (the first commit has just removals, and the second one has actual changes).

## Rationale

This change makes integrations to Pyth in the pull model easier.

## How has this been tested?

- [x] Manually tested the code. Tested two different methods in locally forked network (off eth mainnet) and OP Sepolia.

### DIFF
--- a/target_chains/ethereum/sdk/js/package.json
+++ b/target_chains/ethereum/sdk/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/pyth-evm-js",
-  "version": "2.0.0-alpha",
+  "version": "2.0.0-alpha2",
   "description": "Pyth Network EVM Utils in JS",
   "homepage": "https://pyth.network",
   "author": {


target_chains/ethereum/sdk/js/README.md
```diff
@@ -22,156 +19,60 @@ $ yarn add @pythnetwork/pyth-evm-js
...
-
-Pyth offers a free public endpoint at [https://hermes.pyth.network](https://hermes.pyth.network). However, it is
-recommended to obtain a private endpoint from one of the Hermes RPC providers for more reliability. You can find more
-information about Hermes RPC providers
-[here](https://docs.pyth.network/documentation/pythnet-price-feeds/hermes#public-endpoint).
+import { fillPythUpdate, multicall3Bundler } from "@pythnetwork/pyth-evm-js";
+import { createPublicClient, http } from "viem";
+import { optimismSepolia } from "viem/chains";
+
+async function example() {
```
cprussin: Minor thing but since top-level await exists, you don't need the code here to live in a function, and it kind of distracted me a bit when I was reading through the docs here -- when I saw this my immediate thought was "what is the example function returning and where do I use it" -- I'd just put the body at the top level to minimize the distraction.

target_chains/ethereum/sdk/js/src/filler.ts
```diff
@@ -0,0 +1,218 @@
...
+export type DebugTraceCallConfig = {
+  /** Must be "debug_traceCall" */
+  method: "debug_traceCall";
+  /** Function that takes a Pyth update and original call, returns a single bundled call request.
+   * Common bundlers include multicall3Bundler for combining calls via Multicall3 contract.
+   * The bundler must create a single transaction that executes both the Pyth update and the original call. */
+  bundler: Bundler;
+  /** Maximum number of iterations to find all required price feeds. Default is 5.
+   * Each iteration traces the current transaction to find new Pyth price feed calls. */
+  maxIter: number;
```
aditya520: Why do we need to iterate multiple times to find price feeds?

ali-behjati:
```solidity
    function increment() public returns (int256) {
        // Fetch the latest price from Pyth
        PythStructs.Price memory ethPrice = pyth.getPriceNoOlderThan(0xff61491a931112ddf1bd8147cd1b641375f79f5825126d665480874634fd0ace, 30);
        PythStructs.Price memory btcPrice = pyth.getPriceNoOlderThan(0xe62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43, 30);
    }
```

Consider a call to the function above is given to the method. The problem is that with the first trace call, the execution reverts at the first price fetch request and we only identify ETH price feed id. We need to iterate again, by bundling the update and call to pass that line to get the BTC price feed id.

target_chains/ethereum/sdk/js/README.md
```diff
@@ -22,156 +19,60 @@ $ yarn add @pythnetwork/pyth-evm-js
...
+
+    // Fill Pyth update data using "trace_callMany"
+    const pythUpdate = await fillPythUpdate(
+        client,
+        {
+            to: "0x3252c2F7962689fA17f892C52555613f36056f22",
+            data: "0xd09de08a", // Your transaction calldata
+            from: "0x78357316239040e19fC823372cC179ca75e64b81",
+        },
+        "0x0708325268df9f66270f1401206434524814508b", // Pyth contract address
```
cprussin: Can we export this as a const?  I know that contract manager isn't in a great state but even just manually copying it over seems to be a better experience for now

target_chains/ethereum/sdk/js/README.md
```diff
@@ -22,156 +19,60 @@ $ yarn add @pythnetwork/pyth-evm-js
...
+        "https://hermes.pyth.network", // Hermes endpoint
```
cprussin: Can we make this a const too?  Or better yet, can we make this function just automatically derive these two fields from the `client`, but allow overriding them in the options?

target_chains/ethereum/sdk/js/package.json
```diff
@@ -21,42 +21,37 @@
...
-    "@typescript-eslint/parser": "^5.21.0",
-    "eslint": "^8.14.0",
+    "eslint": "catalog:",
     "jest": "^29.4.1",
     "prettier": "catalog:",
     "ts-jest": "^29.0.5",
-    "typescript": "^4.6.3",
-    "web3": "^1.8.2",
-    "yargs": "^17.4.1"
+    "ts-node": "catalog:",
```
cprussin: Is this dependency used anywhere?

target_chains/ethereum/sdk/js/src/filler.ts
```diff
@@ -0,0 +1,218 @@
...
+  method: "trace_callMany";
+  /** Maximum number of iterations to find all required price feeds. Default is 5.
+   * Each iteration traces the current set of transactions to find new Pyth price feed calls. */
+  maxIter: number;
+};
+
+/**
+ * Union type for tracing configuration options
+ */
+export type Config = DebugTraceCallConfig | TraceCallManyConfig;
```
cprussin: You can make this type a lot more concise:

```ts
export type Config = { maxIter: number } & (
  | { method: "trace_callMany" }
  | { method: "debug_traceCall", bundler: Bundler }
);
```

target_chains/ethereum/sdk/js/src/filler.ts
```diff
@@ -0,0 +1,218 @@
...
+ */
+export async function fillPythUpdate<
+  transport extends Transport,
+  chain extends Chain | undefined,
+>(
+  client: PublicClient<transport, chain>,
+  call: CallRequest,
+  pythContractAddress: Address,
+  hermesEndpoint: string,
+  config?: Config,
```
cprussin, line 142: minor nit but a better way to write the default is:

```ts
export async function fillPythUpdate(
  // ... other args
  config: Config = {
    method: "debug_traceCall",
    bundler: multicall3Bundler,
    maxIter: 5
  }
)
```

target_chains/ethereum/sdk/js/src/index.ts
```diff
@@ -1,175 +1,3 @@
...
-  zetachain_testnet: "0x0708325268dF9F66270F1401206434524814508b",
-  zkfair_testnet: "0xA2aa501b19aff244D90cc15a4Cf739D2725B5729",
-  zksync_era_goerli: "0x8739d5024B5143278E2b15Bd9e7C26f6CEc658F1",
-  zksync_era_sepolia: "0x056f829183Ec806A78c26C98961678c24faB71af",
-  // Testnets (Beta sources)
-  horizen_gobi_testnet: "0xA2aa501b19aff244D90cc15a4Cf739D2725B5729",
-};
+export * from "./filler";
+export * from "./multicall3-bundler";
+export * from "./pyth-abi";
```
cprussin, line 3: As we'd discussed, barrel exports aren't great, I would consider just putting these into the same file -- I know it feels nice to have smaller separated files, but it does add fs read cost at load time for apps that use it.  It's not a huge thing, but it adds up across the whole js package ecosystem, so the more packages that stop using barrel exports, the better.

target_chains/ethereum/sdk/js/src/tracer/debug-trace-call.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Extract Pyth price feed IDs from a transaction call trace.
+ */
+export function extractPythPriceFeedsFromDebugTraceCall(
+  trace: RpcCallTrace,
+  pythContractAddress: Address,
+  ignoreParsingErrors = false,
+): Set<`0x${string}`> {
+  const result = new Set<`0x${string}`>();
```
cprussin: You can use `Address` from viem instead of `0x${string}`

target_chains/ethereum/sdk/js/src/tracer/debug-trace-call.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (priceFeedId !== undefined) {
+        result.add(priceFeedId);
+      }
+    } catch {
+      if (!ignoreParsingErrors) {
+        throw new Error(
+          `Failed to decode calldata: ${trace.input}. Make sure correct Pyth contract address is used.`,
+        );
+      }
+    }
```
cprussin:
```suggestion
    } catch (error: unknown) {
      if (!ignoreParsingErrors) {
        const thrownError = new Error(
          `Failed to decode calldata: ${trace.input}. Make sure correct Pyth contract address is used.`,
        );
        thrownError.cause = error;
        throw thrownError;
      }
    }
```

this way you don't lose the context of the underlying error

target_chains/ethereum/sdk/js/src/filler.ts
ali-behjati: I did it in the beginning but the linter told me not to do it.

ali-behjati: 143:3  error  Do not use an object literal as default for parameter `config`  unicorn/no-object-as-default-parameter

i ignored it though.

target_chains/ethereum/sdk/js/README.md
ali-behjati: screw cursor lol. :D thanks.

target_chains/ethereum/sdk/js/README.md
ali-behjati: i want people to think about this endpoint (and not use the public one if possible)

target_chains/ethereum/sdk/js/src/filler.ts
cprussin: ah I actually was not aware of this rule, TIL

PR #2836, opened 2025-07-07
feat(pyth-lazer) Create schema for lazer agent JRPC endpoint

bplatak:
## Summary
Add a new schema for the JRPC endpoint in `pyth-lazer-agent` for sending updates, and retrieving metadata via the JRPC endpoint. Also move the `Timestamp` logic to `protocol` to reuse across `pyth-lazer`

## Rationale
Adding a new endpoint for publishers to communicate with the `pyth-lazer-agent`.

## How has this been tested?

- [x] Added new tests


lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[serde(flatten)]
+    pub params: JrpcParams,
+    pub id: i64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+#[serde(tag = "method", content = "params")]
+pub enum JrpcParams {
+    #[serde(rename = "send_updates")]
+    SendUpdates(FeedUpdateParams),
```
ali-behjati: i suggest making this an array of updates (to allow batching)

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub id: i64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct ErrorResponse {
+    pub message: String
+}
+
+#[derive(Serialize, Deserialize)]
+struct SymbolMetadata {
```
ali-behjati: let's aim for a map <str -> str> for metadata. i think anything that is usable in the code should not live in metadata (like exponent) but that's a step in the future.

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct FeedUpdateParams {
+    pub feed_id: PriceFeedId,
+    pub source_timestamp: TimestampUs,
+    pub update: UpdateParams,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+#[serde(tag = "type")]
+pub enum UpdateParams {
```
merolish, line 31: Is this just for FeedUpdate or possibly to be used elsewhere?

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[serde(rename = "price")]
+    PriceUpdate {
+        price: Price,
+        best_bid_price: Price,
+        best_ask_price: Price,
+    },
+    #[serde(rename = "funding_rate")]
+    FundingRateUpdate { price: Price, rate: Rate },
+}
```
merolish: I'm wondering if these will need to be optional. I also wonder if we'll ever need to revisit the nonzero requirement on price for any of these cases.

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct GetMetadataParams {
+    filters: Option<Vec<Filter>>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub enum JsonRpcVersion {
+    #[serde(rename = "2.0")]
+    V2,
+}
```
merolish, line 58: Are we doing it this way to break anyone using an older sdk that receives a newer message? As in, we'll add/modify this enum as the sdk evolves?

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub hermes_id: String,
+    #[serde(default, with = "humantime_serde", alias = "interval")]
+    pub interval: Option<Duration>,
+    pub min_channel: String,
+    pub min_publishers: i64,
+    pub name: String,
+    pub pyth_lazer_id: i64,
+    pub schedule: String,
+    pub state: String,
+    pub symbol: String,
```
merolish: Several of these will need to be optional. E.g., non-crypto tickers will not have a CoinMarketCap id, not all tickers will have a pythnet/hermes id (and it'll be going away at some point, I assume).

lazer/sdk/rust/protocol/src/jrpc.rs
keyvankhademi: Agent handles the batching and the communication between publisher and agent is on local network so I think it's better to force publisher to not batch.

lazer/sdk/rust/protocol/src/jrpc.rs
bplatak: This isn't related to _our_ protocol - it's part of the jRPC specs ([see rfc](https://www.jsonrpc.org/specification#response_object))

lazer/sdk/rust/protocol/src/jrpc.rs
ali-behjati: yeah it's fair. just want to make sure a massive rate of update doesn't exhaust the port. some people might not run them on the same machine too although it'll be in a private network.

ali-behjati: just a nit to call it `SendUpdate`

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct PythLazerAgentJrpcV1 {
+    pub jsonrpc: JsonRpcVersion,
+    #[serde(flatten)]
+    pub params: JrpcParams,
+    pub id: i64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+#[serde(tag = "method", content = "params")]
+pub enum JrpcParams {
```
darunrs: Isn't method a top level field? Why is it part of params?

lazer/sdk/rust/protocol/src/jrpc.rs
darunrs: This one should just be FeedUpdate. The goal is probably convert them to FeedUpdate, collect them, and then use them to make a publisher update.

lazer/sdk/rust/protocol/src/jrpc.rs
darunrs: These are marked optional currently both in the transaction schema and also in pyth lazer itself. So, let's keep that for now. I haven't thought too deeply about if some should be marked as required.

lazer/sdk/rust/protocol/src/jrpc.rs
Riateche: It will end up on the top level because of `serde(flatten)`.

Riateche: I would call it `JrpcCall` or something since it not only contains params, but the method name as well.

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[serde(rename = "send_updates")]
```
Riateche: You can use `serde(rename_all = "snake_case")`.

lazer/sdk/rust/protocol/src/jrpc.rs
Riateche: Maybe even `add_update` or `push_update` because "send" implies that it would be sent immediately.

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct Filter {
+    name: Option<String>,
+    asset_type: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct GetMetadataParams {
+    filters: Option<Vec<Filter>>,
```
Riateche: This seems a bit too complex and it's not obvious how the filters are combined. Maybe something like this is sufficient?
```
pub struct GetMetadataParams {
    pub names: Option<Vec<String>>,
    pub asset_types: Option<Vec<String>>,
}
```

lazer/sdk/rust/protocol/src/jrpc.rs
Riateche: I think `BTreeMap<String, serde_value::Value>` would be even better. We should support typed metadata instead of converting everything to string. I also agree about separating exponent, etc from arbitrary metadata.

lazer/sdk/rust/protocol/src/jrpc.rs
```diff
@@ -0,0 +1,238 @@
...
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct PythLazerAgentJrpcV1 {
+    pub jsonrpc: JsonRpcVersion,
+    #[serde(flatten)]
+    pub params: JrpcParams,
+    pub id: i64,
+}
```
Riateche, line 12: Let's use [jsonrpc_types](https://docs.rs/jsonrpc-types/latest/jsonrpc_types/index.html) instead of reimplementing it.

darunrs:
> > These are marked optional currently both in the transaction schema and also in pyth lazer itself. So, let's keep that for now. I haven't thought too deeply about if some should be marked as required.
> 
> Let's have as required here since we don't want publishers to send invalid data.

Would a price with no bid/ask be considered invalid? Some publishers may source them separately, causing them to send the price separately from the bid/ask. I'm not sure making these required helps reduce invalid data much compared to the loss in flexibility.

Riateche:
> Would a price with no bid/ask be considered invalid?

All fields are optional, so an update with any combination of the fields present will still be valid.

> Some publishers may source them separately, causing them to send the price separately from the bid/ask.

That would not be correct because any update overwrites everything from the previous update.

darunrs:
> > Would a price with no bid/ask be considered invalid?
> 
> All fields are optional, so an update with any combination of the fields present will still be valid.
> 
> > Some publishers may source them separately, causing them to send the price separately from the bid/ask.
> 
> That would not be correct because any update overwrites everything from the previous update.

Ah right forgot about the second case. For the first, I agree.

lazer/sdk/rust/protocol/src/jrpc.rs
bplatak: yeah all of this needs to be refactored as per https://github.com/pyth-network/pyth-crosschain/pull/2836#discussion_r2190372851

lazer/sdk/rust/protocol/src/jrpc.rs
bplatak: > separating exponent, etc

Which fields do you consider to not be a part of the arbitrary metadata?

lazer/sdk/rust/protocol/src/jrpc.rs
bplatak: That's right. It's also quite useful in flattening the `params` objects

lazer/sdk/rust/protocol/src/jrpc.rs
keyvankhademi: This package doesn't work really well since we have to manually convert the type to our own struct. This library however could parse into a generic type. https://docs.rs/jrpc-types/0.1.0/jrpc_types/
We will refactor to use this in the next PR.

lazer/sdk/rust/protocol/src/jrpc.rs
bplatak: spoke to @keyvankhademi and we decided to make `best_bid_price`, `best_ask_price` and `price` (in FundingRate) optionals

ali-behjati: are we gonna have ^ fields in the fundingrate update at all :? if not then not having it is easier to understand for publishers. we can change the internals later.

keyvankhademi (approved):

PR #2840, opened 2025-07-09
feat: stylus parse price feed updates fxn

ayushboss:
## Summary

Implemented the parse price feed updates function for the stylus contract, and restructured the architecture of the update functions to deal with that.

## Rationale

parsePriceFeedUpdates is a necessary component of the Pythnet API.

## How has this been tested?

- [x] Added new tests

I have test cases validating these methods.

### DIFF
--- a/target_chains/stylus/contracts/pyth-receiver/src/lib.rs
+++ b/target_chains/stylus/contracts/pyth-receiver/src/lib.rs
@@ -177,7 +177,7 @@ impl PythReceiver {
         update_data: Vec<Vec<u8>>,
     ) -> Result<(), PythReceiverError> {
         for data in &update_data {
-            self.update_price_feeds_internal(data.clone(), Vec::new(), 0, 0, false)?;
+            self.update_price_feeds_internal(data.clone(), 0, 0, false)?;
         }
 
         let total_fee = self.get_update_fee(update_data)?;
@@ -221,16 +221,15 @@ impl PythReceiver {
     fn update_price_feeds_internal(
         &mut self,
         update_data: Vec<u8>,
-        _price_ids: Vec<[u8; 32]>,
         min_publish_time: u64,
         max_publish_time: u64,
-        _unique: bool,
+        unique: bool,
     ) -> Result<Vec<([u8; 32], PriceInfoReturn)>, PythReceiverError> {
         let price_pairs = self.parse_price_feed_updates_internal(
             update_data,
             min_publish_time,
             max_publish_time,
-            false, // check_uniqueness
+            unique,
         )?;
 
         for (price_id, price_return) in price_pairs.clone() {
@@ -266,12 +265,8 @@ impl PythReceiver {
                 }
             }
         }
-        Ok(self.get_total_fee(total_num_updates))
-    }
-
-    fn get_total_fee(&self, total_num_updates: u64) -> U256 {
-        U256::from(total_num_updates).saturating_mul(self.single_update_fee_in_wei.get())
-            + self.transaction_fee_in_wei.get()
+        Ok(U256::from(total_num_updates).saturating_mul(self.single_update_fee_in_wei.get())
+            + self.transaction_fee_in_wei.get())
     }
 
     pub fn get_twap_update_fee(&self, _update_data: Vec<Vec<u8>>) -> U256 {
@@ -312,7 +307,6 @@ impl PythReceiver {
             if store_updates_if_fresh {
                 all_parsed_price_pairs.extend(self.update_price_feeds_internal(
                     data.clone(),
-                    price_ids.clone(),
                     min_allowed_publish_time,
                     max_allowed_publish_time,
                     check_uniqueness,


target_chains/stylus/contracts/pyth-receiver/src/lib.rs
```diff
@@ -193,17 +192,172 @@ impl PythReceiver {
...
+        _price_ids: Vec<[u8; 32]>,
+        min_publish_time: u64,
+        max_publish_time: u64,
+        _unique: bool,
+    ) -> Result<Vec<([u8; 32], PriceInfoReturn)>, PythReceiverError> {
+        let price_pairs = self.parse_price_feed_updates_internal(
+            update_data,
+            min_publish_time,
+            max_publish_time,
+            false, // check_uniqueness
```
tejasbadadare: why not pass `unique` to `parse_price_feed_updates_internal`?

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
```diff
@@ -262,33 +418,40 @@ impl PythReceiver {
...
+                            }
+
+                            let price_info_return = (
+                                U64::from(publish_time),
+                                I32::from_be_bytes(price_feed_message.exponent.to_be_bytes()),
+                                I64::from_be_bytes(price_feed_message.price.to_be_bytes()),
+                                U64::from(price_feed_message.conf),
+                                I64::from_be_bytes(price_feed_message.ema_price.to_be_bytes()),
+                                U64::from(price_feed_message.ema_conf),
+                            );
```
tejasbadadare, line 446: Noticed you're turning the signed ints to bytes and back again. There should be `from`/`try_from` impls for them as well, any reason you're going to bytes and back?

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
```diff
@@ -193,17 +192,172 @@ impl PythReceiver {
...
+                }
+            }
+        }
+        Ok(self.get_total_fee(total_num_updates))
+    }
+
+    fn get_total_fee(&self, total_num_updates: u64) -> U256 {
+        U256::from(total_num_updates).saturating_mul(self.single_update_fee_in_wei.get())
+            + self.transaction_fee_in_wei.get()
+    }
```
tejasbadadare: We should just have `get_update_fee(&self, update_data: Vec<Vec<u8>>)` since the old version that just takes the update data length is [deprecated](https://github.com/pyth-network/pyth-crosschain/blob/02fe8404f7f3cb72ceba57ebe79edff5c460c929/target_chains/ethereum/contracts/contracts/pyth/Pyth.sol#L81-L84)

target_chains/stylus/contracts/pyth-receiver/src/integration_tests.rs
```diff
@@ -342,4 +342,14 @@ mod test {
             multiple_updates_diff_vaa_results()[1]
         );
     }
+
+    #[motsu::test]
+    fn test_multiple_updates_same_id_updates_latest(
```
tejasbadadare, line 347: unfinished tests?

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
```diff
@@ -193,17 +192,172 @@ impl PythReceiver {
...
+        let price_id_fb: FixedBytes<32> = FixedBytes::from(price_id);
+        let recent_price_info = self.latest_price_info.get(price_id_fb);
+        recent_price_info.publish_time.get().to::<u64>()
     }
 
     fn update_price_feeds_internal(
         &mut self,
         update_data: Vec<u8>,
-    ) -> Result<(), PythReceiverError> {
+        _price_ids: Vec<[u8; 32]>,
```
tejasbadadare: Looks like `price_ids` is unused here because we don't pass it in `parse_price_feed_updates_with_config` -- we parse and update _all_ the data, and then filter them using `price_ids`. There was a subtle griefing attack possibility with this behavior in the Pulse use case, but I don't think that's really a factor here and there are other mitigations (`check_update_data_is_minimal`.) 

We can remove the `_price_ids` arg here.

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
ayushboss: the alloy_primitives crate doesn't have a normal From<<i32>> for the I32 type, so this was the cleanest workaround I could think of.

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
ayushboss: Yeah, I went ahead and compressed the two functions. This one was just a helper function for the public facing get_update_fee function though that took the whole update data.

target_chains/stylus/contracts/pyth-receiver/src/integration_tests.rs
ayushboss: The next PR in the line completely cleaned up the test suite and added more, so if it's okay can we circle back to reviewing the tests on that PR?

tejasbadadare: sounds good!

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
tejasbadadare: There is a `try_from` which seems to work in my brief testing, would try that out. Also, it looks like the `from` and `from_be_bytes` methods panic if the conversion fails. I think it would be better to use `try_from` and gracefully catch and propagate any errors. (Can do this in a future PR)

tejasbadadare (approved):

tejasbadadare (approved):
Left a small reply. Nice work!!

target_chains/stylus/contracts/pyth-receiver/src/lib.rs
ayushboss: Ok will do. Will add this into the next PR that I'll send out in a second

PR #2841, opened 2025-07-09
feat(fortuna): Support Postgres DB backend

tejasbadadare:
## Summary
Support both Postgres and SQLite database backends instead of just SQLite. This lets us use a hosted Postgres database that multiple replicas can connect to, instead of a file-based one. We still use an in-memory SQLite for the History tests.

## Rationale
A central database shared between the replicas gives us a consistent storage to serve the API from. 

Some impl details:
- Used a generic `AnyPool` sqlx driver instead of a `Pool<Sqlite>`. It chooses the correct driver at runtime based on the URL passed in.
- Unified the migrations into one.
- Updated DB types and SQL syntax to be cross-compatible with Postgres and SQLite (this means using an integer to store timestamp instead of a Datetime, but afaik there's no way around this due to intricacies in how AnyPool works. Shouldn't be an issue because we handle the type coercion ourselves anyway.)
- AnyPool doesn't support QueryBuilder properly, so we now manually build the select queries in `explorer.rs`, similar to the update queries in `history.rs`.
- Updated docs

## How has this been tested?
- [ ] Added new tests

Tested Fortuna's `/v1/logs` (explorer) API with a SQLite backend and a cloud Postgres backend from Neon. Also tested with multiple replicas pointing to the same DB, which is our intended production configuration.


apps/fortuna/src/history.rs
```diff
@@ -211,28 +216,54 @@ impl From<RequestRow> for Option<RequestStatus> {
...
     write_queue: mpsc::Sender<RequestStatus>,
     _writer_thread: Arc<tokio::task::JoinHandle<()>>,
 }
 
 impl History {
     const MAX_WRITE_QUEUE: usize = 1_000;
     pub async fn new() -> Result<Self> {
-        Self::new_with_url("sqlite:fortuna.db?mode=rwc").await
+        let database_url =
+            std::env::var("DATABASE_URL").unwrap_or_else(|_| DEFAULT_DATABASE_URL.to_string());
```
jayantk, line 228: can this be a config value? it's confusing to have configuration both via a file and via env. I can see this being useful for secrets or something, but this doesn't seem particularly sensitive.

apps/fortuna/src/history.rs
```diff
@@ -211,28 +216,54 @@ impl From<RequestRow> for Option<RequestStatus> {
...
+        // Connect to an in-memory SQLite database
+        // Don't let the pool drop the cxn, otherwise the database will be deleted
+        let pool = AnyPoolOptions::new()
+            .min_connections(1)
+            .max_connections(1)
+            .idle_timeout(None)
+            .max_lifetime(None)
+            .connect("sqlite::memory:")
+            .await?;
+        let migrator = migrate!("./migrations");
```
jayantk: is the `sqlite_migrations` directory used anywhere? if not, delete please

jayantk: (also suggest extracting a string constant for the migrations directory)

apps/fortuna/src/history.rs
tejasbadadare: The `sqlx` tool uses `dotenv` to read the URL from the `.env` file, so we use that in Fortuna as well (we could load it into the env in a different way if needed in k8s.) It helps devex IMO to use the same `.env` file in Fortuna that `sqlx` uses, as sometimes it's necessary to setup/recover migrations using the tool. Also the connection string has credentials in it, so better to consider it a secret rather than a config

apps/fortuna/src/history.rs
tejasbadadare: Ah yep forgot to delete, thanks

apps/fortuna/Cargo.toml
```diff
@@ -46,8 +46,16 @@ chrono = { version = "0.4.38", features = [
...
+sqlx = { version = "0.8", features = [
+  "runtime-tokio",
+  "tls-rustls",
+  "sqlite",
+  "any",
+  "postgres",
+  "chrono",
+] }
 num-traits = "0.2.19"
+dotenv = "0.15.0"
```
ali-behjati, line 58: this seems abandoned (from sqlx docs), can you use dotenvy?

apps/fortuna/src/history.rs
ali-behjati: i hear you here but it's very confusing to have two ways of passing configs. i think moving this to `run.rs` might be better and regardless i recommend logging this (as info level) that you are using this data base url to help people catch it if they don't set it.

tejasbadadare: Yeah fair enough, it can be confusing to have 2 config sources. Since we are ok with exposing the DB credentials, I'll go ahead and unify to config.yaml.

PR #2856, opened 2025-07-16
feat(lazer): improve metadata in protobuf

Riateche:
## Summary

This PR splits old content of `FeedMetadata` into two categories:

1. Essential feed properties that are used for feed aggregation or other functions within Lazer core system (relayer, router, history service): `name, exponent, min_publishers, min_rate, expiry_time, market_schedule, state`. These are now properties of the `Feed` message.
2. Extra metadata used for feed search, documentation, feed catalogs (`symbol, description, asset_type, quote_currency`), and for matching with external systems (`cmc_id, hermes_id`). These fields may be used by publishers and by history service (for feed search API), but are otherwise treated as opaque in the Lazer core system. They are now stored as a collection of dynamically typed key-value pairs. This gives the operations team more freedom to extend and change metadata. 

Other relevant changes in this PR:
 - Rename `is_activated` to `is_enabled_in_shard` (same for other related fields) to clarify the meaning and intended usage, and to avoid confusion with `Feed.state`.
 - Add `DynamicValue` Rust type to deal with metadata without loss of information.

## Rationale

For more flexible metadata.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/lazer/publisher_sdk/proto/state.proto
+++ b/lazer/publisher_sdk/proto/state.proto
@@ -48,10 +48,12 @@ message Publisher {
 
 enum FeedState {
     // Default value. Feeds in this state are not available to consumers.
+    // `COMING_SOON` feeds are expected to become stable in the future.
     COMING_SOON = 0;
     // A fully available feed.
     STABLE = 1;
-    // Inactive feeds are not available to consumers or publishers.
+    // Inactive feeds are not available to consumers.
+    // `INACTIVE` feeds are not expected to become stable again.
     INACTIVE = 2;
 }
 


lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,62 @@ message Publisher {
...
+    optional FeedState state = 107;
+
+
+    // [required] Feed status in the current shard. Disabled feeds will not be visible in
+    // the consumer API for the current shard. This setting should only be used
+    // to migrate a feed from one shard to another.
+    //
+    // If a feed is present in
     // multiple shards, it must only be active in one of them at each time.
     // To enforce this, `pending_activation` and `pending_deactivation` fields
```
ali-behjati: update these field names?

lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,62 @@ message Publisher {
...
-    // new feeds before releasing them. This can also be used to migrate a feed from
-    // one shard to another. If a feed is present in
+    optional google.protobuf.Duration expiry_time = 105;
+    // [required] Market schedule in Pythnet format.
+    optional string market_schedule = 106;
+    // [required] Feed state
+    optional FeedState state = 107;
+
+
+    // [required] Feed status in the current shard. Disabled feeds will not be visible in
```
ali-behjati, line 103: sorry i got a bit lost here with feedState. what cases does it cover that feed state is not :? is it like we add a same feed to all the shards that we need this?

lazer/publisher_sdk/rust/src/dynamic_value.rs
```diff
@@ -0,0 +1,228 @@
...
+        };
+        ProtobufDynamicValue {
+            value: Some(converted),
+            special_fields: Default::default(),
+        }
+    }
+}
+
+impl From<&DynamicValue> for ProtobufDynamicValue {
+    fn from(value: &DynamicValue) -> Self {
```
ali-behjati: i think generally the expectation from `From` is to consume the object.

lazer/publisher_sdk/proto/governance_instruction.proto
```diff
@@ -321,43 +337,68 @@ message UpdateFeed {
...
+    // [optional]
+    optional google.protobuf.Duration min_rate = 104;
+    // [optional]
+    optional google.protobuf.Duration expiry_time = 105;
+    // [optional]
+    optional string market_schedule = 106;
+    // [optional]
+    optional FeedState state = 107;
+    // [optional]
+    optional bool is_enabled_in_shard = 201;
```
keyvankhademi, line 369: I think we should add asset_type here too. Because we will be using asset_type to determine what kind of feed type is expected (e.g funding-rate vs regular price).

lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,62 @@ message Publisher {
 }
 
 enum FeedState {
-    COMING_SOON = 0; // Default value
+    // Default value. Feeds in this state are not available to consumers.
+    COMING_SOON = 0;
+    // A fully available feed.
     STABLE = 1;
+    // Inactive feeds are not available to consumers or publishers.
```
keyvankhademi: it is technically available to publishers now.

lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,62 @@ message Publisher {
...
-    optional google.protobuf.Duration expiry_time = 11;
-    // [required] If true, the feed is visible to the consumers. This can be used to prepare and verify
-    // new feeds before releasing them. This can also be used to migrate a feed from
-    // one shard to another. If a feed is present in
+    optional google.protobuf.Duration expiry_time = 105;
+    // [required] Market schedule in Pythnet format.
+    optional string market_schedule = 106;
+    // [required] Feed state
+    optional FeedState state = 107;
+
```
keyvankhademi, line 101: I think we should add asset_type here too. Because we will be using asset_type to determine what kind of feed type is expected (e.g funding-rate vs regular price).

ali-behjati: let's call it feed type.

keyvankhademi: @ali-behjati are you saying it's different from asset_type and we have both? Because I was thinking we could infer feed type based on asset_type. and multiple asset_types can have the same feed type.

Riateche: I also prefer to have feed type (Price/FundingRate) because those options are what we care about in the implementation. (It's already like this in the transaction protobuf). Asset type has a lot of options and I think it's better to leave it in the dynamic metadata.

lazer/publisher_sdk/proto/state.proto
Riateche: Feed state is a consumer-facing and publisher-facing property that's also used to limit access to the feeds. `is_enabled_in_shard` is a purely internal thing for moving live feeds between shards.

lazer/publisher_sdk/rust/src/dynamic_value.rs
Riateche: Well, it consumes the object, but the object is a reference. I don't think it's particularly unusual. If you look at [std impls](https://doc.rust-lang.org/std/convert/trait.From.html#implementors), there is a bunch of implementations on references. I think it's reasonable to `impl From` for references when conversion from a reference is more efficient than `clone()` followed by a conversion. In this case we can avoid cloning `Vec` and `BTreeMap` unnecessarily.

lazer/publisher_sdk/proto/state.proto
Riateche: What do you mean by "technically available"? Can you suggest a proper description of this state?

lazer/publisher_sdk/proto/state.proto
Riateche: I called it `FeedKind` because `type` is annoying to use in Rust 😅

lazer/publisher_sdk/proto/governance_instruction.proto
```diff
@@ -302,17 +304,31 @@ message SetPublisherActive {
     optional bool is_active = 1;
 }
 
-// Feed is inactive when added, meaning that it will be available to publishers but not to consumers.
+// Add a new feed. Refer to `Feed` message fields documentation.
 message AddFeed {
-    // [required] ID of the feed. Must be unique (within the shard).
+    // [required]
```
darunrs, line 309: Why was this description removed? Is it no longer accurate?

darunrs: If a feed ID needs to change as a result of moving to another shard which already has that ID (further, how do we prevent this from happening?), how would that impact subscribing to the feed? Is this ID separate from the customer/publisher facing feed ID?

lazer/publisher_sdk/proto/governance_instruction.proto
```diff
@@ -302,17 +304,31 @@ message SetPublisherActive {
...
     optional uint32 feed_id = 1;
-    // [required] Feed metadata. Some properties are required (name, exponent, etc.).
-    // Known properties must have the expected type.
-    // Additional arbitrary properties are allowed.
-    // (TODO: document known metadata properties)
-    optional DynamicValue.Map metadata = 2;
-    // IDs of publishers enabled for this feed.
-    repeated uint32 permissioned_publishers = 3;
+    // [required]
+    optional DynamicValue.Map metadata = 3;
```
darunrs, line 312: Any reason we skip field ID 2 if this is not in use yet?

lazer/publisher_sdk/proto/state.proto
darunrs: I think he means publishers can publish to them regardless of the state of the feed? We should change this in the future though. 

Perhaps you can specify they are feeds which can/can't be subscribed to?

lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,74 @@ message Publisher {
...
-    // [required] Feed symbol.
-    optional string symbol = 3;
-    // [required] Feed description.
-    optional string description = 4;
-    // [required] Feed asset type.
-    optional string asset_type = 5;
+// Feed kind determines the set of data fields available in the feed.
+// It also determines the kind of data accepted from publishers for this feed
+// (`PriceUpdate` or `FundingRateUpdate`).
+enum FeedKind {
```
darunrs, line 63: I intended to introduce this in some form. I will use the state enum to guard rail what updates a feed is able to accept.

lazer/publisher_sdk/proto/state.proto
```diff
@@ -45,67 +47,74 @@ message Publisher {
...
+
+
+    // [required] Feed status in the current shard. Disabled feeds will not be visible in
+    // the consumer API for the current shard. This setting should only be used
+    // to migrate a feed from one shard to another.
+    //
+    // If a feed is present in
     // multiple shards, it must only be active in one of them at each time.
-    // To enforce this, `pending_activation` and `pending_deactivation` fields
+    // To enforce this, `enable_in_shard_timestamp` and `disable_in_shard_timestamp` fields
```
darunrs, line 109: To confirm my understanding, Aggregator pulls all shards. So, it is able to handle this switch in logic?

lazer/publisher_sdk/proto/state.proto
keyvankhademi:
```suggestion
    // Inactive feeds are not available to consumers.
```

keyvankhademi: I think the only difference in INACTIVE and COMING_SOON is to inform the end user that inactive feed will never become active.

lazer/publisher_sdk/proto/governance_instruction.proto
```diff
@@ -321,43 +337,68 @@ message UpdateFeed {
...
     // Note: when adding a new variant here, update `Permissions` as well.
     oneof action {
-        UpdateFeedMetadata update_feed_metadata = 101;
-        ActivateFeed activate_feed = 102;
-        DeactivateFeed deactivate_feed = 103;
+        UpdateFeedProperties update_feed_properties = 101;
+        UpdateFeedMetadata update_feed_metadata = 102;
+        EnableFeedInShard enable_feed_in_shard = 103;
+        DisableFeedInShard disable_feed_in_shard = 104;
         google.protobuf.Empty remove_feed = 199;
```
keyvankhademi, line 346: we probably don't need this, since we will make feeds inactive instead of deleting.

lazer/publisher_sdk/proto/state.proto
keyvankhademi: we should use this for guard railing updates

lazer/publisher_sdk/proto/state.proto
ali-behjati: it kind of feels complex but got it. but how do you envision it? on my head i was like we do it as a new feed creation but i think the price feed state is replicated everywhere.

lazer/publisher_sdk/proto/state.proto
ali-behjati: we might let some people consume comin_soon ones too.

lazer/publisher_sdk/proto/state.proto
Riateche: It's not possible to do that deterministically without timestamps because shards have separate transaction queues. Aggregator is only able to achieve deterministic results within a shard because the inputs are always arriving in the same order.

lazer/publisher_sdk/proto/governance_instruction.proto
Riateche: I removed all field comments here because I don't want them to be duplicated with comments on the same fields in `Feed`. Otherwise they could go out of sync.

Feed IDs have to be globally unique. We don't have a way to enforce it in a shard, but we can check it in xc-admin I guess. If a feed is moved to another shard then it will always have the same ID. If the ID is not the same then it's not really the same feed anymore.

lazer/publisher_sdk/proto/state.proto
Riateche: In my view the sequence should be like this:

0. We have a feed live in the old shard.
1. We add a feed with the same feed ID and metadata to the new shard and set `enabled_in_shard = false`.
2. Agent discovers the feed in the new shard and starts publishing the same data to it.
3. We verify that the feed in the new shard is ready to go live (publisher data and aggregation are good).
4. We schedule the switch using `disable_in_shard_timestamp` in the old shard and `enable_in_shard_timestamp` in the new shard.
5. At this timestamp aggregators will stop producing updates for the feed in the old shard and start producing updates for it in the new shard. This makes the switch transparent to consumers, as long as both shards are available to the router that they're connected to.
6. We delete the feed in the old shard.

lazer/publisher_sdk/proto/state.proto
Riateche: Updated comments.

lazer/publisher_sdk/proto/governance_instruction.proto
Riateche: I prefer to have the option to delete them. Having lots of inactive feeds in state could become too costly.

lazer/publisher_sdk/proto/governance_instruction.proto
Riateche: I use the same IDs here as in `Feed`, and 2 is occupied there by `per_publisher` field. However, none of that really matters.

PR #2858, opened 2025-07-16
refactor(fuel): configuring forc and fuel-core upgrades to the Pyth fuel contract

ayushboss:
## Summary

Swaylend Fuel clients asked us to refactor the fuel contract to work with newer versions of the forc and fuel-core libraries (0.68.9 and 0.43.2 respectively). Simply changing the versions resulted in some compile errors, so code changes had to be made.

## Rationale

Client request for support for updated libraries.

## How has this been tested?

- [ ] Added new tests

I refactored the code until forc build worked and the existing integration tests through cargo test all passed.


target_chains/fuel/contracts/pyth-contract/src/main.sw
```diff
@@ -272,7 +272,56 @@ impl PythCore for Contract {
 
     #[storage(read, write), payable]
     fn update_price_feeds(update_data: Vec<Bytes>) {
-        update_price_feeds(update_data)
```
ayushboss: Note that there is duplicated code here in this file with this next chunk. The problem is that that block of code needs to be under a function with a payable attribute, but the payable attributes can only be applied to functions within an Abi.

target_chains/fuel/contracts/pyth-contract/src/main.sw
```diff
@@ -374,7 +374,7 @@ fn update_fee(update_data: Vec<Bytes>) -> u64 {
     total_fee(total_number_of_updates, storage.single_update_fee)
 }
 
-#[storage(read, write), payable]
+#[storage(read, write)]
```
ayushboss, line 377: One thing I wanted a second opinion on: one of the new issues that was introduced was that helper functions cannot be listed as "payable" anymore (only functions within the abi are allowed to have that tag). Leaving the tag in there stops the forc build. Removing the tag allows it to build, and all integration tests still pass. I still wanted to make sure that this is a valid removal (i.e. does the payable trait pass down from its calling function).

ali-behjati: Yeah i believe it is ok, i think Fuel has borrowed some semantics from EVM (just as it is a layer 2) and in EVM entrypoint functions that are not payable always check that msg.value == 0

target_chains/fuel/contracts/pyth-interface/src/pyth_merkle_proof.sw
```diff
@@ -20,8 +20,12 @@ fn node_hash(child_a: Bytes, child_b: Bytes) -> Bytes {
     let mut bytes = Bytes::with_capacity(41);
     bytes.push(MERKLE_NODE_PREFIX);
 
-    let a: b256 = child_a.into();
-    let b: b256 = child_b.into();
+    // let a: b256 = child_a.into();
+    // let b: b256 = child_b.into();
```
ali-behjati: remove?

target_chains/fuel/contracts/pyth-interface/src/pyth_merkle_proof.sw
```diff
@@ -20,8 +20,12 @@ fn node_hash(child_a: Bytes, child_b: Bytes) -> Bytes {
...
+
+    let a: b256 = b256::from_le_bytes(child_a.clone());
+    let b: b256 = b256::from_le_bytes(child_b.clone());
```
ali-behjati: I'm so confused how little endian could work here. generally raw bytes operations translate to big endian on numbers. also in the docs i see the example is similar.

https://fuellabs.github.io/sway/master/std/primitive.b256.html#method.from_be_bytes

If I'm right. I'm wondering how the tests passed with this. if I'm wrong, i like to know how :?

ali-behjati (changes requested):
i think you should use from_be_bytes instead. please let me know what you think and i'll look again.

target_chains/fuel/contracts/pyth-interface/src/pyth_merkle_proof.sw
ayushboss: I think I misinterpreted the structure of the array, BE is definitely right. For some reason the tests don't change state even when I completely mess with the code. I'll try to rebuild things and see if that changes anything.

PR #2859, opened 2025-07-16
feat(lazer): add resilient client in rust

keyvankhademi:
## Summary

This PR updates the rust client to be a resilient client with multiple connections.

## Rationale

We need this both for consumers and for monitor service.

## How has this been tested?

- [ ] Added new tests

Manually ran and tested the connections.


lazer/sdk/rust/client/src/client.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl PythLazerClient {
+    /// Creates a new client instance
+    ///
+    /// # Arguments
+    /// * `endpoints` - A vector of endpoint URLs
+    /// * `access_token` - The access token for authentication
+    /// * `num_connections` - The number of WebSocket connections to maintain
+    pub fn new(
+        endpoints: Vec<String>,
```
Riateche: `Url` type maybe?

lazer/sdk/rust/client/src/client.rs
```diff
@@ -0,0 +1,113 @@
...
+            receivers: Vec::with_capacity(num_connections),
+            backoff,
+        })
+    }
+
+    pub async fn start(&mut self) -> Result<mpsc::Receiver<AnyResponse>> {
+        let (sender, receiver) = mpsc::channel::<AnyResponse>(CHANNEL_CAPACITY);
+
+        for i in 0..self.num_connections {
+            let endpoint = self.endpoints[i % self.endpoints.len()].clone();
```
Riateche, line 73: It will panic if `endpoints` is empty. We should probably check for that in the constructor.

lazer/sdk/rust/client/src/client.rs
```diff
@@ -0,0 +1,113 @@
...
+            access_token,
+            num_connections,
+            ws_connections: Vec::with_capacity(num_connections),
+            receivers: Vec::with_capacity(num_connections),
+            backoff,
+        })
+    }
+
+    pub async fn start(&mut self) -> Result<mpsc::Receiver<AnyResponse>> {
+        let (sender, receiver) = mpsc::channel::<AnyResponse>(CHANNEL_CAPACITY);
```
Riateche: Channel capacity should be configurable by the caller.

lazer/sdk/rust/client/src/client.rs
```diff
@@ -0,0 +1,113 @@
...
+                self.access_token.clone(),
+                self.backoff.clone(),
+                sender.clone(),
+            );
+            self.ws_connections.push(connection);
+            self.receivers.push(receiver);
+        }
+
+        let streams: Vec<_> = self.receivers.drain(..).map(ReceiverStream::new).collect();
+        let mut merged_stream = stream::select_all(streams);
```
Riateche: I think it's easier and more efficient to use a single channel here.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
```diff
@@ -0,0 +1,193 @@
...
+        response_sender: mpsc::Sender<AnyResponse>,
+        request_receiver: &mut mpsc::Receiver<Request>,
+    ) -> Result<()> {
+        let mut last_failure_time = Instant::now();
+
+        loop {
+            if let Err(e) = self.start(response_sender.clone(), request_receiver).await {
+                if last_failure_time.elapsed() > BACKOFF_RESET_DURATION {
+                    self.backoff.reset();
+                }
```
Riateche: This feels like a hack. I think we should reset the backoff object in `start()` when we successfully connect and subscribe.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+
+    pub async fn subscribe(
+        &mut self,
+        ws_connection: &mut PythLazerWSConnection,
+        request: SubscribeRequest,
+    ) -> Result<()> {
+        self.subscriptions.push(request.clone());
+        return ws_connection.subscribe(request).await;
```
Riateche: nit: unneeded return

lazer/sdk/rust/client/src/resilient_ws_connection.rs
```diff
@@ -0,0 +1,193 @@
...
+            .position(|r| r.subscription_id == request.subscription_id)
+        {
+            self.subscriptions.remove(index);
+        } else {
+            warn!(
+                "Unsubscribe called for non-existent subscription: {:?}",
+                request.subscription_id
+            );
+        }
+        return ws_connection.unsubscribe(request).await;
```
Riateche: nit: unneeded return

Riateche (changes requested):
Very nice!

lazer/sdk/rust/client/src/client.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///
+    /// # Arguments
+    /// * `endpoints` - A vector of endpoint URLs
+    /// * `access_token` - The access token for authentication
+    /// * `num_connections` - The number of WebSocket connections to maintain
+    pub fn new(
+        endpoints: Vec<String>,
+        access_token: String,
+        num_connections: usize,
+        backoff: ExponentialBackoff,
```
ali-behjati, line 46: this seems to be part of your public interface. you need to reexport it to downstream consumers. Other thing is to have a wrapper that has what you need (because i see you are not supporting everything there)

lazer/sdk/rust/client/src/client.rs
keyvankhademi: Done

lazer/sdk/rust/client/src/client.rs
keyvankhademi: Fixed.

lazer/sdk/rust/client/src/client.rs
keyvankhademi: Done. Internal channel don't need to be configurable right?

lazer/sdk/rust/client/src/client.rs
keyvankhademi: right. fixed.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
keyvankhademi: I want the behavior to be a little different than that. For example, if client can connect and subscribe, but get disconnected immediately after we should not reset.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
keyvankhademi: fixed.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
keyvankhademi: fixed.

lazer/sdk/rust/client/src/client.rs
keyvankhademi: added a builder wrapper.

Riateche: A builder would also be nice for the `PythLazerClient` itself. You can move `channel_capacity` from `start()` argument to the builder. The builder can also provide default values for everything except `access_token`.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
Riateche: This code means that when the endpoint is down, rather than doing a proper exponential backoff, it will start spamming the endpoint again every 10 seconds. I think what you want can be better achieved with this:
```
pub async fn start(...) {
    //...
    let stream = ws_connection.start().await?;
    let connected_at = Instant::now();
    let mut num_responses = 0;
    let mut confirmed_healthy = false;
    //...
             Ok(response) => {
                 if !confirmed_healthy {
                     num_responses += 1;
                     if num_responses >= CONFIRM_HEALTHY_NUM_RESPONSES &&
                          connected_at.elapsed() > CONFIRM_HEALTHY_DELAY {
                          confirmed_healthy = true;
                          self.backoff.reset();
                      }
                 }
                 sender
                     .send(response)
                     .await
                     .context("Failed to send response")?;
             }
    //...
}





```

keyvankhademi: It only happens when `max_interval` is greater than 10 seconds. Otherwise it will try connecting every `max_interval`. 
I think the best solution is to set BACKOFF_RESET_DURATION to be 2*max_interval.

lazer/sdk/rust/client/src/client.rs
keyvankhademi: good idea.

keyvankhademi: Done.

lazer/sdk/rust/client/src/resilient_ws_connection.rs
keyvankhademi: I fixed it another way. Now we check the duration of the connection and if was connected for more than BACKOFF_RESET_DURATION and timout + second it was connected successfully and we reset the backoff.

Riateche (approved):

lazer/sdk/rust/client/src/backoff.rs
```diff
@@ -0,0 +1,69 @@
+use std::time::Duration;
+
+use backoff::{
+    default::{INITIAL_INTERVAL_MILLIS, MAX_INTERVAL_MILLIS, MULTIPLIER, RANDOMIZATION_FACTOR},
+    ExponentialBackoff, ExponentialBackoffBuilder,
+};
+
+#[derive(Debug)]
+pub struct PythLazerExponentialBackoffBuilder {
```
ali-behjati, line 9: document here as well? maybe default as well?

lazer/sdk/rust/client/src/backoff.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    pub fn build(&self) -> ExponentialBackoff {
+        ExponentialBackoffBuilder::default()
+            .with_initial_interval(self.initial_interval)
+            .with_randomization_factor(self.randomization_factor)
+            .with_multiplier(self.multiplier)
+            .with_max_interval(self.max_interval)
+            .with_max_elapsed_time(None)
+            .build()
+    }
```
ali-behjati, line 68: i'm fine with this but generally my take was reexposing what we use in the API and we still use ExponentialBackoff from the backoff crate. now that you have a builder it makes sense that you have your own wrapper type for the api of the lazer client. later when you convert it you can remove the following error as well: `bail!("max_elapsed_time is not supported in Pyth Lazer client");`

lazer/sdk/rust/client/src/backoff.rs
keyvankhademi: The reason I didn't implement default is that we need to have access token. Default doesn't make sense since it will be in an invalid state.

lazer/sdk/rust/client/src/backoff.rs
keyvankhademi: yeah it's a good improvement. I'll add this + docs in another PR to unblock the progress of monitor for now.

PR #2861, opened 2025-07-17
feat: Support Sui Price Update with Coin Input

darunrs:
## Summary
Adds a new SDK function which takes coins as a function argument. This allows the price update calls to work without accessing the TX gas. 

## Rationale
Request from customer in [this thread](https://dev-forum.pyth.network/t/updatepricefeeds-without-accessing-tx-gas/292). They are using some gas station which supplies the correct amount of gas. But it seems that they require the SDK to not access TX.gas.  

## How has this been tested?

- [ ] Added new tests

Configured the original function to generate the coins and then pass them to the new function. The feed updated correctly.

### DIFF
--- a/target_chains/sui/sdk/js/src/client.ts
+++ b/target_chains/sui/sdk/js/src/client.ts
@@ -6,7 +6,7 @@ import { HexString } from "@pythnetwork/price-service-client";
 import { Buffer } from "buffer";
 
 const MAX_ARGUMENT_SIZE = 16 * 1024;
-type Coin = {
+type NestedTransactionResult = {
   $kind: "NestedResult";
   NestedResult: [number, number];
 };
@@ -112,7 +112,7 @@ export class SuiPythClient {
     tx: Transaction,
     updates: Buffer[],
     packageId: string,
-  ): Promise<any> {
+  ): Promise<NestedTransactionResult> {
     if (updates.length > 1) {
       throw new Error(
         "SDK does not support sending multiple accumulator messages in a single transaction",
@@ -144,7 +144,7 @@ export class SuiPythClient {
     packageId: string,
     feedIds: HexString[],
     priceUpdatesHotPotato: any,
-    coins: Coin[],
+    coins: NestedTransactionResult[],
   ) {
     const priceInfoObjects: ObjectId[] = [];
     let coinId = 0;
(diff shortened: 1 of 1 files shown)


target_chains/sui/sdk/js/src/client.ts
```diff
@@ -104,19 +108,11 @@ export class SuiPythClient {
...
-  async updatePriceFeeds(
+  async verifyVaasAndGetHotPotato(
     tx: Transaction,
     updates: Buffer[],
-    feedIds: HexString[],
-  ): Promise<ObjectId[]> {
-    const packageId = await this.getPythPackageId();
-
+    packageId: string,
+  ): Promise<any> {
```
ali-behjati: please don't use any. if the type is not known and you can extract it later use `unknown`

target_chains/sui/sdk/js/src/client.ts
```diff
@@ -6,6 +6,10 @@ import { HexString } from "@pythnetwork/price-service-client";
 import { Buffer } from "buffer";
 
 const MAX_ARGUMENT_SIZE = 16 * 1024;
+type Coin = {
```
ali-behjati: i wonder whether they had a type for this themselves.

cosmossudofi (approved):

target_chains/sui/sdk/js/src/client.ts
darunrs: Gotcha, will do!

target_chains/sui/sdk/js/src/client.ts
darunrs: Yeah I couldn't find one myself. Their TS docs use the splitCoins function but they never provide an example of what explicit type it is. In their github repo, they have that function return this type: `Extract<Argument, { Result: unknown }> &
	Extract<Argument, { NestedResult: unknown }>[]`

In any case, I figured why not just use the type the IDE supplies me.

darunrs: So there is a [TransactionResult](https://github.com/MystenLabs/ts-sdks/blob/bdfb59f39e2baa32c3e467d55e82eca73a3f0268/packages/typescript/src/transactions/Transaction.ts#L36) type which almost matches this. But we seem to explicitly use the nestedResult version. So I just redefined it and use that. We can use the SDK type if we dont do [hotPotato] and such but I dont know what that would do and don't want to change things so drastically right now.

ali-behjati (approved):

PR #2868, opened 2025-07-18
fix(fortuna): Fix nonce increment issues on high usage chains

jayantk:
## Summary

We are seeing high latency on blast responses. After investigating, @tejasbadadare and I determined that the root cause is that we are incorrectly incrementing the local nonce on occasions where we do not send the transaction to the RPC. This specifically happens when a transaction reverts -- we are incrementing the nonce before identifying that the tx reverts.

This PR adds an estimate_gas call to catch reverts prior to filling the nonce in the transaction.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/src/eth_utils/utils.rs
+++ b/apps/fortuna/src/eth_utils/utils.rs
@@ -196,7 +196,7 @@ pub async fn submit_tx_with_backoff<T: Middleware + NonceManaged + 'static>(
 
 #[allow(clippy::large_enum_variant)]
 pub enum SubmitTxError<T: Middleware + NonceManaged + 'static> {
-    GasUsageEstimateError(ContractError<T>),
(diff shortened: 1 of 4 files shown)


apps/fortuna/src/eth_utils/utils.rs
```diff
@@ -211,8 +211,8 @@ where
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            SubmitTxError::GasUsageEstimateError(e) => {
-                write!(f, "Error estimating gas for reveal: {e:?}")
+            SubmitTxError::GasUsageEstimateError(tx, e) => {
```
jayantk, line 214: drive-by for easier debugging

tejasbadadare (approved):

PR #2869, opened 2025-07-18
fix(pyth-lazer-agent) Allow the relayer endpoint path in config

bplatak:
## Summary
Allow the full URL (including path) to be specified for the relayers. Also bump to 0.1.5 for release

## Rationale
The path structure differs and prod and staging so we can't simply hardcode it as `/v1/transactions` 

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/pyth-lazer-agent/Cargo.toml
+++ b/apps/pyth-lazer-agent/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-agent"
-version = "0.1.5"
+version = "0.2.0"
 edition = "2024"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


apps/pyth-lazer-agent/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-agent"
-version = "0.1.4"
+version = "0.1.5"
```
keyvankhademi:
```suggestion
version = "0.2.0"
```

keyvankhademi (approved):

PR #2880, opened 2025-07-22
@bplatak feat(pyth-lazer-agent) Make the logging format configurable

bplatak:
## Summary
Add an option to switch between different logging formats (json,compact,pretty).

## Rationale
Customer request 

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/pyth-lazer-agent/src/main.rs
+++ b/apps/pyth-lazer-agent/src/main.rs
@@ -16,7 +16,6 @@ mod relayer_session;
 mod websocket_utils;
 
 #[derive(Parser, Deserialize)]
-#[serde(rename_all = "kebab-case")]
 #[command(version)]
 struct Cli {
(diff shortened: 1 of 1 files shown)


apps/pyth-lazer-agent/src/main.rs
```diff
@@ -14,38 +15,64 @@ mod publisher_handle;
 mod relayer_session;
 mod websocket_utils;
 
-#[derive(Parser)]
+#[derive(Parser, Deserialize)]
+#[serde(rename_all = "kebab-case")]
```
ali-behjati: i thought clap parser would do kebab case by default :?

ali-behjati (approved):

apps/pyth-lazer-agent/src/main.rs
bplatak: so it does. learn something new everyday

PR #2887, opened 2025-07-23
refactor: divide lib.rs into governance and pyth_operations modules

ayushboss:
## Summary

- Move governance functions to governance.rs module
- Move price feed and update operations to pyth_operations.rs module
- Keep shared utilities and main struct in lib.rs
- Maintain all existing functionality and public interfaces
- Use internal/wrapper pattern to comply with Stylus SDK constraints

- [ ] Added new tests


tejasbadadare (commented):
Hm I don't think there's much use in pulling the internal impls into `pyth_operations.rs`, imo we should leave those in `lib.rs` to reduce unnecessary indirection. Governance is a logically distinct path so it makes sense to pull that out.

PR #2891, opened 2025-07-24
feat(pyth-lazer) Add support for FundingRateInterval in lazer protos

bplatak:
## Summary
Add a new field to represent funding rate interval to funding rate update messages in protobuf objects and jrpc domain. 

## Rationale
We're adding support for variable funding rate updates in lazer as part of PF-278 

## How has this been tested?

- [x] Added new tests


lazer/publisher_sdk/proto/publisher_update.proto
```diff
@@ -54,9 +54,18 @@ message PriceUpdate {
 
 // feed update containing data relating to funding rate
 message FundingRateUpdate {
+    enum FundingRateInterval {
+      INTERVAL_1H = 0;
+      INTERVAL_4H = 1;
+      INTERVAL_8H = 2;
+    }
```
Riateche: Doesn't seem like a good idea to restrict the possible set of values here. Why not use a Duration?

ali-behjati: yeah, maybe in the future people have half an hour too :D

lazer/sdk/rust/protocol/src/payload.rs
```diff
@@ -52,6 +52,7 @@ pub struct AggregatedPriceFeedData {
     pub confidence: Option<Price>,
     pub funding_rate: Option<Rate>,
     pub funding_timestamp: Option<TimestampUs>,
+    pub funding_rate_interval: Option<u64>,
```
Riateche: We should use `DurationUs` here.

PR #2895, opened 2025-07-24
chore(fortuna) Config API

aditya520:
## How has this been tested?

- [ ] Added new tests


apps/fortuna/src/api/config.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub contract_addr: String,
+    pub reveal_delay_blocks: u64,
+    pub gas_limit: u32,
+    pub fee: u128,
+}
+
+pub async fn get_chain_configs(
+    State(_state): State<ApiState>,
+) -> anyhow::Result<Json<Vec<ChainConfigSummary>>, RestError> {
+    let yaml_content = std::fs::read_to_string("config.yaml").map_err(|e| {
```
tejasbadadare: We need to load the config path from the [RunOptions](https://github.com/pyth-network/pyth-crosschain/blob/955a805e85dc093b9fff0d3f6af1268052ca358c/apps/fortuna/src/config/run.rs#L7), not have it hardcoded :) you'll have to add it to ApiState [here](https://github.com/pyth-network/pyth-crosschain/blob/11f6e6a7dc2466b617ed954d0c89e8d118507e38/apps/fortuna/src/command/run.rs#L57). Then you can access it in get_chain_configs via the `_state` param.

It would be great to add a quick test to the test module in api.rs as well.

tejasbadadare: Also, you could potentially just pass the already parsed config from `run` since it parses the config before starting the API server

apps/fortuna/src/api.rs
```diff
@@ -73,13 +77,17 @@ pub struct ApiState {
     pub metrics: Arc<ApiMetrics>,
 
     pub explorer_metrics: Arc<ExplorerMetrics>,
+
+    /// Parsed configuration
```
tejasbadadare: unnecessary comment

apps/fortuna/src/api.rs
```diff
@@ -73,13 +77,17 @@ pub struct ApiState {
...
+    pub config: Config,
 }
 
 impl ApiState {
     pub async fn new(
         chains: Arc<RwLock<HashMap<ChainId, ApiBlockChainState>>>,
         metrics_registry: Arc<RwLock<Registry>>,
         history: Arc<History>,
+        config: Config,
```
tejasbadadare: let's make it a borrowed `&Config` instead of an owned `Config`.

apps/fortuna/src/api.rs
```diff
@@ -534,4 +575,199 @@ mod test {
...
+
+        let avax_state = BlockchainState {
+            id: "avalanche".into(),
+            network_id: 43114,
+            state: Arc::new(avax_state),
+            contract: avax_read.clone(),
+            provider_address: PROVIDER,
+            reveal_delay_blocks: 2,
+            confirmed_block_status: BlockStatus::Latest,
+        };
```
tejasbadadare: do we actually need to set up these things just to test the chains/config API? isn't the only dependency the Config struct?

apps/fortuna/src/api.rs
aditya520: Why will that be better?

tejasbadadare: Typically you want to borrow (use a reference/pointer to some data instead of copying it,) especially if that data will outlive your function.  Good stuff in this thread: https://www.reddit.com/r/learnrust/s/bgqkueBmRr

PR #2896, opened 2025-07-24
feat(pyth-lazer-agent) Allow conditionally disabling success responses

bplatak:
## Summary

Add support for JSON-RPC notifications. Now if the `id` parameter is omitted the agent will not send acknowledgments for successful updates. Errors will still be sent back to the client, now without any id, but will include the full update to help the client correlate them with the original request.

## Rationale

Publishers see issues due to websocket contention largely caused by the acks being sent back to the publisher and have requested this change. PF-334

## How has this been tested?

- [x] Added new tests


apps/pyth-lazer-agent/README.md
```diff
@@ -1,12 +1,12 @@
 # pyth-lazer-agent
 
 pyth-lazer-agent is intended to be run by Lazer publishers analogous to [pyth-agent](https://github.com/pyth-network/pyth-agent)
-for pythnet publishers. Currently it retains [the existing Lazer publishing interface](https://github.com/pyth-network/pyth-examples/tree/main/lazer/publisher),
+for pythnet publishers. Currently, it retains [the existing Lazer publishing interface](https://github.com/pyth-network/pyth-examples/tree/main/lazer/publisher),
```
merolish, line 4: Not really needed, eh.

apps/pyth-lazer-agent/README.md
```diff
@@ -1,12 +1,12 @@
...
 but will batch and sign transactions before publishing them to Lazer.
 
 ## Keypair
 
-You will need to generate an ed25519 keypair and provide the pubkey to the Lazer team. `solana-keygen` is the recommended utility.
+You will need to generate an ed25519 keypair and provide the pubkey to the Lazer team. `solana-keygen` it is the recommended utility.
```
merolish: Nope.

apps/pyth-lazer-agent/README.md
```diff
@@ -16,9 +16,21 @@ pyth-lazer-agent will need to configure access to this keypair file to sign tran
...
+# Download the cargo package
+cargo install pyth-lazer-agent
+
+# Add .cargo/bin to PATH
+export PATH="$PATH:~/.cargo/bin"
+
+# Run the agent
+pyth-lazer-agent --help
+```
+
```
merolish, line 30: Would also show the actual run usage which is just `pyth-lazer-agent --config /path/to/config.toml`.

apps/pyth-lazer-agent/README.md
```diff
@@ -16,9 +16,21 @@ pyth-lazer-agent will need to configure access to this keypair file to sign tran
...
 ### From source
-Please check [rust-toolchain](rust-toolchain.toml) to see the version of Rust needed to build (currently 1.88).
-You will also need SSL and CA certificates. `cargo build` should then work as usual.
+Please check [rust-toolchain](/rust-toolchain.toml) to see the version of Rust needed to build (currently 1.88).
+You will also need SSL and CA certificates. `cargo build` it should then work as usual.
```
merolish: Nope.

apps/pyth-lazer-agent/README.md
```diff
@@ -42,10 +54,5 @@ publish_interval_duration = "25ms"
 - `relayers_urls`: The Lazer team will provide these.
 - `publish_keypair_path`: The keypair file generated with `solana-keygen` or similar.
 - `authorization_token`: The Lazer team will provide this or instruct that it can be omitted.
-- `listen_address`: The local port the agent will be listening on; can be anything you want.
+- `listen_address`: The local port the agent will be listening to on; can be anything you want.
```
merolish: Nope.

apps/pyth-lazer-agent/config/config.toml
```diff
@@ -1,4 +1,4 @@
-relayer_urls = ["wss://relayer-0.pyth-lazer.dourolabs.app/v1/transaction", "wss://relayer-0.pyth-lazer.dourolabs.app/v1/transaction"]
+relayer_urls = ["wss://relayer.pyth-lazer-staging.dourolabs.app/v1/transaction", "wss://relayer-1.pyth-lazer-staging.dourolabs.app/v1/transaction"]
```
darunrs, line 1: It's probably worth breaking these out as consts so that we can clearly understand they're "STAGING_RELAYER_1_WS_URL" or something.

apps/pyth-lazer-agent/src/jrpc_handle.rs
```diff
@@ -242,6 +189,81 @@ fn filter_symbols(
...
+    request_params: FeedUpdateParams,
+    request_id: Option<i64>,
+) -> anyhow::Result<()> {
+    match lazer_publisher
+        .push_feed_update(request_params.clone().into())
+        .await
+    {
+        Ok(_) => {
+            if let Some(request_id) = request_id {
+                send_json(
```
darunrs, line 204: Do we respond like this on every update or just some? I think acks should be on some timer rather than driven by events from the publisher. Or maybe that can be configured by the publisher. Like they can set it to ack each update when testing, change it to interval when its running well, and disable them when fully running it. What do you think?

darunrs (approved):
Nice! I know many publishers were hesitant to have acks sent over the legacy WS connection as they've all noticed slowdowns so this doesn't totally surprise me.

apps/pyth-lazer-agent/README.md
bplatak: not entirely sure what happened to my linter there 🤔

apps/pyth-lazer-agent/README.md
bplatak: not necessary but technically correct 🤷

apps/pyth-lazer-agent/config/config.toml
bplatak: in config? does toml have support for some consts i'm missing?

apps/pyth-lazer-agent/src/jrpc_handle.rs
bplatak: When `id` is specified we respond to each message as it's processed. We could batch the responses but I'm not sure if there's a big usecase here - most publishers don't seem to care about acks at all and just want to get the errors out. It would also be a little weird trying to fit that within json-rpc specs although not impossible. For now I'd like to keep it as is until somebody asks for it.

apps/pyth-lazer-agent/README.md
bplatak: imho `pyth-lazer-agent --help` already explains how to use it in a lot more detail than we need to here

PR #2899, opened 2025-07-25
feat(lazer): create js solana sdk package

Riateche:
## How has this been tested?

- [ ] Added new tests


lazer/sdk/js-solana/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import type { PythLazerSolanaContract } from "./idl/pyth-lazer-solana-contract.js";
```
ali-behjati, line 1: was the `.js` in the end necessary :?

lazer/contracts/solana/package.json
```diff
@@ -6,6 +6,7 @@
     "fix:format": "prettier --write **/*.*",
     "test:format": "prettier --check **/*.*",
     "test:anchor": "CARGO_TARGET_DIR=\"$PWD/target\" RUSTUP_TOOLCHAIN=nightly-2025-04-15 anchor test",
+    "update-idl": "RUSTUP_TOOLCHAIN=nightly-2025-04-15 anchor build && cp target/types/pyth_lazer_solana_contract.ts ../../sdk/js-solana/src/idl/pyth-lazer-solana-contract.ts && cp target/idl/pyth_lazer_solana_contract.json ../../sdk/js-solana/src/idl/pyth-lazer-solana-contract.json",
```
cprussin: I don't personally like having a task in one package which reaches into and updates another package.  Personally, I'd put this as a `build:idl` sub-task on the `@pythnetwork/pyth-lazer-solana-sdk` package instead, given it's a task that builds files for that package.

If you really want to be pedantic, you leave something like `RUSTUP_TOOLCHAIN=nightly-2025-04-15 anchor build` here, and then add the `cp` commands in `@pythnetwork/pyth-lazer-solana-sdk`, and set up turbo so the latter depends on the former.

lazer/sdk/js-solana/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "scripts": {
+    "build:cjs": "swc src -d dist/cjs --strip-leading-paths --copy-files -C module.type=commonjs && echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json",
+    "build:esm": "swc src -d dist/esm --strip-leading-paths --copy-files -C jsc.experimental.keepImportAttributes=true && echo '{\"type\":\"module\"}' > dist/esm/package.json",
+    "fix:lint": "eslint --fix . --max-warnings 0",
+    "test:lint": "eslint . --max-warnings 0",
+    "test:types": "tsc",
+    "test:format": "prettier --check .",
+    "fix:format": "prettier --write .",
+    "doc": "typedoc --out docs/typedoc src"
```
cprussin: minor nit but I usually name this as `build:docs`

lazer/sdk/js-solana/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "devDependencies": {
+    "@cprussin/eslint-config": "catalog:",
+    "@cprussin/tsconfig": "catalog:",
+    "@swc/cli": "^0.7.8",
+    "@swc/core": "^1.13.2",
+    "@types/node": "^18.19.54",
+    "eslint": "catalog:",
+    "prettier": "catalog:",
+    "typedoc": "^0.26.8",
```
cprussin: I suggest moving these dependencies up to `pnpm-workspace.yaml` and marking them as `catalog:` dependencies; in general I think we should only be using numbered dependencies for places where some package needs to be held back from the rest of the monorepo for some reason.

lazer/sdk/js-solana/src/idl/pyth-lazer-solana-contract.json
```diff
@@ -0,0 +1,464 @@
+{
```
cprussin, line 1: If these files are generated, I suggest not checking them in.  Instead, ensure the build tasks are correctly ordered so the files get generated as part of building this package.  I personally really prefer not to check these files in, since it is effectively a cache and checking them in means you are now susceptible to cache sync issues.

lazer/sdk/js-solana/src/index.ts
cprussin: This is needed when using `nodenext` module resolution, in which typescript does not transform import paths at all.  See https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution for reference.

In general, `nodenext` is the preferred module resolution for anything that supports node (and IMO, any published lib should be using it as it's the most future-proof option).  The short explanation is that nodenext leaves imports alone when transpiling, meaning the import path gets passed directly to node.  Since node requires extensions when using esm, so then does the nodenext module resolution strategy (this also explains why the extension is `.js` and not `.ts`).

For more context, eventually node will change so that type annotations are just ignored at runtime, and at that point typescript will no longer be a transpiler and just a type checker, similar to mypy in python.

lazer/sdk/js-solana/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
```
cprussin: Since you moved to swc, you no longer need this file.

pnpm-workspace.yaml
```diff
@@ -185,3 +186,5 @@ onlyBuiltDependencies:
   - web3
   - web3-bzz
   - web3-shh
+ignoredBuiltDependencies:
```
cprussin: You should add this to `onlyBuiltDependencies`, not `ignoredBuiltDependencies`.  Adding it here will cause it to break in any environment where the binaries aren't already present.

cprussin (changes requested):
Most important change is the change to `pnpm-workspace.yaml` since that will cause swc to be broken in some environments.

lazer/contracts/solana/package.json
Riateche: I moved the update-idl task to the `@pythnetwork/pyth-lazer-solana-sdk` package. I agree that generating IDL as part of the build process would be a cleaner way to do this, but I don't want to introduce anchor as a build dependency to JS packages. We tried to do that before with the Lazer Solana contract tests, and it was constantly causing problems for people.

cprussin: yeah that makes sense

PR #2902, opened 2025-07-26
fix(fortuna): Fix bug in get_request_v2

jayantk:
## Summary

I introduced a bug in this method in the PR when I upgraded fortuna to handle the v2 contract. This bug is breaking the stats on callback errors.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/src/chain/ethereum.rs
+++ b/apps/fortuna/src/chain/ethereum.rs
@@ -274,13 +274,17 @@ impl<T: JsonRpcClient + 'static> EntropyReader for PythRandom<Provider<T>> {
             .get_request_v2(provider_address, sequence_number)
             .call()
             .await?;
-        Ok(Some(reader::Request {
-            provider: request.provider,
-            sequence_number: request.sequence_number,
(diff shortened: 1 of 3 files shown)


apps/fortuna/src/chain/ethereum.rs
```diff
@@ -274,13 +274,17 @@ impl<T: JsonRpcClient + 'static> EntropyReader for PythRandom<Provider<T>> {
...
             .call()
             .await?;
-        Ok(Some(reader::Request {
-            provider: request.provider,
-            sequence_number: request.sequence_number,
-            block_number: request.block_number,
-            use_blockhash: request.use_blockhash,
-            callback_status: reader::RequestCallbackStatus::try_from(request.callback_status)?,
-        }))
+        if request.sequence_number == 0 {
```
jayantk, line 277: get_request returns inactive requests and then it's the caller's responsibility to check. see https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/ethereum/contracts/contracts/entropy/Entropy.sol#L1072

PR #2904, opened 2025-07-28
chore(fortuna)-add-network-id-api

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/fortuna/src/api/config.rs
+++ b/apps/fortuna/src/api/config.rs
@@ -1,5 +1,5 @@
 use {
-    crate::api::{ApiState, RestError},
+    crate::api::{ApiBlockChainState, ApiState, RestError},
     axum::{extract::State, Json},
     serde::Serialize,
 };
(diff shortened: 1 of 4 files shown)


apps/fortuna/src/api/config.rs
```diff
@@ -1,29 +1,35 @@
...
+            _ => 0,
+        };
         configs.push(ChainConfigSummary {
             name: name.clone(),
+            network_id,
             contract_addr: format!("0x{:x}", chain.contract_addr),
             reveal_delay_blocks: chain.reveal_delay_blocks,
             gas_limit: chain.gas_limit,
-            fee: chain.fee,
+            default_fee: chain.fee,
```
tejasbadadare, line 32: what's the default fee going to be used for? shouldn't we show up to date fees using getFeeV2()?

aditya520: Yes, the default fee won't be used. I will use getFeeV2(). I realized it today after our discussion. But keeping it here won't be harmful.

tejasbadadare (approved):

PR #2910, opened 2025-07-30
fix(lazer): include .d.ts files in solana sdk js package

Riateche:
## Summary

- Run tsc to generate .d.ts files
- Update package.json to point at the location of .d.ts files

## Rationale

The package was missing .d.ts files, so it wasn't possible to use it.

## How has this been tested?

- [ ] Added new tests


cprussin (approved):
ah apologies that I missed this

PR #2914, opened 2025-07-30
fix(fuel): further fixes, contract deployment

ayushboss:
## Summary

refactored some of the byte conversions to avoid conversion issues. 

## Rationale

Merkle hashes are 20 hex digits long, and old versions of Sway could autoconvert those to b256s during the merkle parsing and confirmation. The update to fuel get rid of that, so I initially replaced that conversion with a simple from_be_bytes, but that method needs an input with 32 bytes. To remedy that, I padded those merkle proofs before conversion.

Additionally, the beta 5 fuel test net endpoint is not up anymore, so I replaced it with the testnet.fuel.network endpoint in the deployment script.

## How has this been tested?

- [ ] Added new tests

Confirmed deployment on Fuel testnet. Additional diagnostic updates and get_price passes as well

### DIFF
--- a/target_chains/fuel/contracts/pyth-contract/src/main.sw
+++ b/target_chains/fuel/contracts/pyth-contract/src/main.sw
@@ -935,3 +935,4 @@ fn verify_governance_vm(encoded_vm: Bytes) -> WormholeVM {
     set_last_executed_governance_sequence(vm.sequence);
     vm
 }
+
--- a/target_chains/fuel/contracts/tests/utils/interface/wormhole_guardians.rs
+++ b/target_chains/fuel/contracts/tests/utils/interface/wormhole_guardians.rs
@@ -59,10 +59,11 @@ pub(crate) async fn submit_new_guardian_set(
     contract: &PythOracleContract<WalletUnlocked>,
     encoded_vm: Bytes,
 ) -> CallResponse<()> {
-    let result = contract
+    contract
         .methods()
         .submit_new_guardian_set(encoded_vm)
         .call()
         .await
-        .map_err(handle_error).unwrap()
+        .map_err(handle_error)
+        .unwrap()
 }


target_chains/fuel/contracts/pyth-interface/src/pyth_merkle_proof.sw
```diff
@@ -20,8 +20,24 @@ fn node_hash(child_a: Bytes, child_b: Bytes) -> Bytes {
...
+        i += 1;
+    }
+
+    // Append the 20-byte data
+    full_address_child_a.append(child_a);
+    full_address_child_b.append(child_b);
+
+    // Convert to b256
+    let a: b256 = b256::from_be_bytes(full_address_child_a);
+    let b: b256 = b256::from_be_bytes(full_address_child_b);
```
tejasbadadare: How did the b256::from_be_bytes behavior change? Previously it was happy with fewer than 32 bytes as input and it would left-pad zeroes?

ayushboss: so before we updated fuel it used .into(), but i think they removed that conversion trait in the update. afterwards, i switched it to from_be_bytes, and it seemingly passed all the build tests but when it actually did result in errors when i deployed

target_chains/fuel/contracts/pyth-interface/src/data_structures/wormhole_light.sw
```diff
@@ -115,8 +115,21 @@ impl GuardianSetUpgrade {
...
+                full_address_key.push(0u8);
+                j += 1;
+            }
+
+            full_address_key.append(key);
+
+            let key: b256 = b256::from_be_bytes(full_address_key.clone());
+
+
+            new_guardian_set.keys.push(key);
```
tejasbadadare: Can you please add the comments as to why the zero padding is needed (same as pyth_merkle_proof.sw)

target_chains/fuel/contracts/pyth-interface/src/data_structures/wormhole_light.sw
```diff
@@ -115,8 +115,20 @@ impl GuardianSetUpgrade {
         while i < guardian_length {
             let (_, slice) = encoded_upgrade.split_at(index);
             let (key, _) = slice.split_at(20);
-            let key: b256 = b256::from_be_bytes(key.clone());
-            new_guardian_set.keys.push(key.rsh(96));
+            let mut full_address_key = Bytes::with_capacity(32);
+
+            // Sway's from_be_bytes expects a 32-byte array, so we pad with zeros
```
ali-behjati, line 120: looks like this is repeated in several places, maybe a helper function would be good?

PR #2921, opened 2025-08-04
feat: add cached methods

alexcambose:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/PriceFeed/get-feed.tsx
+++ b/apps/insights/src/components/PriceFeed/get-feed.tsx
@@ -1,20 +1,16 @@
 import { notFound } from "next/navigation";
 
-import { Cluster, getFeeds } from "../../services/pyth";
+import { getFeedsCached } from "../../server/pyth";
+import { Cluster } from "../../services/pyth";
 
 export const getFeed = async (params: Promise<{ slug: string }>) => {
   "use cache";
 
-  const [{ slug }, feeds] = await Promise.all([params, getPythnetFeeds()]);
+  const [{ slug }, feeds] = await Promise.all([params, getFeedsCached(Cluster.Pythnet)]);
   const symbol = decodeURIComponent(slug);
   return {
     feeds,
     feed: feeds.find((item) => item.symbol === symbol) ?? notFound(),
     symbol,
   } as const;
-};
-
-const getPythnetFeeds = async () => {
-  "use cache";
-  return getFeeds(Cluster.Pythnet);
 };
--- a/apps/insights/src/server/pyth.ts
+++ b/apps/insights/src/server/pyth.ts
@@ -1,0 +1,11 @@
+import { Cluster, getFeeds, getPublishersForFeed } from "../services/pyth";
+
+export const getFeedsCached = async (cluster: Cluster) => {
+  "use cache";
+  return getFeeds(cluster);
+};
+
+export const getPublishersForFeedCached = async (cluster: Cluster, symbol: string) => {
+  "use cache";
+  return getPublishersForFeed(cluster, symbol);
+};
--- a/apps/insights/src/app/price-feeds/[slug]/layout.ts
+++ b/apps/insights/src/app/price-feeds/[slug]/layout.ts
@@ -2,7 +2,8 @@ import type { Metadata } from "next";
 import { notFound } from "next/navigation";
 import type { ReactNode } from "react";
 
-import { Cluster, getFeeds } from "../../../services/pyth";
+import { getFeedsCached } from "../../../server/pyth";
+import { Cluster } from "../../../services/pyth";
 
 export { PriceFeedLayout as default } from "../../../components/PriceFeed/layout";
 
@@ -19,7 +20,7 @@ export const generateMetadata = async ({
 }: Props): Promise<Metadata> => {
   const [{ slug }, feeds] = await Promise.all([
     params,
-    getFeeds(Cluster.Pythnet),
+    getFeedsCached(Cluster.Pythnet),
   ]);
   const symbol = decodeURIComponent(slug);
   const feed = feeds.find((item) => item.symbol === symbol);
--- a/apps/insights/src/components/Overview/index.tsx
+++ b/apps/insights/src/components/Overview/index.tsx
@@ -9,7 +9,8 @@ import PriceFeedsLight from "./price-feeds-light.svg";
 import PublishersDark from "./publishers-dark.svg";
 import PublishersLight from "./publishers-light.svg";
 import { TabList } from "./tab-list";
-import { Cluster, getFeeds } from "../../services/pyth";
+import { getFeedsCached } from "../../server/pyth";
+import { Cluster } from "../../services/pyth";
 import {
   totalVolumeTraded,
   activeChains,
@@ -23,7 +24,7 @@ import { FormattedDate } from "../FormattedDate";
 import { FormattedNumber } from "../FormattedNumber";
 
 export const Overview = async () => {
-  const priceFeeds = await getFeeds(Cluster.Pythnet);
+  const priceFeeds = await getFeedsCached(Cluster.Pythnet);
   const today = new Date();
   const feedCounts = [
     ...activeFeeds.map(({ date, numFeeds }) => ({
--- a/apps/insights/src/components/PriceFeed/publishers.tsx
+++ b/apps/insights/src/components/PriceFeed/publishers.tsx
@@ -1,17 +1,16 @@
 import { lookup as lookupPublisher } from "@pythnetwork/known-publishers";
 import { notFound } from "next/navigation";
 
+import { PublishersCard } from "./publishers-card";
+import { getFeedsCached, getPublishersForFeedCached } from "../../server/pyth";
 import { getRankingsBySymbol } from "../../services/clickhouse";
 import {
   Cluster,
   ClusterToName,
-  getFeeds,
-  getPublishersForFeed,
 } from "../../services/pyth";
 import { getStatus } from "../../status";
 import { PublisherIcon } from "../PublisherIcon";
 import { PublisherTag } from "../PublisherTag";
-import { PublishersCard } from "./publishers-card";
 
 type Props = {
   params: Promise<{
@@ -28,8 +27,8 @@ export const Publishers = async ({ params }: Props) => {
     pythnetPublishers,
     pythtestConformancePublishers,
   ] = await Promise.all([
-    getFeeds(Cluster.Pythnet),
-    getFeeds(Cluster.PythtestConformance),
+    getFeedsCached(Cluster.Pythnet),
+    getFeedsCached(Cluster.PythtestConformance),
     getPublishers(Cluster.Pythnet, symbol),
     getPublishers(Cluster.PythtestConformance, symbol),
   ]);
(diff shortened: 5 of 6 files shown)


apps/insights/src/server/pyth.ts
```diff
@@ -0,0 +1,71 @@
+import { parse, } from "superjson";
+import { z } from "zod";
+
+import { PUBLIC_URL, VERCEL_AUTOMATION_BYPASS_SECRET } from '../config/server';
+import { Cluster, priceFeedsSchema } from "../services/pyth";
+import { DEFAULT_CACHE_TTL } from "../utils/cache";
+
+// Convenience helpers matching your previous functions
```
fhqvst: Intended comment?

apps/insights/src/config/server.ts
```diff
@@ -1,7 +1,8 @@
 // Disable the following rule because this file is the intended place to declare
 // and load all env variables.
-/* eslint-disable n/no-process-env */
+/* eslint-disable n/no-process-env, @typescript-eslint/no-non-null-assertion, turbo/no-undeclared-env-vars */
```
fhqvst: Is there a `.d.ts` file where we should put the `REDIS_URL` type definition?

apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts
```diff
@@ -0,0 +1,30 @@
+import { NextResponse } from "next/server";
+import { stringify } from "superjson";
+
+import { getFeedsCached } from '../../../../../server/pyth/get-feeds';
+import { Cluster } from "../../../../../services/pyth";
+
+export const GET = async (request: Request, { params }: { params: Promise<{ publisher: string }> }) => {
+  const { publisher } = await params;
+  const { searchParams } = new URL(request.url);
```
Comment (cprussin) [inline apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts]: If you mark the type of `request` as `NextRequest` instead of `Request`, you can use [`request.nextUrl.searchParams`](https://nextjs.org/docs/app/api-reference/functions/next-request#nexturl)

apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts
```diff
@@ -0,0 +1,30 @@
...
+  const cluster = Number.parseInt(searchParams.get("cluster") ?? Cluster.Pythnet.toString()) as Cluster;
```
Comment (cprussin) [inline apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts]: Except in very, very specific situations, we should avoid typecasting; typecasting breaks the type system since it means the runtime values may actually not match what the type system is guaranteeing.  This means that you will have code that typechecks & compiles but fails at runtime.

Instead of typecasting, it's better to create a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

I also generally prefer avoiding serializing enum values; instead I usually prefer to serialize and deserialize to explicit representations, e.g. something like this:

```ts
import { parseCluster } from "../../../../../services/pyth";

// ...

const clusterName = searchParams.get("cluster");
const cluster = clusterName === undefined ? Cluster.Pythnet :  parseCluster(clusterName);
if (cluster === undefined) {
  return NextResponse.json({ error: "Invalid Cluster" }, { status: 400 });
}
```

apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const feeds = await getFeedsCached(cluster);
+
+  const filteredFeeds = feeds.filter((feed) => feed.price.priceComponents.some((c) => c.publisher === publisher));
+  
+  return new Response(stringify(filteredFeeds), {
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  });
```
Comment (cprussin) [inline apps/insights/src/app/api/pyth/get-feeds-for-publisher/[publisher]/route.ts]: You can use [`NextResponse.json`](https://nextjs.org/docs/app/api-reference/functions/next-response#json) to make this a bit more concise

apps/insights/src/app/api/pyth/get-feeds/route.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { z } from 'zod';
+
+import { getFeedsCached } from "../../../../server/pyth/get-feeds";
+import { Cluster, priceFeedsSchema } from "../../../../services/pyth";
+
+export const GET = async (request: Request) => {
+  // get cluster from query params
+  const { searchParams } = new URL(request.url);
+  const excludePriceComponents = searchParams.get("excludePriceComponents") === "true";
+  const cluster = Number.parseInt(searchParams.get("cluster") ?? Cluster.Pythnet.toString()) as Cluster;
```
cprussin: Not necessary to change here, but for the future, I'm a big fan of using Zod to parse any external data; I also find it to be nice for parsing query parameters, [here's an example of one way to do it ](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/app/component-score-history/route.ts#L9-L16) and there libs like https://github.com/theboxer/zod-search-params which basically wrap up this functionality nicely.

Zod is a bit overkill for a case like this, but it's also nice to have consistent patterns for handling any unvalidated data

apps/insights/src/app/api/pyth/get-feeds/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  // check if cluster is valid
+  if (cluster && !Object.values(Cluster).includes(cluster)) {
+    return NextResponse.json({ error: "Invalid cluster" }, { status: 400 });
+  }
+
+  let feeds = await getFeedsCached(cluster) as Omit<z.infer<typeof priceFeedsSchema>[number], 'price'>[];
```
cprussin: We should not typecast here, the typecast will be unnecessary if you just do this like this:

```ts
const feeds = await getFeedsCached(cluster);
return NextResponse.json(stringify(
  excludePriceComponents ? feeds.map(({ price, ...feed }) => feed) : feeds
));
```

apps/insights/src/components/PriceFeed/get-feed.tsx
```diff
@@ -1,20 +1,31 @@
...
+import { DEFAULT_CACHE_TTL } from '../../utils/cache';
 
 export const getFeed = async (params: Promise<{ slug: string }>) => {
-  "use cache";
+  const data = await fetch(`${PUBLIC_URL}/api/pyth/get-feeds?cluster=${Cluster.Pythnet.toString()}&excludePriceComponents=true`, {
+    next: {
+      revalidate: DEFAULT_CACHE_TTL,
+    },
+    headers: {
+      'x-vercel-protection-bypass': VERCEL_AUTOMATION_BYPASS_SECRET,
```
cprussin: It would be helpful to add a comment to explain what this is doing since this is a less-commonly used feature

apps/insights/src/config/server.ts
```diff
@@ -56,3 +58,29 @@ export const SOLANA_RPC =
...
+  if (IS_PRODUCTION_SERVER) {
+    const productionUrl = demand("VERCEL_PROJECT_PRODUCTION_URL");
+    return `https://${productionUrl}`;
+  } else if (IS_PREVIEW_SERVER) {
+    const previewUrl = demand("VERCEL_URL");
+    return `https://${previewUrl}`;
+  } else {
+    return `http://localhost:3003`;
+  }
+})();
```
cprussin: I think you can actually get away without this if you use the request's origin header for the domain when you need to make internal requests, it will also be a bit less error-prone since the origin header should always be accurate & resolvable

apps/insights/src/server/pyth/get-metadata.ts
```diff
@@ -0,0 +1,13 @@
+
+import { clients, Cluster } from '../../services/pyth';
+import { memoryOnlyCache } from '../../utils/cache';
+
+const getPythMetadata = async (cluster: Cluster) => {
+  return clients[cluster].getData();
+};
+
+export const getPythMetadataCached = memoryOnlyCache.define(
```
cprussin: Very minor nit but IMO I'd prefer not exposing these functions as `get<thing>Cached` and would prefer just exposing `get<thing>` -- a caller generally does not care about caching and the extra verbosity on saying that the thing is cached is unneeded and distracting.  The only case it's actually useful is if there's a use case for using an uncached method, which currently does not exist -- and if that ever arises it's going to be much less common than using the cached methods, so a caller should be explicitly opting in to using uncached methods, rather than explicitly opting in to using cached methods.

apps/insights/src/server/pyth.ts
```diff
@@ -0,0 +1,71 @@
...
+) {
+  const data = await fetch(`${PUBLIC_URL}/api/pyth/get-publishers/${encodeURIComponent(symbol)}?cluster=${cluster.toString()}`, {
+    next: {
+      revalidate: DEFAULT_CACHE_TTL,
+    },
+    headers: {
+      'x-vercel-protection-bypass': VERCEL_AUTOMATION_BYPASS_SECRET,
+    },
+  });
+  return data.json() as Promise<string[]>;
```
cprussin: We should not typecast, instead write a zod schema and validate.

General rule of thumb is don't ever typecast unless you absolutely have to, which usually means you're hitting a limit with the typescript typesystem (usually something around dealing with dependent types) OR you'r dealing with a very hot path where parsing with zod is unreasonably expensive AND you have fenced the unsafe code to ensure that it will never be possible to violate the typecast at runtime

apps/insights/src/server/pyth.ts
```diff
@@ -0,0 +1,71 @@
...
+      'x-vercel-protection-bypass': VERCEL_AUTOMATION_BYPASS_SECRET,
+    },
+  });
+  
+  if(!data.ok) {
+    return undefined;
+  }
+
+  const dataJson = await data.text();
+  const feed: z.infer<typeof priceFeedsSchema>[0] = parse(dataJson);
```
cprussin: In general the explicit type declaration is unnecessary and I usually recommend against it, because allowing the type system to automatically unify types is a great tool for refactoring and makes it very easy to find where code actually breaks down -- e.g. if you changed the type of this stuff, this function may actually not need any changes other than type declaration changes, but the actual breaking change would likely be somewhere else in the code.  By writing the explicit declarations and not inferring, you would now have to change these things before you can even find where unification is actually failing.

apps/insights/src/utils/cache.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const transformer = {
+  serialize: stringify,
+  deserialize: parse,
+};
+
+export const DEFAULT_CACHE_TTL = 86_400; // 24 hours
+export const DEFAULT_CACHE_STALE = 86_400; // 24 hours
+
+export const redisCache: ACDCache = createCache({
```
cprussin, line 15: Nice, very clever

apps/insights/package.json
```diff
@@ -28,10 +28,12 @@
     "@pythnetwork/known-publishers": "workspace:*",
     "@react-hookz/web": "catalog:",
     "@solana/web3.js": "catalog:",
+    "async-cache-dedupe": "^3.0.0",
```
cprussin: We should stick to using [catalog dependencies](https://pnpm.io/catalogs) as it makes handling upgrades across the monorepo much easier if this package were to be used in other apps / packages

cprussin (changes requested):
Mostly looking really good.  My main feedback is to get rid of typecasting wherever possible.  Other than that I recommend adding some comments to explain why using internal fetches is necessary here and the steps you've tried to avoid doing that since it's not really intuitive why we landed on this architecture as a necessity, and it would be easy for someone to not understand it and try to refactor it away without realizing the issues

apps/insights/src/config/server.ts
cprussin: @alexcambose I think this comment still stands, I would get rid of this config and just use the request origin when building urls

cprussin (approved):

PR #2931, opened 2025-08-07
feat: add hermes client

keyvankhademi:
## Summary

This PR adds a resilient hermes client that will be published to cargo

## Rationale

We use hermes in multiple services, including hermes publisher and circuit breaker in lazer. This crate allows us to easily get data from hermes while having resiliency.

## How has this been tested?

- [ ] Added new tests

Manually ran the code.

### DIFF
--- a/apps/hermes/client/rust/Cargo.toml
+++ b/apps/hermes/client/rust/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "pyth-hermes-client"
+version = "0.0.1"
+edition = "2021"
+description = "A Rust client for Pyth Hermes"
+license = "Apache-2.0"
+
+[dependencies]
+pyth-sdk = { version = "0.8.0" }
+tokio = { version = "1", features = ["full"] }
+tokio-tungstenite = { version = "0.20", features = ["native-tls"] }
+futures-util = "0.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+anyhow = "1.0"
+tracing = "0.1"
+url = "2.4"
+derive_more = { version = "1.0.0", features = ["from"] }
+backoff = { version = "0.4.0", features = ["futures", "tokio"] }
+ttl_cache = "0.5.1"
+
+
+[dev-dependencies]
+bincode = "1.3.3"
+ed25519-dalek = { version = "2.1.1", features = ["rand_core"] }
+hex = "0.4.3"
+libsecp256k1 = "0.7.1"
+bs58 = "0.5.1"
+alloy-primitives = "0.8.19"
+tracing-subscriber = { version = "0.3.19", features = ["env-filter", "json"] }
--- a/apps/hermes/client/rust/src/backoff.rs
+++ b/apps/hermes/client/rust/src/backoff.rs
@@ -1,0 +1,118 @@
+//! Exponential backoff implementation for Pyth Lazer client.
+//!
+//! This module provides a wrapper around the [`backoff`] crate's exponential backoff functionality,
+//! offering a simplified interface tailored for Pyth Lazer client operations.
+
+use std::time::Duration;
+
+use backoff::{
+    default::{INITIAL_INTERVAL_MILLIS, MAX_INTERVAL_MILLIS, MULTIPLIER, RANDOMIZATION_FACTOR},
+    ExponentialBackoff, ExponentialBackoffBuilder,
+};
+
+/// A wrapper around the backoff crate's exponential backoff configuration.
+///
+/// This struct encapsulates the parameters needed to configure exponential backoff
+/// behavior and can be converted into the backoff crate's [`ExponentialBackoff`] type.
+#[derive(Debug)]
+pub struct HermesExponentialBackoff {
+    /// The initial retry interval.
+    initial_interval: Duration,
+    /// The randomization factor to use for creating a range around the retry interval.
+    ///
+    /// A randomization factor of 0.5 results in a random period ranging between 50% below and 50%
+    /// above the retry interval.
+    randomization_factor: f64,
+    /// The value to multiply the current interval with for each retry attempt.
+    multiplier: f64,
+    /// The maximum value of the back off period. Once the retry interval reaches this
+    /// value it stops increasing.
+    max_interval: Duration,
+}
+
+impl From<HermesExponentialBackoff> for ExponentialBackoff {
+    fn from(val: HermesExponentialBackoff) -> Self {
+        ExponentialBackoffBuilder::default()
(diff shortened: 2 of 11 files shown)


tejasbadadare (commented):
Looks like this client is focused specifically on the websockets endpoint. Should we rename it to `pyth-hermes-websockets-client` instead of `pyth-hermes-client` which implies that it's a client that implements the [Hermes OpenAPI spec](https://hermes.pyth.network/docs/)?

Also, i believe we want to push the public to use the SSE endpoint, not WS. If we release a client for the (undocumented) WS endpoint it could be confusing for end users. Maybe we keep it privately published?

Sidenote, I actually had Devin [generate a rust client a while ago from the OpenAPI spec](https://github.com/pyth-network/pyth-crosschain/pull/2686) plus SSE support but never got around to merging and publishing it.

keyvankhademi:
> Looks like this client is focused specifically on the websockets endpoint. Should we rename it to `pyth-hermes-websockets-client` instead of `pyth-hermes-client` which implies that it's a client that implements the [Hermes OpenAPI spec](https://hermes.pyth.network/docs/)?
> 
> Also, i believe we want to push the public to use the SSE endpoint, not WS. If we release a client for the (undocumented) WS endpoint it could be confusing for end users. Maybe we keep it privately published?
> 
> Sidenote, I actually had Devin [generate a rust client a while ago from the OpenAPI spec](https://github.com/pyth-network/pyth-crosschain/pull/2686) plus SSE support but never got around to merging and publishing it.

We might add more features to it in future depending on our/clients needs. I don't think we should deprecate the WS anymore as we have seen it has benefits over SSE; that's why we have changed it to ws in circuit breaker.

apps/hermes/client/rust/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "pyth-hermes-client"
+version = "0.0.1"
+edition = "2021"
+description = "A Rust client for Pyth Hermes"
+license = "Apache-2.0"
+
+[dependencies]
+pyth-sdk = { version = "0.8.0" }
```
darunrs, line 9: Should we instead read the hermes structs since it seems it defines things (metadata) that isn't in pyth-sdk?

apps/hermes/client/rust/src/backoff.rs
```diff
@@ -0,0 +1,118 @@
...
+    default::{INITIAL_INTERVAL_MILLIS, MAX_INTERVAL_MILLIS, MULTIPLIER, RANDOMIZATION_FACTOR},
+    ExponentialBackoff, ExponentialBackoffBuilder,
+};
+
+/// A wrapper around the backoff crate's exponential backoff configuration.
+///
+/// This struct encapsulates the parameters needed to configure exponential backoff
+/// behavior and can be converted into the backoff crate's [`ExponentialBackoff`] type.
+#[derive(Debug)]
+pub struct HermesExponentialBackoff {
```
darunrs, line 18: What's the value add of defining a wrapper struct around expo backoff? I don't seem to see any additional features or behavior?

apps/hermes/client/rust/src/client.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// - The `endpoints` vector is empty
+    ///
+    pub fn new(
+        endpoints: Vec<Url>,
+        num_connections: usize,
+        backoff: HermesExponentialBackoff,
+        timeout: Duration,
+        channel_capacity: usize,
+    ) -> Result<Self> {
+        if endpoints.is_empty() {
```
darunrs, line 119: We should add a check to ensure num_connections is also zero.

apps/hermes/client/rust/src/ws_connection.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub price: HermesPrice,
+    pub ema_price: HermesPrice,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<HermesPriceFeedMetadata>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub vaa: Option<String>,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone, Hash)]
+pub struct HermesPrice {
```
darunrs, line 77: Can you provide doc comments there explaining that price and publish time can be 0 if the feed is disabled, and that the publish time can also be when a market closes, and doesn't change? It's useful behavior to document since it changes how the data can be handled.

apps/hermes/client/rust/Cargo.toml
keyvankhademi: i tried that, it doesn't work out of the box and require changes to the hermes server, which I don't think is worth doing.

apps/hermes/client/rust/src/backoff.rs
keyvankhademi: There is a feature that we don't support `max_elapsed_time` iirc.

apps/hermes/client/rust/src/client.rs
keyvankhademi: this causes panic. num connections being 0 i think doesn't panic.

apps/hermes/client/rust/src/ws_connection.rs
keyvankhademi: That's a hermes behavior and I think should be documented in hermes api docs.

tejasbadadare:
> > Looks like this client is focused specifically on the websockets endpoint. Should we rename it to `pyth-hermes-websockets-client` instead of `pyth-hermes-client` which implies that it's a client that implements the [Hermes OpenAPI spec](https://hermes.pyth.network/docs/)?
> > Also, i believe we want to push the public to use the SSE endpoint, not WS. If we release a client for the (undocumented) WS endpoint it could be confusing for end users. Maybe we keep it privately published?
> > Sidenote, I actually had Devin [generate a rust client a while ago from the OpenAPI spec](https://github.com/pyth-network/pyth-crosschain/pull/2686) plus SSE support but never got around to merging and publishing it.
> 
> We might add more features to it in future depending on our/clients needs. I don't think we should deprecate the WS anymore as we have seen it has benefits over SSE; that's why we have changed it to ws in circuit breaker.

Yeah true if we are relying on it for Lazer, we should undeprecate it. I'd wanna understand why we initially hid it from public use though and pushed people to use SSE. cc @ali-behjati

apps/hermes/client/rust/src/client.rs
darunrs: It could cause strange invisible behavior such as the client running but doing nothing. Given this is a customer facing function, I figure it's good to communicate to them any behavior we don't expect.

PR #2935, opened 2025-08-07
fix: rename hermes rust client to

keyvankhademi:
the old name already existed


tejasbadadare (approved):

tejasbadadare:
Shall we add a description to the toml that indicates this is the official one?

keyvankhademi:
> Shall we add a description to the toml that indicates this is the official one?

Will do if we decide to promote it for public use. For I think we should keep it internal, since the features are limited and is not tested well.

PR #2940, opened 2025-08-11
feat: drop jito bundle size and improve packing

guibescos:
## Summary

Before this change, the instruction packing algorithm was assuming every transaction had the Jito tip, this was leading to suboptimal packing (in reality only the first transaction of each bundle has the jito tip).

Rather than giving knowledge of `jitoBundleSize` to `addInstruction`, I decided to hardcode `JITO_BUNDLE_SIZE` to the maximum of 5 everywhere (why would anyone want to send bundles smaller than the maximum?). Now `addInstruction` can know which transactions may have a Jito tip transfer. 

## Rationale

This allows the scheduler to land 2 updates per transaction like before the last Merkle Proof size increase.

## How has this been tested?

- [ ] Added new tests

I run the scheduler locally with the prod config.

### DIFF
--- a/target_chains/solana/sdk/js/solana_utils/src/transaction.ts
+++ b/target_chains/solana/sdk/js/solana_utils/src/transaction.ts
@@ -192,7 +192,7 @@ export class TransactionBuilder {
             this.transactionInstructions.length - 1
           ].instructions,
           instruction,
-          this.transactionInstructions.length % JITO_BUNDLE_SIZE === 0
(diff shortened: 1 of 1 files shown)


target_chains/solana/sdk/js/solana_utils/src/transaction.ts
```diff
@@ -188,7 +192,11 @@ export class TransactionBuilder {
             this.transactionInstructions.length - 1
           ].instructions,
           instruction,
-          buildJitoTipInstruction(this.payer, 1),
+          this.transactionInstructions.length % JITO_BUNDLE_SIZE === 0
```
tejasbadadare: A comment explaining that the tip is only required for the first tx in the bundle would be helpful for future readers

PR #2944, opened 2025-08-11
feat(pyth-lazer-agent) Allow deduplicating updates within each batch

bplatak:
## Summary
Add an option to deduplicate updates within each of the batches before sending them over to lazer. The dedup logic will keep all distinct updates only removing any consecutive duplicates.

## Rationale
Reduce pressure on the relayers for publishers that emit updates without any changes at frequency higher than the publish interval 

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/apps/pyth-lazer-agent/config/config.toml
+++ b/apps/pyth-lazer-agent/config/config.toml
@@ -1,5 +1,5 @@
-relayer_urls = ["ws://localhost:10001/v1/transaction"]
-publish_keypair_path = "/tmp/keypair.json"
+relayer_urls = ["wss://relayer.pyth-lazer-staging.dourolabs.app/v1/transaction", "wss://relayer-1.pyth-lazer-staging.dourolabs.app/v1/transaction"]
+publish_keypair_path = "/path/to/solana/id.json"
 listen_address = "0.0.0.0:8910"
 publish_interval_duration = "25ms"
 authorization_token="token1"
--- a/apps/pyth-lazer-agent/src/lazer_publisher.rs
+++ b/apps/pyth-lazer-agent/src/lazer_publisher.rs
@@ -13,7 +13,6 @@ use pyth_lazer_publisher_sdk::transaction::{
     Ed25519SignatureData, LazerTransaction, SignatureData, SignedLazerTransaction,
 };
 use solana_keypair::read_keypair_file;
-use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
@@ -133,11 +132,10 @@ impl LazerPublisherTask {
             return Ok(());
         }
 
-        let updates = if self.config.enable_update_deduplication {
-            deduplicate_feed_updates(&self.pending_updates.drain(..).collect())?
-        } else {
-            self.pending_updates.drain(..).collect()
-        };
+        let mut updates = self.pending_updates.drain(..).collect();
+        if self.config.enable_update_deduplication {
+            deduplicate_feed_updates(&mut updates);
+        }
 
         let publisher_update = PublisherUpdate {
             updates,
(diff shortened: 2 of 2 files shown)


apps/pyth-lazer-agent/config/config.toml
```diff
@@ -1,5 +1,5 @@
-relayer_urls = ["wss://relayer.pyth-lazer-staging.dourolabs.app/v1/transaction", "wss://relayer-1.pyth-lazer-staging.dourolabs.app/v1/transaction"]
-publish_keypair_path = "/path/to/solana/id.json"
+relayer_urls = ["ws://localhost:10001/v1/transaction"]
+publish_keypair_path = "/tmp/keypair.json"
```
keyvankhademi: in future we should have keys in the repo for local testing.

apps/pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -173,13 +180,37 @@ impl LazerPublisherTask {
     }
 }
 
+fn deduplicate_feed_updates(feed_updates: &Vec<FeedUpdate>) -> Result<Vec<FeedUpdate>> {
+    let mut deduped_feed_updates = Vec::new();
+    let mut last_feed_update = HashMap::new();
+
+    // assume that feed_updates is already sorted by ts (within feed_update_id groups)
```
keyvankhademi:
```suggestion
    // assume that feed_updates is already sorted by timestamp for each feed_update.feed_id
```

apps/pyth-lazer-agent/src/lazer_publisher.rs
```diff
@@ -173,13 +180,37 @@ impl LazerPublisherTask {
...
+
+        if let Some(update) = feed_update.update.as_ref() {
+            if let Some(last_update) = last_feed_update.get(&feed_id) {
+                if update == last_update {
+                    continue;
+                }
+            }
+
+            deduped_feed_updates.push(feed_update.clone());
+            last_feed_update.insert(feed_id, update.clone());
```
keyvankhademi: I think we are keeping the update with the lowest timestamp, shouldn't we keep the highest timestamp?

keyvankhademi: I would insert all in the map and collect it to Vec at the end.

keyvankhademi (changes requested):
I think we should keep the latest timestamp

apps/pyth-lazer-agent/src/lazer_publisher.rs
darunrs: You can also use [dedupe_by_key](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by_key) on the std vec.

bplatak: > shouldn't we keep the highest timestamp?

I don't think we should. My reasoning was that we generally care about lowest-latency so it's more fair to keep the first value we've seen and remove consecutive duplicates - that way we capture the 'earliest' timestamp in the batch (may not affect quality metrics but it's still a better representation of what the publisher does). 

In contrast keeping "highest" timestamp is kind of like a "most recently seen" cache. If we do that, is there even a point in recording intra-batch history instead f just returning last seen?

What's your rationale for keeping the last rather than first? What's the benefit?

merolish: Given that this is intended for real-time consumption, I'd think we want to reflect the latest timestamp that this data is accurate as-of, but I'm not entirely clear. If so, yeah, for a single publisher stream, why not just send the most recently seen per batch.

bplatak: > why not just send the most recently seen per batch.

I don't want to censor data as much as I don't want to invent it. I'm OK with removing truly duplicate values but I'm not comfortable deleting things that actually contain information.

Maybe we should just do a simple deduplication on (feed_id, source_timestamp)?

bplatak: IMHO the data is accurate "since" the first time it was seen, "until" the end of the aggregate window (at which point the publisher needs to retransmit unchanged-entries) or, notionally, until they send a changed value (for which we also need to know the earliest occurance). Knowing what the "most recent" occurrence of a value within an agg window doesn't give us any useful information. Knowing the "first time" the value showed up in a batch let's us better understand the timing characteristics of each publisher

keyvankhademi: The reason I was suggesting the latest timestamp was because of price expiry. When we do lowest timestamp, the price will expire faster than it should if they don't send another update. However, I think you are right about keeping the information about the earliest time a publisher sent us a new price is more important. Let's keep the lowest timestamp.

apps/pyth-lazer-agent/src/jrpc_handle.rs
```diff
@@ -299,6 +299,7 @@ pub mod tests {
             publish_keypair_path: Default::default(),
             publish_interval_duration: Default::default(),
             history_service_url: None,
+            enable_update_deduplication: false,
         };
 
         println!("{:?}", get_metadata(config).await.unwrap());
```
darunrs, line 305: Is there a reason this is a println and not an info log?

bplatak: we don't have tracing setup properly in unit tests making println the easiest choice. this is just for that one manual test that's only ever run locally during dev to grab example data

PR #2950, opened 2025-08-13
feat(entropy-explorer): implement some feedback items

cprussin:
This commit implements the following items to address feedback:

- Use fortuna-staging for testnet chains
- Make "All Mainnet Chains" and "All Testnet Chains" separate options
- Use slugs instead of chain ids in query params
- Fix warning about table missing row header
- Clarify what the callback error message is

### DIFF
--- a/apps/entropy-explorer/src/entropy-deployments.tsx
+++ b/apps/entropy-explorer/src/entropy-deployments.tsx
@@ -6,10 +6,11 @@ export type EntropyDeployment = {
   name: string;
   icon: string;
   isTestnet: boolean;
+  chainId: number;
 };
 
 export const EntropyDeployments = {
-  [10]: {
+  "op-mainnet": {
     address: "0xdF21D137Aadc95588205586636710ca2890538d5",
     name: "OP Mainnet",
     rpc: "https://optimism.llamarpc.com",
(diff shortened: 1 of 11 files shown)


apps/entropy-explorer/src/entropy-deployments.tsx
```diff
@@ -6,46 +6,51 @@ export type EntropyDeployment = {
   name: string;
   icon: string;
   isTestnet: boolean;
+  chainId: number;
 };
 
 export const EntropyDeployments = {
-  [10]: {
+  "op-mainnet": {
```
aditya520, line 13: You could keep chain id for lookup. We can fetch using `https://fortuna-staging.dourolabs.app/v1/chains/configs`, and get all chain-ids as well

cprussin: we do still have the chain IDs, they are just moved into the `chainId` fields of the objects since the human-readable slug is now the primary way to index the object.  The issue with `https://fortuna-staging.dourolabs.app/v1/chains/configs` is that it doesn't include many fields that we need for the explorer, so until we move some more stuff into contract manager, it's not really a usable object to us; the best we could do is still maintain this mapping but maybe remove items if they aren't in https://fortuna-staging.dourolabs.app/v1/chains/configs for some reason.

PR #2953, opened 2025-08-13
Remove 1ms fixed rate from lazer fixed rate channels

danimhr:
## Summary

Remove 1 ms Fixed-Rate Channel and Add Real-Time Channel Support

## Rationale

Previously, the Pyth Lazer used 1ms fixed-rate channels, pushing updates every millisecond regardless of whether the price changed. This created unnecessarily high load, as actual price changes often occur less frequently and at variable intervals.
With this change, the fixed 1 ms updates are replaced by a real-time, event-driven channel:
   - If the price changes, an update is sent immediately.
   - If there’s no change, an update is sent at a maximum interval of 50 ms.
This approach significantly reduces load while ensuring prices are still updated promptly and reliably.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/sdk/rust/protocol/src/router.rs
+++ b/lazer/sdk/rust/protocol/src/router.rs
@@ -12,6 +12,7 @@ use {
     rust_decimal::{prelude::FromPrimitive, Decimal},
     serde::{de::Error, Deserialize, Serialize},
     std::{
+        cmp::Ordering,
         fmt::Display,
         num::NonZeroI64,
         ops::{Add, Deref, DerefMut, Div, Sub},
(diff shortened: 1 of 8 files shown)


lazer/sdk/rust/protocol/src/router.rs
```diff
@@ -206,9 +207,24 @@ pub enum JsonBinaryEncoding {
...
+            Channel::FixedRate(rate) => rate.duration().as_micros(),
+            Channel::RealTime => FixedRate::MIN.duration().as_micros(),
+        };
+        let rate_right = match other {
+            Channel::FixedRate(rate) => rate.duration().as_micros(),
+            Channel::RealTime => FixedRate::MIN.duration().as_micros(),
+        };
+        Some(rate_left.cmp(&rate_right))
+    }
 }
```
keyvankhademi, line 228: shouldn't we do Ord with real_time being strictly less than 50ms? we might want to allow 50ms but not real time for some symbols.

keyvankhademi (approved):

lazer/sdk/rust/protocol/src/router.rs
keyvankhademi: i guess this is also because we are temporarily treating real_time as 50ms?

danimhr: Yes yes

PR #2957, opened 2025-08-17
fix(insights): remove suspense from SearchButton

alexcambose:
## Summary
Removed Suspense from extraCta to fix server boundary crash in header. This does fix the error but it's not super optimal as we're loosing the `<SearchButtonImpl isLoading />` state.

## Rationale
I think the issue is that Nextjs can’t stream Suspense data inside props passed to a Client Component;
```ts
  <AppShell (client)
    ...
    extraCta={
      <Suspense fallback={<SearchButtonImpl isLoading />}>
        <SearchButton />
      </Suspense>
    }
  >
    {children}
  </AppShell>
```
And it throws `The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering.`

Another possible way to fix this would be to just fetch the data on the client side with useState/useEffect.

Tried a few other ways, but couldn't really get it to either work or not throw the same error.

## How has this been tested?

- [ ] Added new tests


cprussin (commented):
I'm not sure but I think this will cause more perf issues because it means that all pages will have to delay for the async stuff in `<SearchButton>` before rendering the layout.

@alexcambose I suggest you first merge your perf improvements PR, then rebase this on top and do some testing to ensure it doesn't cause more perf issues.

We might have to think of another way to implement the global search if we can't use suspense.

alexcambose:
@cprussin I agree. Do you think it would make sense to statically fetch/generate the data at build time and then inject it into these components? Wondering how often things will change in relation to the search.

PR #2961, opened 2025-08-18
fix: added headers invariant

alexcambose:
## How has this been tested?

- [ ] Added new tests


apps/insights/src/utils/absolute-url.ts
```diff
@@ -30,20 +30,11 @@ export async function absoluteUrl(pathname: string) {
     // if we have a host and a proto, we can construct the origin
     if (host && proto) origin = `${proto}://${host}`;
   } catch {
-    // headers() is unavailable
+    throw new Error("Headers are not available in this context. Please check where are you calling this function from.");
```
cprussin: I would just remove the try/catch and let `headers()` throw directly.  Adding this here just adds indirection and removes some of the underlying error context.

apps/insights/src/utils/absolute-url.ts
```diff
@@ -30,20 +24,11 @@ export async function absoluteUrl(pathname: string) {
...
-    if (IS_PRODUCTION_SERVER) {
-      const productionUrl = demand("VERCEL_PROJECT_PRODUCTION_URL");
-      origin = `https://${productionUrl}`;
-    } else if (IS_PREVIEW_SERVER) {
-      const previewUrl = demand("VERCEL_URL");
-      origin = `https://${previewUrl}`;
-    } else {
-      origin = "http://localhost:3003";
-    }
+    throw new Error("No origin found! This should never happen.");
```
cprussin: The way the code is structured makes it hard to understand what this invariant is actually doing.  I'd restructure and simplify this as follows:

```ts
export const getHost = async () => {
  const nextHeaders = await headers();
  const xfHost = nextHeaders.get("x-forwarded-host")?.split(",")[0]?.trim();
  const host = xfHost ?? nextHeaders.get("host") ?? undefined;
  if (host === undefined) {
    throw new NoHostError();
  } else {
    const proto =
      nextHeaders.get("x-forwarded-proto")?.split(",")[0]?.trim() ??
      (host?.startsWith("localhost") ? "http" : "https");
    return `${proto}://${host}`;
  }
}

class NoHostError extends Error {
  constructor() {
    super("Request had neither an `x-forwarded-host` header nor a `host` header");
    this.name = "NoHostError";
  }
}
```

alexcambose: This looks good!

PR #2962, opened 2025-08-18
chore: update default db config

alexcambose:
## Summary

Updated default db config to point to new cluster.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/config/server.ts
+++ b/apps/insights/src/config/server.ts
@@ -45,11 +45,12 @@ export const AMPLITUDE_API_KEY = defaultInProduction(
   "AMPLITUDE_API_KEY",
   "6faa78c51eff33087eb19f0f3dc76f33",
 );
+
 export const CLICKHOUSE = {
   url:
     process.env.CLICKHOUSE_URL ??
     "https://ryf56suqpb.eu-west-2.aws.clickhouse.cloud:8443",
-  username: process.env.CLICKHOUSE_USERNAME ?? "ui_user",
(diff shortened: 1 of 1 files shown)


apps/insights/src/config/server.ts
```diff
@@ -48,8 +48,8 @@ export const AMPLITUDE_API_KEY = defaultInProduction(
 export const CLICKHOUSE = {
   url:
     process.env.CLICKHOUSE_URL ??
-    "https://oxcuvjrqq7.eu-west-2.aws.clickhouse.cloud:8443",
-  username: process.env.CLICKHOUSE_USERNAME ?? "insights",
+    "https://ryf56suqpb.eu-west-2.aws.clickhouse.cloud:8443",
+  username: process.env.CLICKHOUSE_USERNAME ?? "ui_user",
```
cprussin: Let's not use a shared user across UI apps.  Instead we should create a user for each app explicitly.  That way if something is misbehaving, it will be easier to track it down.  It will also be easier to lock down permissions to just what an app needs.

tejasbadadare: Oh! Didn't know we already had a granular user for insights hub. I think we can stick to the same one then since its more specific than "ui_user." We can use ui_user for web_api.

cprussin: lol @tejasbadadare just saw your comment

alexcambose: Yep makes sense

PR #2967, opened 2025-08-19
feat: rounded property to toggle

alexcambose:
## Summary

https://github.com/user-attachments/assets/b460fde1-bc1f-4b7c-a331-e06df9488a64

## Rationale

Added a rounded option for the toggle that will be used for the chart timeframe buttons.

## How has this been tested?

- [ ] Added new tests


packages/component-library/src/SingleToggleGroup/index.tsx
```diff
@@ -63,6 +66,7 @@ export const SingleToggleGroup = ({
           )}
           data-size="sm"
           data-variant="ghost"
+          data-rounded={rounded ? true : undefined}
```
cprussin, line 69: This should be `rounded ? "" : undefined` (the idea is that we want data attributes that are boolean to either be present or not, rather than have the value `true`, to be consistent with how react-aria does things)

PR #2972, opened 2025-08-20
feat(lazer/sui): state and initializer

tejasbadadare:
## Summary

- Add a contract state module. The singleton `State` consists of the current set of trusted signers.
  - Add getters
  - Add setter - `state::update_trusted_signer` can be used to upsert trusted signer info

- Add an `init`. Called automatically at module publish time.  
  - Mints the singleton `AdminCap` and transfers ownership is transferred to the deployer. Only the admin cap owner can call `state::update_trusted_signer`.
    - This is done by specifying `&AdminCap` as the first parameter. The type system guarantees that only the admin cap owner can obtain a reference (borrow) to it, ensuring only the admin is capable of calling the function. By naming the parameter `_`, the cap reference is immediately consumed. 
      - See: https://move-book.com/programmability/capability
    - In the future, the admin cap should be transferred by the deployer to a custodian controlled by governance.
  - Creates the singleton state and shares it publicly. This moves the object into global storage and allows anyone to reference and read it. Updating it is only possible via the `state:update_trusted_signer` function.

Next up:
- Upgradeability & versioning
- Error handling

## How has this been tested?

- [x] Added new tests


lazer/contracts/sui/sources/state.move
```diff
@@ -0,0 +1,214 @@
...
+    let mut i: u64 = 0;
+    while (i < (len as u64)) {
+        let info_ref = vector::borrow(signers, i);
+        if (*public_key(info_ref) == *target) {
+            return option::some(i)
+        };
+        i = i + 1
+    };
+    option::none()
+}
```
merolish, line 94: Ah, so just a linear search.

lazer/contracts/sui/sources/state.move
```diff
@@ -0,0 +1,214 @@
...
+/// - If the pubkey isn't found, it is added as a new trusted signer with the given expires_at.
+public fun update_trusted_signer(_: &AdminCap, s: &mut State, pubkey: vector<u8>, expires_at: u64) {
+    assert!(vector::length(&pubkey) as u64 == ED25519_PUBKEY_LEN, EInvalidPubkeyLen);
+
+    let mut maybe_idx = find_signer_index(&s.trusted_signers, &pubkey);
+    if (expires_at == 0) {
+        if (option::is_some(&maybe_idx)) {
+            let idx = option::extract(&mut maybe_idx);
+            // Remove by swapping with last (order not preserved), discard removed value
+            let _ = vector::swap_remove(&mut s.trusted_signers, idx);
```
merolish, line 62: So there's no ordering of these signers that we care about.

tejasbadadare: Yep

lazer/contracts/sui/sources/state.move
tejasbadadare: Yeah, we only have 2 trusted signers right now :)

lazer/contracts/sui/sources/admin.move
```diff
@@ -0,0 +1,32 @@
...
+/// It has the `drop` ability, allowing it to be consumed immediately after use.
+/// See: https://move-book.com/programmability/one-time-witness
+public struct ADMIN has drop {}
+
+
+/// Initializes the module. Called at publish time.
+/// Creates and transfers ownership of the singular AdminCap capability to the deployer.
+/// Only the AdminCap owner can update the trusted signers.
+fun init(otw: ADMIN, ctx: &mut TxContext) {
+    assert!(types::is_one_time_witness(&otw), 1);
```
ali-behjati: not sure you need it. the OTW guarantees that it's unique. Based on the docs usage of this function seems useful for generics.

lazer/contracts/sui/sources/pyth_lazer.move
```diff
@@ -5,21 +5,34 @@ use pyth_lazer::i64::Self;
...
+/// The `PYTH_LAZER` resource serves as the one-time witness.
+/// It has the `drop` ability, allowing it to be consumed immediately after use.
+/// See: https://move-book.com/programmability/one-time-witness
+public struct PYTH_LAZER has drop {}
+
+/// Initializes the module. Called at publish time.
+/// Creates and shares the singular State object.
+/// AdminCap is created and transferred in admin::init via a One-Time Witness.
+fun init(otw: PYTH_LAZER, ctx: &mut TxContext) {
+    assert!(types::is_one_time_witness(&otw), 1);
```
ali-behjati: ditto.

lazer/contracts/sui/sources/pyth_lazer.move
```diff
@@ -5,21 +5,34 @@ use pyth_lazer::i64::Self;
...
 // standalone verify signature function
 
+/// The `PYTH_LAZER` resource serves as the one-time witness.
+/// It has the `drop` ability, allowing it to be consumed immediately after use.
+/// See: https://move-book.com/programmability/one-time-witness
+public struct PYTH_LAZER has drop {}
+
+/// Initializes the module. Called at publish time.
+/// Creates and shares the singular State object.
+/// AdminCap is created and transferred in admin::init via a One-Time Witness.
```
ali-behjati, line 28: Hmmm it's not here right?

lazer/contracts/sui/sources/admin.move
```diff
@@ -0,0 +1,32 @@
...
+    let cap = AdminCap { id: object::new(ctx) };
```
ali-behjati, line 18: AdminCap can be created in other places by mistake no? If you want the admin cap to be only made by one, you need to enforce it via a type (like the cap retaining the witness, or having a phantom data or something).

lazer/contracts/sui/sources/pyth_lazer.move
tejasbadadare: Yeah it's happening in the admin module's init, but it should happen within the same publish PTB

lazer/contracts/sui/sources/admin.move
tejasbadadare: I guess you can instantiate AdminCap directly within this module in another place. I like the idea of the cap retaining the witness, ill give that a shot.

tejasbadadare:
@ali-behjati Do you think we'll use the same wormhole-based governance system for Lazer contract administration? Currently the AdminCap is just transferred to the deployer, but we will want to instead trade a valid governance VAA for an ephemeral capability that can modify the state, like how the Pyth contract does it. We of course dont have the lazer governance message structure yet. 

I'm thinking the following:
- I can at least lock the UpgradeCap within the State so that the deployer can't also upgrade, only the state can upgrade itself via future governance 
- I'll ensure State can only be created once by having its ctor consume the DeployerCap. The AdminCap will still be owned by the deployer until work on governance is done. 
- FWIW, the existing pyth contract also can't guard against new logic adding the ability to edit state. Any friend of state can call [assert_latest_only](https://github.com/pyth-network/pyth-crosschain/blob/27bba80a8cb231254692660073877c6932adbabc/target_chains/sui/contracts/sources/state.move#L187) and obtain a LatestOnly cap that lets them edit state. Currently governance and the "create price feeds" codepaths call it but others could be introduced in the future.

lazer/contracts/sui/sources/admin.move
tejasbadadare: FWIW, the existing pyth contract also can't guard against new logic adding the ability to edit state. Any friend of state can call [assert_latest_only](https://github.com/pyth-network/pyth-crosschain/blob/27bba80a8cb231254692660073877c6932adbabc/target_chains/sui/contracts/sources/state.move#L187) and obtain a LatestOnly cap that lets them edit state. Currently governance and the "create price feeds" codepaths call it but others could be introduced in the future.

If the AdminCap retains the witness, doesn't the vuln just move to the witness? i.e. some codepath can be introduced in the future that can instantiate the witness, allowing instantiation of the AdminCap

tejasbadadare:
@ali-behjati Okay today's convo on the future plan help elucidate this for me.

How the Pyth contract works is: 
1. When the package is deployed for the first time, it issues a DeployerCap to the tx sender. This DeployerCap and the framework-provided UpgradeCap are "traded in" to initialize the state, ensuring the state is created once and only once, even across contract upgrades.
3. During the state initialization, the DeployerCap is destroyed and the UpgradeCap is transferred to the contract itself (gets stored in the singular State.) This means that only the contract can upgrade itself. The only codepath to trigger that is through governance. 

For this iteration of the Lazer contract:
- We don't have a remote executor to transfer the ownership to 
- We don't have the governance messages defined for signer rotations or upgrades.

Thus, I currently can't have the contract store its own UpgradeCap since then we'd have no way of upgrading it after deploying.

The implementation in this PR basically gives UpgradeCap and AdminCap to the deployment tx sender, giving it full control. 
I think that's kind of all we can do until we define and implement governance messages for Lazer. We can transfer the AdminCap to some DAO wallet until we get there? Wdyt.

lazer/contracts/sui/sources/admin.move
ali-behjati: Hmmm you can't get the witness out of AdminCap without deconstructing it, right?

Regardless, I was thinking about the problem you are trying to solve (not being able to instantiate AdminCap twice) and I'm not sure we really need to solve it. We generally trust ourselves and not the others and in the future might have good reasons to have two or divide it in two capabilities.

The usage of witness also seems to be where you want external people to behave in a certain way and not us. (like someone is creating a Coin and we want to impose some certain characteristic there)

PR #2982, opened 2025-08-22
feat(lazer/sui): verify signatures against trusted signers

tejasbadadare:
## Summary

Verify signature against trusted signers from state.
Next up i'll break up the parse function a bit so I can add better error handling.

## How has this been tested?

- [x] Added new tests


lazer/contracts/sui/sources/state.move
```diff
@@ -1,13 +1,11 @@
 module pyth_lazer::state;
 
-use pyth_lazer::admin::AdminCap;
-use pyth_lazer::admin;
+use pyth_lazer::admin::{Self, AdminCap};
 
-const ED25519_PUBKEY_LEN: u64 = 32;
+const SECP256K1_COMPRESSED_PUBKEY_LEN: u64 = 33;
```
merolish, line 5: Was this mislabeled?

lazer/contracts/sui/sources/pyth_lazer.move
```diff
@@ -162,14 +216,16 @@ public fun parse_and_verify_le_ecdsa_update(update: vector<u8>): Update {
...
+                        option::some(option::some(funding_rate_interval)),
+                    );
                 } else {
                     feed.set_funding_rate_interval(option::some(option::none()));
                 }
             } else {
                 // When we have an unknown property, we do not know its length, and therefore
                 // we cannot ignore it and parse the next properties.
-                abort 0 // FIXME: return more granular error messages
+                abort EInvalidUpdate // FIXME: return more granular error messages
```
merolish, line 228: Is this no longer a fixme?

lazer/contracts/sui/sources/pyth_lazer.move
```diff
@@ -31,13 +35,62 @@ fun init(_: PYTH_LAZER, ctx: &mut TxContext) {
...
+/// # Arguments
+/// * `s` - The pyth_lazer::state::State
+/// * `clock` - The sui::clock::Clock
+/// * `signature` - The ECDSA signature bytes (little endian)
+/// * `payload` - The message payload that was signed
+///
+/// # Errors
+/// * `ESignerNotTrusted` - The recovered public key is not in the trusted signers list
+/// * `ESignerExpired` - The signer's certificate has expired
+public fun verify_le_ecdsa_message(
```
ali-behjati: do we need to make it public :? maybe public in crate be enough (so we can test it, but users don't use it)

lazer/contracts/sui/sources/state.move
tejasbadadare: Gave a more accurate name

lazer/contracts/sui/sources/pyth_lazer.move
tejasbadadare: Still a FIXME since there are several other places in the parsing where the code can revert. Will be tackling this next.

lazer/contracts/sui/sources/pyth_lazer.move
tejasbadadare: Sure, figured it might be handy for users, but yeah i can't think of a scenario where they'd wanna validate without parsing.

PR #2993, opened 2025-08-27
feat(insights): conformance reports

alexcambose:
## Summary

Conformance reports for publishers and feeds.

## Rationale

Adding missing conformance reports form the old legacy website.

## How has this been tested?

- [ ] Added new tests


apps/insights/src/components/ConformanceReport/conformance-report.module.scss
```diff
@@ -0,0 +1,4 @@
+.conformanceReport {
+  display: flex;
+  gap: 0.5rem;
```
cprussin: Please use `theme.spacing(2)` instead

apps/insights/src/components/ConformanceReport/conformance-report.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [isGeneratingReport, setIsGeneratingReport] = useState(false);
+  const { open } = useAlert();
+  const downloadReportForFeed = useDownloadReportForFeed();
+  const downloadReportForPublisher = useDownloadReportForPublisher();
+  const logger = useLogger();
+
+  /**
+   * Download the conformance report for the given symbol or publisher
+   */
+  const downloadReport = async () => {
```
cprussin: We should wrap this in `useCallback`

apps/insights/src/components/ConformanceReport/conformance-report.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (props.publisher) {
+      return downloadReportForPublisher({
+        publisher: props.publisher,
+        cluster: props.cluster,
+        interval: timeframe,
+      });
+    }
+  };
```
cprussin: This API feels awkward to me; why not just take `onPress` as a prop to the component?

apps/insights/src/components/ConformanceReport/conformance-report.tsx
```diff
@@ -0,0 +1,104 @@
...
+        });
+        logger.error(error);
+      })
+      .finally(() => {
+        setIsGeneratingReport(false);
+      });
+  };
+
+  if (props.isLoading) {
+    return <Skeleton width={100} />;
```
cprussin: You can use `<Select isloading />` and `<Button isLoading />` to give it a nicer loading skeleton UI and have less visual changes when the load completes

apps/insights/src/components/ConformanceReport/use-download-report-for-publisher.tsx
```diff
@@ -0,0 +1,228 @@
...
+const getFeeds = async (cluster: (typeof CLUSTER_NAMES)[number]) => {
+  const url = new URL(`/api/pyth/get-feeds`, globalThis.window.origin);
+  url.searchParams.set("cluster", cluster);
+  const data = await fetch(url);
+  const rawData = await data.text();
+  const parsedData = parse(rawData);
+  return priceFeedsSchema.element.array().parse(parsedData);
+};
+
+const PublisherQualityScoreSchema = z.object({
```
cprussin: Super minor but these should not be capitalized since they aren't types

apps/insights/src/components/ConformanceReport/use-download-report-for-publisher.tsx
```diff
@@ -0,0 +1,228 @@
...
+  cluster: (typeof CLUSTER_NAMES)[number],
+  publisher: string,
+  interval: Interval,
+) => {
+  const { date, intervalDays } = getRankingDateAndIntervalDays(
+    new Date(),
+    interval,
+  );
+  const quantityRankUrl = new URL(
+    `/publisher_ranking?publisher=${publisher}&cluster=${cluster}`,
```
cprussin: We should use `searchParams` instead of string interpolation just in case something ever changes with one of these and causes subtle escaping issues

apps/insights/src/components/ConformanceReport/use-download-report-for-publisher.tsx
```diff
@@ -0,0 +1,228 @@
...
+    WEB_API_BASE_URL,
+  );
+  quantityRankUrl.searchParams.set("cluster", cluster);
+  quantityRankUrl.searchParams.set("publisher", publisher);
```
cprussin, line 103: Oh you have the search params here; why have both these and the string interpolation?

apps/insights/src/components/PublisherTag/index.module.scss
```diff
@@ -28,7 +28,7 @@
   .name {
     color: theme.color("heading");
 
-    @include theme.text("base", "medium");
+    @include theme.text("lg", "medium");
```
cprussin: I think this could make the search look weird since publishers & feeds will show up differently, take a look and make sure it looks ok

apps/insights/src/schemas/pyth.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from "zod";
```
cprussin: Personally I'm not a big fan of organizing files by the type of file, I prefer functional domain based organization; in particular I'd prefer (and expect) to find this scheme with the code that fetches the data that the schema validates.

cprussin (approved):
Nice, looks great!

apps/insights/src/schemas/pyth.ts
alexcambose: Changed to a separate `price-feeds-schema.ts` file.

what if the schema is used in different places/files? 

For example, we use the schema in api routes, pyth functions (eg. getFeeds()), http request functions, etc.

apps/insights/src/components/ConformanceReport/conformance-report.tsx
alexcambose: actually this is not used, deleted

apps/insights/src/components/ConformanceReport/use-download-report-for-publisher.tsx
alexcambose: ah overlooked that

PR #2994, opened 2025-08-27
chore(api-reference) Fix Input fieulds

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/api-reference/src/apis/evm/parse-price-feed-updates-unique.tsx
+++ b/apps/api-reference/src/apis/evm/parse-price-feed-updates-unique.tsx
@@ -53,6 +53,8 @@ export const parsePriceFeedUpdatesUnique = writeApi<
       type: ParameterType.HexArray,
       description:
         "The price update data for the contract to verify. Fetch this data from [Hermes API](https://hermes.pyth.network/docs/#/rest/latest_price_updates).",
+      defaultValue:
+        "0x504e41550100000003b801000000040d00cea20e5677f66ed178e9410ddd8280617c06921916e8fd4b71e597d7f6c6d0a14daf3bb3e1a0d8c9e051c8d0................",
     },
     {
       name: "priceId",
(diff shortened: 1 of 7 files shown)


apps/api-reference/src/apis/evm/parse-price-feed-updates-unique.tsx
```diff
@@ -53,6 +53,8 @@ export const parsePriceFeedUpdatesUnique = writeApi<
       type: ParameterType.HexArray,
       description:
         "The price update data for the contract to verify. Fetch this data from [Hermes API](https://hermes.pyth.network/docs/#/rest/latest_price_updates).",
+      defaultValue:
+        "0x504e41550100000003b801000000040d00cea20e5677f66ed178e9410ddd8280617c06921916e8fd4b71e597d7f6c6d0a14daf3bb3e1a0d8c9e051c8d0................",
```
alexcambose, line 57: curious whats this

apps/api-reference/src/components/EvmApi/parameter.ts
```diff
@@ -21,6 +22,12 @@ export const TRANSFORMS: Partial<
   [ParameterType.IntArray]: (value) => [value],
 };
 
+export const getPlaceHolder = <Name extends string>(
+  parameter: Parameter<Name>,
+) => {
+  return parameter.defaultValue ?? PLACEHOLDERS[parameter.type];
+};
+
```
alexcambose: don't think we really make use of the generic here so this could simplified to
```ts
export function getPlaceHolder(parameter: Parameter): string {
  return parameter.defaultValue ?? PLACEHOLDERS[parameter.type];
}
```

apps/api-reference/src/apis/evm/parse-price-feed-updates-unique.tsx
aditya520: This is `updateData` / signed message/ VAA from our hermes. This is just a hardcoded example.

nidhi-singh02: How about we add a comment in the code ?

nidhi-singh02 (commented):
Build is failing, due to formatting. Pls check.

nidhi-singh02 (approved):
lgtm, a nit to add comments for updateData.

PR #3002, opened 2025-08-28
chore(dev-hub)-header Improvements

aditya520:
## Summary

Note: 
I managed to add the search in the header, but search server is not giving me any results.
Need help in fixing that.

## How has this been tested?

- [ ] Added new tests


apps/developer-hub/src/components/search-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+        style={{
+          background: "transparent",
+          border: "1px solid #ccc",
+          padding: "8px 16px",
+          borderRadius: "6px",
+          cursor: "pointer",
+        }}
+      >
+        Search
+      </button>
```
cprussin: you should use the `<Button>` component from the component library, e.g. try to match something like this from IH: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/Root/search-button.tsx#L73-L99

apps/developer-hub/src/components/search-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+        onClick={() => {
+          setOpen(true);
+        }}
+        variant="outline"
+        size="sm"
+        rounded
+        beforeIcon={<MagnifyingGlass size={16} />}
+      >
+        Search
+      </Button>
```
cprussin: I suggest putting in both variants in https://github.com/pyth-network/pyth-crosschain/blob/56465a9499bca6f59d7c140ebf6061bf7eb1aabf/apps/insights/src/components/Root/search-button.tsx#L73-L99 along with the css so that you get the responsive button size on smaller screen sizes (maybe we should move that into the component library or make it a default part of the shell...)

alexcambose: I'll take this

alexcambose: [feat: search button component](https://github.com/pyth-network/pyth-crosschain/pull/3004#top)

apps/insights/src/components/Root/search-button.tsx
```diff
@@ -73,29 +73,12 @@ const ResolvedSearchButton = (props: ResolvedSearchButtonProps) => {
...
-      {...props}
-    >
-      Search
-    </Button>
-  </div>
+  <SearchButtonComponent
+    size="sm"
+    {...props}
+    smallScreenContent={<SearchShortcutText />}
+    largeScreenContent={<SearchShortcutText />}
```
cprussin: We can probably move the shortcut text into the shared component too since it's usually going to be the same, maybe make it optionally overridable cc @alexcambose

cprussin (approved):

PR #3011, opened 2025-09-01
feat(IH): improve search

alexcambose:
## Summary

Added a dedicated package for searching and sorting feeds/publishers by different keys. This will also rank the results by the best search match.
Also made sure we're searching by description and feed id as well.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/Root/search-button.tsx
+++ b/apps/insights/src/components/Root/search-button.tsx
@@ -181,13 +181,13 @@ const SearchDialogContents = ({
   const results = useMemo(() => {
     const filteredFeeds = matchSorter(feeds, search, {
       keys: ["displaySymbol", "symbol", "description", "priceAccount"],
-    })
-      .entries()
-      .map(([symbol, feed]) => ({
-        type: ResultType.PriceFeed as const,
-        id: symbol,
-        ...feed,
-      }));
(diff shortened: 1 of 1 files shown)


apps/insights/src/components/Root/search-button.tsx
```diff
@@ -170,58 +172,38 @@ const SearchDialogContents = ({
...
+    const filteredPublishers = matchSorter(publishers, search, {
+      keys: ["publisherKey", "name"],
+    }).map((publisher) => ({
+      type: ResultType.Publisher as const,
+      id: [ClusterToName[publisher.cluster], publisher.publisherKey].join(":"),
+      ...publisher,
+    }));
+
+    if (type === ResultType.PriceFeed) {
+      return [...filteredFeeds];
```
cprussin: No need to spread into a new array

apps/insights/src/components/Root/search-button.tsx
```diff
@@ -170,58 +172,38 @@ const SearchDialogContents = ({
...
+    }
+    if (type === ResultType.Publisher) {
+      return [...filteredPublishers];
```
cprussin: Same comment

PR #3014, opened 2025-09-02
feat(lazer/js/sdk): QSP auth for browser support

tejasbadadare:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/sdk/js/src/socket/websocket-pool.ts
+++ b/lazer/sdk/js/src/socket/websocket-pool.ts
@@ -1,3 +1,5 @@
+import type { ClientRequestArgs } from "node:http";
+
 import TTLCache from "@isaacs/ttlcache";
 import type { ErrorEvent } from "isomorphic-ws";
 import WebSocket from "isomorphic-ws";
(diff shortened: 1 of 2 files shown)


lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -8,6 +8,77 @@ import type { Request, Response } from "../protocol.js";
...
+ */
+function addAuthentication(
+  url: string,
+  token: string,
+  wsOptions: any = {}
+): { endpoint: string; wsOptions: any } {
+  if (isBrowser()) {
+    // Browser: Add token as query parameter
+    const urlObj = new URL(url);
+    urlObj.searchParams.set("ACCESS_TOKEN", token);
```
cprussin, line 69: If this is available as a query param, why use bearer auth at all?  Seems like if you have to support query param auth it's simpler to just always use the query param and never set the header...

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -8,6 +8,77 @@ import type { Request, Response } from "../protocol.js";
...
+ * 
+ * @param url - The WebSocket URL
+ * @param token - The authentication token
+ * @param wsOptions - Existing WebSocket options
+ * @returns Object containing the modified endpoint and wsOptions
+ */
+function addAuthentication(
+  url: string,
+  token: string,
+  wsOptions: any = {}
```
cprussin: In general you shouldn't ever use `any`; here you should type this correctly to match the types that are valid as options here.  If you must, you can use `unknown`, but I don't think that makes sense in this case.

cprussin (approved):

lazer/sdk/js/src/socket/websocket-pool.ts
tejasbadadare: Yeahh, considered that, but we prefer people use the header whenever possible to avoid leaking the token in logs/some nginx somewhere/etc.

cprussin: ah yeah that's fair

PR #3017, opened 2025-09-04
feat(lazer/sui/js): Init Lazer Sui SDK

tejasbadadare:
## Summary
Create a Lazer Sui SDK to enable users to create txns to parse and verify Lazer payloads on-chain.
- Current limitation: Since the Lazer Sui contract doesn't support upgradeability yet, the `packageId` isn't automatically handled by the SDK. This will be added once we handle upgrades in the Sui contract.  

## Testing
- [ ] Added new tests
  - Tested local build, will e2e test when the Lazer Sui contract is ready to test on devnet


lazer/contracts/sui/sdk/js/examples/FetchAndVerifyUpdate.ts
```diff
@@ -0,0 +1,117 @@
+import { SuiClient } from "@mysten/sui/client";
+import { Transaction } from "@mysten/sui/transactions";
+import { SuiLazerClient } from "../src/client";
+import { PythLazerClient, Request } from "@pythnetwork/pyth-lazer-sdk";
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+
+async function getOneLeEcdsaUpdate(urls: string[], token: string | undefined) {
+  const config: Parameters<typeof PythLazerClient.create>[0] = {
```
cprussin: The annotation here should be unnecessary, just let typescript infer & validate

lazer/contracts/sui/sdk/js/examples/FetchAndVerifyUpdate.ts
```diff
@@ -0,0 +1,117 @@
...
+    urls,
+    token: token ?? "",
+    numConnections: 1,
+  };
+  const lazer = await PythLazerClient.create(config);
+
+  const subscription: Request = {
```
cprussin: You should probably not use the name `Request`, since there's a standard global Web Requests API and whenever you see `Request` in js code you generally expect it to mean https://developer.mozilla.org/en-US/docs/Web/API/Request

lazer/contracts/sui/sdk/js/examples/FetchAndVerifyUpdate.ts
```diff
@@ -0,0 +1,117 @@
...
+      description: "Lazer authentication token",
+    })
+    .help()
+    .parseAsync();
+
+  if (process.env.SUI_KEY === undefined) {
+    throw new Error(`SUI_KEY environment variable should be set to your Sui private key in hex format.`);
+  }
+
+  const lazerUrls = args.lazerUrls.split(",");
```
cprussin: yargs supports actual arrays, probably best to use it

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,32 @@
+import { bcs } from "@mysten/sui/bcs";
+import { SuiClient } from "@mysten/sui/client";
+import { Transaction } from "@mysten/sui/transactions";
+import { SUI_CLOCK_OBJECT_ID } from "@mysten/sui/utils";
+
+const MAX_ARGUMENT_SIZE = 16 * 1024;
+
+export type ObjectId = string;
```
cprussin: I would advise against newtypes in typescript, there's nothing that actually enforces the typing here unless you use [branded types](https://egghead.io/blog/using-branded-types-in-typescript), which imo is not worth the complexity, and the risk of using newtypes like this is I could do something like this:

```ts
type ObjectId = string;

const someFunction = (objectid: ObjectId) => void;

// ...

type ServerName = string;

const serverName: ServerName = "foobar";
someFunction(serverName);
```

Typescript will not complain here, and devs get the false belief that the newtype is somehow enforcing constraints about the value.  At the end of the day, it just adds cognitive overhead and places where the code can get very confusing.

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class SuiLazerClient {
```
cprussin: I would not use a class for this.  There isn't anything stateful and classes are problematic for a whole host of reasons that we've discussed.  If you must store state down the line, it makes sense to add a class at that time -- but keep the class small and contained.  Typescript is not a fundamentally object oriented language; most code should not live in classes (and doing so causes issues like breaking tree shaking, etc)

lazer/contracts/sui/sdk/js/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./client";
```
cprussin: Generally re-exports just add bundle latency; in a case like this there's no reason to even have `index.ts`.  Just point the package.json entry points to `client.ts` instead.

lazer/contracts/sui/sdk/js/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@pythnetwork/pyth-lazer-sui-js",
+  "version": "0.1.0",
+  "description": "TypeScript SDK for the Pyth Lazer Sui contract",
+  "license": "Apache-2.0",
+  "type": "module",
+  "main": "lib/index.js",
+  "module": "lib/index.js",
+  "types": "lib/index.d.ts",
```
cprussin: We should be using [subpath exports](https://nodejs.org/api/packages.html#subpath-exports); see for instance https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/package.json#L9-L19 as a reference

lazer/contracts/sui/sdk/js/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "sideEffects": false,
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
+    "clean": "rimraf lib",
+    "fix": "pnpm prettier --write src && pnpm eslint --fix src --ext .ts",
+    "lint": "eslint src --ext .ts",
+    "typecheck": "tsc -p tsconfig.json --noEmit",
+    "prepublishOnly": "pnpm build",
+    "example:fetch-and-verify": "tsx examples/FetchAndVerifyUpdate.ts"
+  },
```
cprussin: you need to make sure your scripts are named so that they will hook into turbo so that CI works correctly and the turbo commands can traverse your package like any other in the monorepo; in particular you should have a `fix:lint`, `fix:format`, `test:lint`, and `test:format`.  See for instance https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/package.json#L26-L33.

Note a lot of the legacy packages in the monorepo still have varying conventions here and aren't consistent; I'm trying to update this as I touch stuff.  If you're using another package as a reference, it should be the component library, which is regularly maintained and has no historical baggage.

lazer/contracts/sui/sdk/js/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "dependencies": {
+    "@mysten/sui": "^1.3.0",
```
cprussin: I generally suggest sticking to catalog dependencies unless you have a really good reason to diverge; using the catalog makes it easier to keep everything up to date by default and avoid bitrot

lazer/contracts/sui/sdk/js/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../../../../../tsconfig.base.json",
```
cprussin: I suggest using [my canned tsconfig](https://www.cprussin.net/modules/_cprussin_tsconfig.html) which we're using on all new packages and is much, much stricter than the tsconfig base you're using here (I plan to remove the one you're using here eventually once I can migrate packages over)

lazer/contracts/sui/sdk/js/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "0.1.0",
+  "description": "TypeScript SDK for the Pyth Lazer Sui contract",
+  "license": "Apache-2.0",
+  "type": "module",
+  "main": "lib/index.js",
+  "module": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "sideEffects": false,
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
```
cprussin: Note this build will not dual-export your package in cjs & esm.  This is generally important because different runtimes in javascript will use different module systems.  You can refer to e.g. [the Fogo sessions sdk](https://github.com/fogo-foundation/fogo-sessions/blob/main/packages/sessions-sdk-ts/package.json#L41-L42) for a good example of how you can configure dual-exporting correctly

lazer/contracts/sui/sdk/js/eslint.config.js
```diff
@@ -0,0 +1,8 @@
+import { base } from "@cprussin/eslint-config";
+
+export default [
+  ...base,
+  {
+    ignores: ["eslint.config.js", "lib", "src/**/*.js"],
```
cprussin: You should not ignore your eslint config from linting.  You should also not have js files in your `src` directory at all, so I would not ignore those either, because ignoring them is likely masking some underlying issue that's causing them to be built there incorrectly in the first place.

cprussin (changes requested):
I'd suggest referring to a newer package like the [component library](https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library) or [the sessions sdk](https://github.com/fogo-foundation/fogo-sessions/tree/main/packages/sessions-sdk-ts) for a good example of scaffolding a new package.  It looks like a lot of what's in this PR is copied out of legacy packages that aren't well configured and have a lot of small issues.  Let me know if it would help to pair up on this and I'm happy to lend a hand

lazer/contracts/sui/sdk/js/src/client.ts
tejasbadadare: Sounds good. I made it a class since i'll need to store state soon (pyth and wormhole packageIds,) but i'll introduce the class later when i need to.

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,28 @@
+import { bcs } from "@mysten/sui/bcs";
+import { Transaction } from "@mysten/sui/transactions";
+import { SUI_CLOCK_OBJECT_ID } from "@mysten/sui/utils";
+
+const MAX_ARGUMENT_SIZE = 16 * 1024;
```
cprussin: Super minor and not worth changing, but one thing I really like to do with consts like this is break them down with units so it's super easy to understand where the numbers come from, e.g. something like this:

```ts
const ONE_KIBIBYTE_IN_BYTES = 1024;
const MAX_ARGUMENT_SIZE_IN_BYTES = 16 * ONE_KIBIBYTE_IN_BYTES;
```

(I'm assuming that's where 1024 comes from but I don't know where 16 comes from)

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function addParseAndVerifyLeEcdsaUpdateCall(opts: {
+  tx: Transaction;
+  packageId: string;
+  stateObjectId: string;
+  updateBytes: Buffer;
```
cprussin: Note that `Buffer` is not available in browsers without a polyfill, you might want to consider using `Uint8Array` instead which is a superclass of `Buffer` and would make the code more easily isomorphic, if that's something you care about and you aren't using anything specific to `Buffer` (I don't think you are)

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { tx, packageId, stateObjectId, updateBytes } = opts;
+  const [updateObj] = tx.moveCall({
+    target: `${packageId}::pyth_lazer::parse_and_verify_le_ecdsa_update`,
+    arguments: [
+      tx.object(stateObjectId),
+      tx.object(SUI_CLOCK_OBJECT_ID),
+      tx.pure(
+        bcs
+          .vector(bcs.U8)
+          .serialize([...updateBytes], { maxSize: MAX_ARGUMENT_SIZE })
```
cprussin: I'm guessing that if you used `Uint8Array` then you wouldn't have to spread into a new array here

lazer/contracts/sui/sdk/js/src/client.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const MAX_ARGUMENT_SIZE = 16 * 1024;
+
+export function addParseAndVerifyLeEcdsaUpdateCall(opts: {
+  tx: Transaction;
+  packageId: string;
+  stateObjectId: string;
+  updateBytes: Buffer;
+}) {
+  const { tx, packageId, stateObjectId, updateBytes } = opts;
```
cprussin: Super minor but in case you aren't aware, you can destructure as part of the function signature e.g.

```ts
export const addParseAndVerifyLeEcdsaUpdateCall = (
  { tx, packageId, stateObjectId, updateBytes }: {
    tx: Transaction;
    packageId: string;
    stateObjectId: string;
    updateBytes: Buffer;
  }
) => {
  // ...
}

File: lazer/contracts/sui/sdk/js/eslint.config.js [hunk 17]
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line n/no-unpublished-import
```
cprussin: This wouldn't be necessary if you excluded `eslint.config.js` from the `files` array in the `package.json` -- basically ensure that `files` just contains `lib/**/*` since everything else is dev tooling

lazer/contracts/sui/sdk/js/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "license": "Apache-2.0",
+  "type": "module",
+  "engines": {
+    "node": "22"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/esm/client.d.ts",
+      "default": "./dist/esm/client.js"
+    }
```
cprussin: Since you're building both esm & cjs you need to export them both too:

```json
{
  "exports": {
    ".": {
      "import":  {
        "types": "./dist/esm/client.d.ts",
        "default": "./dist/esm/client.js"
      },
      "require":  {
        "types": "./dist/cjs/client.d.ts",
        "default": "./dist/cjs/client.js"
      }
    }
  }
}

cprussin (approved):
Make sure to change the subpath exports so that you're also exporting the cjs you build but other than that looks great!

lazer/contracts/sui/sdk/js/src/client.ts
tejasbadadare: Ahhh ok cool. Sui examples use Buffer but idt it's actually needed by their API.

lazer/contracts/sui/sdk/js/src/client.ts
tejasbadadare: Actually realized this might be an old requirement imposed by WH, idt we need it anymore

ali-behjati (approved):
LGTM

PR #3018, opened 2025-09-04
feat: update search on feeds table

alexcambose:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/PriceFeeds/price-feeds-card.tsx
+++ b/apps/insights/src/components/PriceFeeds/price-feeds-card.tsx
@@ -14,24 +14,25 @@ import type {
 } from "@pythnetwork/component-library/Table";
 import { Table } from "@pythnetwork/component-library/Table";
 import { useLogger } from "@pythnetwork/component-library/useLogger";
-import { useQueryState, parseAsString } from "nuqs";
+import { matchSorter } from "match-sorter";
+import { parseAsString, useQueryState } from "nuqs";
 import type { ReactNode } from "react";
 import { Suspense, useCallback, useMemo } from "react";
-import { useFilter, useCollator } from "react-aria";
+import { useCollator } from "react-aria";
 
-import styles from "./price-feeds-card.module.scss";
 import { useQueryParamFilterPagination } from "../../hooks/use-query-param-filter-pagination";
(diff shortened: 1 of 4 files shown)


apps/insights/src/components/PriceFeeds/price-feeds-card.tsx
```diff
@@ -83,14 +83,19 @@ const ResolvedPriceFeedsCard = ({ priceFeeds, ...props }: Props) => {
...
       const field = column === "assetClass" ? "assetClass" : "displaySymbol";
       return (
         (direction === "descending" ? -1 : 1) *
         collator.compare(a[field], b[field])
       );
     },
+    (items, search) => {
+      return matchSorter(items, search, {
+        keys: ["displaySymbol", "symbol", "description", "key"],
+      });
```
fhqvst, line 97: Is there any chance this call would be blocking? Perhaps the array is small enough that this isn't a problem

apps/insights/src/hooks/use-query-param-filter-pagination.ts
```diff
@@ -14,8 +14,9 @@ import { useCallback, useMemo } from "react";
 
 export const useQueryParamFilterPagination = <T>(
   items: T[],
-  predicate: (item: T, term: string) => boolean,
+  predicate: (item: T, search: string) => boolean,
   doSort: (a: T, b: T, descriptor: SortDescriptor) => number,
+  doMutate: (items: T[], search: string) => T[],
```
fhqvst, line 19: Could probably have deleted `predicate` and `doSort` at this point since `doMutate` is a more open/powerful API anyway

apps/insights/src/components/PriceFeeds/price-feeds-card.tsx
alexcambose: it's a few thousands items so I don't think it's going to impact too much

apps/insights/src/hooks/use-query-param-filter-pagination.ts
alexcambose: That's fair and I think at some point we can remove that. I didn't want to do a big refactor this time as it's used in a few places.

PR #3019, opened 2025-09-04
feat(lazer): add allowed feeds per publisher, aggregated prices and feature flags to state

Riateche:
## Summary

* Add allowed feed IDs to publisher state data.
* Add feature flags to state. They can be used in the future to alter the aggregator's behavior while keeping it deterministic.
* Add aggregated data to feed state.
* Add governance actions for new fields (except aggregated data).

## Rationale

Implementing state and governance.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/publisher_sdk/proto/state.proto
+++ b/lazer/publisher_sdk/proto/state.proto
@@ -46,6 +46,8 @@ message State {
     repeated Publisher publishers = 8;
     // List of governance sources.
     repeated GovernanceSourceState governance_sources = 9;
+    // Currently active feature flags. Feature flags influence the aggregator's behavior.
+    repeated string feature_flags = 10;
 }
 
 // An item of the state describing a publisher.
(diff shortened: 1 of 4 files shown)


danimhr:
@Riateche 
I assumed you'll update the version later

merolish (approved):

lazer/publisher_sdk/proto/state.proto
```diff
@@ -58,6 +60,8 @@ message Publisher {
     repeated bytes public_keys = 3;
     // [required] If true, the publisher is active, i.e. it's allowed to publish updates.
     optional bool is_active = 4;
+    // IDs of the feeds for which the publisher's price will be included in the aggregate.
+    repeated uint32 allowed_feed_ids = 5;
```
ali-behjati, line 64: i was wondering which one makes more sense, allowed_feed_ids per publishers or allowed_pub_ids per price feeds?

Riateche: It's a symmetric many-to-many relationship, so it's fine in either case. In the memory representation we can also reverse it if needed. Having a property on publisher makes governance actions look more natural.

tejasbadadare: Looks like line 140 anticipated it being added in Feed, outdated now:
`    // TODO: list of permissioned publisher IDs.`

PR #3023, opened 2025-09-04
chore(dev-docs) Entropy Guide Edits - 1

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/components/EntropyTable/entropy-api-data-fetcher.tsx
+++ b/apps/developer-hub/src/components/EntropyTable/entropy-api-data-fetcher.tsx
@@ -12,10 +12,6 @@ const ApiChainConfigSchema = z.object({
   default_fee: z.number(),
 });
 
-type ApiChainConfig = z.infer<typeof ApiChainConfigSchema>;
-
-const entropyDeploymentsSchema = z.array(ApiChainConfigSchema);
-
 export type EntropyDeployment = {
   address: string;
   delay: string;
@@ -26,13 +22,11 @@ export type EntropyDeployment = {
   nativeCurrency?: string;
 };
 
-const getChainData = (network_id: number) => {
+function getChainData(network_id: number) {
   return Object.values(chains).find((chain) => chain.id === network_id);
-};
+}
 
-const transformChainData = (
-  chain: ApiChainConfig,
-): [string, EntropyDeployment] => {
+const apiChainConfigToEntrySchema = ApiChainConfigSchema.transform((chain) => {
   const viemChainData = getChainData(chain.network_id);
 
   const configOverride = EntropyDeploymentsConfig[chain.network_id];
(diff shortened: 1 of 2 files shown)


apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -0,0 +1,125 @@
+"use client";
```
cprussin, line 1: Not something you need to worry about @aditya520 but just a note to self, we should probably look into using async server components from markdown to see if it's possible cc @alexcambose

apps/developer-hub/src/components/EntropyTable/entropy-api-data-fetcher.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export const fetchEntropyDeployments = async (
+  url: string,
+): Promise<Record<string, EntropyDeployment>> => {
+  try {
+    const response = await fetch(url);
+    const apiData = entropyDeploymentsSchema.parse(await response.json());
+
+    return Object.fromEntries(apiData.map((item) => transformChainData(item)));
```
cprussin: FYI that you can simplify this a lot using [zod transforms](https://v3.zod.dev/?id=transform) which you can inject at different places in the schema to simplify the transformation logic

apps/developer-hub/src/components/EntropyTable/entropy-api-data-fetcher.tsx
```diff
@@ -0,0 +1,72 @@
...
+  } catch (error_) {
+    throw new Error(String(error_));
```
cprussin: In general you shouldn't catch and re-throw.  At best, this will gobble the stack trace.  At worst, you'll stringify something that has no valid string representation AND eat the stack trace, meaning you'll get nonsensical errors and no idea where they came from.

Just remove the `try` / `catch` entirely here if you don't have anything useful to do with the error except re-throw

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <p>Failed to fetch the list of entropy contracts.</p>
+        </InfoBox>
+      );
+    }
+    case StateType.Loaded: {
+      return <EntropyTableContent chains={state.chains} />;
+    }
+    default: {
+      throw new Error(`Unknown state type: ${typeof state}`);
+    }
```
cprussin: This is not necessary and you should not have it as it will break typescript's ability to do exhaustivity checks (e.g. if you add a state, you want typescript to tell you that you aren't handling it.  If you have a default case, typescript doesn't know you aren't handling it and thinks you're handling all cases, and it won't tell you)

cprussin (approved):
Overall LGTM, most important thing to change before you merge is to make sure you don't have a `default` case anywhere that you're using the sum types pattern as it breaks exhaustivity checks, and that's one of the main reasons to use that pattern in the first place

apps/developer-hub/src/components/EntropyTable/index.tsx
aditya520: Fair point.

PR #3026, opened 2025-09-05
feat(pyth-solana-receiver-sdk): upgrade anchor-lang to 0.31.1

tejasbadadare:
## Summary

- Upgrade `pyth-solana-receiver-sdk`'s anchor-lang dependency to version 0.31.1, as requested by several ecosystem teams.
  - Unfortunately, as a result, we have to pull the SDK out of the `target_chains/solana` cargo workspace since anchor-lang=0.31.1 conflicts with other dependencies in the workspace, namely the pinned version of `wormhole-core-bridge-solana`.
  - The crates in the workspace that depend on `pyth-solana-receiver-sdk` have been pinned to previous version `0.6.1` from crates-io, which is the last version that is compatible with the WH bridge dependency (since it uses the workspace dep of `anchor-lang=0.28.0`.)
  - To avoid clashing versions of `pythnet-sdk` brought in by the crates-io `pyth-solana-receiver-sdk`, the dependency has been patched to reference the local path.

## How has this been tested?

- [ ] Added new tests
  - Builds and tests locally
  - Tested in an example consumer app: https://github.com/guibescos/anchor-pyth/tree/v0.31.1

### DIFF
--- a/.github/workflows/ci-solana-contract.yml
+++ b/.github/workflows/ci-solana-contract.yml
@@ -41,7 +41,6 @@ jobs:
         run: cargo-test-sbf
         if: success() || failure()
       - name: Run sdk tests
-        # Skip doctests with --lib since they are currently broken
-        run: cargo test --lib --package pyth-solana-receiver-sdk
+        run: cargo test --package pyth-solana-receiver-sdk
         working-directory: target_chains/solana/pyth_solana_receiver_sdk
         if: success() || failure()
--- a/target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
+++ b/target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
@@ -104,7 +104,7 @@ impl TwapUpdate {
     /// Get a `TwapPrice` from a `TwapUpdate` account for a given `FeedId` no older than `maximum_age` with a specific window size.
     ///
     /// # Example
-    /// ```
+    /// ```ignore
     /// use pyth_solana_receiver_sdk::price_update::{get_feed_id_from_hex, TwapUpdate};
     /// use anchor_lang::prelude::*;
     ///
@@ -213,7 +213,7 @@ impl PriceUpdateV2 {
     /// Please read the documentation for [`VerificationLevel`] for more information.
     ///
     /// # Example
-    /// ```
+    /// ```ignore
     /// use pyth_solana_receiver_sdk::price_update::{get_feed_id_from_hex, VerificationLevel, PriceUpdateV2};
     /// use anchor_lang::prelude::*;
     ///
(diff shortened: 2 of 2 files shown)


target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
```diff
@@ -2,7 +2,7 @@ pub use pythnet_sdk::messages::{FeedId, PriceFeedMessage};
 use {
     crate::{check, error::GetPriceError},
     anchor_lang::prelude::{borsh::BorshSchema, *},
-    solana_program::pubkey::Pubkey,
+    anchor_lang::solana_program::pubkey::Pubkey,
```
guibescos: nit: doesn't the prelude already have `Pubkey`

tejasbadadare: Yeah it does, lemme see if we can just remove this import

tejasbadadare: Yep it works 👍

.github/workflows/ci-solana-contract.yml
```diff
@@ -41,5 +41,7 @@ jobs:
         run: cargo-test-sbf
         if: success() || failure()
       - name: Run sdk tests
-        run: cargo test --package pyth-solana-receiver-sdk
+        # Skip doctests with --lib since they are currently broken
+        run: cargo test --lib --package pyth-solana-receiver-sdk
+        working-directory: target_chains/solana/pyth_solana_receiver_sdk
```
guibescos, line 45: nit: I think I like better disabling the tests individually by doing:

```rust
    /// # Example
    /// ```ignore
    /// use pyth_solana_receiver_sdk::price_update::{get_feed_id_from_hex, PriceUpdateV2};
    /// use anchor_lang::prelude::*;
```

target_chains/solana/pyth_solana_receiver_sdk/src/price_update.rs
```diff
@@ -324,17 +323,16 @@ pub mod tests {
...
     };
 
     #[test]
     fn check_size() {
-        assert!(
-            PriceUpdateV2::discriminator().len() + borsh0_10::get_packed_len::<PriceUpdateV2>()
-                == PriceUpdateV2::LEN
-        );
+        #[allow(deprecated)]
+        // TODO: borsh0_10 is deprecated, v1::get_packed_len should be used in the future
```
guibescos: maybe         
`#[allow(deprecated, reason = "borsh0_10 is deprecated, v1::get_packed_len should be used in the future")]
`

PR #3031, opened 2025-09-08
Chore(dev hub) entropy improvements

aditya520:
## How has this been tested?

- [ ] Added new tests


apps/developer-hub/src/components/IntegrationCard/index.tsx
```diff
@@ -22,35 +26,59 @@ const colorClasses = {
...
-          className={`w-8 h-8 rounded-lg ${colors.bg} flex items-center justify-center`}
+          className={`flex h-8 w-8 items-center justify-center rounded-md ${colors.bg}`}
         >
-          <div className={`w-4 h-4 ${colors.text}`}>{icon}</div>
+          <div className={`h-4 w-4 ${colors.text}`}>{icon}</div>
         </div>
         <h3
-          className={`text-lg font-semibold text-gray-900 dark:text-white ${colors.hoverText}`}
+          className={`m-0 text-base font-semibold text-[var(--color-fd-foreground)] ${colors.hoverText}`}
         >
```
alexcambose: Would be great to use classnames with `cn()` where we merge classes together.
Not something to change here but for the future.

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -120,3 +129,80 @@ const getEntropyDeployments = async (
...
+
+  return feesByChain;
+}
+
+function formatWeiFixed18(value: bigint): string {
+  const intPart = value / 1_000_000_000_000_000_000n;
+  const fracPart = value % 1_000_000_000_000_000_000n;
+  const fracStr = fracPart.toString().padStart(18, "0");
+  return `${intPart.toString()}.${fracStr}`;
+}
```
alexcambose: not super sure what this does but I think you can use some build in methods from viem, 
https://viem.sh/docs/utilities/formatEther#formatether
https://viem.sh/docs/utilities/formatGwei#formatgwei

apps/developer-hub/src/components/IntegrationCard/index.tsx
```diff
@@ -22,35 +26,59 @@ const colorClasses = {
...
+    : { href };
 
   return (
-    <a
-      href={href}
-      className="block group bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-700 rounded-lg p-6 hover:border-gray-300 dark:hover:border-gray-600 transition-colors"
+    <Wrapper
+      {...(wrapperProps as
+        | { href: string }
+        | { href: string; target: string; rel: string })}
```
alexcambose: do we need the typecasting here? `as | {...} | {...}`. I feel like this should work without

apps/developer-hub/content/docs/entropy/best-practices.mdx
```diff
@@ -53,5 +24,55 @@ function mapRandomNumber(
...
+In the following example, `mapRandomNumber` is used to generate 6 random attributes for a character.
+
+<DynamicCodeBlock
+  lang="solidity"
+  code={`\
+function generateAttributes(bytes32 randomNumber) internal {
+  int256 strength = mapRandomNumber(
+    keccak256(abi.encodePacked(randomNumber, "strength")),
+    15,
+    20
```
alexcambose: feel free to ignore but I'm curious if this would work with just doing the triple quotes like ```solidity .... 

Feels a bit cumbersome to have to use <DynamicCodeBlock .../> every time we write code

alexcambose:
Nice!

apps/developer-hub/content/docs/entropy/meta.json
```diff
@@ -7,19 +7,22 @@
...
-    "protocol-design",
-    "current-fees",
-    "error-codes",
-    "examples",
     "request-callback-variants",
-    "create-your-first-entropy-app"
+    "error-codes",
+    "[Example Applications](https://github.com/pyth-network/pyth-examples/tree/main/entropy)",
+    "[Entropy Explorer](https://entropy-explorer.pyth.network/)",
+    "[Fortuna API Reference](https://fortuna.dourolabs.app/docs/)",
```
jayantk, line 23: I think the icons for these links don't clearly indicate that they open a new page. I think the up-right arrow should go after the text.

apps/developer-hub/content/docs/entropy/index.mdx
```diff
@@ -5,29 +5,63 @@ icon: DiceSix
...
 **Pyth Entropy** is an on-chain random number generator (RNG) designed for developers who need fair, unbiased, and cryptographically secure randomness.
 Whether you're building a blockchain game, NFT mint, lottery, or simulation, Entropy delivers randomness that is:
 
-- **Trustless & verifiable** - built on commit-reveal(TODO: link to commit-reveal).
-- **Low-latency** - randomness available within a few blocks(TODO: link to latency).
-- **Easy to integrate** - Permissionless Integration, Visual Tx Explorer(TODO: link to explorer).
-- **Cost-efficient** - designed for scalable production use(TODO: link to fees).
+- **Trustless & verifiable** - built on [commit-reveal protocol](./entropy/protocol-design).
+- **Low-latency** - randomness available within a [few blocks](./entropy/contract-addresses).
+- **Easy to integrate** - Permissionless Integration, including [Visual Tx Explorer](https://entropy-explorer.pyth.network/).
```
jayantk: I suggest "Easy to use" and the benefit should be something like "get started in 5 minutes, no registration or setup required". "Permissionless integration" isn't clear.

apps/developer-hub/content/docs/entropy/whats-new-entropyv2.mdx
```diff
@@ -54,11 +54,10 @@ Improved contract efficiency reduces overall gas costs for entropy requests.
 ## Migration Guide
 
 If you're upgrading from Entropy v1 to v2:
-(TODO: Add links to the interface)
```
jayantk, line 57: the content on this page needs some work imo

1. Multiple Request Variants -- this doesn't seem like a benefit. I think the real user benefit is "configurable gas limits". Forget about explaining the other variants of Request Callback -- no one cares about anything besides the default variant and the one that lets you set a gas limit.

2. Improved Fee structure -- this is wrong. these fee calculation methods correspond to request methods. Just delete this section

3. Entropy Debugger / Enhanced Callback Status -- theses should be the same item. Also should be called Entropy Explorer. Point here being that there's a tool that people can use to see the status of their request / callback.

5. Gas optimization -- i don't think we actually did this

apps/developer-hub/content/docs/entropy/create-your-first-entropy-app.mdx
```diff
@@ -1,108 +1,366 @@
...
 In this tutorial we will implement and deploy a coin flip contract which will use entropy to generate a random output.
 
 ## Preliminaries
 
 Before we start, please make sure you have the following tools installed:
 
-- [Foundry](https://book.getfoundry.sh/getting-started/installation)
-- [Node.js](https://nodejs.org/en/download) (version >= v18.0.0)
+- [Foundry](https://book.getfoundry.sh/getting-started/installation).
+- [Node](https://nodejs.org/en/download). Run `node -v{:js}` to confirm. You should get an output with version >= `v18.0.0`.
```
jayantk, line 14: the {:js} postfix is appearing in the guide. this is true for all similar instances below

apps/developer-hub/content/docs/entropy/create-your-first-entropy-app.mdx
```diff
@@ -1,108 +1,366 @@
...
 
-Create a directory named `coin-flip` and initialize a new Foundry project:
+Create a directory named `coin-flip{:bash}` in your filesystem.
+We will use this directory as the working directory for the rest of the tutorial.
+Let's initialize a new project in `coin-flip{:bash}` by running `forge init contracts{:bash}`.
 
-```bash
+This will create a new directory in `coin-flip{:bash}` named `contracts/src`, which will contain the smart contract code.
+
+<DynamicCodeBlock
```
jayantk: the copy icon on these should be visible by default. Not clear to the reader that these are easily copyable right now.

apps/developer-hub/content/docs/entropy/create-your-first-entropy-app.mdx
```diff
@@ -1,108 +1,366 @@
...
-- [Node.js](https://nodejs.org/en/download) (version >= v18.0.0)
+- [Foundry](https://book.getfoundry.sh/getting-started/installation).
+- [Node](https://nodejs.org/en/download). Run `node -v{:js}` to confirm. You should get an output with version >= `v18.0.0`.
 
 ## Getting Started
 
-Create a directory named `coin-flip` and initialize a new Foundry project:
+Create a directory named `coin-flip{:bash}` in your filesystem.
+We will use this directory as the working directory for the rest of the tutorial.
+Let's initialize a new project in `coin-flip{:bash}` by running `forge init contracts{:bash}`.
```
jayantk: when you're prefacing a code block like this, the right thing to do is to end the sentence with a colon not a period

also, the explanation of what the code block does should be *after* the code block, not before. So

let's initialize a new project:

```
code here
```

These commands will ...

jayantk: many instances of the colon thing on this page

apps/developer-hub/content/docs/entropy/create-your-first-entropy-app.mdx
```diff
@@ -1,108 +1,366 @@
...
-## Smart Contract Implementation
+## Implementation
 
-Replace the content of `contracts/src/Counter.sol` with the following coin flip contract:
+Create a new file `CoinFlip.sol{:solidity}` in `contracts/src` directory and add the following code into it to start.
 
-```solidity
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+<DynamicCodeBlock lang="solidity" code={`\
```
jayantk: FYI the indentation of these code snippets is broken

apps/developer-hub/content/docs/entropy/generate-random-numbers-evm.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Generate Random Numbers onchain
 description: Learn how to integrate Pyth Entropy to generate random numbers in your dapp
+icon: Code
```
jayantk, line 4: why is onchain lowercase

also, isn't it On-Chain ?

jayantk: the install the SDK thing may also need the forge remapping instructions if someone is using foundry

apps/developer-hub/content/docs/entropy/generate-random-numbers-evm.mdx
```diff
@@ -104,42 +107,43 @@ import { IEntropyV2 } from "@pythnetwork/entropy-sdk-solidity/IEntropyV2.sol";
...
-bytes32 randomNumber
-) internal override {
-// Implement your callback logic here.
-}
-
-// This method is required by the IEntropyConsumer interface.
-// It returns the address of the entropy contract which will call the callback.
-function getEntropy() internal view override returns (address) {
-return address(entropy);
-}
```
jayantk, line 142: couple comments below this

entropyCallback function -- this is missing some syntax higlighting

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -120,3 +129,80 @@ const getEntropyDeployments = async (
...
+
+    let isCancelled = false;
+
+    async function loadInBatches() {
+      const batchSize = 5;
+      for (let i = 0; i < entries.length; i += batchSize) {
+        if (isCancelled) return;
+        const batch = entries.slice(i, i + batchSize);
+        const results = await Promise.allSettled(
+          batch.map(async ([name, d]) => {
```
cprussin: We should avoid using single-letter variable names, it makes the code unnecessarily hard to understand

apps/developer-hub/content/docs/entropy/best-practices.mdx
```diff
@@ -1,46 +1,17 @@
...
-2. The callback function contains logic that throws an error
-
-If you're not receiving a callback, you can manually invoke it to identify the specific issue. This allows you to:
-
-- See if the transaction fails and why
-- Check the gas usage against the chain's callback gas limit
-- Debug your callback implementation
-
-For detailed instructions on how to manually invoke and debug callbacks, refer to the [Debug Callback Failures](debug-callback-failures) guide.
+import { DynamicCodeBlock } from "fumadocs-ui/components/dynamic-codeblock";
```
jayantk: this page really shouldn't be called Best Practices. It kind of made sense before with the gas usage stuff, but now it's like some tips on how to transform the entropy result into something usable. TBH it could probably be a How-To Guide at this point.

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -120,3 +129,80 @@ const getEntropyDeployments = async (
...
+            if (!d.rpc) return [name, undefined] as const;
+            try {
+              const client = getClient(d.rpc);
+              const fee = await client.readContract({
+                address: d.address as `0x${string}`,
```
cprussin: Is this typecasting necessary?  It's always best to avoid typecasting; if the underlying types don't already guarantee the string format matches this then you should use a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to validate (and handle the case where it doesn't match) rather than typecasting.

apps/developer-hub/content/docs/entropy/contract-addresses.mdx
```diff
@@ -1,12 +1,19 @@
...
+title: Contract & Provider Addresses
 description: Pyth Entropy contract addresses on EVM networks
+icon: FileText
 ---
 
 import { EntropyTable } from "../../../src/components/EntropyTable";
 
 ## Mainnets
 
+> ⚠️ **Warning**
```
jayantk: this warning callout isn't visually salient enough. there's a better warning callout in the Generate Random Numbers guide

apps/developer-hub/content/docs/entropy/contract-addresses.mdx
```diff
@@ -16,10 +23,14 @@ The Entropy contract is deployed on the following mainnet chains:
 The default provider on mainnet has a reveal delay to avoid changes on the outcome of the Entropy request because of block reorgs.
 The reveal delay shows how many blocks should be produced after the block including the request transaction in order to reveal and submit a callback transaction.
 
-The default provider fulfills the request by sending a transaction with a gas limit as mentioned in above table. Entropy callbacks the consumer as part of this transaction.
+The default provider fulfills the request by sending a transaction with a gas limit as mentioned in above table or as set by the user in the request.
+Entropy callbacks the consumer as part of this transaction.
 
 ## Testnets
 
+> ℹ️ **Note**
```
jayantk: same comment on this info callout

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -120,3 +129,80 @@ const getEntropyDeployments = async (
...
+              return [name, undefined] as const;
+            }
+          }),
+        );
+
+        const next: Record<string, bigint | undefined> = {};
+        for (const r of results) {
+          if (r.status === "fulfilled") {
+            const [name, fee] = r.value;
+            next[name] = fee as bigint;
```
cprussin: We should parse instead of typecasting, `zod` can do this or you can just do `if (typeof fee === "bigint")`

apps/developer-hub/content/docs/entropy/contract-addresses.mdx
```diff
@@ -1,12 +1,19 @@
 ---
-title: Contract Addresses
+title: Contract & Provider Addresses
```
jayantk: the name of this item should definitely mention fees or suggest that fee information is here.

apps/developer-hub/src/components/EntropyTable/index.tsx
```diff
@@ -120,3 +129,80 @@ const getEntropyDeployments = async (
...
+          }
+        }
+        if (Object.keys(next).length > 0) {
+          setFeesByChain((prev) => ({ ...prev, ...next }));
+        }
+      }
+    }
+
+    void loadInBatches();
```
cprussin: Personally, I really dislike using `void` for this, for a few reasons:

1. It's not clear what your intention is
2. Generally, ignoring errors is incorrect and not what you should do.

If you _really_ wanted to ignore errors, I recommend doing it like this:

```ts
loadInBatches().catch(() => { /* no-op on error */ });
```

That way, it's obvious you truly intended to disregard errors.  However, I suggest you add a proper error handler, at the very least logging the errors to the console.

apps/developer-hub/src/components/IntegrationCard/index.tsx
```diff
@@ -22,35 +26,59 @@ const colorClasses = {
...
   icon,
   title,
   description,
   colorScheme = "blue",
-}: IntegrationCardProps) => {
+  external,
+  showArrow = true,
+}: IntegrationCardProps) {
   const colors = colorClasses[colorScheme];
+  const Wrapper = external ? "a" : Link;
```
cprussin: I strongly recommend against doing this kind of pattern.  It's generally unneeded for avoiding repetition, and doing things like this causes messy typing, and it generally makes the code harder to predict & understand.  I used to use this pattern and have nearly always moved away from it.

Instead, what I recommend doing is something along these lines:

```tsx
const MyComponent = ({ external, ...props }) =>
  external
    ? <a {...commonProps(props)} />
    : <Link {...commonProps(props)} />

const commonProps = (props) => ({
  href: props.href,
  className: "not-prose ...",
  children: (
    <div>
      ...
    </div>
  )
});

File: apps/developer-hub/src/components/IntegrationCard/index.tsx (hunk 3 above)
Comment (cprussin) [reply apps/developer-hub/src/components/IntegrationCard/index.tsx]: this typecasting would go away if you follow my previous suggestion

File: apps/developer-hub/src/components/IntegrationCard/index.tsx (hunk 1 above)
Comment (cprussin) [reply apps/developer-hub/src/components/IntegrationCard/index.tsx]: agreed with alex but note that our standard class name builder library is `clsx`, not `cn` (they are effectively the same but just stick to `clsx` so we don't add another unneeded dependency)

File: apps/developer-hub/src/components/Root/theme.css [hunk 23]
```diff
@@ -334,3 +334,16 @@
...
+  line-height: 1.5;
+}
+
+:where(.prose) :where(p, li) {
+  line-height: 1.5;
+}
+
+:where(.prose) :where(h1, h2, h3, h4, h5, h6) {
+  line-height: 1.2;
+}
```
cprussin, line 349: I very strongly recommend against setting global styles; instead we should be scoping styles to components.

apps/developer-hub/package.json
```diff
@@ -22,17 +22,22 @@
...
     "@phosphor-icons/react": "catalog:",
     "@pythnetwork/component-library": "workspace:*",
+    "@pythnetwork/contract-manager": "workspace:*",
+    "@pythnetwork/entropy-sdk-solidity": "workspace:*",
     "@react-hookz/web": "catalog:",
     "clsx": "catalog:",
     "fumadocs-core": "catalog:",
     "fumadocs-mdx": "catalog:",
     "fumadocs-ui": "catalog:",
+    "katex": "^0.16.22",
```
cprussin: Better to use catalog dependencies

apps/developer-hub/src/components/IntegrationCard/index.tsx
```diff
@@ -22,35 +26,59 @@ const colorClasses = {
...
 
   return (
-    <a
-      href={href}
-      className="block group bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-700 rounded-lg p-6 hover:border-gray-300 dark:hover:border-gray-600 transition-colors"
+    <Wrapper
+      {...(wrapperProps as
+        | { href: string }
+        | { href: string; target: string; rel: string })}
+      className="not-prose group no-underline grid h-full grid-rows-[auto_1fr] gap-3 rounded-xl border bg-[var(--color-fd-card)] border-[var(--color-fd-border)] p-5 md:p-6 shadow-sm outline-none transition-colors hover:border-[var(--color-fd-accent)] focus-visible:ring-2 focus-visible:ring-[var(--color-fd-accent)]"
```
cprussin: I realize this isn't newly added (and I also realize I should have noticed this a few PRs ago), but I think using tailwind for all these styles is a mistake.  We don't use tailwind in the component library or design system; we're using sass with `theme.scss` in the component library exporting the core design tokens and shared styles.

I realize it will be a challenge for you to convert this to sass, but I also think it's a bad precedence to get in the habit of writing tailwind as it will make this app harder to integrate with the core design system & component library.

Tailwaind is used internally by fuma docs so there will be some places we have to use tailwind to interface, but in general I don't think any custom components are a place where it makes sense to use tailwaind.

cprussin (approved):
Overall lgtm, I have some small suggestions.

apps/developer-hub/src/components/EntropyTable/index.tsx
aditya520: Oh yes, Wow, how can I miss that.

apps/developer-hub/content/docs/entropy/best-practices.mdx
nidhi-singh02: this is the same thing that we were discussing yesterday @aditya520 as in don't see any difference in the look and feel of these two methods.

apps/developer-hub/src/components/IntegrationCard/index.tsx
alexcambose: ah yess, clsx

apps/developer-hub/src/components/EntropyTable/index.tsx
aditya520: Yeah, I was avoiding doing `console.errors`, but I see your point.

apps/developer-hub/src/components/Root/theme.css
aditya520: @alexcambose 
Can you please fix this in your next PR?

apps/developer-hub/src/components/EntropyTable/index.tsx
cprussin: console errors are reasonable if you have nothing else in place.  Better is to use `useLogger` which is set up for you by the app shell; this does just currently do `console.error` but it centralizes the logging so we can add error reporting down the line whenever we decide to do so

apps/developer-hub/content/docs/entropy/best-practices.mdx
aditya520: This component support [shiki](https://shiki.style/) styles. I want to experiment with them in future.

apps/developer-hub/content/docs/entropy/meta.json
aditya520: I agree, but that is the fumadocs default behaviour. I would need to override that component byone of our component, which will take some time.

apps/developer-hub/content/docs/entropy/whats-new-entropyv2.mdx
aditya520: Enhanced callback status should be a separate section IMO. The new flag passed under callback event enables user to have a better safeguard against callback failures.

apps/developer-hub/content/docs/entropy/create-your-first-entropy-app.mdx
aditya520: Yeah, that needed to be fixed. @alexcambose I would need your help here.

apps/developer-hub/content/docs/entropy/generate-random-numbers-evm.mdx
aditya520: It should be Onchain, on-chain

aditya520: It is present.

PR #3032, opened 2025-09-08
feat: impl Ord and Eq for FeedUpdate

keyvankhademi:
## Summary

This PR implements Ord and Eq for FeedUpdate and all the required traits for them.

## Rationale

We need it for the fix in agent deduplication logic

## How has this been tested?

- [ ] Added new tests


lazer/publisher_sdk/rust/src/lib.rs
```diff
@@ -147,3 +147,95 @@ impl From<pyth_lazer_protocol::api::Channel> for state::Channel {
...
+            other.best_ask_price,
+        ))
+    }
+}
+
+impl Eq for FundingRateUpdate {}
+
+impl Ord for FundingRateUpdate {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        (
```
ali-behjati:
```
self.price
    .cmp(&other.price)
    .then(self.best_bid_price.cmp(&other.best_bid_price))
    .then(self.best_ask_price.cmp(&other.best_ask_price))
```

lazer/publisher_sdk/rust/src/lib.rs
```diff
@@ -147,3 +147,95 @@ impl From<pyth_lazer_protocol::api::Channel> for state::Channel {
...
+                other.rate,
+                other
+                    .funding_rate_interval
+                    .as_ref()
+                    .map(|duration| (duration.seconds, duration.nanos)),
+            ))
+    }
+}
+
+impl PartialOrd for FundingRateUpdate {
```
Riateche: It's incorrect to ignore a field (`special_fields` in this case) in `PartialOrd` if it's used in `PartialEq`. From the `PartialOrd` docs:

> The methods of this trait must be consistent with each other and with those of [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html). The following conditions must hold:
>     a == b if and only if partial_cmp(a, b) == Some(Equal).

In general, `PartialOrd` and `Ord` are rarely implemented for structs unless their semantic meaning defines an order. I think it's not warranted in this case. Perhaps this issue could be handled in a better way in the agent itself? Maybe use `Hash` instead of `Ord`?

merolish: Looks like this is to support [updated dedup logic](https://github.com/pyth-network/pyth-crosschain/pull/3033/files#diff-414976ec6301d7377953dab09948628ed00c868872cc7228a8e5692143b4be03R183-R193). The agent queues up vec of updates from one or more connected publishers and then every publish interval, batches and signs them as a transaction to send to the relayers.

Perhaps the agent could queue up per feed id, and we could sort with a lighter-weight comparator that just looks at timestamp?

keyvankhademi: Valid points. I'll handle it without these traits.

PR #3033, opened 2025-09-08
fix: deduplication logic in agent

keyvankhademi:
## Summary

This PR fixes the deduplication logic in agent.

## Rationale

There is a bug in the current deduplication logic, `dedup_by_key` only removes the consecutive duplicated. It doesn't work if there an update for another feed in the middle of the 2 duplicated update for a feed.

We also improve the logic to only includes at most 1 update per feed. This is done by getting the last distinct update, while prioritizing the lower timestamps. This is because the aggregator effectively ignores the previous updates by immediately overriding them with new values.

## How has this been tested?

- [x] Added new tests

Wrote a test that would fail with the current deduplication logic.

### DIFF
--- a/apps/pyth-lazer-agent/Cargo.toml
+++ b/apps/pyth-lazer-agent/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-agent"
-version = "0.4.1"
+version = "0.4.2"
 edition = "2024"
 description = "Pyth Lazer Agent"
 license = "Apache-2.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5639,7 +5639,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-lazer-agent"
-version = "0.4.1"
+version = "0.4.2"
 dependencies = [
  "anyhow",
  "backoff",


danimhr:
@keyvankhademi The ci is failing :-?

bplatak:
@keyvankhademi this wasn't a bug - we made a deliberate decision at the time not to censor any information sent to us by publishers and instead handle this in lazer after the updates are persisted to the db. But yes this will, at least with the publisher the dedup was originally added for, have a big reduction in the number of sent updates.

bplatak (approved):

apps/pyth-lazer-agent/Cargo.toml
```diff
@@ -7,8 +7,8 @@ license = "Apache-2.0"
 repository = "https://github.com/pyth-network/pyth-crosschain"
 
 [dependencies]
-pyth-lazer-publisher-sdk = "0.3.0"
-pyth-lazer-protocol = "0.10.1"
+pyth-lazer-publisher-sdk = { path = "../../lazer/publisher_sdk/rust" }
```
bplatak: did you mean to leave those as a source dependency?

keyvankhademi: i will use the correct version after releasing the new sdk.

PR #3035, opened 2025-09-09
feat(lazer): add history api client for symbols

Riateche:
## Summary

- rename PythLazerClient to PythLazerStreamClient
- add history client for fetching symbols

I wasn't sure what would be the most convenient API for fetching symbol updates, so I added a few different methods:

- `all_symbols_metadata` - fetches only once.
- `all_symbols_metadata_handle` - returns an ArcSwap that will be continuously updated in background. Convenient for migrating old code, but there is no way to get a notification when something changed.
- `all_symbols_metadata_fault_tolerant_handle` - same as above, but returns an empty value if the initial fetch failed. Useful for relayer: while it needs symbols for some things, it can still work without them.
- `all_symbols_metadata_stream` - the most flexible option: get a receiver that receives new symbols.

There is also a caching feature: to avoid hard dependency on history service, we can store last known symbols locally and fetch them on startup.

## Rationale

To remove `symbols.toml` and use symbols for the state-aware history service.

## How has this been tested?

- [ ] Added new tests

Tested in tilt.

### DIFF
--- a/lazer/sdk/rust/client/src/history_client.rs
+++ b/lazer/sdk/rust/client/src/history_client.rs
@@ -36,7 +36,7 @@ pub struct PythLazerHistoryClientConfig {
     pub request_timeout: Duration,
     /// Path to the cache directory that can be used to provide latest data if history service is unavailable.
     pub cache_dir: Option<PathBuf>,
-    /// Capacity of communication channels created by this client.
+    /// Capacity of communication channels created by this client. It must be above zero.
     #[serde(default = "default_channel_capacity")]
     pub channel_capacity: usize,
 }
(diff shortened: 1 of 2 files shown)


lazer/sdk/rust/client/src/history_client.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Fetch metadata for all symbols as a receiver.
+    ///
+    /// Returns an error if the initial fetch failed.
+    /// On a successful return, the channel will always contain the initial data that can be fetched
+    /// immediately from the returned receiver.
+    /// You should continuously poll the receiver to receive updates.
+    ///
+    /// Panics if the buffer capacity is 0.
```
keyvankhademi: since we are returning a `Result` let's return Err instead of panicing.

lazer/sdk/rust/client/examples/symbols.rs
```diff
@@ -0,0 +1,28 @@
...
+        update_interval: Duration::from_secs(5),
+        ..Default::default()
+    });
+    let feeds = client.all_symbols_metadata_handle().await?;
+
+    loop {
+        println!("feeds len: {}", feeds.symbols().len());
+        println!("feed 1: {:?}", feeds.symbols().get(&PriceFeedId(1)));
+        sleep(Duration::from_secs(15)).await;
+    }
```
keyvankhademi, line 27: would be nice to have an example for stream too.

lazer/sdk/rust/client/src/history_client.rs
Riateche: Done.

lazer/sdk/rust/client/examples/symbols.rs
Riateche: Added.

PR #3044, opened 2025-09-12
chore(developer-hub): Oracle Integrity Staking(OIS) section

nidhi-singh02:
## How has this been tested?

- [ ] Added new tests


apps/developer-hub/source.config.ts
```diff
@@ -27,27 +26,7 @@ export const docs = defineDocs({
...
-      ],
-      inline: "tailing-curly-colon",
-      themes: {
-        light: "github-light",
-        dark: "github-dark",
-      },
-    },
     remarkPlugins: [remarkMath],
-    rehypePlugins: [rehypeKatex, rehypeCode],
+    rehypePlugins: (v) => [rehypeKatex, ...v],
```
nidhi-singh02: exactly as from fuma docs, this will preserve the default plugins. 
Do we need rehypeCodeOptions here ? @aditya520 as you added this.

https://fumadocs.dev/docs/ui/markdown/math

apps/developer-hub/content/docs/meta.json
```diff
@@ -1,3 +1,9 @@
 {
-  "pages": ["pyth-core", "pyth-token", "express-relay", "entropy"]
+  "pages": [
+    "pyth-core",
+    "pyth-token",
```
nidhi-singh02: would be changed later

apps/developer-hub/source.config.ts
aditya520: Yes, let's keep `rehypCode`. It is needed from inline code parsing

pnpm-lock.yaml
```diff
@@ -399,7 +399,7 @@ importers:
         version: 2.2.0(react@19.1.0)
       '@next/third-parties':
         specifier: 'catalog:'
-        version: 15.3.2(next@15.5.0(@babel/core@7.27.1)(@opentelemetry/api@1.9.0)(babel-plugin-react-compiler@19.1.0-rc.1)(react-dom@19.1.0(react@19.1.0))(react@19.1.0)(sass@1.86.1))(react@19.1.0)
+        version: 15.3.2(next@15.5.0(@babel/core@7.27.1)(@opentelemetry/api@1.9.0)(react-dom@19.1.0(react@19.1.0))(react@19.1.0)(sass@1.86.1))(react@19.1.0)
```
cprussin: This file likely shouldn't h ave changes since you didn't modify any dependencies; you should revert and try `pnpm i` again.  If you still get changes, it's because someone checked in code without remembering to update this file and it's OK to commit.

cprussin (dismissed):

pnpm-lock.yaml
nidhi-singh02: Thanks, done. Will keep it mind for the next time.

apps/developer-hub/source.config.ts
nidhi-singh02: Sure, I kept the above file as it was earlier from main, build for developer hub was failing https://vercel.com/pyth-network/developer-hub/9yaQyRHagqAtFosEumwBL9aUd44D?filter=errors

Fixed it, kept rehypeCodeOptions as well.

PR #3049, opened 2025-09-12
feat(pyth-lazer-protocol)!: enable subscribing by either price_feed_ids or symbols

tejasbadadare:
## Summary
- This PR adds a `symbols: Option<Vec<String>>` optional request parameter to `SubscriptionParams`, `LatestPriceRequest`, `PriceRequest` and makes `price_feed_ids` an Option instead of being required. This enables specifying feeds by symbol rather than ID if desired.
  - From an API standpoint this change should be backward compatible.
- `ignore_price_feed_ids` is backward-compatibly renamed to `ignore_price_feeds` so that it can be used regardless of whether the feeds are specified by `symbols` or `price_feed_ids`.
- Once this lands in the protocol crate, i'll add support for resolving the `symbols` into IDs in the router.

## Rationale
We want to support fetching latest price/subscribing to feeds via symbols, not just IDs. This is more user friendly for people trying out the system, as it eliminates the need to look up the mapping from symbol to ID. Paired with the new symbology, users should be able to discover feeds predictably.

## How has this been tested?

- [ ] Added new tests


lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -215,12 +221,28 @@ impl<'de> Deserialize<'de> for SubscriptionParams {
 
 impl SubscriptionParams {
     pub fn new(value: SubscriptionParamsRepr) -> Result<Self, &'static str> {
-        if value.price_feed_ids.is_empty() {
-            return Err("no price feed ids specified");
+        if value.price_feed_ids.is_none() && value.symbols.is_none() {
+            return Err("either price feed ids or symbols must be specified");
+        }
```
Riateche, line 226: I think we should also forbid passing both feed ids and symbols at the same time.

tejasbadadare: Yeah makes sense. I was debating between this and union-ing them in the server. But yeah i'll go ahead forbid passing both feed ids and symbols at the same time.

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -195,8 +200,9 @@ pub struct SubscriptionParamsRepr {
     #[serde(default = "default_parsed")]
     pub parsed: bool,
     pub channel: Channel,
-    #[serde(default)]
-    pub ignore_invalid_feed_ids: bool,
+    // "ignoreInvalidFeedIds" was renamed to "ignoreInvalidFeeds". "ignoreInvalidFeedIds" is still supported for compatibility.
+    #[serde(default, alias = "ignoreInvalidFeedIds")]
+    pub ignore_invalid_feeds: bool,
```
merolish, line 205: So this can be omitted altogether (defaulting to false) or one of ignoreInvalidFeedIds or ignoreInvalidFeeds can be specified? What about both?

merolish:
No objections. Guess there's no way around being messy with an API change that needs to retain compatibility.

lazer/sdk/rust/protocol/src/api.rs
tejasbadadare: Yeah this applies to the deserializer (e.g., from a JSON payload). Fields named `ignoreInvalidFeedIds` (the old name) will still be accepted, and will be aliased/renamed to the new `ignoreInvalidFeeds`. I'm actually not sure what happens if both are provided -- i assume serde would fail to deserialize the payload with some kind of "duplicate keys" error.

Riateche (dismissed):

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -181,7 +184,9 @@ impl<'de> Deserialize<'de> for Channel {
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SubscriptionParamsRepr {
-    pub price_feed_ids: Vec<PriceFeedId>,
+    // Either price feed ids or symbols must be specified.
+    pub price_feed_ids: Option<Vec<PriceFeedId>>,
+    pub symbols: Option<Vec<String>>,
```
danimhr, line 189: I think this is a new field here. Shouldn't we add default value to None, to make sure api is backward compatible?
Maybe it's okay by default but I just wanna make sure if that's the case

danimhr (dismissed):

lazer/sdk/rust/protocol/src/api.rs
Riateche: serde_json already defaults to `None` for an `Option` field if it's absent, so there is no need for `#[serde(default)]` in this case.

merolish (approved):

PR #3050, opened 2025-09-15
fix(dev hub): ui fixes

alexcambose:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/components/CopyAddress/index.tsx
+++ b/apps/developer-hub/src/components/CopyAddress/index.tsx
@@ -5,16 +5,31 @@ import { Link } from "@pythnetwork/component-library/Link";
 
 import styles from "./index.module.scss";
 
-const CopyAddress = ({ address, url }: { address: string; url?: string }) => {
+const truncate = (value: string, maxLength?: number) => {
+  if (!maxLength) {
+    return value;
+  }
+  return `${value.slice(0, maxLength)}...${value.slice(-maxLength)}`;
+};
+
+const CopyAddress = ({
+  address,
+  maxLength,
+  url,
+}: {
+  address: string;
+  maxLength?: number;
+  url?: string;
+}) => {
   return url ? (
     <div className={styles.address}>
       <Link href={url} target="_blank" rel="noreferrer">
-        {address}
+        {truncate(address, maxLength)}
       </Link>
       <CopyButton text={address} iconOnly />
     </div>
   ) : (
-    <CopyButton text={address}>{address}</CopyButton>
+    <CopyButton text={address}>{truncate(address, maxLength)}</CopyButton>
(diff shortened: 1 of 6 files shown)


packages/component-library/src/AppShell/index.tsx
```diff
@@ -30,36 +30,52 @@ type Tab = ComponentProps<typeof MainNavTabs>["tabs"][number] & {
...
+  ...props
+}: AppShellRootProps & AppBodyProps) => (
+  <AppShellRoot
+    enableAccessibilityReporting={enableAccessibilityReporting}
+    amplitudeApiKey={amplitudeApiKey}
+    googleAnalyticsId={googleAnalyticsId}
+    providers={providers}
+  >
+    <AppBody {...props} />
+  </AppShellRoot>
```
alexcambose, line 76: Split the AppShell into 2 components so we can have the fumadocs provider after `<body>` but before the App Shell header

apps/developer-hub/src/components/Root/theme.css
```diff
@@ -347,3 +347,14 @@
 :where(.prose) :where(h1, h2, h3, h4, h5, h6) {
   line-height: 1.2;
 }
+
+.prose p + ul,
+.prose p + ol {
```
alexcambose: this reduces list paddings

also there's no cleaner way to change the default styles other than globally styling cc https://github.com/pyth-network/pyth-crosschain/pull/3031#discussion_r2334832374

apps/developer-hub/src/components/CopyAddress/index.tsx
```diff
@@ -5,16 +5,31 @@ import { Link } from "@pythnetwork/component-library/Link";
...
+}: {
+  address: string;
+  maxLength?: number;
+  url?: string;
+}) => {
   return url ? (
     <div className={styles.address}>
       <Link href={url} target="_blank" rel="noreferrer">
-        {address}
+        {truncate(address, maxLength)}
```
cprussin: Memoize this?

apps/developer-hub/src/components/Root/theme.css
cprussin: :(

cprussin: Maybe add a comment around all this to explain why we're using global styles?  The thing to avoid is some dev coming along, seeing this, and thinking it's OK to use global styles

cprussin: Maybe even move this stuff to a separate file and import it, something like `fumadocs-global-styles-override.scss` with a big scary comment at the top

cprussin (approved):
very nice @alexcambose , thanks

PR #3057, opened 2025-09-17
chore(developer-hub): reward calculator and image import in ois

nidhi-singh02:
## Summary

- Reward calculator similar to how we have in the documentation.
- Image import and keeping image colocated to where they are used.

## How has this been tested?

- [ ] Added new tests


apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const [delegatorStake, setDelegatorStake] = useState(300);
+  const [maxCap, setMaxCap] = useState(500);
+  const [delegatorFee, setDelegatorFee] = useState(20);
+  const [rewardRate, setRewardRate] = useState(10);
+
+  const [publisherReward, setPublisherReward] = useState(0);
+  const [delegatorReward, setDelegatorReward] = useState(0);
+  const [publisherRewardRate, setPublisherRewardRate] = useState(0);
+  const [delegatorRewardRate, setDelegatorRewardRate] = useState(0);
+
```
nidhi-singh02: this file is referred from the reward simulator of the documentation.

apps/developer-hub/content/docs/oracle-integrity-staking/slashing-rulebook.mdx
```diff
@@ -6,6 +6,12 @@ description: >-
 ---
 
 import DualModeImage from "../../../src/components/DualModeImage";
+import OISSlashingInvestigationDark from "./images/ois_slashing_investigation_dark.png";
+import OISSlashingInvestigationLight from "./images/ois_slashing_investigation_light.png";
+import OISSlashingPoolPreDark from "./images/ois_slashing_pool_pre_dark.png";
+import OISSlashingPoolPreLight from "./images/ois_slashing_pool_pre_light.png";
+import OISSlashingPoolPostDark from "./images/ois_slashing_pool_post_dark.png";
+import OISSlashingPoolPostLight from "./images/ois_slashing_pool_post_light.png";
```
nidhi-singh02, line 14: using image import and moved the images from the public folder inside the folder where they are used.

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+    calculateRewards();
+  }, [publisherStake, delegatorStake, maxCap, delegatorFee, rewardRate]);
+
+  return (
+    <Card
+      variant="secondary"
+      title="Reward Simulator"
+      nonInteractive
+      className={clsx(styles.card)}
```
alexcambose: `clsx` is useful when we want to combine different class names together, if it's just one class it's okay to just do `className={styles.card}`.

apps/developer-hub/src/components/RewardSimulator/index.module.scss
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.resultsGrid {
+  display: grid;
+  grid-template-columns: 1fr;
+  gap: theme.spacing(6);
+
+  @media (width >= 640px) {
+    grid-template-columns: 1fr 1fr;
+  }
```
alexcambose: please use the mixing we have: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/src/components/Home/results.module.scss#L6-L8

source: https://github.com/pyth-network/pyth-crosschain/blob/main/packages/component-library/src/theme.scss#L898-L901

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+      setDelegatorReward(Number(finalDelegatorReward.toFixed(2)));
+      setPublisherRewardRate(
+        Number(((finalPublisherReward * 100) / publisherStake).toFixed(2)),
+      );
+      setDelegatorRewardRate(
+        Number(((finalDelegatorReward * 100) / delegatorStake).toFixed(2)),
+      );
+    };
+
+    calculateRewards();
```
alexcambose: we don't need a separate function here, you can just write the code in the top level block.
Usually we use a function when it's async code, which is not in this case

alexcambose (changes requested):

apps/developer-hub/src/components/RewardSimulator/index.module.scss
```diff
@@ -0,0 +1,85 @@
...
+  margin-top: theme.spacing(6);
+}
+
+.inputGrid {
+  display: grid;
+  grid-template-columns: 1fr;
+  gap: theme.spacing(6);
+  padding: theme.spacing(4);
+
+  @media (width >= 640px) {
```
cprussin: We should use `@theme.breakpoint`

apps/developer-hub/src/components/RewardSimulator/index.module.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.inputGroup {
+  display: flex;
+  flex-direction: column;
+  gap: theme.spacing(2);
+}
+
+.input {
+  padding: theme.spacing(2) theme.spacing(3);
+  border: 1px solid var(--color-fd-border);
```
cprussin: In any custom components, I'd suggest we stick to colors from the theme, not colors from fumadocs (they should match but the theme is the actual source of truth, and we try to make fumadocs match the theme as much as possible)

apps/developer-hub/src/components/RewardSimulator/index.module.scss
```diff
@@ -0,0 +1,85 @@
...
+  color: var(--color-fd-muted-foreground);
+}
+
+.resultLabel {
+  font-weight: theme.font-weight("medium");
+}
+
+.resultValue {
+  font-family:
+    ui-monospace, SFMono-Regular, Consolas, "Liberation Mono", Menlo, monospace;
```
cprussin: We should not use custom fonts in specific components.  I actually don't know if we have a global monospace font, but the correct way to do this would be to  make sure we have one and to reuse it, not to define a new font-family

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+      const finalPublisherReward = publisherRewardBase + delegatorFeeAmount;
+
+      setPublisherReward(Number(finalPublisherReward.toFixed(2)));
+      setDelegatorReward(Number(finalDelegatorReward.toFixed(2)));
+      setPublisherRewardRate(
+        Number(((finalPublisherReward * 100) / publisherStake).toFixed(2)),
+      );
+      setDelegatorRewardRate(
+        Number(((finalDelegatorReward * 100) / delegatorStake).toFixed(2)),
+      );
```
cprussin: Generally when you always set a bunch of state at the same time, there isn't a good reason to use separate state, it just makes the code messier.  It's better to just combine these into an object and do something like:

```tsx
const [rewards, setRewards] = useState({ publisher: 0, delegator: 0, publisherRate: 0, delegatorRate: 0 });

useEffect(() => {
  ...
  setRewards({
    publisher: ...,
    delegator: ...,
    publisherRate: ...,
    delegatorRate: ...
  });
}, [])
```

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+            id="publisher-stake"
+            type="number"
+            value={publisherStake}
+            onChange={(e) => {
+              setPublisherStake(Number(e.target.value));
+            }}
+            className={styles.input ?? ""}
+            min="0"
+          />
+        </div>
```
cprussin: It is not idiomatic to be using effects and controlled inputs for this kind of thing.  See e.g. https://medium.com/@nimitwadhwa722/useeffect-most-abused-react-hook-640ff616e362 -- this is a class of React patterns called "effect abuse" and it's a problem because effects can be very error prone (it's very easy to stack effects in a way that causes render problems or to forget a effect trigger, or add too many effect triggers, and these things can be extremely hard to debug and notice) and can be very hard to understand (how would a reader have any expectation that changing the `publisherStake` state triggers an effect that changes some other state, if they're looking down here at this part of the code?)

Rather than doing it using effects and controlled input, my suggestion is to use a form and attach `onChange` callbacks which trigger a calculation based on the form state.  A simplified version would look like this:

```tsx
const MyComponent = () => {
  const [rewards, setRewards] = useState(0);
  const doSubmit = useCallback((e) => e.target.form.requestSubmit(), []);
  const recalculateRewards = useCallback((e) => {
    e.preventDefault();
    setRewards(new FormData(e.target).get("foo"));
  }, []);

  return (
    <form onSubmit={recalculateRewards} onChange={doSubmit}>
      <input name="foo" />
      <div>
        Rewards are: {rewards}
      </div>
    </form>
  );
}
```

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: addressed.

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </div>
+
+      <div className={styles.resultsSection}>
+        <div className={styles.resultsGrid}>
+          <div className={styles.resultGroup}>
+            <h4 className={styles.resultTitle}>Calculated Rewards</h4>
+            <div className={styles.resultValues}>
+              <p className={styles.resultItem}>
+                <span className={styles.resultLabel}>
+                  Publisher Reward (
```
cprussin: Usually this amount of nesting of elements with only one child is unnecessary and indicates poor css use, probably worth revisiting this structure?

apps/developer-hub/src/components/RewardSimulator/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+                  <MathExpression>
+                    R<Sup>p</Sup>
```
cprussin, line 181: You should probably use a `<dl>` here

apps/developer-hub/src/components/RewardSimulator/index.module.scss
nidhi-singh02: done.

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: fair point, I just copied this code from the docs. Addressed it.

apps/developer-hub/src/components/RewardSimulator/index.module.scss
nidhi-singh02: sure, using colors from theme.

apps/developer-hub/src/components/RewardSimulator/index.module.scss
nidhi-singh02: We don't have a global monospace font, I have created a font family for monospace in theme. and using it.

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: using useState and setRewards without use effect, as per another comment.

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: sure, using dl

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: Refactored the structure,

apps/developer-hub/src/components/RewardSimulator/index.tsx
nidhi-singh02: Thanks for suggesting the recommended pattern, going to stay away from useEffects, addressed this. 

Reusing the code from the doc was not a good idea.

apps/developer-hub/src/components/RewardSimulator/index.module.scss
```diff
@@ -0,0 +1,86 @@
...
+
+.resultTerm {
+  font-size: theme.font-size("sm");
+  color: theme.color("muted");
+  font-weight: theme.font-weight("medium");
+  margin: 0;
+}
+
+.resultValue {
+  font-family: theme.font-family("monospace");
```
alexcambose, line 82: can you try using `var(--font-mono)` - this is from tailwind https://tailwindcss.com/docs/font-family and it might work.
If it does then we don't need a separate font-family mapping in the ui-components

nidhi-singh02: perfect, using it. Removed the font family I created in the component library theme.

alexcambose (dismissed):

apps/developer-hub/src/components/RewardSimulator/index.module.scss
cprussin: Actually I disagree with this, I don't think we should be using classes out of tailwind. It won't port to anywhere else and only works here because of fumadocs

nidhi-singh02: ummm should I revert in that case ? @cprussin @alexcambose

nidhi-singh02: as discussed, reverted this change.

packages/component-library/src/theme.scss
```diff
@@ -30,6 +30,21 @@ $font-weight: (
...
+    "consolas",
+    "Liberation Mono",
+    "menlo",
+    "monospace",
+  ),
+);
+
+@function font-family($family) {
+  @return map-get-strict($font-family, $family);
+}
```
cprussin, line 46: I think this is great but at some point we should probably have a specific font we ship for mono text, cc @partyparrotgreg not sure if you already have one in mind

PR #3060, opened 2025-09-17
feat: HIP-3 oracle pusher

merolish:
## Summary

HIP-3 pusher application draft.

## Rationale

Serve as market data provider for HIP-3 deployers.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/.github/workflows/docker-hip-3-pusher.yml
+++ b/.github/workflows/docker-hip-3-pusher.yml
@@ -1,0 +1,55 @@
+name: Build and Push hip-3-pusher Image
+on:
+  push:
+    tags:
+      - hip-3-pusher-v*
+  pull_request:
+    paths:
+      - "apps/hip-3-pusher/**"
+  workflow_dispatch:
+    inputs:
+      dispatch_description:
+        description: "Dispatch description"
+        required: true
+        type: string
+permissions:
+  contents: read
+  id-token: write
+  packages: write
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: pyth-network/hip-3-pusher
+jobs:
+  hip-3-pusher-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set image tag to version of the git tag
+        if: ${{ startsWith(github.ref, 'refs/tags/hip-3-pusher-v') }}
+        run: |
+          PREFIX="refs/tags/hip-3-pusher-"
+          VERSION="${GITHUB_REF:${#PREFIX}}"
+          echo "IMAGE_TAG=${VERSION}" >> "${GITHUB_ENV}"
+      - name: Set image tag to the git commit hash
+        if: ${{ !startsWith(github.ref, 'refs/tags/hip-3-pusher-v') }}
+        run: |
+          echo "IMAGE_TAG=${{ github.sha }}" >> "${GITHUB_ENV}"
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: metadata_hip_3_pusher
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      - name: Build and push server docker image
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
+        with:
+          context: .
+          file: "./apps/hip-3-pusher/Dockerfile"
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.metadata_hip_3_pusher.outputs.tags }}
+          labels: ${{ steps.metadata_hip_3_pusher.outputs.labels }}
--- a/apps/hip-3-pusher/.gitignore
+++ b/apps/hip-3-pusher/.gitignore
@@ -1,0 +1,10 @@
+# Python-generated files
+__pycache__/
+*.py[oc]
+build/
+dist/
+wheels/
+*.egg-info
+
+# Virtual environments
+.venv
--- a/apps/hip-3-pusher/Dockerfile
+++ b/apps/hip-3-pusher/Dockerfile
@@ -1,5 +1,5 @@
 # Use a Python image with uv pre-installed
-FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+FROM ghcr.io/astral-sh/uv:python3.13-trixie-slim
 
 # Install the project into `/app`
 WORKDIR /app
--- a/apps/hip-3-pusher/Dockerfile
+++ b/apps/hip-3-pusher/Dockerfile
@@ -1,0 +1,30 @@
+# Use a Python image with uv pre-installed
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+
+# Install the project into `/app`
+WORKDIR /app
+
+# Enable bytecode compilation
+ENV UV_COMPILE_BYTECODE=1
+
+# Copy from the cache instead of linking since it's a mounted volume
+ENV UV_LINK_MODE=copy
+
+# Ensure installed tools can be executed out of the box
+ENV UV_TOOL_BIN_DIR=/usr/local/bin
+
+# Install the project's dependencies using the lockfile and settings
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project --no-dev
+
+# Then, add the rest of the project source code and install it
+# Installing separately from its dependencies allows optimal layer caching
+COPY src/ ./src/
+COPY config/ ./config/
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --no-dev
+
+# Run the app by default
+CMD ["uv", "run", "src/main.py", "--config", "config/config.toml"]
--- a/apps/hip-3-pusher/Dockerfile
+++ b/apps/hip-3-pusher/Dockerfile
@@ -13,19 +13,19 @@ ENV UV_LINK_MODE=copy
 # Ensure installed tools can be executed out of the box
 ENV UV_TOOL_BIN_DIR=/usr/local/bin
 
-COPY uv.lock .
-COPY pyproject.toml .
+COPY apps/hip-3-pusher/uv.lock .
+COPY apps/hip-3-pusher/pyproject.toml .
 
 # Install the project's dependencies using the lockfile and settings
 RUN --mount=type=cache,target=/root/.cache/uv \
-    --mount=type=bind,source=uv.lock,target=uv.lock \
-    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=apps/hip-3-pusher/uv.lock,target=uv.lock \
+    --mount=type=bind,source=apps/hip-3-pusher/pyproject.toml,target=pyproject.toml \
     uv sync --locked --no-install-project --no-dev
 
 # Then, add the rest of the project source code and install it
 # Installing separately from its dependencies allows optimal layer caching
-COPY src/ ./src/
-COPY config/ ./config/
+COPY apps/hip-3-pusher/src/ ./src/
+COPY apps/hip-3-pusher/config/ ./config/
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync --locked --no-dev
 
--- a/apps/hip-3-pusher/Dockerfile
+++ b/apps/hip-3-pusher/Dockerfile
@@ -13,6 +13,9 @@ ENV UV_LINK_MODE=copy
 # Ensure installed tools can be executed out of the box
 ENV UV_TOOL_BIN_DIR=/usr/local/bin
 
+COPY uv.lock .
+COPY pyproject.toml .
+
 # Install the project's dependencies using the lockfile and settings
 RUN --mount=type=cache,target=/root/.cache/uv \
     --mount=type=bind,source=uv.lock,target=uv.lock \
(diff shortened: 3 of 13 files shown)


tejasbadadare (approved):

apps/hip-3-pusher/README.md
tejasbadadare, line 1: Pls dont forget to add a readme!

apps/hip-3-pusher/pyproject.toml
```diff
@@ -0,0 +1,15 @@
...
+requires-python = ">=3.13"
+dependencies = [
+    "asn1crypto>=1.5.1",
+    "boto3>=1.40.31",
+    "cryptography>=45.0.7",
+    "hyperliquid-python-sdk>=0.19.0",
+    "loguru>=0.7.3",
+    "toml>=0.10.2",
+    "websockets>=15.0.1",
+]
```
tejasbadadare, line 15: Using >= opens us to to breaking major version changes. Let's use ^ (minor version updates allowed) or ~= (patch version updates allowed).

apps/hip-3-pusher/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "hip-3-pusher"
+version = "0.1.0"
+description = "Add your description here"
```
tejasbadadare, line 4: TODO for later

apps/hip-3-pusher/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    --mount=type=bind,source=apps/hip-3-pusher/uv.lock,target=uv.lock \
+    --mount=type=bind,source=apps/hip-3-pusher/pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project --no-dev
+
+# Then, add the rest of the project source code and install it
+# Installing separately from its dependencies allows optimal layer caching
+COPY apps/hip-3-pusher/src/ ./src/
+COPY apps/hip-3-pusher/config/ ./config/
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --no-dev
```
tejasbadadare, line 30: Great that we have proper layer caching :)

apps/hip-3-pusher/.gitignore
tejasbadadare, line 1: Also add `**/config.toml` to avoid accidentally pushing secrets

apps/hip-3-pusher/src/price_state.py
```diff
@@ -0,0 +1,11 @@
...
+    """
+    Maintain latest prices seen across listeners and publisher.
+    """
+    def __init__(self):
+        self.latest_oracle_price = None
+        self.latest_mark_price = None
+        self.lazer_base_price = None
+        self.lazer_quote_price = None
+        self.hermes_base_price = None
+        self.hermes_quote_price = None
```
tejasbadadare, line 11: Let's add typehints for these attributes so readers and editors can figure out what they are supposed to be

```python
class PriceState:
    latest_oracle_price: SomeType
    latest_oracle_price: SomeType
    ...
```

apps/hip-3-pusher/src/main.py
```diff
@@ -0,0 +1,47 @@
...
+    parser = argparse.ArgumentParser(description="hip3-agent command-line arguments")
+    parser.add_argument(
+        "-c", "--config",
+        required=True,
+        help="hip3-agent config file",
+    )
+    config_path = parser.parse_args().config
+    with open(config_path, "r") as config_file:
+        config = toml.load(config_file)
+    return config
```
tejasbadadare, line 23: Let's make a config type and put it in its own file. That way we fail fast if we load a bad config, and its statically known what the config contains. Pydantic is great for this bc it runs validations on the inputs when it deserializes into the target type. Sample from their docs: 

```python
from pydantic import BaseModel
import tomllib

class DatabaseConfig(BaseModel):
    host: str
    port: int

class ServiceConfig(BaseModel):
    debug: bool
    name: str

class Config(BaseModel):
    database: DatabaseConfig
    service: ServiceConfig

# Load TOML
"""
# config.toml
[database]
host = "localhost"
port = 5432

[service]
debug = true
name = "myapp"
"""
with open("config.toml", "rb") as f:
    raw = tomllib.load(f)

cfg = Config(**raw)

print(cfg.database.host)   # localhost
print(cfg.service.debug)   # True
```

apps/hip-3-pusher/src/lazer_listener.py
```diff
@@ -0,0 +1,87 @@
...
+        await asyncio.gather(*(self.subscribe_single(router_url) for router_url in self.router_urls))
+
+    async def subscribe_single(self, router_url):
+        while True:
+            try:
+                await self.subscribe_single_inner(router_url)
+            except websockets.ConnectionClosed:
+                logger.error("Connection to {} closed; retrying", router_url)
+            except Exception as e:
+                logger.exception("Error on {}: {}", router_url, e)
```
tejasbadadare, line 44: Check out [tenacity](https://tenacity.readthedocs.io/en/latest/) for decorators that provide patterns like this. It also lets you easily set up backoffs and failovers.

apps/hip-3-pusher/src/lazer_listener.py
```diff
@@ -0,0 +1,87 @@
...
+            await ws.send(json.dumps(subscribe_request))
+            logger.info("Sent Lazer subscribe request to {}", self.router_urls[0])
+
+            # listen for updates
+            async for message in ws:
+                try:
+                    data = json.loads(message)
+                    self.parse_lazer_message(data)
+                except json.JSONDecodeError as e:
+                    logger.error("Failed to decode JSON message: {}", e)
```
tejasbadadare, line 63: TODO for later -- reconnect on no data received for some time

apps/hip-3-pusher/config/config.toml
```diff
@@ -0,0 +1,23 @@
...
+[kms]
+enable_kms = false
+key_path = "/path/to/kms_key.txt"
+aws_region_name = "ap-northeast-1"
+
+[lazer]
+router_urls = ["wss://pyth-lazer-0.dourolabs.app/v1/stream", "wss://pyth-lazer-1.dourolabs.app/v1/stream"]
+api_key = "lazer_api_key"
+base_feed_id = 1   # BTC
+quote_feed_id = 8  # USDT
```
tejasbadadare, line 18: What's the quote asset used for?

apps/hip-3-pusher/config/config.toml
```diff
@@ -0,0 +1,23 @@
...
+enable_publish = false
+
+[kms]
+enable_kms = false
+key_path = "/path/to/kms_key.txt"
+aws_region_name = "ap-northeast-1"
+
+[lazer]
+router_urls = ["wss://pyth-lazer-0.dourolabs.app/v1/stream", "wss://pyth-lazer-1.dourolabs.app/v1/stream"]
+api_key = "lazer_api_key"
```
tejasbadadare, line 16: consider loading this from the env

apps/hip-3-pusher/src/lazer_listener.py
```diff
@@ -0,0 +1,87 @@
...
+
+    async def subscribe_single_inner(self, router_url):
+        headers = {
+            "Authorization": f"Bearer {self.api_key}",
+        }
+
+        async with websockets.connect(router_url, additional_headers=headers) as ws:
+            subscribe_request = self.get_subscribe_request(1)
+
+            await ws.send(json.dumps(subscribe_request))
```
tejasbadadare, line 54: If we get back a subscription failure it might be good to wrap that in a non-retryable exception so we fail fast instead of eternally retrying

apps/hip-3-pusher/src/hyperliquid_listener.py
```diff
@@ -0,0 +1,33 @@
...
+    See https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/websocket
+    """
+    def __init__(self, config: dict, price_state: PriceState):
+        self.market_symbol = config["hyperliquid"]["market_symbol"]
+        url = TESTNET_API_URL if config["hyperliquid"].get("use_testnet", True) else MAINNET_API_URL
+        self.info = Info(base_url=url)
+        self.price_state = price_state
+
+    def subscribe(self):
+        self.info.subscribe({"type": "activeAssetCtx", "coin": self.market_symbol}, self.on_activeAssetCtx)
```
tejasbadadare, line 21: Needs the retry wrapper that exists for the Lazer and Hermes feeds. It might be worthwhile to extract all that out into a LIstener superclass.

apps/hip-3-pusher/src/hyperliquid_listener.py
```diff
@@ -0,0 +1,33 @@
...
+    def on_activeAssetCtx(self, message):
+        """
+        Parse oraclePx and markPx from perp context update
+
+        :param message: activeAssetCtx websocket update message
+        :return: None
+        """
+        ctx = message["data"]["ctx"]
+        self.price_state.latest_oracle_price = ctx["oraclePx"]
+        self.price_state.latest_mark_price = ctx["markPx"]
```
tejasbadadare, line 32: We should also store the timestamp so we know how stale the price is in case the data source disconnects and have to serve stale prices. It could also inform fallback behavior (it can choose the most recent price from the backup data sources.)

apps/hip-3-pusher/src/kms_signer.py
tejasbadadare, line 1: Will hold off on reviewing this file since its WIP

apps/hip-3-pusher/src/main.py
```diff
@@ -0,0 +1,47 @@
...
+    config = load_config()
+
+    price_state = PriceState()
+    publisher = Publisher(config, price_state)
+    hyperliquid_listener = HyperliquidListener(config, price_state)
+    lazer_listener = LazerListener(config, price_state)
+    hermes_listener = HermesListener(config, price_state)
+
+    # TODO: Probably pull this out of the sdk.
+    hyperliquid_listener.subscribe()
```
tejasbadadare, line 37: Do you know if the HL `subscribe()` function uses asyncio? Asking because the other components use asyncio which is convenient because its a single-threaded event loop, and there's no locking/synchronization needed on `price_state`. But if the HL SDK isn't using that, we might have to worry about synchronizing `price_state`.

apps/hip-3-pusher/src/publisher.py
```diff
@@ -0,0 +1,76 @@
...
+        self.publish_interval = float(config["hyperliquid"]["publish_interval"])
+        self.kms_signer = None
+        self.enable_kms = False
+        self.use_testnet = config["hyperliquid"].get("use_testnet", True)
+
+        if config["kms"]["enable_kms"]:
+            self.enable_kms = True
+            oracle_account = None
+            kms_key_path = config["kms"]["key_path"]
+            kms_key_id = open(kms_key_path, "r").read().strip()
```
tejasbadadare, line 30: use Pathlib's `read_text` for FS operations like this (it also closes the file for you)

apps/hip-3-pusher/src/lazer_listener.py
```diff
@@ -0,0 +1,87 @@
...
+            logger.debug("price_feeds: {}", price_feeds)
+            for feed_update in price_feeds:
+                feed_id = feed_update.get("priceFeedId", None)
+                price = feed_update.get("price", None)
+                if feed_id is None or price is None:
+                    continue
+                if feed_id == self.base_feed_id:
+                    self.price_state.lazer_base_price = price
+                if feed_id == self.quote_feed_id:
+                    self.price_state.lazer_quote_price = price
```
tejasbadadare, line 85: what are we using base_price and quote_price for? i noticed we don't push them anywhere in publisher.py.

apps/hip-3-pusher/src/hyperliquid_listener.py
tejasbadadare: Nvm, i see you do this in #3061 ✅

.github/workflows/docker-hip-3-pusher.yml
```diff
@@ -0,0 +1,55 @@
+name: Build and Push hip-3-pusher Image
+on:
+  push:
+    tags:
+      - hip-3-pusher-v*
+  pull_request:
+    paths:
+      - "apps/hip-3-pusher/**"
```
tejasbadadare, line 8: Probably don't want to build and push whenever a PR is opened/updated. Would remove this trigger

merolish: The push part actually doesn't happen on PRs for whatever reason. Somebody on the team explained it to me at some point when I was working on agent.

apps/hip-3-pusher/Dockerfile
merolish: Yeah, that was courtesy of the [uv docker example](https://github.com/astral-sh/uv-docker-example/blob/main/Dockerfile). Astral is crushing it.

apps/hip-3-pusher/.gitignore
merolish: Intent is for the config.toml in git to be an example. Locally I use separately named config files, and of course k8s mounts our deployment configs.

apps/hip-3-pusher/src/price_state.py
merolish: I need to get better at this in general -- I came of age long before type hints were widespread.

apps/hip-3-pusher/src/main.py
merolish: Good call, effectively matching what we do in Rust.

apps/hip-3-pusher/src/lazer_listener.py
merolish: Ah yes, I'd forgotten that.

apps/hip-3-pusher/config/config.toml
merolish: You mean env var or from a 1pw secret?

apps/hip-3-pusher/src/lazer_listener.py
merolish: Good call.

apps/hip-3-pusher/src/hyperliquid_listener.py
merolish: Right, I think I need to scrap the sdk here and handle the connect/subscription manually, don't think this sdk function is resilient or even gives feedback at all.

apps/hip-3-pusher/src/kms_signer.py
merolish: Unfortunately yes, hopefully I can get it to match, it's unfortunately a bit of reverse engineering both KMS and the sdk and eth libraries.

apps/hip-3-pusher/src/main.py
merolish: As per other comment, planning to do this myself.

apps/hip-3-pusher/src/lazer_listener.py
merolish: HL oracle prices are denominated in USDT, so for pyth prices I'm dividing BTCUSD / USDTUSD.

PR #3061, opened 2025-09-18
feat(hip-3-pusher): Price fallback logic and staleness checks

merolish:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/hip-3-pusher/config/config.toml
+++ b/apps/hip-3-pusher/config/config.toml
@@ -1,3 +1,5 @@
+stale_price_threshold_seconds = 5
+
 [hyperliquid]
 market_name = ""
 market_symbol = "BTC"
@@ -12,12 +14,16 @@ key_path = "/path/to/kms_key.txt"
 aws_region_name = "ap-northeast-1"
 
 [lazer]
-router_urls = ["wss://pyth-lazer-0.dourolabs.app/v1/stream", "wss://pyth-lazer-1.dourolabs.app/v1/stream"]
+lazer_urls = ["wss://pyth-lazer-0.dourolabs.app/v1/stream", "wss://pyth-lazer-1.dourolabs.app/v1/stream"]
 api_key = "lazer_api_key"
 base_feed_id = 1   # BTC
+base_feed_exponent = -8
 quote_feed_id = 8  # USDT
+quote_feed_exponent = -8
 
 [hermes]
-urls = ["wss://hermes.pyth.network/ws"]
-base_id = "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"   # BTC
-quote_id = "2b89b9dc8fdf9f34709a5b106b472f0f39bb6ca9ce04b0fd7f2e971688e2e53b"  # USDT
+hermes_urls = ["wss://hermes.pyth.network/ws"]
+base_feed_id = "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"   # BTC
+base_feed_exponent = -8
+quote_feed_id = "2b89b9dc8fdf9f34709a5b106b472f0f39bb6ca9ce04b0fd7f2e971688e2e53b"  # USDT
+quote_feed_exponent = -8
--- a/apps/hip-3-pusher/src/hyperliquid_listener.py
+++ b/apps/hip-3-pusher/src/hyperliquid_listener.py
@@ -1,4 +1,5 @@
 from loguru import logger
+import time
 
 from hyperliquid.info import Info
 from hyperliquid.utils.constants import TESTNET_API_URL, MAINNET_API_URL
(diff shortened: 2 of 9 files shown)


tejasbadadare (approved):

apps/hip-3-pusher/src/hyperliquid_listener.py
```diff
@@ -28,6 +29,7 @@ def on_activeAssetCtx(self, message):
         :return: None
         """
         ctx = message["data"]["ctx"]
-        self.price_state.latest_oracle_price = ctx["oraclePx"]
-        self.price_state.latest_mark_price = ctx["markPx"]
-        logger.debug("on_activeAssetCtx: oraclePx: {} marketPx: {}", self.price_state.latest_oracle_price, self.price_state.latest_mark_price)
+        self.price_state.hl_oracle_price = ctx["oraclePx"]
+        self.price_state.hl_mark_price = ctx["markPx"]
+        logger.debug("on_activeAssetCtx: oraclePx: {} marketPx: {}", self.price_state.hl_oracle_price, self.price_state.hl_mark_price)
+        self.price_state.latest_hl_timestamp = time.time()
```
tejasbadadare, line 35: Would recommend storing the timestamp together with the price in one type to enforce they are updated together. something like 
```python
class PriceUpdate:
    asset_id: str
    price: float
    timestamp: int
```

apps/hip-3-pusher/src/price_state.py
```diff
@@ -1,11 +1,73 @@
...
+        self.lazer_quote_exponent = config["lazer"]["quote_feed_exponent"]
+        self.latest_lazer_timestamp = now
+
         self.hermes_base_price = None
+        self.hermes_base_exponent = config["hermes"]["base_feed_exponent"]
         self.hermes_quote_price = None
+        self.hermes_quote_exponent = config["hermes"]["quote_feed_exponent"]
+        self.latest_hermes_timestamp = now
+
+    def get_current_oracle_price(self):
```
tejasbadadare, line 31: Worth thinking through how this evolves as we need to support pushing multiple prices. Also, we might want to have this orchestration in a different module and split up the state to be managed by the individual data sources instead of a single mono-state.

apps/hip-3-pusher/src/price_state.py
```diff
@@ -1,11 +1,73 @@
...
+            time_diff = now - self.latest_lazer_timestamp
+            if time_diff < self.stale_price_threshold_seconds:
+                return self.get_lazer_price()
+            else:
+                logger.error("Lazer price stale by {} seconds", time_diff)
+        else:
+            logger.error("Lazer base/quote prices not received yet")
+
+        logger.error("All prices missing or stale!")
+        return None
```
tejasbadadare, line 63: Any reason we fall back to Hermes before Lazer?

apps/hip-3-pusher/src/publisher.py
```diff
@@ -47,15 +47,21 @@ def __init__(self, config: dict, price_state: PriceState):
...
+            if not oracle_px:
+                logger.error("No valid oracle price available!")
+                return
+            else:
+                logger.debug("Current oracle price: {}", oracle_px)
+                oracle_pxs[self.market_symbol] = oracle_px
+
             mark_pxs = []
+            #if self.price_state.hl_mark_price:
+            #    mark_pxs.append({self.market_symbol: self.price_state.hl_mark_price})
```
tejasbadadare, line 62: Why not publish the mark prices?

apps/hip-3-pusher/config/config.toml
```diff
@@ -12,12 +14,16 @@ key_path = "/path/to/kms_key.txt"
...
 
 [hermes]
-urls = ["wss://hermes.pyth.network/ws"]
-base_id = "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"   # BTC
-quote_id = "2b89b9dc8fdf9f34709a5b106b472f0f39bb6ca9ce04b0fd7f2e971688e2e53b"  # USDT
+hermes_urls = ["wss://hermes.pyth.network/ws"]
+base_feed_id = "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"   # BTC
+base_feed_exponent = -8
+quote_feed_id = "2b89b9dc8fdf9f34709a5b106b472f0f39bb6ca9ce04b0fd7f2e971688e2e53b"  # USDT
+quote_feed_exponent = -8
```
tejasbadadare, line 29: We'll eventually need to split out the price config into its own file and source the metadata (like exponent) from the data sources themselves.

apps/hip-3-pusher/src/price_state.py
merolish: No, I think I just mixed up the order you guys suggested for whatever reason. Dunno if currently we trust one more than the other.

apps/hip-3-pusher/src/publisher.py
merolish: They told us not to for now. Note that this will also include the actual HIP-3's market data in the calculation.

merolish: These are just mark prices for the HyperCore perps which are useful data points but separate.

apps/hip-3-pusher/config/config.toml
merolish: Agreed, will need to scale this into a general set of prices and calcs.

PR #3065, opened 2025-09-19
feat(pyth-lazer-protocol): add validations for price request types, add unknown_symbols to InvalidFeedSubscriptionDetails

tejasbadadare:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/sdk/rust/client/Cargo.toml
+++ b/lazer/sdk/rust/client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-client"
-version = "8.0.0"
+version = "8.1.0"
 edition = "2021"
 description = "A Rust client for Pyth Lazer"
 license = "Apache-2.0"
--- a/lazer/sdk/rust/protocol/src/api.rs
+++ b/lazer/sdk/rust/protocol/src/api.rs
@@ -16,7 +16,8 @@ use crate::{
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct LatestPriceRequest {
+pub struct LatestPriceRequestRepr {
+    // Either price feed ids or symbols must be specified.
     pub price_feed_ids: Option<Vec<PriceFeedId>>,
     pub symbols: Option<Vec<String>>,
     pub properties: Vec<PriceFeedProperty>,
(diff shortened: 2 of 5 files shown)


lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -50,6 +118,73 @@ pub struct PriceRequest {
...
+        D: serde::Deserializer<'de>,
+    {
+        let value = PriceRequestRepr::deserialize(deserializer)?;
+        Self::new(value).map_err(Error::custom)
+    }
+}
+
+impl PriceRequest {
+    pub fn new(value: PriceRequestRepr) -> Result<Self, &'static str> {
+        if value.price_feed_ids.is_none() && value.symbols.is_none() {
```
Riateche: I think it's better to move out the repeating parts of code to a common verification function.

lazer/sdk/rust/client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-client"
-version = "8.0.0"
```
danimhr: On the main branch, this version is `8.0.1` but it's `8.0.0` in the PR diff.
Not sure why this happens, but can we check if it's fine?

tejasbadadare: Yeahh I need to clean up some of these conflicts. Will fix

tejasbadadare: Oh i see what you mean. Yeah that's weird... usually GH has a bit of eventual consistency with these things but it should have updated by now. Gonna pull in & merge latest main, that should fix it.

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -32,9 +33,55 @@ pub struct LatestPriceRequest {
...
+
+impl<'de> Deserialize<'de> for LatestPriceRequest {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = LatestPriceRequestRepr::deserialize(deserializer)?;
+        Self::new(value).map_err(Error::custom)
+    }
+}
```
darunrs, line 48: Is there a reason that this was manually implemented? It looks functionally very similar to the actual trait. This follows for the other trait impls below.

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -32,9 +33,55 @@ pub struct LatestPriceRequest {
...
+
+impl LatestPriceRequest {
+    pub fn new(value: LatestPriceRequestRepr) -> Result<Self, &'static str> {
```
darunrs, line 51: Is there a significance in the error being a static string here, instead of using anyhow Error types?

lazer/sdk/rust/protocol/src/api.rs
tejasbadadare: It's handling two decoupled tasks, deserializing the input representation and applying the validation rules in `LatestPriceRequest::new`. It gives some type safety - once you have a PriceRequest, you know it's valid.

lazer/sdk/rust/protocol/src/api.rs
tejasbadadare: I don't think so, but was conforming with the existing pattern in the file :)

lazer/sdk/rust/protocol/src/api.rs
```diff
@@ -16,7 +16,8 @@ use crate::{
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct LatestPriceRequest {
+pub struct LatestPriceRequestRepr {
```
keyvankhademi, line 19: I think you should increase the protocol version here too.

PR #3071, opened 2025-09-19
Chore(dev hub) openapi exploration

aditya520:
## How has this been tested?

- [ ] Added new tests


apps/developer-hub/scripts/generate-docs.ts
```diff
@@ -1,15 +1,50 @@
...
+        contentDir: "./content/docs/openapi",
+      },
+      items: Object.keys(products).map((productName) => ({
+        path: `${productName}/index.mdx`,
+      })),
+    },
+  });
+}
+
+function getProductName(title: string): string {
```
alexcambose, line 39: nit I don't think you need the return type here `: string {`

apps/developer-hub/scripts/generate-docs.ts
```diff
@@ -1,15 +1,50 @@
...
-  includeDescription: true,
-});
+      if (output.type === "operation") {
+        const operation =
+          document.paths?.[output.item.path]?.[output.item.method];
+        const operationId =
+          operation?.operationId ??
+          output.item.path.replaceAll(/[^a-zA-Z0-9]/g, "_");
+        return `${productName}/${operationId}`;
+      }
```
alexcambose, line 23: can you add a comment here with why we need to do this

apps/developer-hub/scripts/generate-docs.ts
```diff
@@ -1,15 +1,50 @@
...
-// const fortunaSchema = schemas[fortunaOpenApiUrl];
-// const fortunaStagingSchema = schemas[fortunaStagingOpenApiUrl];
+export async function generateDocs() {
+  await generateFiles({
+    input: openapi,
+    output: outDir,
+    per: "operation",
+    name: (output, document) => {
+      // Extract product name from the OpenAPI document info
+      const productName = getProductName(document.info.title || "unknown");
```
alexcambose, line 14: `"unknown"` seems to be repeated here and in `getProductName`, we can make that cleaner

apps/developer-hub/scripts/generate-docs.ts
aditya520: We will have more openApi schemas in future. So this will help me in future to keep me in separate folders.

alexcambose (approved):

PR #3073, opened 2025-09-22
fix(dev-hub): sidebar and table of contents positioning

alexcambose:
## Summary

Making dev hub sidebar and table of contents sticky for better UX

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/components/search-button.tsx
+++ b/apps/developer-hub/src/components/search-button.tsx
@@ -20,14 +20,15 @@ export const SearchButton = () => {
         largeScreenContent="Search"
         onClick={handleSearch}
       />
-      {open && createPortal(
-        <DefaultSearchDialog
-          open={open}
-          onOpenChange={setOpen}
-          api="/api/search"
-        />,
-        document.body
-      )}
+      {open &&
+        createPortal(
+          <DefaultSearchDialog
+            open={open}
+            onOpenChange={setOpen}
+            api="/api/search"
+          />,
+          document.body,
+        )}
     </>
   );
 };


packages/component-library/src/AppShell/index.module.scss
```diff
@@ -33,9 +32,7 @@
       background: theme.color("background", "primary");
       border-top-left-radius: calc(var(--offset) * theme.border-radius("xl"));
       border-top-right-radius: calc(var(--offset) * theme.border-radius("xl"));
-      overflow: hidden auto;
-      transform: scale(calc(100% - (var(--offset) * 5%)));
```
alexcambose, line 37: @cprussin I removed these, for some reason it was messing with the position fixed/sticky and we weren't able to fix elements to the top of the page. 
I'm not sure what these were doing or why we had them.

cprussin: This is used when opening the mobile drawer to offset the page underneath it.  It does need to be here -- I'm surprised you're having issues with sticky positioning as we are actively using sticky positioning on table headers in IH to stick table headers to the top of the page....

packages/component-library/src/AppShell/index.module.scss
```diff
@@ -14,7 +14,6 @@
   }
 
   .body {
-    background: black;
```
cprussin, line 17: This is also needed to make the mobile drawer work correctly

packages/component-library/src/AppShell/index.module.scss
alexcambose: Reverted

alexcambose:
Table of contents is still not fixed but spent too much time on this and the most obvious is making the sidebar sticky which works fine

packages/component-library/src/AppShell/index.module.scss
```diff
@@ -33,9 +33,9 @@
       background: theme.color("background", "primary");
       border-top-left-radius: calc(var(--offset) * theme.border-radius("xl"));
       border-top-right-radius: calc(var(--offset) * theme.border-radius("xl"));
+      height: 100dvh;
```
cprussin, line 36: I think this change is safe but I'd be careful with it and make sure to test across screen sizes, especially with the interactions with overlays / drawers

cprussin (dismissed):

packages/component-library/src/AppShell/index.module.scss
alexcambose: actually this is just reverting to the initial version, think just the ordering changed here

cprussin: ah yes I missed the deleted line below, nice ok

cprussin (approved):

PR #3075, opened 2025-09-23
feat(lazer/js/sdk): add REST endpoints and examples

tejasbadadare:
## Summary

- Add Lazer REST API support to `PythLazerClient`. Previously, the client was only focused on the WS streaming API. This PR adds support for our other consumer endpoints and makes the SDK more useful.
  - Add `get_symbols()` -- fetch feed metadata, with optional `query` and `asset` filters
  - Add `get_latest_price()` -- fetch latest prices by ID or symbol
  - Add `get_price()` -- fetch historical prices by ID or symbol

- Update config. Now that the client also handles REST endpoints and not just WS, expanded the configuration to account for these. The config is now:
  - ```typescript
    export type LazerClientConfig = {
      token: string;
      metadataServiceUrl?: string; // Metadata URL with default (currently history service)
      priceServiceUrl?: string; // Price URL with default (currently router)
      logger?: Logger; // Optional logger
      webSocketPoolConfig?: WebSocketPoolConfig; // Old config (WS params)
    };
    ```
  - I named the URL params based on the product surface they expose rather than internal service names, since these will change once the endpoints are unified under an API gateway.

- Add more examples to showcase different APIs. Check it out: https://asciinema.org/a/T4rLhqb3EAt2tM7dB1JtrngNA
  - `pnpm run example:streaming`: streams prices over WS, parses them to human-friendly format using the exponent, and renders them in-place on the terminal to showcase the fast update rate
  - `pnpm run example:symbols`: get all symbols, search by symbol name and/or asset type
  - `pnpm run example:history`: get latest price and timestamped price

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/sdk/js/examples/streaming.ts
+++ b/lazer/sdk/js/examples/streaming.ts
@@ -117,6 +117,7 @@ client.unsubscribe(1);
 client.unsubscribe(2);
 client.unsubscribe(3);
 
+// Clear screen and move cursor to top
 process.stdout.write("\u001B[2J\u001B[H");
 console.log("🛑 Shutting down Pyth Lazer demo after 30 seconds...");
 console.log("👋 Goodbye!");


lazer/sdk/js/examples/streaming.ts
```diff
@@ -0,0 +1,115 @@
...
+  jsonBinaryEncoding: "hex",
+});
+
+await new Promise((resolve) => setTimeout(resolve, 30_000));
+
+client.unsubscribe(1);
+client.unsubscribe(2);
+client.unsubscribe(3);
+
+process.stdout.write("\u001B[2J\u001B[H");
```
keyvankhademi, line 121: what is this?

tejasbadadare: It clears the screen haha. I'll leave a comment

lazer/sdk/js/examples/util.ts
```diff
@@ -0,0 +1,131 @@
...
+    const timeAgo = Math.round(Date.now() - feed.lastUpdate.getTime());
+
+    const symbolName = symbolsMap.get(Number(feed.priceFeedId));
+    const displayName = symbolName
+      ? `Feed ID: ${feed.priceFeedId.toString()} (${symbolName})`
+      : `Feed ID: ${feed.priceFeedId.toString()}`;
+
+    console.log(`\u001B[36m${(index + 1).toString()}. ${displayName}\u001B[0m`);
+    console.log(
+      `   💰 Price: \u001B[32m$${readablePrice.toLocaleString("en-US", { minimumFractionDigits: 2, maximumFractionDigits: 8 })}\u001B[0m`,
```
darunrs, line 52: nit: I feel like if we want to commit to doing things like coloring our logs, it's better to add something like chalk. Seeing ANSI codes just feels weird to me.

lazer/sdk/js/examples/streaming.ts
```diff
@@ -0,0 +1,125 @@
...
+      break;
+    }
+    case "binary": {
+      // Print out the binary hex messages if you want:
+      // if ("solana" in message.value) {
+      //   console.info("solana message:", message.value.solana?.toString("hex"));
+      // }
+      // if ("evm" in message.value) {
+      //   console.info("evm message:", message.value.evm?.toString("hex"));
+      // }
```
darunrs, line 65: Remove?

lazer/sdk/js/examples/streaming.ts
```diff
@@ -0,0 +1,125 @@
+/* eslint-disable no-console */
+/* eslint-disable @typescript-eslint/no-empty-function */
+/* eslint-disable unicorn/prefer-top-level-await */
+
+import { renderFeeds, refreshFeedDisplay } from "./util.js";
+import type { JsonUpdate } from "../src/index.js";
+import { PythLazerClient } from "../src/index.js";
+
+// Ignore debug messages
+console.debug = () => {};
```
darunrs, line 10: Should we be doing this at a file level? I feel like we should respect the log level that is passed into the execution call.

lazer/sdk/js/src/protocol.ts
```diff
@@ -97,3 +97,63 @@ export const FORMAT_MAGICS_LE = {
   LE_ECDSA: 1_296_547_300,
   LE_UNSIGNED: 1_499_680_012,
 };
+
```
darunrs, line 100: Hopefully maintaining this isn't a big problem. I wonder if we should set up something where we define the types in Rust somewhere once, and then derive types from it for other languages. Maybe @cprussin has some ideas. This is probably fine for now.

lazer/sdk/js/examples/util.ts
tejasbadadare: Yeahh didn't wanna add a dependency just for this since its not really service logs or a core part of the package.

lazer/sdk/js/examples/streaming.ts
tejasbadadare: Left it in as a hint for people who want to use the binary message

lazer/sdk/js/examples/streaming.ts
tejasbadadare: Yeah there is no `console.setLevel` unfortunately, this is a common workaround to set "log level" when using `console` directly as a logger. Functionally its the same as passing in a `ts-log` instance or something, but again its just an example so didn't want to prescribe things like which logger to pass in.

lazer/sdk/js/src/protocol.ts
tejasbadadare: Yeah it's a good point, i don't have a strong opinion here -- in the Hermes client, we get the types from the hermes /docs OpenAPI json, which seems as good a place as any to consider the source of truth for types. We don't have OpenAPI support yet in router yet though. There's some interesting projects out there for this as well: https://github.com/Aleph-Alpha/ts-rs

PR #3085, opened 2025-09-25
feat: Allow date range selection in charts

fhqvst:
## Summary

Initially based on #2986, this PR adds support for "resolution" and "lookback" selection in the chart component.

image

## How has this been tested?

- [ ] Added new tests


apps/insights/src/app/historical-prices/route.ts
```diff
@@ -1,17 +1,102 @@
...
+  if (!parsed.success) {
+    return new Response(`Invalid params: ${parsed.error.message}`, {
+      status: 400,
+    });
+  }
+
+  const { symbol, publisher, cluster, from, to, resolution } = parsed.data;
+
+  try {
+    checkMaxDataPointsInvariant(from, to, resolution);
```
cprussin: Personally, I'm not a fan of functions with hidden side effects.  I also think the implementation of `checkMaxDataPointsInvariant` is unnecessarily awkward -- I prefer to avoid the unnecessary use of `let` bindings, and it's less concise than it needs to be.  Additionally, the current implementation is using [exceptions as control flow](https://wiki.c2.com/?DontUseExceptionsForFlowControl) which I generally try to avoid.

How something like this instead?

```ts
if ((to - from) / RESOLUTION_TO_SECONDS[resolution] > MAX_DATA_POINTS) {
  return new Response("Unsupported resolution for date range", { status: 400 });
}

// ...

const ONE_MINUTE_IN_SECONDS = 60;
const ONE_HOUR_IN_SECONDS = 60 * ONE_MINUTE_IN_SECONDS;
const RESOLUTION_TO_SECONDS = {
  "1 MINUTE": ONE_MINUTE_IN_SECONDS,
  "5 MINUTE": 5 * ONE_MINUTE_IN_SECONDS,
  "1 HOUR": ONE_HOUR_IN_SECONDS,
  "1 DAY":  24 * ONE_HOUR_IN_SECONDS,
};

File: apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx [hunk 2]
```diff
@@ -0,0 +1,102 @@
...
+import {
+  LOOKBACK_TO_RESOLUTION,
+  LOOKBACKS,
+  RESOLUTION_TO_LOOKBACK,
+  RESOLUTIONS,
+  useChartLookback,
+  useChartResolution,
+} from "./use-chart-toolbar";
+
+const ENABLE_RESOLUTION_SELECTOR = false;
```
cprussin, line 18: I'm not sure there's a good reason to feature-gate this but if you're going to, I suggest you do by putting the flag in `config/server.ts` and allow an env var to override it

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export const ChartToolbar = () => {
+  const logger = useLogger();
+  const [lookback, setLookback] = useChartLookback();
+  const [resolution, setResolution] = useChartResolution();
+
+  const handleLookbackChange = useCallback(
+    (newValue: Key) => {
```
cprussin: I don't think there's any reason to use `Key` here -- if you make the resolutions an enum or a string union, the types in `<Select>` should work out that you can take the enum / union here and you don't have to type guard or cast

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+const useChartElem = (symbol: string, feedId: string) => {
+  const logger = useLogger();
+  const [lookback] = useChartLookback();
+  const [resolution] = useChartResolution();
+  const chartContainerRef = useRef<HTMLDivElement | null>(null);
+  const chartRef = useRef<ChartRefContents | undefined>(undefined);
+  const isBackfilling = useRef(false);
+  const priceFormatter = usePriceFormatter();
+  const resolutionRef = useRef(resolution);
```
cprussin: What's the point of this ref and why not just use `resolution` directly everywhere?

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,490 @@
...
+      url.searchParams.set("resolution", resolution);
+      url.searchParams.set("cluster", "pythnet");
+
+      abortControllerRef.current = new AbortController();
+      abortControllerRef.current.signal.addEventListener("abort", () => {
+        isBackfilling.current = false;
+      });
+
+      fetch(url, { signal: abortControllerRef.current.signal })
+        .then(async (data) => historicalDataSchema.parse(await data.json()))
```
cprussin: I don't think we should mix async/await and then, just move the schema parsing down to the next `.then` instead

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,490 @@
...
+          const currentHistoricalPriceData = chartRef.current.price
+            .data()
+            .filter((d) => isLineData(d));
+          const currentHistoricalConfidenceHighData =
+            chartRef.current.confidenceHigh.data().filter((d) => isLineData(d));
+          const currentHistoricalConfidenceLowData =
+            chartRef.current.confidenceLow.data().filter((d) => isLineData(d));
+
+          const newHistoricalPriceData = data.map((d) => ({
+            time: Number(d.timestamp) as UTCTimestamp,
```
cprussin: We should do this transformation in `historicalDataSchema` instead

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,490 @@
...
+        .then(async (data) => historicalDataSchema.parse(await data.json()))
+        .then((data) => {
+          if (!chartRef.current) {
+            return;
+          }
+
+          // Get the current historical price data
+          const currentHistoricalPriceData = chartRef.current.price
+            .data()
+            .filter((d) => isLineData(d));
```
cprussin: I'm not clear on why you need to filter these arrays, can you add a brief comment explaining why `isLineData` is necessary and what kind of data the `data()` function might return which wouldn't be line data?

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,490 @@
...
+        timeVisible: true,
+        secondsVisible: true,
+      },
+      localization: {
+        priceFormatter: priceFormatter.format,
+      },
+    });
+
+    const confidenceHigh = chart.addSeries(AreaSeries, confidenceConfig);
+    const confidenceLow = chart.addSeries(AreaSeries, confidenceConfig);
```
cprussin, line 247: Since you moved to using an AreaSeries and don't need to draw the two lines, could you compress this into a single series centered on the price instead of a series above and one below?

apps/insights/src/components/PriceFeed/Chart/use-chart-toolbar.tsx
```diff
@@ -0,0 +1,58 @@
+import { parseAsStringLiteral, useQueryState } from "nuqs";
+
+export const RESOLUTIONS = ["1s", "1m", "5m", "1H", "1D"] as const;
+export type Resolution = (typeof RESOLUTIONS)[number];
+
+export const LOOKBACKS = ["1m", "1H", "1D", "1W", "1M"] as const;
+export type Lookback = (typeof LOOKBACKS)[number];
```
cprussin: I really don't like the name "lookbacks" for this -- if we do the auto-resolution adjust, this is more accurately a "snap to" than a lookback.  Even now I think this is effectively behaving as a "snap to"

apps/insights/src/services/clickhouse.ts
```diff
@@ -336,15 +336,47 @@ export const _getPublisherAverageScoreHistory = async ({
...
+  const clusterClause =
+    cluster === "pythtest"
+      ? "cluster IN {clusters: Array(String)}"
+      : "cluster = {cluster: String}";
+
+  if (publisher) {
+    queryParams.publisher = publisher;
+  }
+  if (cluster === "pythtest") {
+    queryParams.clusters = ["pythnet", "pythtest-conformance"];
```
cprussin: I'm confused why we would fetch from both clusters when pythtest is specified?  It also does not look like the cluster is ever passed in so why even have this as a config?

apps/insights/src/services/clickhouse.ts
```diff
@@ -354,20 +386,23 @@ export const getHistoricalPrices = async ({
...
           FROM prices
-          WHERE cluster = 'pythnet'
-          AND symbol = {symbol: String}
-          AND version = 2
-          AND time > fromUnixTimestamp(toInt64({until: String})) - INTERVAL 5 MINUTE
-          AND time < fromUnixTimestamp(toInt64({until: String}))
-          AND publisher = ''
-          GROUP BY time
-          ORDER BY time ASC
+          PREWHERE
```
cprussin, line 375: Hmm I'm not sure I understand the reason for splitting the where clause and using prewhere here, does postgres not optimize this query well?

fhqvst: to really get to the bottom here we should probably measure a couple different queries and see which one is faster. 

that said, the `(cluster, publisher, symbol)` combo follows the specified physical sort order of the table, and so forcing the engine to filter by that first _should_ be some kind of guarantee that we're always doing lookups as fast as possible.

slight chance that this is a premature optimization, but it's so small i figured adding it couldnt hurt

apps/insights/src/services/clickhouse.ts
fhqvst: so this in an artifact from [here](https://github.com/pyth-network/web-api/blob/main/web_api/db.py#L106-L108), but for some reason i've messed it up (pythnet instead of pythtest)... and tbh i'm not even sure we wanna do this at all.

i'll remove the addition of `pythtest-conformance` so that cluster can only be `pythnet` or `pythtest`, and we can add back any conformance-related stuff later.

apps/insights/src/components/PriceFeed/Chart/use-chart-toolbar.tsx
fhqvst: yep yep, i don't have a strong opinion here. another candidate was `INTERVAL`, but i found that it could be easily confused with `RESOLUTION`. what do you think?

apps/insights/src/components/PriceFeed/Chart/chart.tsx
fhqvst: i don't think we can do this if we want the purple background between the lines, but i can have another look

apps/insights/src/components/PriceFeed/Chart/chart.tsx
fhqvst: let me double check this, but i think `price.data()` returns an intersection of many types of `Data` (something like `AreaData | LineData | ...` which is why i needed that type guard

apps/insights/src/components/PriceFeed/Chart/chart.tsx
fhqvst: no point - i think we can delete this now

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
fhqvst: iirc it did _not_ wanna play nice with enums, but i'll have another look

i _think_ the right thing to do is for `Select` to take a `K` generic (perhaps one that extend `string | number`)? maybe not

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
fhqvst: this is a client component though i'd probably need to put it in the closest parent page.  either way this should be temporary and i'll probably delete this flag over the next few days.

apps/insights/src/app/historical-prices/route.ts
fhqvst: how would you feel about doing `checkMaxDataPointsInvariant(...): boolean` instead?

```
if (!checkMaxDataPointsInvariant()) {
   return new Response("Unsupported resolution")
}
```

cprussin: I think it's reasonable to extract to a separate function but `checkMaxDataPointsInvariant` is IMO a pretty bad name, how about either `getNumDataPoints(to, from, resolution) > MAX_DATA_POINTS` or maybe `isNumDataPointsOverMax(to, from, resolution)`?  I personally think `getNumDataPoints` is simpler / more obvious.

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
cprussin: ah, fair enough!

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
cprussin: hmm I think enums should be OK since we're using them elsewhere e.g. [here](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/entropy-explorer/src/components/Home/search-controls.tsx#L109-L155), and [here](https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/Publishers/publishers-card.tsx#L268-L282) is an example of where we use it with a union in case you choose that.

I wouldn't use a `string | number` here; IMO a select should nearly always map to an enum or a union -- there are a restricted set of valid values and `string | number` implies a freeform input.  Not only is the input not freeform so `string | number` doesn't accurately describe the values it produces, but also if that's what you use then you'll unnecessarily need to do validation & type checking OR type casting.  Representing things as an enum/union is both more accurate and will type check correctly without additional checks.

apps/insights/src/components/PriceFeed/Chart/chart.tsx
cprussin: ah yeah understood

apps/insights/src/components/PriceFeed/Chart/use-chart-toolbar.tsx
cprussin: the best I have in mind is `SNAP_TO_PERIODS`.  I'm sure there's a better name, but `INTERVAL` doesn't seem clear to me either.  Maybe @alexcambose has some ideas?

cprussin: `QUICK_SELECT_WINDOWS` maybe?

apps/insights/src/services/clickhouse.ts
cprussin: Ah so to be clear `pythtest` is not a cluster name, we sometimes mix the names `pythtest` and `pythtest-conformance` but `pythtest-conformance` is the proper name.

I don't think you will ever have a good reason to display a price chart for the pythtest-conformance aggregate -- the aggregate price in pythtest-conformance doesn't actually mean anything since the publishers use that environment to test their feeds and so the data quality is wildly inaccurate.  I'd probably just get rid of this.  But if we find there is in fact some reason to bring this back, the clusters will be `pythnet` and `pythtest-conformance`, not `pythtest` which does not in fact correspond to a real cluster.

apps/insights/src/services/clickhouse.ts
cprussin: Yeah that's fine, it doesn't bother me to have it if you think it could help with perf (and has no risk of having the opposite impact by accident), but I also doubt this is going to do anything differently than what the query plan optimizer would do anyways...

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
fhqvst: ah right, the problem here is with `SingleToggleGroup`, not `Select`.

as far as i can tell, the `SingleToggleGroup` does not take a generic, and thus we need the type guard.

apps/insights/src/services/clickhouse.ts
fhqvst: thanks, I'll just hard-code it to `pythnet`.

apps/insights/src/app/historical-prices/route.ts
```diff
@@ -1,17 +1,96 @@
...
+      diff = diff / 3600;
+      break;
+    }
+    case "1 DAY": {
+      diff = diff / 86_400;
+      break;
+    }
+  }
+
+  return diff;
```
cprussin: This function still feels quite messy to me in a lot of small nitpicky ways:

- The `let` binding is not needed and I really dislike using let bindings unless mutability is really a requirement
- It's not instantly obvious where the magic numbers come from -- assigning to clearly named constants would make it more clear
- `diff / 60 / 5` requires thinking about operator precedence to calculate properly, again assigning the denominator to a constant would alleviate this
- The actual equation is not changing in any switch branch -- it's just the value of the denominator -- but the structure of the code doesn't make the structure of the equation obvious

None of these are a huge deal, but lots of cognitive overhead nits for what is fundamentally a very simple function so I'd clean it up

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const handleResolutionChanged = useCallback(
+    (resolution: Resolution) => {
+      setResolution(resolution).catch((error: unknown) => {
+        logger.error("Failed to update resolution", error);
+      });
+      setQuickSelectWindow(RESOLUTION_TO_QUICK_SELECT_WINDOW[resolution]).catch(
+        (error: unknown) => {
+          logger.error("Failed to update quick select window", error);
+        },
+      );
```
cprussin, line 34: I'm not sure I agree that changing the resolution should snap to a window -- I think I'd expect that changing the resolution keeps the window I'm in, or maybe zooms out or in around the center point of my current window if needed

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -0,0 +1,489 @@
...
+        .then((rawData) => rawData.json())
+        .then((jsonData) => {
+          if (!chartRef.current) {
+            return;
+          }
+
+          const data = historicalDataSchema.parse(jsonData);
+
+          // Get the current historical price data
+          // Note that .data() returns (WhitespaceData | LineData)[], hence the type cast
```
cprussin: What is `WhitespaceData`?  Would be helpful to indicate why we know the compiler is wrong here and why we feel safe that we'll only actually have `LineData`

apps/insights/src/components/PriceFeed/Chart/use-chart-toolbar.tsx
```diff
@@ -0,0 +1,69 @@
...
+      return 3_600_000;
+    }
+    case "1D": {
+      return 86_400_000;
+    }
+    case "1W": {
+      return 604_800_000;
+    }
+    case "1M": {
+      return 2_629_746_000;
```
cprussin: Total nit, and I'm sure these numbers are correct, but they aren't instantly obviously so.  I always like being very explicit about these constants:

```ts
const ONE_SECOND_IN_MS = 1000;
const ONE_MINUTE_IN_MS = 60 * ONE_SECOND_IN_MS;
const ONE_HOUR_IN_MS = 60 * ONE_MINUTE_IN_MS;
const ONE_DAY_IN_MS = 24 * ONE_HOUR_IN_MS;
const ONE_WEEK_IN_MS = 7 * ONE_DAY_IN_MS;
const ONE_MONTH_IN_MS = 30 * ONE_DAY_IN_MS;

const QUICK_SELECT_WINDOW_TO_MS = {
  1m: ONE_MINUTE_IN_MS,
  1H: ONE_HOUR_IN_MS,
  ...
}
```

I know it's a bit redundant and pedantic, but IMO it's much simpler to trust the numbers and know where they come from.  It also resolves ambiguity around things like the 1 month value, which could be defined different ways (did you get 2,629,746,000 by calculating a month as 4 weeks, or 30 days, or 31 days, or some other definition?).  I've seen a surprising number of bugs by someone having a slightly incorrect number in a place like this and no one catching it because it's not instantly obvious  that it's correct / incorrect.

cprussin (approved):

apps/insights/src/components/PriceFeed/Chart/chart-toolbar.tsx
fhqvst: Right, this is just how trading view behaves. If we decide to remove resolution selection completely (which it sounds like we might), then this problem goes away on its own.

apps/insights/src/app/historical-prices/route.ts
fhqvst: Not sure what to change here tbh - how about this?


```tsx
const SECONDS_PER_MINUTE = 60
const SECONDS_PER_FIVE_MINUTES = 5 * SECONDS_PER_MINUTE;
const SECONDS_PER_HOUR = 60 * SECONDS_PER_MINUTE;
const SECONDS_PER_DAY = 24 * SECONDS_PER_HOUR;

/**
 * Computes the number of data points are rendered for a given resolution, in a given date range.
 */
function getNumDataPoints(
  to: number,
  from: number,
  resolution: "1 SECOND" | "1 MINUTE" | "5 MINUTE" | "1 HOUR" | "1 DAY",
) {
  const diff = to - from; // seconds
  switch (resolution) {
    case "1 SECOND": {
      return diff;
    }
    case "1 MINUTE": {
      return diff / SECONDS_PER_MINUTE;
    }
    case "5 MINUTE": {
      return diff / SECONDS_PER_FIVE_MINUTES;
    }
    case "1 HOUR": {
      return diff / SECONDS_PER_HOUR;
    }
    case "1 DAY": {
      return diff / SECONDS_PER_DAY;
    }
  }
}
```

cprussin: I'd do this:

```ts
type Resolution = "1 SECOND" | "1 MINUTE" | "5 MINUTE" | "1 HOUR" | "1 DAY";

const ONE_MINUTE_IN_SECONDS = 60
const ONE_HOUR_IN_SECONDS = 60 * ONE_MINUTE_IN_SECONDS;

const SECONDS_IN_ONE_PERIOD: Record<Resolution, number> = {
  "1 SECOND": 1,
  "1 MINUTE": ONE_MINUTE_IN_SECONDS,
  "5 MINUTE": 5 * ONE_MINUTE_IN_SECONDS,
  "1 HOUR": ONE_HOUR_IN_SECONDS,
  "1 DAY": 24 * ONE_HOUR_IN_SECONDS
}

const getNumDataPoints = (from: number, to: number, resolution: Resolution) =>
  (to - from) / SECONDS_IN_ONE_PERIOD[resolution];
```

PR #3088, opened 2025-09-26
feat(component-library): table grid

alexcambose:
## Summary

Adding an initial version of ag-grid based grid.

## How has this been tested?

- [ ] Added new tests


packages/component-library/src/TableGrid/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      return {
+        ...colDef,
+        // the types in ag-grid are `any` for the cellRenderers which is throwing an error here
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        cellRenderer: isLoading
+          ? (colDef.loadingCellRenderer ?? SkeletonCellRenderer)
+          : colDef.cellRenderer,
+      };
+    });
+  }, [colDefs, isLoading]);
```
fhqvst: IMO we should try to avoid mapping colDefs as much as we possibly can, since it implicitly means we've added an abstraction layer on top of AG grid, and in my experience this quickly becomes very messy.

For this specific use case is there any chance we can leverage the [`cellRendererSelector`](https://www.ag-grid.com/javascript-data-grid/component-cell-renderer/#dynamic-component-selection) + `defaultColDef` instead?

Also - is it really correct that we should swap out the renderers of all cells instead of showing a [loading overlay](https://www.ag-grid.com/javascript-data-grid/overlays/#loading-overlay)?

packages/component-library/src/TableGrid/table-grid-props.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type TableGridProps<TData extends Record<string, unknown>> = {
+  rowData: TData[];
+  colDefs: ExtendedColDef<TData>[];
+  isLoading?: boolean;
+  cardProps?: Omit<CardProps<"div">, "children" | "footer"> & {
+    nonInteractive?: true;
+  };
+  pagination?: boolean;
+} & Omit<AgGridReactProps<TData>, "rowData" | "defaultColDef" | "columnDefs">;
```
fhqvst: As mentioned in the other comment, we should probably avoid abstracting the AG grid API for as long as we can - it will be very difficult if not impossible to change this later.

For example, adding an `isLoading` here while still allowing to pass AG grid's own `loading` param is confusing, so we should choose one or the other!

Now, if we choose to keep `isLoading`, it means we have a custom param that might have undocumented side effects, meaning I need to look in the `TableGrid` component source code to understand exactly what it does.

packages/component-library/src/TableGrid/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const [pageSize, setPageSize] = useState(10);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [totalPages, setTotalPages] = useState(1);
+
+  const defaultColDef = useMemo(() => {
+    return {
+      cellRenderer: SkeletonCellRenderer,
+      flex: 1,
+    };
+  }, []);
```
fhqvst: Even though the `useMemo` dep array is empty, we should probably put this as a plain constant outside of the component. I actually don't know how AG grid behaves if the `defaultColDef` reference changes.

packages/component-library/src/TableGrid/index.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+      description={value}
+    />
+  </div>
+);
+
+const FeedCellRendererLoading = () => (
+  <div style={{ height: "100%", display: "flex", alignItems: "center" }}>
+    <SymbolPairTag isLoading />
+  </div>
+);
```
fhqvst, line 75: Very interesting discussion to be had here as well regarding what the right abstractions are.

- Should `priceCellRenderer` inherit from a `numberCellRenderer`?
    - Perhaps there should even be a `priceColDef`?
        - And perhaps there should even be a `priceColDefFactory`, to which I can pass parameters like `colorizeNumbers`?
            - ...and so on ad infinitum (this was only 3 layers of abstraction deep!)

packages/component-library/src/TableGrid/index.tsx
alexcambose: I wanted to have a nicer loading state rather than disabling the whole thing - especially if the grid is full screen height. DIdn't know about about [cellRendererSelector](https://www.ag-grid.com/javascript-data-grid/component-cell-renderer/#dynamic-component-selection) I'll try it out.

packages/component-library/src/TableGrid/index.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  --ag-background-color: #{theme.color("background", "primary")};
+  --ag-header-background-color: var(--ag-background-color);
+  --ag-foreground-color: #{theme.color("paragraph")};
+  --ag-accent-color: #{theme.color("button", "outline", "background", "active")};
+  --ag-header-font-size: #{theme.font-size("xs")};
+  --ag-header-font-weight: #{theme.font-weight("medium")};
+  --ag-cell-font-size: #{theme.font-size("xs")};
+  --ag-wrapper-border: none;
+  --ag-cell-text-color: #{theme.color("paragraph")};
+  --ag-data-font-size: #{theme.font-size("base")};
```
cprussin, line 14: nice!

packages/component-library/src/TableGrid/index.stories.tsx
alexcambose: Yeah I think with these kind of things the best way is let things as simple as possible and then create the abstraction as needed, when we start duplicating the code or having a similar logic in 2-3 places.

packages/component-library/src/TableGrid/index.tsx
alexcambose: Later edit: actually cellRendererSelector is on a per cell basis so we would still be doing the same mapping over colDefs, and also it might not work as its a function that uses the row data, so we might need to recreate this function that returns different renderers based on the loading state.
 
I propose we leave it like this for now and if it causes issue then we can remove it favor of the default loading overlay. It's very surprising it's so difficult to have a loading skeleton instead of the default loading overlay  - which is a much better UX.

fhqvst: There's also a [`loadingOverlayComponent`](https://www.ag-grid.com/javascript-data-grid/grid-options/#reference-overlays-loadingOverlayComponent) param we can pass to the grid btw, but maybe you've already looked at that.

alexcambose: yeah, I'd say let's go with the current approach, more than happy to change it later if it's causing issues

PR #3089, opened 2025-09-26
fix: skeleton fill

alexcambose:
## Summary

Fixed the fill property on the skeleton

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/packages/component-library/src/Skeleton/index.tsx
+++ b/packages/component-library/src/Skeleton/index.tsx
@@ -23,10 +23,7 @@ export const Skeleton = ({
     {...(width &&
       !fill && { style: { "--skeleton-width": width } as CSSProperties })}
     data-fill={fill ? "" : undefined}
-    className={clsx(
-      styles.skeleton,
-      className,
-    )}
+    className={clsx(styles.skeleton, className)}
   >
     <span className={clsx(styles.skeletonInner, className)} {...props}>
       <Label>{label ?? "Loading"}</Label>


packages/component-library/src/Skeleton/index.tsx
```diff
@@ -11,22 +11,28 @@ type Props = Omit<ComponentProps<"span">, "children"> & {
...
+  fill,
+  ...props
+}: Props) => (
+  <span
+    data-fill-width={width === undefined ? "" : undefined}
+    {...(width &&
+      !fill && { style: { "--skeleton-width": width } as CSSProperties })}
+    className={clsx(
+      styles.skeleton,
+      { [styles.fullSkeleton ?? ""]: fill },
```
cprussin: Super minor nit but I generally prefer doing variations as data attributes instead, it plays very nicely with react-aria:

```tsx
data-full={fill ? "" : undefined}
```

packages/component-library/src/Skeleton/index.tsx
```diff
@@ -11,22 +11,28 @@ type Props = Omit<ComponentProps<"span">, "children"> & {
...
+      { [className ?? ""]: fill },
```
cprussin: I think this class name should be added in regardless of whether `fill` is true right?  I think the way I wrote this before was just wrong

cprussin (approved):
Minor suggestions, feel free to disregard

PR #3090, opened 2025-09-27
feat(components-library) Social Links Drawer

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/packages/component-library/src/Header/index.module.scss
+++ b/packages/component-library/src/Header/index.module.scss
@@ -219,4 +219,87 @@
       }
     }
   }
+
+  .communityLinkList {
+    display: flex;
+    flex-flow: column nowrap;
+    gap: theme.spacing(4);
+
+    .title {
+      @include theme.text("lg", "medium");
+
+      color: theme.color("heading");
+    }
+
+    .communityGrid {
+      display: grid;
+      grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+      gap: theme.spacing(3);
+      width: 100%;
+
+      @include theme.breakpoint("sm") {
+        grid-template-columns: repeat(auto-fit, minmax(160px, 1fr));
+        gap: theme.spacing(2);
+      }
+    }
+
+    .communityCard {
+      aspect-ratio: 1;
+      min-height: 160px;
+      padding: theme.spacing(6);
+      display: flex;
+      align-items: center;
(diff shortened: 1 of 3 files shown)


packages/component-library/src/Header/index.module.scss
```diff
@@ -219,4 +219,87 @@
...
+
+      @include theme.breakpoint("sm") {
+        grid-template-columns: repeat(auto-fit, minmax(160px, 1fr));
+        gap: theme.spacing(2);
+      }
+    }
+
+    .communityCard {
+      aspect-ratio: 1;
+      min-height: 160px;
```
cprussin, line 248: Please stick to `theme.spacing` for spacing units

packages/component-library/src/Header/index.module.scss
```diff
@@ -219,4 +219,87 @@
...
+      justify-content: center;
+      text-align: center;
+      transition:
+        transform 0.2s ease,
+        box-shadow 0.2s ease;
+
+      &:hover {
+        transform: translateY(-2px);
+        box-shadow: 0 8px 25px rgb(0 0 0 / 15%);
+      }
```
cprussin, line 261: This hover styling appears nowhere else in the design system; it's inconsistent and we should make this match everything else

packages/component-library/src/Header/index.module.scss
```diff
@@ -219,4 +219,87 @@
...
+      align-items: center;
+      justify-content: center;
+      width: theme.spacing(12);
+      height: theme.spacing(12);
+      color: theme.color("states", "data", "normal");
+
+      svg {
+        width: 100%;
+        height: 100%;
+      }
```
cprussin, line 290: This styling should be unneeded, since you're using phosphor icons the default width and height are set to `1em`.  This means that the right way to style them is to set `font-size: theme.spacing(12)` on `.communityIcon` instead of setting `width` and `height` and having the styling here.

cprussin (changes requested):
The code is pretty solid!  However, the styling is off -- it doesn't match the link styling elsewhere in the design system, and the cards are huge at smaller screen sizes, e.g. here's a smaller browser window:

Image

And here's a mobile screen:

Image

I suggest we have @partyparrotgreg take a quick pass at this and match his suggestion.  I think that card-style links probably aren't the right presentation here.

PR #3091, opened 2025-09-29
Chore(dev hub) move pf pro

aditya520:
## How has this been tested?

- [ ] Added new tests


apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+  CurrencyDollar as MoneyWavyIcon,
+} from "@phosphor-icons/react/dist/ssr";
+
+import { ProductCard } from "../../ProductCard";
+
+export function PriceFeedsLandingPage() {
+  return (
+    <div className="max-w-7xl mx-auto px-4 sm:px-6 lg:px-8 py-8">
+      <div className="mb-12">
+        <h1 className="text-4xl font-bold text-gray-900 dark:text-white mb-6">
```
alexcambose: We're using css modules instead of tailwind so we should keep that

apps/developer-hub/src/components/ProPriceFeedTable/fetcher.ts
```diff
@@ -0,0 +1,29 @@
+import { z } from "zod";
+
+import { PRICE_FEEDS_PRO_API_URL } from "./constants";
+
+export const PriceFeedsProPriceIdSchema = z.object({
```
alexcambose: nit we don't uppercase the first character of the schema

```suggestion
export const priceFeedsProPriceIdSchema = z.object({
```

apps/developer-hub/src/components/ProPriceFeedTable/fetcher.ts
```diff
@@ -0,0 +1,29 @@
...
+  description: string;
+  exponent: number;
+  name: string;
+  pyth_lazer_id: number;
+  symbol: string;
+};
+
+export async function fetchPriceFeedsProPriceIdMetadata(): Promise<
+  PriceFeedsProPriceIdMetadata[]
+> {
```
alexcambose: instead of redefining the type (same as schema) you can do `z.infer<typeof priceFeedsProPriceIdSchema>`

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+    { id: "proId", name: "Pyth Pro Id" },
+    { id: "exponent", name: "Exponent" },
+  ];
+
+  const rows: RowConfig<Col>[] = filteredFeeds.map((priceFeed) => ({
+    id: `${String(priceFeed.pyth_lazer_id)}-${priceFeed.symbol}`,
+    data: {
+      assetType: priceFeed.asset_type,
+      description: (
+        <span className={styles.description ?? ""}>
```
alexcambose:
```suggestion
        <span className={styles.description}>
```

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+    return priceFeeds.filter((feed) => {
+      const searchLower = search.toLowerCase();
+      return (
+        feed.symbol.toLowerCase().includes(searchLower) ||
+        feed.name.toLowerCase().includes(searchLower) ||
+        feed.description.toLowerCase().includes(searchLower) ||
+        feed.pyth_lazer_id.toString().includes(searchLower)
+      );
+    });
+  }, [priceFeeds, search]);
```
alexcambose, line 105: We have this searching lib that's quite nice https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/Root/search-button.tsx#L146-L153 would be great to use this

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+        value={searchRaw}
+        onChange={updateSearch}
+        className="w-full p-2 mb-4 border border-gray-300 rounded-md"
+      />
+      <Table<Col>
+        className={styles.table ?? ""}
+        label="Pyth Pro price feed IDs"
+        columns={columns}
+        rows={rows}
+        isLoading={false}
```
alexcambose: I feel like the ux would be better if we use the loading state on the table, instead of displaying a spinner and then switching to a table

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+      return (
+        feed.symbol.toLowerCase().includes(searchLower) ||
+        feed.name.toLowerCase().includes(searchLower) ||
+        feed.description.toLowerCase().includes(searchLower) ||
+        feed.pyth_lazer_id.toString().includes(searchLower)
+      );
+    });
+  }, [priceFeeds, search]);
+
+  const columns: ColumnConfig<Col>[] = [
```
alexcambose: nit this is a static var so you can move it outside of the component

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <Input
+        aria-label="Search price feeds"
+        role="searchbox"
+        type="text"
+        placeholder="Search by symbol, name, description, or pyth pro id..."
+        value={searchRaw}
+        onChange={updateSearch}
+        className="w-full p-2 mb-4 border border-gray-300 rounded-md"
+      />
+      <Table<Col>
```
alexcambose, line 132: the page seems to be lagging when all the feeds are displayed, I think it would be nice to have some sort of pagination. I reccomend using a similar approach to here: 
this manages filtering, sorting and pagination https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/PriceFeeds/price-feeds-card.tsx#L84 (https://insights.pyth.network/price-feeds)
this is the bottom pagination component: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/PriceFeeds/price-feeds-card.tsx#L285-L292

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  }, []);
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setSearch(searchRaw.trim());
+    }, 300);
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [searchRaw]);
```
alexcambose, line 93: this is good, but I think the main heavy thing we do here is about UI rendering the whole list, sorting through 2000-3000 js objects is not that heavy.
If we have pagination this is not needed anymore

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.tsx
aditya520: Yes, I always forget that.

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
aditya520: You mean showing an empty table in the loading state?

alexcambose: yes

apps/developer-hub/src/components/ProductCard/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  description,
+  features,
+  ctaText,
+  href,
+  className = "",
+  quickActions = [],
+}: ProductCardProps) {
+  return (
+    <div
+      className={`bg-white dark:bg-gray-800 rounded-lg border border-gray-200 dark:border-gray-700 p-6 shadow-sm hover:shadow-md transition-shadow flex flex-col h-full ${className}`}
```
alexcambose, line 38: this is also tailwind, also can we please use clsx for appending classNames

apps/developer-hub/content/docs/price-feeds/pro/index.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<InfoBox variant="info">Pyth Pro was previously known as Pyth Lazer.</InfoBox>
+
+Pyth Pro delivers customizable, enterprise-grade price data directly from first-party publishers.
+Subscribers can configure their price feeds, update schedules, and usage rights for display or redistribution.
+The service is delivered through standard APIs for seamless integration.
+
+---
+
+<div className="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-3">
```
cprussin, line 20: I think we should avoid tailwind and use sass for anything we write -- tailwind is used internally by fumadocs but all of this would break if we were to e.g. move to a different markdown renderer.

We should probably see if we can configure the fumadocs built-in tailwind to not look at the `content` folder...

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.module.scss
```diff
@@ -0,0 +1,120 @@
...
+    padding-right: theme.spacing(6);
+  }
+}
+
+@media (width >= 1024px) {
+  .container {
+    padding-left: theme.spacing(8); /* lg:px-8 */
+    padding-right: theme.spacing(8);
+  }
+}
```
cprussin, line 22: You should be nesting these inside `.container`, and you should use `theme.breakpoint` for these, e.g. this:

```scss
.container {
  max-width: ...;

  @include theme.breakpoint("sm") {
    ...
  }
}
```

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.module.scss
```diff
@@ -0,0 +1,120 @@
+@use "@pythnetwork/component-library/theme";
+
+.container {
+  max-width: 80rem; /* ~ max-w-7xl */
```
cprussin, line 4: Please stick to `theme.spacing` for size units

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.module.scss
```diff
@@ -0,0 +1,120 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  padding: theme.spacing(8) theme.spacing(4) theme.spacing(8) theme.spacing(4); /* px-4 */ /* py-8 */
```
cprussin, line 7: This is more concise as `padding: theme.spacing(8) theme.spacing(4);`

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.module.scss
```diff
@@ -0,0 +1,120 @@
...
+    padding-right: theme.spacing(8);
+  }
+}
+
+.section {
+  margin-bottom: theme.spacing(12); /* mb-12 */
+}
+
+.title {
+  font-size: 2.25rem; /* text-4xl */
```
cprussin, line 29: Please use `@include theme.text()` for font sizes & weights etc

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.module.scss
```diff
@@ -0,0 +1,120 @@
...
+}
+
+.link {
+  color: theme.color("accent");
+  font-weight: 500; /* font-medium */
+  text-decoration: none;
+}
+
+.link:hover {
+  color: var(--color-blue-800);
```
cprussin, line 109: Don't color vars --  these shouldn't even exist and I think they're defined internally in fumadocs or something -- stick to colors from `theme.color`

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+        <p className={styles.body}>
+          Pyth offers two main versions of price feeds, each optimized for
+          different use cases:
+        </p>
+
+        <div className={styles.gridTwo}>
+          <ProductCard
+            badge="Pro"
+            badgeColor="bg-blue-600"
+            icon={<LightningIcon size={24} />}
```
cprussin, line 40: Minor nit, but generally I don't like passing sizing to icons from html, I instead prefer setting `font-size` in css -- this is after all part of styling, not content.

apps/developer-hub/src/components/Pages/PriceFeedsLandingPage/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+        <h2 className={styles.h2}>Product Options</h2>
+        <p className={styles.body}>
+          Pyth offers two main versions of price feeds, each optimized for
+          different use cases:
+        </p>
+
+        <div className={styles.gridTwo}>
+          <ProductCard
+            badge="Pro"
+            badgeColor="bg-blue-600"
```
cprussin, line 39: We should not be passing around tailwind class names like this -- as I'd mentioned above, tailwind is internal to fumadocs and we shouldn't use it in stuff we write.

lazer/contracts/solana/Cargo.lock
```diff
@@ -3293,7 +3293,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-lazer-protocol"
-version = "0.14.0"
```
cprussin, line 3296: There should not be changes to this file in this PR, is this intentional?

apps/developer-hub/src/components/ProPriceFeedTable/index.tsx
cprussin: This is also a very strange way to do debouncing / throttling and it's a bit of an abuse of effects, try using [react-hookz](https://react-hookz.github.io/web/?path=/docs/state-usethrottledstate--example) instead, it's quite nice

cprussin (changes requested):
We should not be using tailwind for manually created things, also please review the css to make sure you're using the theme correctly and not falling back to manual configs.

PR #3093, opened 2025-09-29
fix(lazer): use config values in history client

Riateche:
## Summary

Remove constants and use values from config

## How has this been tested?

- [ ] Added new tests


darunrs (dismissed):

darunrs (commented):
Are the other config attr like cache dir, urls, and capacity also read? I only see changes for two of the attributes.

darunrs (approved):
Ok sorry. Read through the current crosschain code. LGTM

PR #3100, opened 2025-10-01
feat(hip-3-pusher): AWS client config, backup write support, minor fixes

merolish:
## Summary
- Use AWS env vars for kms client
- Support hypothetical backup write urls if they ever become available
- Minor fixes to metrics and file reads

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/hip-3-pusher/src/pusher/kms_signer.py
+++ b/apps/hip-3-pusher/src/pusher/kms_signer.py
@@ -79,7 +79,7 @@ class KMSSigner:
         signature = self.sign_l1_action(
             action=action,
             nonce=timestamp,
-            is_mainnet= self.use_testnet,
+            is_mainnet=not self.use_testnet,
         )
         return self._send_update(action, signature, timestamp)
 
--- a/apps/hip-3-pusher/src/pusher/publisher.py
+++ b/apps/hip-3-pusher/src/pusher/publisher.py
@@ -25,6 +25,7 @@ class Publisher:
         self.publish_interval = float(config.hyperliquid.publish_interval)
         self.use_testnet = config.hyperliquid.use_testnet
         self.push_urls = config.hyperliquid.push_urls
+        logger.info("push urls: {}", self.push_urls)
 
         self.kms_signer = None
         self.enable_kms = False


apps/hip-3-pusher/src/pusher/config.py
```diff
@@ -30,6 +28,7 @@ class HermesConfig(BaseModel):
 
 class HyperliquidConfig(BaseModel):
     hyperliquid_ws_urls: list[str]
+    backup_push_urls: Optional[list[str]] = Field(default_factory=list)
```
tejasbadadare: Might be cleaner to just specify `push_urls` as a list, with default `["https://api.hyperliquid.xyz"]` (or the testnet equivalent). The default behavior should be to broadcast to all URLs.

apps/hip-3-pusher/src/pusher/kms_signer.py
```diff
@@ -6,51 +7,56 @@
...
-        url = TESTNET_API_URL if use_testnet else MAINNET_API_URL
-        self.oracle_publisher_exchange: Exchange = Exchange(wallet=None, base_url=url)
-        self.client = self._init_client(config)
+    def __init__(self, config: Config, publisher_exchanges: list[Exchange]):
+        self.use_testnet = config.hyperliquid.use_testnet
+        self.publisher_exchanges = publisher_exchanges
 
+        # AWS client and public key load
+        self.client = _init_client()
+        self._load_public_key(config.kms.key_path)
```
tejasbadadare: If we fail to load the public key at init time, lets throw a meaningful error message so its easy to spot the bad config.

apps/hip-3-pusher/src/pusher/kms_signer.py
```diff
@@ -67,15 +73,24 @@ def set_oracle(self, dex, oracle_pxs, all_mark_pxs, external_perp_pxs):
...
+    def _send_update(self, action, signature, timestamp):
+        for exchange in self.publisher_exchanges:
+            try:
+                return exchange._post_action(
+                    action=action,
+                    signature=signature,
+                    nonce=timestamp,
+                )
+            except Exception as e:
+                logger.exception("perp_deploy_set_oracle exception for endpoint: {} error: {}", exchange.base_url, e)
```
tejasbadadare: i'd use `repr(e)` instead of just `e` as that will also print the error type which is sometimes very useful.

apps/hip-3-pusher/src/pusher/publisher.py
```diff
@@ -79,18 +81,38 @@ def publish(self):
...
+                return exchange.perp_deploy_set_oracle(
+                    dex=self.market_name,
+                    oracle_pxs=oracle_pxs,
+                    all_mark_pxs=all_mark_pxs,
+                    external_perp_pxs=external_perp_pxs,
+                )
+            except Exception as e:
+                logger.exception("perp_deploy_set_oracle exception for endpoint: {} error: {}", exchange.base_url, e)
+
+        return None
```
tejasbadadare: We've seen in Fortuna that sometimes RPCs will take very long to return (~30s) when the network is congested. It would be to specify a timeout and move on to the next RPC in the list if one is doing that.

apps/hip-3-pusher/src/pusher/publisher.py
```diff
@@ -79,18 +81,38 @@ def publish(self):
...
+
+    def _handle_response(self, response):
+        if response is None:
+            logger.error("Push API call failed")
+            self.metrics.failed_push_counter.add(1, self.metrics_labels)
+            return
```
tejasbadadare: It's better to not conflate None with an error state. With the the current control flow, you can't tell whether the response is somehow actually None (could be possible since its a third party function we're calling), or it's None because of some exception (and the exc info is lost at this point), or because of some other control flow (like self.enable_publish is False.)

apps/hip-3-pusher/src/pusher/publisher.py
```diff
@@ -79,18 +81,38 @@ def publish(self):
...
+
+        logger.debug("publish: push response: {} {}", response, type(response))
```
tejasbadadare, line 122: Is it possible to track downtime? I.e. how many seconds over the 3s SLA we took to land an update. This might not be the right place, but it would be great to know that number

apps/hip-3-pusher/src/pusher/publisher.py
```diff
@@ -20,28 +21,28 @@ class Publisher:
...
-            oracle_account = None
-            self.kms_signer = KMSSigner(config)
-        else:
-            oracle_pusher_key_path = config.hyperliquid.oracle_pusher_key_path
-            oracle_pusher_key = open(oracle_pusher_key_path, "r").read().strip()
+        self.kms_signer = None
+        self.enable_kms = False
+        oracle_account = None
+        if not config.kms.enable_kms:
+            oracle_pusher_key = Path(config.hyperliquid.oracle_pusher_key_path).read_text().strip()
```
tejasbadadare, line 34: You can guarantee that these paths exist in the pydantic model by using [FilePath](https://docs.pydantic.dev/2.2/usage/types/file_types/) field type

apps/hip-3-pusher/src/pusher/kms_signer.py
```diff
@@ -6,51 +7,56 @@
...
+        self.client = _init_client()
+        self._load_public_key(config.kms.key_path)
+
+    def _load_public_key(self, key_path: str):
         # Fetch public key once so we can derive address and check recovery id
-        key_path = config.kms.key_path
-        self.key_id = open(key_path, "r").read().strip()
-        self.pubkey_der = self.client.get_public_key(KeyId=self.key_id)["PublicKey"]
+        self.key_id = Path(key_path).read_text().strip()
+        pubkey_der = self.client.get_public_key(KeyId=self.key_id)["PublicKey"]
```
tejasbadadare: idea for unit tests: it would be fantastic to use [moto](https://www.learnaws.org/2021/02/21/test-aws-kms-boto/) to emulate KMS, and write a test that validates that your method of signing transactions works correctly. It would be valuable imo since the signing logic is partially ripped from the HL SDK (nice work on that btw 🙂) 

Moto is great because it's a drop-in mock that's super easy to set up and can emulate pretty much any AWS service.
Example: https://www.learnaws.org/2021/02/21/test-aws-kms-boto/

apps/hip-3-pusher/config/config.toml
```diff
@@ -13,9 +13,6 @@ enable_publish = false
 [kms]
 enable_kms = false
 key_path = "/path/to/aws_kms_key_id.txt"
```
tejasbadadare: nit: lets name this config something quite specific like "aws_kms_key_id_path" since there are a lot of key and key-adjacent names floating around

merolish: Done, good call.

apps/hip-3-pusher/src/pusher/config.py
merolish: Nice, tweaked this with a model validator.

apps/hip-3-pusher/src/pusher/publisher.py
merolish: Lucky for us their python sdk exposes a timeout param, so we don't have to do anything more clever.

apps/hip-3-pusher/src/pusher/publisher.py
merolish: Syncing across instances would be more of a list, but certainly can do within an instance.

apps/hip-3-pusher/src/pusher/publisher.py
merolish: Ah yes, made this more pythonic with exceptions.

apps/hip-3-pusher/src/pusher/publisher.py
merolish: Added a metric in any case.

apps/hip-3-pusher/src/pusher/kms_signer.py
merolish: So, it appears that moto doesn't currently support external key import the way localstack does, in which case I can't do a full regression test against the local private key/sdk case. You can generate a key but never have access to it.

apps/hip-3-pusher/src/pusher/kms_signer.py
```diff
@@ -6,51 +6,61 @@
...
+        self.use_testnet = config.hyperliquid.use_testnet
+        self.publisher_exchanges = publisher_exchanges
+
+        # AWS client and public key load
+        self.client = _init_client()
+        try:
+            self._load_public_key(config.kms.aws_kms_key_id_path)
+        except Exception as e:
+            logger.exception("Failed to load public key from KMS; it might be incorrectly configured; error: {}", repr(e))
+            exit()
```
tejasbadadare, line 39: Invoking `exit()` isn't ideal (often times this doesn't work the way you expect, esp in threaded/async workloads.) Lets just rethrow a chained exception. It should crash the program as expected. Something like 
```python
except Exception as e:
    logger.exception(...)
    throw Exception("Failed to load public key from KMS") from e
```

PR #3106, opened 2025-10-03
fix(entropy): Only mark callbacks as failed if enough gas was provided, regardless of revert code

jayantk:
## Summary

This PR removes the check that allows us to mark callbacks as failed (even if enough gas wasn't provided) if they revert with a non-empty return code.

## Rationale

When we wrote this initally, we thought that all out-of-gas reverts from calling contracts would bubble up and be caught by the `ret.length == 0` condition. However, we have found that some contracts catch out-of-gas reverts and then revert with a different error code. This means we can end up calling these callbacks without enough gas. 

Simple fix here is to require enough gas to be provided regardless of the revert code.

## How has this been tested?

- [x] Added new tests

### DIFF
--- a/target_chains/ethereum/contracts/test/Entropy.t.sol
+++ b/target_chains/ethereum/contracts/test/Entropy.t.sol
@@ -1113,6 +1113,16 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents, EntropyEventsV2 {
             userRandomNumber
         );
 
+        // If the callback reverts, the Entropy reveal also reverts unless
+        // provided enough gas to pass on.
+        vm.expectRevert();
+        random.revealWithCallback{gas: defaultGasLimit - 1000}(
+            provider1,
+            assignedSequenceNumber,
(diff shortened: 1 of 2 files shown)


target_chains/ethereum/contracts/test/Entropy.t.sol
```diff
@@ -1113,6 +1113,16 @@ contract EntropyTest is Test, EntropyTestUtils, EntropyEvents, EntropyEventsV2 {
...
 
+        // If the callback reverts, the Entropy reveal also reverts unless
+        // provided enough gas to pass on.
+        vm.expectRevert();
+        random.revealWithCallback{gas: defaultGasLimit - 1000}(
+            provider1,
+            assignedSequenceNumber,
+            userRandomNumber,
+            provider1Proofs[assignedSequenceNumber]
+        );
```
jayantk, line 1124: this test used to fail before the change

tejasbadadare (approved):

PR #3111, opened 2025-10-07
feat(lazer-protocol): add Metadata V3 response types

tejasbadadare:
## Summary
Add types for output representations of the Metadata v3 API.
- `FeedResponseV3`: Returned by the `/v3/feeds` APIs
- `AssetResponseV3`: Returned by the `/v3/assets` APIs
- Existing `/v1/symbols` API will continue to return the `history_service::api::SymbolResponse` from Lazer internals for the time being. In the future, this type should be unified with the duplicated one in the protocol crate (`pyth_lazer_protocol::jrpc::SymbolMetadata`, used by History client). 

## How has this been tested?

- [x] Added new tests
  - Added basic ser/de roundtrip tests, added validation tests for SymbolV3 format

### DIFF
--- a/lazer/sdk/rust/protocol/src/metadata.rs
+++ b/lazer/sdk/rust/protocol/src/metadata.rs
@@ -1,8 +1,7 @@
 //! Types describing Lazer's metadata APIs.
 
 use crate::time::{DurationUs, TimestampUs};
-use crate::FeedKind;
-use crate::{symbol_state::SymbolState, PriceFeedId};
+use crate::PriceFeedId;
 use serde::{Deserialize, Serialize};
 
 /// The pricing context or type of instrument for a feed.
@@ -91,7 +90,7 @@ pub struct FeedResponseV3 {
     pub min_publishers: u16,
     /// Status of the feed.
     /// Example: `"active"`
-    pub state: SymbolState,
+    pub state: String,
     /// High-level asset class. One of crypto, fx, equity, metal, rates, nav, commodity, funding-rate.
     /// Should be one of the values in the AssetClass enum.
     /// Example: `"crypto"`
@@ -113,7 +112,7 @@ pub struct FeedResponseV3 {
     pub feed_expiry: Option<TimestampUs>,
     /// The nature of the data produced by the feed.
     /// Examples: `"price"`, `"fundingRate"`
-    pub feed_kind: FeedKind,
+    pub feed_kind: String,
 }
 
 /// Asset metadata as returned by the v3 metadata API.
@@ -151,7 +150,7 @@ mod tests {
 
     #[test]
     fn test_feed_response_v3_json_serde_roundtrip() {
-        use crate::{symbol_state::SymbolState, FeedKind, PriceFeedId};
+        use crate::PriceFeedId;
 
         let symbol = SymbolV3::new(
             "pyth".to_string(),
@@ -173,14 +172,14 @@ mod tests {
             exponent: -8,
             update_interval: Some(DurationUs::from_secs_u32(10)),
             min_publishers: 3,
-            state: SymbolState::Stable,
+            state: "stable".to_string(),
             asset_type: "crypto".to_string(),
             cmc_id: Some(1),
             pythnet_id: "e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"
                 .to_string(),
             nasdaq_symbol: None,
             feed_expiry: None,
-            feed_kind: FeedKind::Price,
+            feed_kind: "price".to_string(),
         };
 
         // Test JSON serialization


keyvankhademi (approved):

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Metal
+    Metal,
+    /// Rates
+    Rates,
+    /// Net Asset Value
+    Nav,
+    /// Commodity
+    Commodity,
+    /// Funding rate
+    FundingRate,
```
ali-behjati: is this really an asset class?

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Net Asset Value
+    Nav,
+    /// Time-weighted average price
+    Twap,
+}
+
+/// High-level asset class.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum AssetClass {
```
ali-behjati, line 34: One of the reasons for opting in a dynamic metadata was to not go through a code change when new asset classes are added to the system. I know that users (on both sides) rely on these, and that's probably why you opted for explicit definition here. But if that's the case, you might actually remove any dynamic field and make everything very explicit. Being in the middle (some explicit metadata, some implicit dynamic) is probably worse.

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Foreign exchange
+    Fx,
+    /// Equity
+    Equity,
+    /// Metal
+    Metal,
+    /// Rates
+    Rates,
+    /// Net Asset Value
+    Nav,
```
ali-behjati: This also doesn't seem to be an asset class, it's a feed/instrument type.

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Unique integer identifier for a feed. Known as `pyth_lazer_id` in V1 API.
+    /// Example: `1`
+    pub id: PriceFeedId,
+    /// Short feed name.
+    /// Example: `"Bitcoin / US Dollar"`
+    pub name: String,
+    /// Unique human-readable identifier for a feed.
+    /// Format: `source.instrument_type.base/quote`
+    /// Examples: `"pyth.spot.btc/usd"`, `"pyth.redemptionrate.alp/usd"`, `"binance.fundingrate.btc/usdt"`, `"pyth.future.emz5/usd"`
+    pub symbol: String,
```
ali-behjati, line 61: i recommend opting for a specific type with validation here to ensure it's correct.

q: what if the asset has no quote asset? (such as rates)

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Status of the feed.
+    /// Example: `"active"`
+    pub state: SymbolState,
+    /// High-level asset class. One of crypto, fx, equity, metal, rates, nav, commodity, funding-rate.
+    /// Example: `"crypto"`
+    pub asset_type: AssetClass,
+    /// CoinMarketCap asset identifier.
+    /// Example: `"123"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cmc_id: Option<String>,
```
ali-behjati: i wonder whether you considered @jayantk suggestion on having a nested struct as "external mappings". It's probably more difficult on the DB layer.

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Example: `"ADSK"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub nasdaq_symbol: Option<String>,
+    /// ISO datetime after which the feed will no longer produce prices because the underlying market has expired.
+    /// Example: `"2025-10-03T11:08:10.089998603Z"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub feed_expiry: Option<String>,
+    /// The nature of the data produced by the feed.
+    /// Examples: `"price"`, `"fundingRate"`
+    pub feed_kind: FeedKind,
```
ali-behjati: Hmmm so we do have a feedkind and an instrument type. I wonder whether we can merge them.

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Example: `"e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43"`
+    pub pythnet_id: String,
+    /// Nasdaq symbol identifier.
+    /// Example: `"ADSK"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub nasdaq_symbol: Option<String>,
+    /// ISO datetime after which the feed will no longer produce prices because the underlying market has expired.
+    /// Example: `"2025-10-03T11:08:10.089998603Z"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub feed_expiry: Option<String>,
```
ali-behjati: Use a datetime type or an integer based timestamp (i bet we should have something similar in our code)

lazer/sdk/rust/protocol/src/metadata.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Example: `"Bitcoin"`
+    pub full_name: String,
+    /// High-level asset class.
+    /// Example: `"crypto"`
+    pub class: AssetClass,
+    /// More granular categorization within class.
+    /// Example: `"stablecoin"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub subclass: Option<String>,
+    /// Primary or canonical listing exchange, when applicable.
```
ali-behjati, line 137: Hmmm, what if there are multiple?
i suggest renaming the field to `primary_exchange` or something similar here. not sure `listing_exchange` is clear.

lazer/sdk/rust/protocol/src/metadata.rs
Riateche: I think we should stick with our decision and use fully dynamic metadata in the protocols. In Rust that would be `BTreeMap<String, serde_value::Value>`. We can revisit it later if we feel like the metadata structure is very stable and future-proof, but I doubt that it will happen soon.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Hmmm.. we currently consider it an asset class in the V1 API, but i agree that conceptually it doesn't really fit. We can introduce an `AssetClassV3` that excludes these. 

But then the question is what we'd set for `asset_class` for funding rate feeds. Maybe it's reasonable to leave it as FundingRate? Or we can set it to the underlying asset's class. Will check with @YaserJazouane on this.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Yeah i did consider it, but i figured that a nested struct in the dynamic metadata opened the door to making its own table to ensure proper types in the DB. FIgured it was simpler to leave it flat.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: I think they're capturing slightly different things.
Afaik `feed_kind` is meant to be the coarsest discriminator of feeds, which we use to trigger different aggregation logic (price feeds, non price feeds like funding rates). Whereas `instrument_type` is a bit more specific, and describes the pricing context (spot price, averaged price, etc.).

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Yeah my main goal in adding these explicit types is to ensure end users can depend on a stable API contract across different versions. I.e. the different types like the existing `SymbolResponse` and the new `FeedResponseV3` can handle the differences between v1 and v3 representations derived from the same internal dynamic metadata representation.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Yeah same thing as FundingRate above.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: A type with validation to enforce the string format is a good idea. 
Not sure about a feed with no quote asset. If such a feed is valid it poses a deeper problem with the data model, since it revolves around feed being an asset/quote pair. Fwiw currently funding rates are USD denominated.

tejasbadadare: Okay caught up with @YaserJazouane on this and yep we should make the quote asset optional. There are cases like yields, indices, fundingrates that aren't quoted in any currency. Our existing funding rate example is denominated in USD which is incorrect.

tldr:
- We're amending the format to be `source.instrument_type.base[/quote]` where the quote asset is optional
- We're making `quote_asset_id` optional

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Consensus on this topic was that the two fields capture different things as mentioned above, so planning on keeping it this way.

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Moved NAV to instrument_type ✅

lazer/sdk/rust/protocol/src/metadata.rs
tejasbadadare: Caught up, yes funding rate is not an asset class. It'll be a feed_kind (as it already is).

lazer/sdk/rust/protocol/src/metadata.rs
darunrs: We've managed to eliminate most Enums from the payload. Considering FeedKind is used for aggregation, could we also make this a String in this payload? Just to not complicate adding new aggregation types. It's likely the case that a new feed kind requires dev effort to implement first, so we can safely convert the existing Enum to String for these responses, with an expectation that Governance would do the same when initially adding to state.

tejasbadadare: Agreed, forgot about this one! will update

tejasbadadare: We are going to add `beta` soon to SymbolState as well -- I'll also make that a string.

PR #3113, opened 2025-10-07
chore(contract_manager): upgrade contract script fix 

nidhi-singh02:
## Summary

The upgrade contract script was failing as the build artifacts are generated via foundry(forge) now, so the bytecode is not a string, it is instead an object that has "object" property.

Fixed it for entropy, price feeds and one more place where it is used.

## How has this been tested?

- [ ] Added new tests


jayantk (approved):
thank you 🙇

PR #3117, opened 2025-10-09
fix(insights): when feeds aren't trading, show last valid price

cprussin:
## Summary

When feeds are not currently trading, we would previously still show the price point published to pythnet.  However this is misleading in IH since it looks like the feed is still live and updating.  With this PR, we now only show ticking prices when feeds are trading -- when they aren't we instead fall back to `previousPrice` / `previousTimestamp`.

We also enter whitespace data into the chart for points where the feed is not trading -- to be honest there are probably more intuitive ways to account for this in the chart (e.g. maybe we should just skip these points and default the chart to show a window including the last published point or something), but this is a reasonable stopgap that correctly displays what we saw in pythnet, while removing confusion around stale data looking like it's still being published.  cc @fhqvst , this is something to think of in future iterations of the chart.

## Rationale

Resolves some confusing behavior of IH

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/PriceFeed/Chart/chart.tsx
+++ b/apps/insights/src/components/PriceFeed/Chart/chart.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { PriceStatus } from "@pythnetwork/client";
 import { useLogger } from "@pythnetwork/component-library/useLogger";
 import { useResizeObserver, useMountEffect } from "@react-hookz/web";
 import {
@@ -12,7 +13,9 @@ import type {
   IChartApi,
   ISeriesApi,
   LineData,
+  Time,
   UTCTimestamp,
+  WhitespaceData,
 } from "lightweight-charts";
 import {
   AreaSeries,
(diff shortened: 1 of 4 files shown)


apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -451,18 +496,27 @@ const getColors = (container: HTMLDivElement, resolvedTheme: string) => {
...
+) {
+  const unique = new Map<Time, LineData | WhitespaceData>();
 
   for (const a of as) {
-    unique.set(a.time as number, a);
+    unique.set(a.time, a);
   }
   for (const b of bs) {
-    unique.set(b.time as number, b);
+    unique.set(b.time, b);
```
fhqvst, line 509: Wonder if we need to add some check here so that if the timestamp for a `LineData` and a `WhitespaceData` are equal, the `LineData` wins.

I.e., if we're using 1 day as resolution, and the status goes down and up within that same day, we'll be guaranteed to see the price instead of whitespace.

Perhaps this could/should be done in the Clickhouse query even? Not sure...

PR #3129, opened 2025-10-14
Chore(dev hub) entropy improvements

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/app/(docs)/[section]/layout.tsx
+++ b/apps/developer-hub/src/app/(docs)/[section]/layout.tsx
@@ -5,15 +5,14 @@ import type { ReactNode } from "react";
 
 import { docsOptions } from "../../../config/layout.config";
 
-export default function Layout({
+export default async function Layout({
   children,
   params,
 }: {
   children: ReactNode;
-  params: { section: string };
+  params: Promise<{ section: string }>;
 }) {
-  const { section } = params;
-  const options = { ...docsOptions };
+  const { section } = await params;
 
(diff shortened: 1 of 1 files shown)


apps/developer-hub/src/app/(docs)/[section]/layout.tsx
```diff
@@ -1,9 +1,43 @@
...
-export default function Layout({ children }: { children: ReactNode }) {
+export default function Layout({
+  children,
+  params,
+}: {
+  children: ReactNode;
+  params: { section: string };
+}) {
+  const { section } = params;
   const options = { ...docsOptions };
```
Comment (alexcambose) [inline apps/developer-hub/src/app/(docs)/[section]/layout.tsx]: actually I don't think we need this, we should be able to pass `<DocsLayout {...docsOptions}>{children}</DocsLayout>` directly

apps/developer-hub/src/app/(docs)/[section]/layout.tsx
```diff
@@ -1,9 +1,43 @@
...
 
 import { docsOptions } from "../../../config/layout.config";
 
-export default function Layout({ children }: { children: ReactNode }) {
+export default function Layout({
+  children,
+  params,
+}: {
+  children: ReactNode;
+  params: { section: string };
```
Comment (alexcambose) [inline apps/developer-hub/src/app/(docs)/[section]/layout.tsx]: NextJS needs a promise here so if you do `params: Promise<{ section: string }>;` and await below, it should also fix the build issues

alexcambose (approved):

PR #3130, opened 2025-10-14
chore(upgrade) Upgrade Viem & Add Chain Overrides

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/contracts/solana/Cargo.lock
+++ b/lazer/contracts/solana/Cargo.lock
@@ -3293,7 +3293,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-lazer-protocol"
-version = "0.18.1"
+version = "0.14.0"
 dependencies = [
  "anyhow",
  "byteorder",
@@ -3312,7 +3312,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-lazer-solana-contract"
-version = "0.7.1"
+version = "0.6.0"
 dependencies = [
  "anchor-lang",
  "bytemuck",


apps/api-reference/src/components/EvmProvider/chain-overrides.ts
```diff
@@ -74,4 +74,42 @@ export const chainOverrides: readonly Chain[] = [
...
+    },
+    testnet: false,
+    blockExplorers: {
+      default: {
+        name: "Campscout",
+        url: "https://camp.cloud.blockscout.com/",
+      },
+    },
+  }),
+  defineChain({
```
benduran, line 96: for the uninitiated / new folks (hi, that's me 👋 ), what is `defineChain()` doing? a cursory glance looks like it's setting up some metadata to display chains in the docsite, is that correct? is it powering this dropdown when playing with the APIs?
CleanShot 2025-10-14 at 14 27 01

apps/api-reference/src/components/EvmProvider/chain-overrides.ts
```diff
@@ -0,0 +1,115 @@
...
+ *    - id: chainId. id in EvmChains.json is the name
+ *    - name: Chain display name
+ *    - nativeCurrency: \{ name, symbol, decimals \}
+ *    - rpcUrls: \{ default: \{ http: ['...'] \} \}
+ *    - testnet: true/false (optional, defaults to false for mainnet)
+ *    - blockExplorers: \{ default: \{ name, url \} \} (optional)
+ */
+export const chainOverrides: readonly Chain[] = [
+  // Example chain override (remove or replace with actual chains):
+  // defineChain({
```
benduran, line 20: nitpick: should these `defineChain()` usages that are commented out be removed, for brevity?

lazer/contracts/solana/Cargo.lock
```diff
@@ -3293,7 +3293,7 @@ dependencies = [
 
 [[package]]
 name = "pyth-lazer-protocol"
-version = "0.14.0"
+version = "0.18.1"
```
cprussin: This should not be touched with your PR, you should revert this file

apps/api-reference/src/components/EvmProvider/chain-overrides.ts
aditya520: Hey @benduran , you are absolutely right. It is powering the chain list you showed above in  the dropdown. 
Initially we used to fetch the chainIds (The numeric identifier of every chain) from [EvmChains.json](https://github.com/pyth-network/pyth-crosschain/blob/main/contract_manager/store/chains/EvmChains.json) and then find those chain in viem.sh [chainlist](https://github.com/wevm/viem/tree/main/src/chains/definitions) for extra data needed. 

We figured out that many chains are not present in the viem chainlist, so this `chain-override` file follows the same type: [defineChain](https://github.com/wevm/viem/tree/main/src/chains/definitions). The code will check the override file and then find the remaining chains from viem chains.

apps/api-reference/src/components/EvmProvider/chain-overrides.ts
aditya520: This is just an example. Alot of times non-tech BD folks update some chains, so I kept it as an example.

lazer/contracts/solana/Cargo.lock
aditya520: My bad. My Cargo analyzer automatically updates it.

PR #3135, opened 2025-10-15
chore(lazer) Added Feed struct

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/contracts/evm/src/PythLazerLib.sol
+++ b/lazer/contracts/evm/src/PythLazerLib.sol
@@ -519,7 +519,7 @@ library PythLazerLib {
             );
     }
 
-    // Requested (applicability) helpers — property included in this update
+    // Requested helpers — property included in this update
     function isPriceRequested(
         PythLazerStructs.Feed memory feed
     ) public pure returns (bool) {
@@ -613,7 +613,11 @@ library PythLazerLib {
     function getPrice(
         PythLazerStructs.Feed memory feed
     ) public pure returns (int64) {
-        require(hasPrice(feed), "Price does not exist");
+        require(
+            isPriceRequested(feed),
+            "Price is not requested for the timestamp"
+        );
+        require(hasPrice(feed), "Price is not present for the timestamp");
         return feed._price;
     }
 
@@ -621,7 +625,14 @@ library PythLazerLib {
     function getBestBidPrice(
         PythLazerStructs.Feed memory feed
     ) public pure returns (int64) {
-        require(hasBestBidPrice(feed), "Best bid price does not exist");
+        require(
+            isBestBidPriceRequested(feed),
+            "Best bid price is not requested for the timestamp"
+        );
+        require(
+            hasBestBidPrice(feed),
+            "Best bid price is not present for the timestamp"
+        );
         return feed._bestBidPrice;
     }
 
@@ -629,7 +640,14 @@ library PythLazerLib {
     function getBestAskPrice(
         PythLazerStructs.Feed memory feed
     ) public pure returns (int64) {
-        require(hasBestAskPrice(feed), "Best ask price does not exist");
+        require(
+            isBestAskPriceRequested(feed),
+            "Best ask price is not requested for the timestamp"
+        );
+        require(
+            hasBestAskPrice(feed),
+            "Best ask price is not present for the timestamp"
+        );
         return feed._bestAskPrice;
     }
 
@@ -637,7 +655,14 @@ library PythLazerLib {
     function getPublisherCount(
         PythLazerStructs.Feed memory feed
     ) public pure returns (uint16) {
-        require(hasPublisherCount(feed), "Publisher count does not exist");
+        require(
+            isPublisherCountRequested(feed),
+            "Publisher count is not requested for the timestamp"
+        );
+        require(
+            hasPublisherCount(feed),
+            "Publisher count is not present for the timestamp"
+        );
         return feed._publisherCount;
     }
 
@@ -645,7 +670,11 @@ library PythLazerLib {
     function getExponent(
         PythLazerStructs.Feed memory feed
     ) public pure returns (int16) {
-        require(hasExponent(feed), "Exponent does not exist");
+        require(
+            isExponentRequested(feed),
+            "Exponent is not requested for the timestamp"
+        );
+        require(hasExponent(feed), "Exponent is not present for the timestamp");
         return feed._exponent;
     }
 
@@ -653,7 +682,14 @@ library PythLazerLib {
     function getConfidence(
         PythLazerStructs.Feed memory feed
     ) public pure returns (uint64) {
-        require(hasConfidence(feed), "Confidence does not exist");
+        require(
+            isConfidenceRequested(feed),
+            "Confidence is not requested for the timestamp"
+        );
+        require(
+            hasConfidence(feed),
+            "Confidence is not present for the timestamp"
+        );
         return feed._confidence;
     }
 
@@ -661,7 +697,14 @@ library PythLazerLib {
     function getFundingRate(
         PythLazerStructs.Feed memory feed
     ) public pure returns (int64) {
-        require(hasFundingRate(feed), "Funding rate does not exist");
+        require(
+            isFundingRateRequested(feed),
+            "Funding rate is not requested for the timestamp"
+        );
+        require(
+            hasFundingRate(feed),
+            "Funding rate is not present for the timestamp"
+        );
         return feed._fundingRate;
     }
 
@@ -669,7 +712,14 @@ library PythLazerLib {
     function getFundingTimestamp(
         PythLazerStructs.Feed memory feed
     ) public pure returns (uint64) {
-        require(hasFundingTimestamp(feed), "Funding timestamp does not exist");
+        require(
+            isFundingTimestampRequested(feed),
+            "Funding timestamp is not requested for the timestamp"
+        );
+        require(
+            hasFundingTimestamp(feed),
+            "Funding timestamp is not present for the timestamp"
+        );
         return feed._fundingTimestamp;
     }
 
@@ -678,8 +728,12 @@ library PythLazerLib {
         PythLazerStructs.Feed memory feed
     ) public pure returns (uint64) {
         require(
+            isFundingRateIntervalRequested(feed),
+            "Funding rate interval is not requested for the timestamp"
+        );
+        require(
             hasFundingRateInterval(feed),
-            "Funding rate interval does not exist"
+            "Funding rate interval is not present for the timestamp"
         );
         return feed._fundingRateInterval;
     }


lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +105,285 @@ library PythLazerLib {
...
+                    revert("Unknown property");
+                }
+            }
+
+            // Store feed in update
+            update.feeds[i] = feed;
+        }
+
+        // Ensure we consumed all bytes
+        require(pos == payload.length, "Payload not fully consumed");
```
tejasbadadare: Imo a more useful error would be "Payload has extra unknown bytes"

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +105,285 @@ library PythLazerLib {
...
+    }
+
+    // Helper functions for existence checks
+
+    /// @notice Check if price exists
+    function hasPrice(
+        PythLazerStructs.Feed memory feed
+    ) public pure returns (bool) {
+        return (feed.existsFlags & PythLazerStructs.PRICE_EXISTS) != 0;
+    }
```
tejasbadadare, line 435: Should we create getters for the properties that enforce checking existence? It would ensure users dont assume a property like `price` always exists. Something like `getPrice(feed) -> (bool, int64)`, it could return `(exists, value)`. It's an alternative to storing the type as `Option<Option<PropertyType>>`.

aditya520: There are chances that user can ignore the `exist` flag. Plus returning a bool will consume more gas.

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +107,286 @@ library PythLazerLib {
...
+                    property ==
+                    PythLazerStructs.PriceFeedProperty.FundingRateInterval
+                ) {
+                    (feed.fundingRateInterval, pos) = parseFeedValueUint64(
+                        payload,
+                        pos
+                    );
+                    if (feed.fundingRateInterval != 0)
+                        feed.existsFlags |= PythLazerStructs
+                            .FUNDING_INTERVAL_EXISTS;
```
Riateche: Funding rate, timestamp and interval are encoded differently from prices. Zero is a valid value for them, so they have an extra byte in front for presence (1 or 0). See serialization logic [here](https://github.com/pyth-network/pyth-crosschain/blob/main/lazer/sdk/rust/protocol/src/payload.rs).

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+    // Bitmap constants for Feed.existsFlags
+    uint32 constant PRICE_EXISTS = 0x01; // bit 0
+    uint32 constant BEST_BID_EXISTS = 0x02; // bit 1
+    uint32 constant BEST_ASK_EXISTS = 0x04; // bit 2
+    uint32 constant PUBLISHER_COUNT_EXISTS = 0x08; // bit 3
+    uint32 constant EXPONENT_EXISTS = 0x10; // bit 4
+    uint32 constant CONFIDENCE_EXISTS = 0x20; // bit 5
+    uint32 constant FUNDING_RATE_EXISTS = 0x40; // bit 6
+    uint32 constant FUNDING_TS_EXISTS = 0x80; // bit 7
+    uint32 constant FUNDING_INTERVAL_EXISTS = 0x100; // bit 8
```
Riateche:
```suggestion
    uint32 constant FUNDING_TIMESTAMP_EXISTS = 0x80; // bit 7
    uint32 constant FUNDING_RATE_INTERVAL_EXISTS = 0x100; // bit 8
```
For naming consistency

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    struct Update {
+        uint64 timestamp;
+        Channel channel;
+        Feed[] feeds;
+    }
+
+    // Bitmap constants for Feed.existsFlags
+    uint32 constant PRICE_EXISTS = 0x01; // bit 0
```
Riateche: Can you write it like this in Solidity? That would avoid the need for comments.
```
uint32 constant PRICE_EXISTS = 1 << 0;
uint32 constant BEST_BID_EXISTS = 1 << 1; 
uint32 constant BEST_ASK_EXISTS = 1 << 2;
```

lazer/contracts/evm/test/PythLazer.t.sol
```diff
@@ -71,4 +73,396 @@ contract PythLazerTest is Test {
...
+    /// @notice Test parsing single feed with all 9 properties
+    function test_parseUpdate_singleFeed_allProperties() public pure {
+        bytes[] memory properties = new bytes[](9);
+        properties[0] = buildProperty(0, encodeInt64(100000000)); // price
+        properties[1] = buildProperty(1, encodeInt64(99000000)); // bestBid
+        properties[2] = buildProperty(2, encodeInt64(101000000)); // bestAsk
+        properties[3] = buildProperty(3, encodeUint16(5)); // publisherCount
+        properties[4] = buildProperty(4, encodeInt16(-8)); // exponent
+        properties[5] = buildProperty(5, encodeInt64(50000)); // confidence
+        properties[6] = buildProperty(6, encodeInt64(123456)); // fundingRate
```
Riateche, line 183: This is not correct because there is a presence byte (as noted above).

It would be good to fetch real price and funding rate updates from Lazer staging and parse them in the tests.

lazer/contracts/evm/src/PythLazerLib.sol
aditya520: ohhh. I missed that. Thank you. Why didn't we do this for price?

lazer/contracts/evm/src/PythLazerLib.sol
tejasbadadare: As it stands they would have to first call `hasPrice` before using the price, which to me sounds like a greater chance of being missed by implementors since they would have to know about that method and intentionally use it. On the other hand, the exists flag more clearly signals the possibility that the field may not exist, and one would have to acknowledge this to ignore it. But that's just my opinion

aditya520: They can just call `getPrice()`. It will return an error if price is zero.

Riateche (approved):

lazer/contracts/evm/src/PythLazer.sol
```diff
@@ -105,7 +107,30 @@ contract PythLazer is OwnableUpgradeable, UUPSUpgradeable {
...
+    /// @return signer The address of the signer
+    /// @return parsedUpdate The parsed Update struct with all feeds and properties
+    function verifyAndParseUpdate(
+        bytes calldata update
+    )
+        external
+        payable
+        returns (
+            bytes calldata payload,
+            address signer,
```
ali-behjati: is there a need to return these?

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+        Confidence,
+        FundingRate,
+        FundingTimestamp,
+        FundingRateInterval
+    }
+
+    struct Feed {
+        // Slot 1: 4 + 4 + 8 + 8 + 8 = 32 bytes (fully packed)
+        uint32 feedId;
+        uint32 existsFlags; // Bitmap: bit 0-31 for up to 32 properties
```
ali-behjati: just want to raise that exceeding this might break our upgradability for Feed struct.

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+        int64 bestBidPrice;
+        int64 bestAskPrice;
+        // Slot 2: 8 + 8 + 8 + 8 = 32 bytes (fully packed)
+        int64 confidence;
+        int64 fundingRate;
+        uint64 fundingTimestamp;
+        uint64 fundingRateInterval;
+        // Slot 3: 2 + 2 = 4 bytes (28 bytes wasted, but unavoidable)
+        uint16 publisherCount;
+        int16 exponent;
```
ali-behjati: people can directly use these right? i recommend you rename it to `maybe<field>` to make it clear for people that this is an optional field (and they should use a getter)

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +107,301 @@ library PythLazerLib {
...
+                // Read property ID
+                PythLazerStructs.PriceFeedProperty property;
+                (property, pos) = parseFeedProperty(payload, pos);
+
+                // Parse value and set flag based on property type
+                // Price Property
+                if (property == PythLazerStructs.PriceFeedProperty.Price) {
+                    (feed.price, pos) = parseFeedValueInt64(payload, pos);
+                    if (feed.price != 0)
+                        feed.existsFlags |= PythLazerStructs.PRICE_EXISTS;
```
ali-behjati: Just to be clear, it seems you do not distinguish between lack of a property vs the property being None (as in Lazer doesn't know the price). Is that intentional?

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +107,301 @@ library PythLazerLib {
...
+        PythLazerStructs.Feed memory feed
+    ) public pure returns (bool) {
+        return
+            (feed.existsFlags &
+                PythLazerStructs.FUNDING_RATE_INTERVAL_EXISTS) != 0;
+    }
+
+    // Safe getter functions (revert if property doesn't exist)
+
+    /// @notice Get price (reverts if not exists)
```
ali-behjati, line 612: please update the comments on the failure modes. what does "not exists" mean?

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+        BestAskPrice,
+        PublisherCount,
+        Exponent,
+        Confidence,
+        FundingRate,
+        FundingTimestamp,
+        FundingRateInterval
+    }
+
+    struct Feed {
```
ali-behjati, line 35: add comments either here on what it is and how it should be used.

lazer/contracts/evm/src/PythLazerStructs.sol
```diff
@@ -0,0 +1,57 @@
...
+        // Slot 1: 4 + 4 + 8 + 8 + 8 = 32 bytes (fully packed)
+        uint32 feedId;
+        uint32 existsFlags; // Bitmap: bit 0-31 for up to 32 properties
+        int64 price;
+        int64 bestBidPrice;
+        int64 bestAskPrice;
+        // Slot 2: 8 + 8 + 8 + 8 = 32 bytes (fully packed)
+        int64 confidence;
```
ali-behjati: confidence is always positive, uint64

lazer/contracts/evm/src/PythLazerLib.sol
```diff
@@ -101,4 +107,301 @@ library PythLazerLib {
...
+                ) {
+                    (feed.exponent, pos) = parseFeedValueInt16(payload, pos);
+                    feed.existsFlags |= PythLazerStructs.EXPONENT_EXISTS;
+
+                    // Confidence Property
+                } else if (
+                    property == PythLazerStructs.PriceFeedProperty.Confidence
+                ) {
+                    (feed.confidence, pos) = parseFeedValueInt64(payload, pos);
+                    if (feed.confidence != 0)
```
ali-behjati: it can be 0 too.

lazer/contracts/evm/test/PythLazer.t.sol
ali-behjati, line 1: i recommend adding testing via some fixtures from Lazer as well to make sure everything works properly cross-environment.

lazer/contracts/evm/src/PythLazer.sol
aditya520: not sure. following the same pattern as [verifyUpdate](https://github.com/pyth-network/pyth-crosschain/blob/e7049752d4060c83ea8087d934df8364f76e9ee0/lazer/contracts/evm/src/PythLazer.sol#L74).

lazer/contracts/evm/src/PythLazerLib.sol
aditya520: Yeah, that was intentional. From a user standpoint, if the price is unknown by the lazer, we shouldn't even mention that property exist for timestamp. 
I don't have a strong opinion on this. I took the easier approach from both development and user standpoint.

lazer/contracts/evm/src/PythLazerLib.sol
aditya520: If the price doesn't exist, confidence doesn't exist. Which means the None is parsed as 0. Should we just make confidence throw error when price is zero?

lazer/contracts/evm/src/PythLazerStructs.sol
aditya520: Do you think if we will have more than 32 properties? 
Should I make it 64 or 128?

lazer/contracts/evm/src/PythLazerStructs.sol
aditya520: I am replacing them like `_price`. I believe that are better than `maybePrice`.

lazer/contracts/evm/src/PythLazerStructs.sol
aditya520: I changed it to 128.

lazer/contracts/evm/src/PythLazerLib.sol
Riateche: I think it's important to keep the distinction. A consumer contract may require the presence of a field so that it always uses the value if it's available in Lazer, but it can also have an alternative code path that's only used if Lazer doesn't have a value for this timestamp (which the payload guarantees in that case).

lazer/contracts/evm/src/PythLazerLib.sol
Riateche: We currently use the same `Price` type for confidence so it cannot be zero. Zero value here means it's not present. I think it's incorrect though. We should have created a `PriceDelta` type for confidence. But until we decide to change it in Lazer, we have to follow the existing logic here.

lazer/contracts/evm/src/PythLazer.sol
ali-behjati: payload is being used there. @Riateche why were we returning the signer there? my suggestion is removing it. it's a friction to a user (as in they think was there any reason it was returned and they can't find any answer).

Riateche: I don't remember. I think it's fine to remove it.

aditya520: I removed it from `verifyAndParseUpdate()`. If I remove it from `verifyUpdate()`, it will change the method signature, hence it won't be backward compatible anymore. WDYT? @ali-behjati @Riateche

PR #3136, opened 2025-10-15
feat: make price formatting feed-exponent aware

fhqvst:
## Summary

In this PR we make use of the feed's `exponent` in all places where we display its price or confidence intervals. 

## Rationale

While slightly less visually pleasing, this change should prevent users from seeing any inconsistencies as a result of rounding.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/PriceFeed/Chart/chart.tsx
+++ b/apps/insights/src/components/PriceFeed/Chart/chart.tsx
@@ -69,7 +69,6 @@ const useChartElem = (symbol: string, feedId: string) => {
   const chartContainerRef = useRef<HTMLDivElement | null>(null);
   const chartRef = useRef<ChartRefContents | undefined>(undefined);
   const isBackfilling = useRef(false);
-  const priceFormatter = usePriceFormatter();
   const abortControllerRef = useRef<AbortController | undefined>(undefined);
   // Lightweight charts has [a
   // bug](https://github.com/tradingview/lightweight-charts/issues/1649) where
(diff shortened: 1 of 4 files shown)


benduran (approved):
**LGTM**
LGTM

apps/insights/src/hooks/use-price-formatter.ts
```diff
@@ -1,40 +1,38 @@
...
+      bigNumberFormatter,
+      smallNumberFormatter,
+      exponentBasedFormatter,
+      decimals,
+    ],
   );
   return useMemo(() => ({ format }), [format]);
 };
-
-const formatToSubscriptNumber = (numString: string) => {
```
cprussin, line 19: I think you should consider keeping this formatter, for some cryptos the pricing & CI are very very small fractions and you can have very long streams of zeroes which are very hard to read.  This formatter resolves that issue quite nicely and can be applied on top of the `exponentBasedFormatter`.

screenshot-2025-10-17-095310

fhqvst: Makes sense - I've added it back now!

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -79,6 +78,9 @@ const useChartElem = (symbol: string, feedId: string) => {
   const whitespaceData = useRef<Set<WhitespaceData>>(new Set());
 
   const { current: livePriceData } = useLivePriceData(Cluster.Pythnet, feedId);
+  const priceFormatter = usePriceFormatter(livePriceData?.exponent, {
+    subscriptZeros: false,
```
cprussin, line 82: Is there a good reason why we'd turn these off?  Seems like it's a strictly better UX right?

fhqvst: Not when comparing the numbers to other numbers, such as in the y-axis of the price chart. In that case the better user experience is right-aligned numbers with a fixed set of decimals.

cprussin: ah yeah that makes sense, thanks for clarifying

PR #3137, opened 2025-10-15
feat: Remove cryptocurrency-icons

fhqvst:
## Summary

In this PR we remove the `cryptocurrency-icons` package completely, so to only use the more generic asset-specific icons.

## Rationale

There have been asks to update some of the currency icons, but it is difficult for us to keep them all up to date, which is why we'd rather use the generic icons instead.

## How has this been tested?

- [ ] Added new tests


apps/insights/src/components/PriceFeed/header.tsx
```diff
@@ -100,10 +95,7 @@ const PriceFeedHeaderImpl = (props: PriceFeedHeaderImplProps) => (
                 description: props.feed.product.description,
                 displaySymbol: props.feed.product.display_symbol,
                 icon: (
-                  <PriceFeedIcon
-                    assetClass={props.feed.product.asset_type}
-                    symbol={props.feed.symbol}
-                  />
+                  <PriceFeedIcon assetClass={props.feed.product.asset_type} />
```
benduran, line 98: nit: I see `<PriceFeedIcon assetClass={props.feed.product.asset_type} />` in a lot of places. We could consider keeping things DRY and hoisting this up to a variable, since it seems it's most the same implementation everywhere:
```typescript
const priceFeedIcon = <PriceFeedIcon assetClass={props.feed.product.asset_type} />

// then, later, down below
icon: priceFeedIcon
```

This way, if we need to change this component for whatever reason, we only need to change it in one place 🙂

apps/api-reference/src/apis/evm/get-update-fee.tsx
```diff
@@ -1,6 +1,3 @@
-import Btc from "cryptocurrency-icons/svg/color/btc.svg";
-import Eth from "cryptocurrency-icons/svg/color/eth.svg";
```
cprussin, line 2: Hmm I can't recall what the fallback looks like for api-reference, is there a reasonable generic icon in place?

fhqvst: There isn't, but I'd argue showing a generic icon here isn't helpful anyway since the purpose of the icon in this case is to distinguish BTC from ETH.

PR #3138, opened 2025-10-15
feat: Use benchmarks API for historical data

fhqvst:
## Summary

With this PR we now consume historical price data from the Benchmarks API instead of using our own hand-rolled Clickhouse query.

One downside is that we don't have confidence intervals in the Benchmarks API...

## Rationale

Our clients' users have complained that they're seeing inconsistent data in insights hub vs. the data on our clients' pages. With this change we now make sure that the data our clients are consuming is the same one we're using on insights hub, meaning there shouldn't be any inconsistencies.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/PriceFeed/Chart/chart.tsx
+++ b/apps/insights/src/components/PriceFeed/Chart/chart.tsx
@@ -184,15 +184,7 @@ const useChartElem = (symbol: string, feedId: string) => {
             return;
           }
 
-          const benchmarkData = benchmarksApiResponseSchema.parse(jsonData);
-
-          // Transform OHLC data to our format using close prices
-          const data = benchmarkData.t.map((timestamp, i) => ({
-            time: timestamp as UTCTimestamp,
-            price: benchmarkData.c[i], // Use close price
-            confidence: 0, // No confidence data from benchmarks API
-            status: PriceStatus.Trading,
-          }));
+          const data = benchmarksApiResponseSchema.parse(jsonData);
 
           // Get the current historical price data
           // Note that .data() returns (WhitespaceData | LineData)[], hence the type cast.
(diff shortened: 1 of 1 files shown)


apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -177,7 +182,15 @@ const useChartElem = (symbol: string, feedId: string) => {
...
-          const data = historicalDataSchema.parse(jsonData);
+          const benchmarkData = benchmarksApiResponseSchema.parse(jsonData);
+
+          // Transform OHLC data to our format using close prices
+          const data = benchmarkData.t.map((timestamp, i) => ({
+            time: timestamp as UTCTimestamp,
+            price: benchmarkData.c[i], // Use close price
+            confidence: 0, // No confidence data from benchmarks API
+            status: PriceStatus.Trading,
+          }));
```
cprussin: I'd do this using a zod transform

cprussin (changes requested):
The change is overall good!  But I think there's a few more things before we ship it:

1. I think the `/historical-prices` endpoint is now dead code, can you check and clean it up if so?
2. @ali-behjati can you tell us how hard it would be to add CI to the benchmarks endpoint?  Is it feasible?  I really hate losing CI from the historical charts, plus it does seem reasonable that we might want to have CI in the benchmarks apis.

apps/insights/src/components/PriceFeed/Chart/chart.tsx
```diff
@@ -177,7 +182,15 @@ const useChartElem = (symbol: string, feedId: string) => {
             return;
           }
 
-          const data = historicalDataSchema.parse(jsonData);
+          const benchmarkData = benchmarksApiResponseSchema.parse(jsonData);
+
+          // Transform OHLC data to our format using close prices
+          const data = benchmarkData.t.map((timestamp, i) => ({
+            time: timestamp as UTCTimestamp,
+            price: benchmarkData.c[i], // Use close price
```
cprussin: I don't love this -- I think we will need to put more thought in here.  If we aren't showing OHLC candles, I think the chart will be misleading.  Plus, we should have full resolution (individual points) when zoomed all the way in, so OHLC doesn't really make sense at that resolution.  Let's grab some time to chat through this

PR #3139, opened 2025-10-15
fix(lazer-sdk): implemented function to convert all non-string WS frames to an array buffer, then leverage an isomorphic buffer implementation

benduran:
# Partially-resolves [PFI-101](https://linear.app/douro-labs/issue/PFI-101/fix-browser-support-in-lazer-js-sdk)

## Summary

As the title suggests, there was a bug in the lazer Websocket client, which was built and intended to built use isomorphically across browser and server-side JavaScript environments.

However, the current implementation made assumptions that the `Buffer` class would be available in all environments, when it is not something that exists in the browser.

As such, this PR adds the following:
- Initial conversion of all non-string WebSocket frames to an `ArrayBuffer`
- Then, pass the created ArrayBuffer into newly-created `IsomorphicBuffer.from()` function
This allows all existing logic to remain in place, minimizing changes or needing to create approximations of functions but against the `ArrayBuffer` class.

## Rationale

This feels like the least risky way to fix the issue with isomorphism. This PR _**does not**_ address the auth token being provided in the `QueryString` at WSS open time. This will occur in another PR, as that will have some business logic changes that change how connections are created and destroyed.

## How has this been tested?

- [ ] Added new tests

### Working in Node

https://github.com/user-attachments/assets/5944172a-0e6b-41a9-80b5-13c9533cdb67

### Working in the browser

https://github.com/user-attachments/assets/ed6404a2-a192-43a7-9d66-e962412968bd

### DIFF
--- a/lazer/sdk/js/src/util/env-util.ts
+++ b/lazer/sdk/js/src/util/env-util.ts
@@ -10,7 +10,10 @@ const g = globalThis as Partial<{
  * @returns true if in a worker of some kind, false if otherwise
  */
 export function envIsServiceOrWebWorker() {
-  return typeof WorkerGlobalScope !== "undefined" && g.self instanceof WorkerGlobalScope;
+  return (
+    typeof WorkerGlobalScope !== "undefined" &&
+    g.self instanceof WorkerGlobalScope
+  );
 }
 
 /**


lazer/sdk/js/src/client.ts
```diff
@@ -46,53 +48,55 @@ export class PythLazerClient {
...
-            value.leUnsigned = data.subarray(pos, pos + len);
-          } else if (magic == FORMAT_MAGICS_LE.JSON) {
-            value.parsed = JSON.parse(
-              data.subarray(pos + UINT32_NUM_BYTES, pos + len).toString(),
-            ) as ParsedPayload;
-          } else {
-            throw new Error("unknown magic: " + magic.toString());
-          }
-          pos += len;
+      const value: BinaryResponse = { subscriptionId };
```
benduran, line 139: this code is unchanged, except for whitespace / indentation

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -7,6 +9,53 @@ import { dummyLogger } from "ts-log";
...
+ * 
+ * Browsers don't support custom headers for WebSocket connections, so if a browser is detected,
+ * the token is added as a query parameter instead. Else, the token is added as an Authorization header.
+ * 
+ * @param url - The WebSocket URL
+ * @param token - The authentication token
+ * @param wsOptions - Existing WebSocket options
+ * @returns Object containing the modified endpoint and wsOptions
+ */
+function addAuthentication(
```
benduran: this branch was a fork off of [this PR](https://github.com/pyth-network/pyth-crosschain/pull/3014), which added the query-string variation for authentication.

This is an okay short-term solution. Longer term, for improved security, we will need to update the back-and-forth between the client and server to allow for unauthd connections, but require the first message to be the token.

this behavior will be incoming in a follow-up PR

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -7,25 +9,72 @@ import { dummyLogger } from "ts-log";
...
+): AddAuthenticationReturnType {
+  if (envIsBrowser() || envIsWorker()) {
+    // Browser: Add token as query parameter
+    const urlObj = new URL(url);
+    urlObj.searchParams.set("ACCESS_TOKEN", token);
+
+    // For browsers, filter out wsOptions since headers aren't supported
+    return {
+      endpoint: urlObj.toString(),
+      wsOptions: undefined,
```
cprussin: Is there any reason to even have the key here?

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -7,25 +9,72 @@ import { dummyLogger } from "ts-log";
...
+    };
+  }
+  // Node.js: Add Authorization header
+  return {
+    endpoint: url,
+    wsOptions: {
+      ...wsOptions,
+      headers: {
+        ...(wsOptions.headers as Record<string, string> | undefined),
```
cprussin: Why is the typecast needed?

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -104,6 +151,7 @@ export class WebSocketPool {
         rws.onError = config.onError;
       }
       // Handle all client messages ourselves. Dedupe before sending to registered message handlers.
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
cprussin: I don't think we should ignore this rule -- there's a very good reason for this rule, which is that something that returns a promise is being passed to something that does not expect a promise.  The promise will get triggered, but any exceptions that result will be unhandled and potentially result in a broken system state.

We should always add a `.catch()` whenever interfacing between async and non-async code to avoid unhandled exceptions implicitly resulting in a broken state, even if that catch is a no op (which at least signals your intention is to explicitly ignore exceptions)

lazer/sdk/js/src/socket/resilient-websocket.ts
```diff
@@ -155,7 +155,7 @@ export class ResilientWebSocket {
 
     this.heartbeatTimeout = setTimeout(() => {
       this.logger.warn("Connection timed out. Reconnecting...");
-      this.wsClient?.terminate();
+      this.wsClient?.close();
```
tejasbadadare: Looks like `close()` is a clean closure using the closing handshake, whereas `terminate()` forcefully closes the socket. In this case we're closing the connection because the server failed to heartbeat in time. I think terminate works better here since the server hanged/broke protocol. (This is shown in the ws library [example](https://github.com/websockets/ws#how-to-detect-and-close-broken-connections) too.)

lazer/sdk/js/src/socket/websocket-pool.ts
tejasbadadare: Discussed offline, we're gonna punt on this because we are okay with querystring auth for now, and the vast majority of users will use header based auth from a Node env.

lazer/sdk/js/src/client.ts
```diff
@@ -112,73 +48,71 @@ export class PythLazerClient {
    * or a binary response containing EVM, Solana, or parsed payload data.
    */
   addMessageListener(handler: (event: JsonOrBinaryResponse) => void) {
-    const wsp = this.getWebSocketPool();
-    wsp.addMessageListener((data: WebSocket.Data) => {
-      if (typeof data == "string") {
+    this.wsp.addMessageListener(async (data: WebSocket.Data) => {
```
cprussin: I don't think it's safe to make this async if `addMessageListener` doesn't expect an async function (for the same reason above, it won't catch exceptions) -- although I notice you didn't add an eslint override so does `addMessageListener` handle async callbacks?

lazer/sdk/js/src/client.ts
```diff
@@ -112,73 +48,71 @@ export class PythLazerClient {
...
-        pos += UINT32_NUM_BYTES;
-        if (magic != BINARY_UPDATE_FORMAT_MAGIC_LE) {
-          throw new Error("binary update format magic mismatch");
-        }
-        // TODO: some uint64 values may not be representable as Number.
-        const subscriptionId = Number(
-          data.subarray(pos, pos + UINT64_NUM_BYTES).readBigInt64BE(),
-        );
-        pos += UINT64_NUM_BYTES;
+        return;
```
cprussin, line 122: This is a nit, but something we've recently discussed in some team conversations -- personally I really dislike the early return pattern.  It has many similar drawbacks to a goto -- if I'm reading code in the large chunk of functionality below, it's not clear from the control structure that execution may not have reached that point for different reasons or that different constraints have been applied to the types of values, and as code grows the early return makes it much harder to understand the actual control structure or type refinement at play.

It's also error prone, as code is refactored and grows over time you have to be aware of context that is not obvious or encoded in either the types (for conditions that typescript can't narrow) OR the control structure of the code in order to understand how the code is executing, which leads to incorrect assumptions when code is refactored or extended.

I'd much rather have some extra control flow nesting -- for one, the control flow is obvious, and the branching structure clearly shows the refinements /  filters that have been applied to get to a point in the codebase.  For two, it's harder to mask nasty control structure that should be refactored when that nasty control structure is not hidden behind early returns, so I tend to find that code that avoids early returns tends to get refactored and have better abstractions because you feel the pain more acutely of having the wrong abstractions.

lazer/sdk/js/src/socket/resilient-websocket.ts
benduran: Thanks for the response.

the terminate method doesn’t appear to exist on the browser-based implementation, which feels like a bug with the isomorphic-ws library.

you can see an uncaught exception if you setup a test page and don’t subscribe to anything, which will trigger the web socket timeout, then throw an unsought exception at runtime.

my recommendation would be to call close, but provide a meaningful code min the close call.

you’ll see that the concept of termination isn’t something that exists in the native websocket api: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#instance_methods

I can upload a video of the runtime failure in the morning

lazer/sdk/js/src/client.ts
```diff
@@ -112,73 +48,71 @@ export class PythLazerClient {
...
+        } else {
+          throw new Error("unknown magic: " + magic.toString());
         }
-        handler({ type: "binary", value });
-      } else {
-        throw new TypeError("unexpected event data type");
+        pos += len;
       }
+      handler({ type: "binary", value });
+      return;
```
cprussin: Is the return here needed?  it's the last line of the callback no?

lazer/sdk/js/src/util.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Detects if the code is running in a regular DOM or Web Worker context.
+ * @returns true if running in a DOM or Web Worker context, false if running in Node.js
+ */
+export function envIsBrowser() {
+  // disabling this rule for this line, as the rule is reporting a false positive (it's a bug)
+  // for the typeof check for the existence of the querySelector function
+
```
cprussin: looks like eslint decided this rule wasn't actually in use and removed the disable....

lazer/sdk/js/src/util.ts
```diff
@@ -0,0 +1,61 @@
...
+  return typeof window !== 'undefined' && Boolean(window.navigator.userAgent);
```
cprussin: I think a more correct way to do this is with `typeof globalThis.window === "undefined"`, because [node will support `globalThis.navigator.userAgent` existing](https://nodejs.org/api/globals.html#navigatoruseragent) in the future, but it will never have a window.

lazer/sdk/js/src/util.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Utility function for detecting whether the current JS execution
+ * is some server-side context, either Node.js or Bun
+ * (since they both have top-level process objects)
+ */
+export function envIsNode() {
+  return typeof globalThis !== 'undefined' && typeof globalThis.process === 'object';
```
cprussin: Should this not just be `!envIsBrowser()`?

lazer/sdk/js/src/socket/resilient-websocket.ts
tejasbadadare: Ahh, another tricky non-isomorphism :) Yeah, terminate() goes a layer below the websocket API and just closes the socket on the OS level. Makes sense that the browser-based implementation doesn't (can't?) expose that. 

And yep i'm aware of the timeout runtime exception, but that actually isn't because the WS heartbeat times out, it's a forceful disconnect from the server. The intention there is to disconnect clients that haven't subscribed to anything and are just sitting around taking up a socket. (Yes it should ideally be a graceful d/c.) 

I'd just want to validate that `close()` is able to successfully close the connection even when the server doesn't respond to the close handshake. It's important because we want to ensure Node-based users don't regress and are able to reconnect if a Lazer node is overloaded/flaky.

benduran: I'll double check things on the browser-side. Worst-case scenario, we can do something like the following:
```
if (typeof this.wsClient?.terminate === 'function') this.wsClient.terminate();
else this.wsClient.close(SOME_CODE_HERE);
```

lazer/sdk/js/src/client.ts
benduran: update for those following along from home: we chatted about this offline and agreed that we're happy to leave the proposed code as-is, as the early return is code a style paradigm that has its pros and cons, and its preference depends on your background as an engineer (I've spent a lot of time around Go engineers, and have written a lot of Go, myself, and have gotten used to reading control flows in this style).

Either way, the feedback and civil discussion is extremely appreciated. Thank you!

lazer/sdk/js/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@pythnetwork/pyth-lazer-sdk",
   "version": "4.0.0",
   "description": "Pyth Lazer SDK",
+  "engines": {
+    "node": ">=20"
```
benduran: added this, per discussion over in our Slack thread about missing an engine and having bogus linting errors thrown

lazer/sdk/js/src/client.ts
benduran: comment is out of date to the current code. resolving conversation 👌

lazer/sdk/js/src/util.ts
benduran: done!

lazer/sdk/js/src/util.ts
benduran: Done in the latest changes. thanks!

lazer/sdk/js/src/client.ts
benduran: `addMessageListener` supports either sync or async functions now 👌

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: done by adding a simple, wrapped function that voids out the return value from the async func ([code here](https://github.com/pyth-network/pyth-crosschain/blob/bduran/lazer-js-sdk/browser-support/lazer/sdk/js/src/socket/websocket-pool.ts#L123))

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: code no longer exists so comment is out of date. resolving

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: I think this comment might be out of date now? can you confirm, when you have a moment?

lazer/sdk/js/src/socket/resilient-websocket.ts
benduran: Addressed this in 2e3f822a35bfa9481e42eacc104815d15976c3d5 by using `terminate()`, if it's available, and falling back to `close()` with a custom closure code (which is exported, in case folks want to sniff out why the socket is closing).

Resolving thread 👍

lazer/sdk/js/src/socket/resilient-websocket.ts
```diff
@@ -26,7 +29,9 @@ export class ResilientWebSocket {
   private maxRetryDelayMs: number;
   private logAfterRetryCount: number;
 
-  wsClient: undefined | WebSocket;
+  wsClient:
+    | undefined
+    | (Omit<WebSocket, "terminate"> & Partial<Pick<WebSocket, "terminate">>);
```
cprussin: Nice, might want to leave a comment explaining this type though since it won't be obvious to folks who don't know the typescript utility types as well

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -65,16 +70,28 @@ export class WebSocketPool {
     const numConnections = config.numConnections ?? DEFAULT_NUM_CONNECTIONS;
 
     for (let i = 0; i < numConnections; i++) {
-      const url = urls[i % urls.length];
+      let url = urls[i % urls.length];
```
cprussin: Total nit, but I really dislke using a let binding for this -- it's not strictly required that you have mutability here, and the fact that you have to do this is due to constructing a mixture of values within the same code path.  I think this would be more concise and simpler to understand if written something like this:

```ts
const baseUrl = urls[i % urls.length];
const isBrowserOrWorker = envIsBrowserOrWorker();
const url = isBrowserOrWorker ? addAccessToken(url) : url;
const wsOptions: ResilientWebSocketConfig["wsOptions"] = {
    ...config.rwsConfig?.wsOptions,,
    ...(isBrowserOrWorker && { headers: { Authorization: `Bearer ${token}` } })
};
```

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -104,7 +121,9 @@ export class WebSocketPool {
         rws.onError = config.onError;
       }
       // Handle all client messages ourselves. Dedupe before sending to registered message handlers.
-      rws.onMessage = pool.dedupeHandler;
+      rws.onMessage = (data) => {
+        void pool.dedupeHandler(data);
```
cprussin: Personally I don't like using `void` to discard errors because it's not clear that discarding errors is the intention, when I choose to discard errors I prefer to write it as:

```ts
pool.dedupeHandler(data).catch(() => { /* no-op on error */ });
```

That way it's very obvious to future readers that the intention was to discard errors, and it wasn't an accident.

Also I'd probably just at least log out the error message rather than swallowing it silently; unless you expect for errors to happen and if they do it's not important, you probably at least want a message in the event that something breaks unexpectedly.

lazer/sdk/js/src/socket/websocket-pool.ts
```diff
@@ -144,11 +163,14 @@ export class WebSocketPool {
...
-        ? data
-        : Buffer.from(data as Buffer).toString("hex");
+  dedupeHandler = async (data: WebSocket.Data): Promise<void> => {
+    let cacheKey = "";
+    if (typeof data === "string") {
+      cacheKey = data;
+    } else {
+      const buff = await IsomorphicBuffer.fromWebsocketData(data);
+      cacheKey = buff.toString("hex");
+    }
```
cprussin: Similar nit -- this can be trivially written without using a mutable binding and I strongly prefer to avoid unneeded mutable bindings when not strictly required:

```ts
const cacheKey = typeof data === "string" ? data : (
  (await IsomorphicBuffer.fromWebsocketData(data)).toString("hex")
);
```

If you don't like the ternary (I don't love it either), you can extract to a `getCacheKey` function too, which avoids the mutable binding AND decomposes the function a bit too which is nice.

lazer/sdk/js/src/util/buffer-util.ts
```diff
@@ -0,0 +1,42 @@
...
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import { Buffer as BrowserBuffer } from "buffer";
+
+import type { Data } from "isomorphic-ws";
+
+const { Buffer: PossibleBuiltInBuffer } = globalThis as Partial<{
+  Buffer: typeof Buffer;
+}>;
+
+const BufferClassToUse = PossibleBuiltInBuffer ?? BrowserBuffer;
```
cprussin: I'm not sure I understand the need to do the typecast `globalThis` like you have above, could this be expressed simpler as:

```ts
const BufferClassToUse = 'Buffer' in globalThis ? globalThis.Buffer : BrowserBuffer;
```

?

lazer/sdk/js/src/util/buffer-util.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class IsomorphicBuffer extends BufferClassToUse {
```
cprussin: I don't think there's any good reason to use an es6 class here -- there is no state being captured and only a static method here.   Conceptually you are only exporting functions that have no instances or state, but if this were to get extended, then it would degrade the quality of tree-shaking since many bundlers cannot tree shake parts of es6 classes at all, and even those that can have very major limitations in how they can do it.

I would just go back to what you had before and export `fromWebsocketData` as a naked function not encapsulated in a class.

lazer/sdk/js/src/util/buffer-util.ts
```diff
@@ -0,0 +1,42 @@
...
+      const arrBuffer = new ArrayBuffer(data.length);
+      const v = new Uint8Array(arrBuffer);
+      for (const [i, item] of data.entries()) {
+        v[i] = item;
+      }
+      return BufferClassToUse.from(arrBuffer);
+    }
+    throw new TypeError(
+      "unexpected event data type found when IsomorphicBuffer.fromWebsocketData() called",
+    );
```
cprussin: Can this case be prevented entirely using typescript exhaustivity checking?

lazer/sdk/js/src/util/env-util.ts
```diff
@@ -0,0 +1,34 @@
...
+}>;
+
+/**
+ * Detects if this code is running within any Service or WebWorker context.
+ * @returns true if in a worker of some kind, false if otherwise
+ */
+export function envIsServiceOrWebWorker() {
+  const possiblyInAWorker =
+    typeof WorkerGlobalScope !== "undefined" && g.self !== undefined;
+  return possiblyInAWorker && g.self instanceof WorkerGlobalScope;
```
cprussin: Can this be minorly more concisely written as:

```ts
return WorkerGlobalScope !== "undefined" && g.self instanceof WorkerGlobalScope;
```

Isn't the check for `g.self !== undefined` redundant?

lazer/sdk/js/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@pythnetwork/pyth-lazer-sdk",
   "version": "4.0.0",
```
tejasbadadare: Pls increment the package version (probably a major version bump since we restricted the engine to node >= 22)

tejasbadadare (approved):
Awesome work!! Pls increment version before merging :)

lazer/sdk/js/src/socket/resilient-websocket.ts
benduran: done in d4cf06be69d3e5609a1c312cb621e99dd2e3c046

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: done in 20e664aab4b618d602b2f550f4ac38715e3322d1. thanks!

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: crafted a slightly more meaningful error message and am now throwing that, in the off chance it occurs (unlikely). done in 8ec22ae61f70992c1846f511473ade1248e01623

lazer/sdk/js/src/util/buffer-util.ts
benduran: yes, this is much better. I was overthinking things a bit 😅

lazer/sdk/js/src/socket/websocket-pool.ts
benduran: done in 7aebe1f1b2c2709ca9de50a2e0b09efc8496d4b5

lazer/sdk/js/src/util/buffer-util.ts
benduran: also done in 7aebe1f1b2c2709ca9de50a2e0b09efc8496d4b5

great suggestion, btw

lazer/sdk/js/src/util/buffer-util.ts
benduran: done in 7aebe1f1b2c2709ca9de50a2e0b09efc8496d4b5,. captured all cases indicated by the typings on the `WebSocket.Data` interface ([changes in this file, if you're curious](https://github.com/pyth-network/pyth-crosschain/blob/bduran/lazer-js-sdk/browser-support/lazer/sdk/js/src/util/buffer-util.ts#L17))

lazer/sdk/js/src/util/env-util.ts
benduran: done in 8baec64433940479ac9079f537be0c782207f205

lazer/sdk/js/package.json
benduran: done in 27d0b5d6cd38467830e075b9185b228122b60d13

thanks!

PR #3140, opened 2025-10-15
feat(lazer/publisher-sdk): add asset state and governance messages

tejasbadadare:
## Summary
- Add Asset governance proto messages
- Asset follows similar structure to Feed. The only static field is `asset_id`, the rest of the fields will be dynamic inside `metadata`.
- There isn't a real need to scope Assets to specific shards since they are global reference data. As such, no shard related  fields are included in Asset.
- These values will be ignored by the governance handling code, so this is backward compatible and safe. We will only use these instructions when state and governance handling code has been updated to load these objects.

## Rationale
- Needed to add Assets to Lazer state

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/lazer/publisher_sdk/rust/Cargo.toml
+++ b/lazer/publisher_sdk/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-publisher-sdk"
-version = "0.18.0"
+version = "0.19.0"
 edition = "2021"
 description = "Pyth Lazer Publisher SDK types."
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


darunrs (approved):
Nice!

lazer/publisher_sdk/rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pyth-lazer-publisher-sdk"
-version = "0.17.0"
+version = "0.17.1"
```
Riateche: Adding new fields is a breaking change.
```suggestion
version = "0.18.0"
```

PR #3147, opened 2025-10-21
fix(hip-3-pusher): Multisig update support (only 1/N, no KMS)

merolish:
## Summary

Support multisig oracle update as per https://hyperliquid.gitbook.io/hyperliquid-docs/hypercore/multi-sig .

## Rationale

This allows us to provide our own wallet for updating without ever having access to the deployer oracle account. They can grant and revoke access at will.

## How has this been tested?

- [ ] Added new tests


apps/hip-3-pusher/src/pusher/publisher.py
```diff
@@ -29,19 +30,28 @@ def __init__(self, config: Config, price_state: PriceState, metrics: Metrics):
...
+            # TODO: Add KMS/multisig support
+            if config.multisig.enable_multisig:
+                raise Exception("KMS/multisig not yet supported")
+
             self.enable_kms = True
             self.kms_signer = KMSSigner(config, self.publisher_exchanges)
 
+        if config.multisig.enable_multisig:
+            if not config.multisig.multisig_address:
+                raise Exception("Multisig enabled but missing multisig address")
```
tejasbadadare, line 52: You can enforce this directly in the pydantic model. Doing as much validation as possible in the model itself aligns with pydantic's [philosophy](https://docs.pydantic.dev/latest/concepts/models/#:~:text=on%20the%20model.-,Validation,-%E2%80%94%20a%20deliberate%20misnomer) of "if i can construct the object, i know it's valid."

```python
class MultisigConfig(BaseModel):
    enable_multisig: bool
    multisig_address: Optional[str] = None

    @model_validator(mode="after")
    def validate_address_exists_if_multisig_enabled(self) -> 'MultisigConfig':
        if self.enable_multisig and self.multisig_address is None:
            raise ValueError("`multisig_address` must be provided when `enable_multisig` is True")
        return self
```

tejasbadadare (approved):
Nice work! Suggested a minor improvement but nothing that blocks merging

PR #3151, opened 2025-10-26
fix(insights): use RPC as source of truth for publisher list

cprussin:
## Summary

Use the RPC as the source of truth for the publisher list & number of active feeds for insights hub.

## Rationale

Previously we were using clickhouse to drive this data, however clickhouse can be inaccurate or delayed for a number of reasons.

When I originally built this I misunderstood and thought publishers would only show up in the rpc if they had a data point for the last slot, but in fact publishers show up regardless of if they are actively publishing or not.

So this PR moves to the more accurate source of truth, the chain itself.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/insights/src/components/Publishers/publishers-card.tsx
+++ b/apps/insights/src/components/Publishers/publishers-card.tsx
@@ -155,9 +155,7 @@ const ResolvedPublishersCard = ({
           textValue: publisher.name ?? id,
           prefetch: false,
           data: {
-            ranking:
-              // eslint-disable-next-line unicorn/no-null
-              ranking === undefined ? null : <Ranking>{ranking}</Ranking>,
+            ranking: ranking !== undefined && <Ranking>{ranking}</Ranking>,
             name: (
               <PublisherTag
                 publisherKey={id}
@@ -177,11 +175,9 @@ const ResolvedPublishersCard = ({
                 {activeFeeds}
               </Link>
             ),
-            averageScore:
-              // eslint-disable-next-line unicorn/no-null
-              averageScore === undefined ? null : (
-                <Score score={averageScore} width={PUBLISHER_SCORE_WIDTH} />
-              ),
+            averageScore: averageScore !== undefined && (
+              <Score score={averageScore} width={PUBLISHER_SCORE_WIDTH} />
+            ),
           },
         }),
       ),
--- a/apps/insights/src/services/pyth/get-publishers-for-cluster.ts
+++ b/apps/insights/src/services/pyth/get-publishers-for-cluster.ts
@@ -5,27 +5,29 @@ import { redisCache } from "../../cache";
 const _getPublishersByFeedForCluster = async (cluster: Cluster) => {
   const data = await getPythMetadata(cluster);
   const result: Record<string, string[]> = {};
-  for (const key of data.productPrice.keys()) {
-    const price = data.productPrice.get(key);
-    result[key] =
-      price?.priceComponents.map(({ publisher }) => publisher.toBase58()) ?? [];
+  for (const [key, price] of data.productPrice.entries()) {
+    result[key] = price.priceComponents.map(({ publisher }) =>
+      publisher.toBase58(),
+    );
   }
   return result;
 };
 
+/**
+ * Given a cluster, this function will return a record which maps each
+ * permissioned publisher to the list of price feed IDs for which that publisher
+ * is permissioned.
+ */
 const _getFeedsByPublisherForCluster = async (cluster: Cluster) => {
   const data = await getPythMetadata(cluster);
   const result: Record<string, string[]> = {};
-  for (const symbol of data.productPrice.keys()) {
-    const price = data.productPrice.get(symbol);
-    if (price !== undefined) {
-      for (const component of price.priceComponents) {
-        const publisherKey = component.publisher.toBase58();
-        if (result[publisherKey] === undefined) {
-          result[publisherKey] = [symbol];
-        } else {
-          result[publisherKey].push(symbol);
-        }
+  for (const [symbol, price] of data.productPrice.entries()) {
+    for (const component of price.priceComponents) {
+      const publisherKey = component.publisher.toBase58();
+      if (result[publisherKey] === undefined) {
+        result[publisherKey] = [symbol];
+      } else {
+        result[publisherKey].push(symbol);
       }
     }
   }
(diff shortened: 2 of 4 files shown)


apps/insights/src/services/pyth/get-publishers-for-cluster.ts
```diff
@@ -13,7 +13,31 @@ const _getPublishersForCluster = async (cluster: Cluster) => {
...
+        if (result[publisherKey] === undefined) {
+          result[publisherKey] = [symbol];
+        } else {
+          result[publisherKey].push(symbol);
+        }
+      }
+    }
+  }
+  return result;
+};
```
alexcambose, line 35: might be worth adding a comment here explaining what this does, kinda hard to figure out unless you have an idea of the payload from the rpc.

apps/insights/src/server/pyth.ts
```diff
@@ -98,3 +79,22 @@ export const getFeedForSymbolRequest = async ({
     ? undefined
     : priceFeedsSchema.element.parse(parsedData);
 };
+
+const fetchPythData = async (
```
alexcambose, line 83: 👍 nice

apps/insights/src/app/api/pyth/get-publishers/route.ts
```diff
@@ -0,0 +1,25 @@
...
+  const cluster = clusterSchema.safeParse(
+    request.nextUrl.searchParams.get("cluster"),
+  );
+
+  return cluster.success
+    ? NextResponse.json(await getPublishers(cluster.data))
+    : new Response("Invalid params", { status: 400 });
+};
+
+const clusterSchema = z
```
benduran, line 18: (not something to address in this PR, but more of a general Q): should we consider colocating all of our Zod schemas for easier searching and referencing? I've seen them sprinkled throughout the codebase(s), which is convenient for immediate use but perhaps not the most reusable or portable

apps/insights/src/components/Publishers/publishers-card.tsx
```diff
@@ -144,7 +155,9 @@ const ResolvedPublishersCard = ({
           textValue: publisher.name ?? id,
           prefetch: false,
           data: {
-            ranking: <Ranking>{ranking}</Ranking>,
+            ranking:
+              // eslint-disable-next-line unicorn/no-null
+              ranking === undefined ? null : <Ranking>{ranking}</Ranking>,
```
benduran: nit: to eliminate the use of the eslint rule disable line (and assuming that the `ranking` prop is setup to properly handle `ReactNode` types), you can turn this into a Boolean statement, which React should handle the same as any nullish value:
```tsx
ranking: ranking !== undefined && <Ranking>{ranking></Ranking>
```

benduran: Ditto for any other lines that are similar to this one in this PR

apps/insights/src/services/pyth/get-publishers-for-cluster.ts
```diff
@@ -13,7 +13,31 @@ const _getPublishersForCluster = async (cluster: Cluster) => {
...
 };
 
-export const getPublishersForCluster = redisCache.define(
-  "getPublishersForCluster",
-  _getPublishersForCluster,
-).getPublishersForCluster;
+const _getFeedsByPublisherForCluster = async (cluster: Cluster) => {
+  const data = await getPythMetadata(cluster);
+  const result: Record<string, string[]> = {};
+  for (const symbol of data.productPrice.keys()) {
```
benduran: q: why the preference here for `.keys()` instead of `.entries()`, which returns a tuple of `[symbol, price]`?

apps/insights/src/app/api/pyth/get-publishers/route.ts
alexcambose: I think this would make sense for some schemas, the ones used for validating data api side and also UI side.
In this particular case I feel like it makes more sense to keep the schema as close as possible to where it is used.

apps/insights/src/components/Publisher/layout.tsx
```diff
@@ -471,6 +465,33 @@ const ActiveFeedsCardImpl = (props: ActiveFeedsCardImplProps) => (
   </StatCard>
```
alexcambose, line 465: nit but reading this I find it strange that after the component has finished loading the data can still be undefined. Maybe `activeFeeds` should be `0` instead and we can get rid of this check?

apps/insights/src/components/Publishers/publishers-card.tsx
alexcambose: We need to do a similar thing here: https://github.com/pyth-network/pyth-crosschain/blob/main/apps/insights/src/components/Publisher/performance.tsx#L67-L69

Image

apps/insights/src/app/api/pyth/get-publishers/route.ts
cprussin: yeah I agree, I think it really depends on how the schema will be used.  For most of these schemas, they are only used where we ingest data from some external system and would never be reused outside of consuming that external system, so it makes sense to me to keep the schemas colocated with the code to consume the system.  But if we were to have a use case for a schema that's reused in different ways, I definitely think reorganizing makes sense.

apps/insights/src/components/Publishers/publishers-card.tsx
cprussin: nice catch, will do!

apps/insights/src/services/pyth/get-publishers-for-cluster.ts
cprussin: ahh great catch as well, no good reason.  I will update

apps/insights/src/services/pyth/get-publishers-for-cluster.ts
cprussin: yep, good call, will do

apps/insights/src/components/Publisher/layout.tsx
cprussin: Hmm well there might be a better way to name these things but `activeFeeds` comes from data in clickhouse and is `undefined` if there is no ranking information.  I guess I could be convinced otherwise, but it seems to me that if we don't have rankings for a publisher, it is more correct to show nothing than to show 0.  Let's chat offline to decide.

apps/insights/src/components/Publishers/publishers-card.tsx
cprussin: ahh nice catch

PR #3158, opened 2025-10-28
fix(insights): use web3.js Connection instead of PythConnection

cprussin:
Upon investigation, it seems the `PythConnection` from `@pythnetwork/client` is extremely inefficient for a few reasons:

1. `PythConnection` requires loading _all_ account info when creating the connection class, so that it can build a mapping from price account to program account.  However IH never used this mapping in practice, so all of that was wasted work.  This caused major issues for page load performance as loading all account info from pythnet was a ton of parsing which locked up the browser for multiple seconds.

2. `PythConnection` did not expose a mechanism to remove unused subscriptions

In doing this I also removed all the live prices contexts since none of that is needed any more, as well as the call to initialize prices with the last available price as that too looks unnecessary and redundant.

This change should _drastically_ improve performance in IH, especially during page load.

### DIFF
--- a/apps/insights/src/hooks/use-live-price-data.tsx
+++ b/apps/insights/src/hooks/use-live-price-data.tsx
@@ -3,53 +3,34 @@
 import type { PriceData } from "@pythnetwork/client";
 import { useLogger } from "@pythnetwork/component-library/useLogger";
 import { PublicKey } from "@solana/web3.js";
-import type { ComponentProps } from "react";
-import {
-  use,
-  createContext,
-  useEffect,
-  useCallback,
-  useState,
-  useMemo,
-  useRef,
-} from "react";
+import { useEffect, useState, useMemo } from "react";
 
-import {
-  Cluster,
-  subscribe,
-  getAssetPricesFromAccounts,
-} from "../services/pyth";
-
-const LivePriceDataContext = createContext<
-  ReturnType<typeof usePriceData> | undefined
->(undefined);
-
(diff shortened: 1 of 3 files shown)


apps/insights/src/hooks/use-live-price-data.tsx
```diff
@@ -3,53 +3,34 @@
...
+    const subscriptionId = subscribe(
+      cluster,
+      new PublicKey(feedKey),
+      ({ data }) => {
+        setData((prev) => ({ current: data, prev: prev.current }));
+      },
+    );
     return () => {
-      removeSubscription(feedKey, setData);
+      unsubscribe(cluster, subscriptionId).catch((error: unknown) => {
```
benduran, line 26: q: should we consider a failure to unsubscrive as a more critical failure and alert the user to reload their page (or similar)?

cprussin: No, at the worst case it means that the websocket will be continuing to send some data that we ignore.  The way the web3.js Connection object works is that it [updates all subscriptions](https://github.com/solana-foundation/solana-web3.js/blob/maintenance/v1.x/src/connection.ts#L6452) when creating a new subscription or deleting a subscription, so the websocket unsubscribe message will be retried by the next time any change in the subscribed feeds happens anyways.  So even the problem with continuing to receive unnecessary ws messages would resolve itself by just navigating around the app.

PR #3160, opened 2025-10-29
Bduran/dual publish

benduran:
## Summary

* Moved to using a new dual-publishing library called [ts-duality](https://www.npmjs.com/package/@better-builds/ts-duality), which handles the whole song-and-dance of compiling, generating types and dual writing ESM and CJS output formats (if needed).
    * this library was actually built in the last week, initially internally, to try and solve this problem (which is a common one in OSS-land, but hasn't been solved well yet). Now that the project is further along, it has been published under the `beta` npm tag, where I have been updating it.
* Deleted any bespoke `tsconfig.json` files that drifted from the repo best practices.
* via `ts-duality`, generated `tsconfig.build.json` files, which are exclusively used for building packages.
* went through various files, by hand, and updated TypeScript typing issues that were causing build failures as a result of moving to the stricter TypeScript configurations

## Rationale

we have many brilliant engineers here, but only a few who have been deep in the TypeScript and JavaScript weeds, especially with regards to package build and distribution practices. This PR is an initial step to clamp on a single "paved-path" way of doing things here.

## How has this been tested?

- [ ] Added new tests
    - no logic changes have been made to any existing code. only changes have been solely typing-related or `package.json#exports` related

Tested building and running the following apps (which are powered by NextJS). Will check more off as more tests and changes are made. Anything with a ~~strikethrough~~ is unrelated and likely not a JS / TS project. Verification involved verifying the apps loaded up without issue. Functionality-wise, they haven't been extensively tested.
* [X] api-reference
* ~~[ ] argus~~
* [x] developer-hub
    * NOTE: This doesn't seem to load anything more than an empty page, but it does run
* [X] entropy-debugger
* [X] entropy-explorer
* [X] entropy-tester
* ~~[ ] fortuna~~
* ~~[ ] hermes~~
* ~~[ ] hip-3-pusher~~
* [X] insights
* [X] price_pusher
* ~~[ ] pyth-lazer-agent~~
* ~~[ ] quorum~~
* [X] staking

## Picture of all builds passing locally
CleanShot 2025-10-30 at 14 00 11

### DIFF
--- a/apps/staking/jest.config.js
+++ b/apps/staking/jest.config.js
@@ -1,4 +1,3 @@
 import { defineJestConfigForNextJs } from "@pythnetwork/jest-config/define-next-config";
-("@pythnetwork/jest-config");
 
 export default defineJestConfigForNextJs();


contract_manager/scripts/@types/contractManagerConfig.d.ts
ali-behjati, line 1: dumb question, what is this?

apps/pyth-lazer-agent/.gitignore
ali-behjati, line 1: i'm wondering why we have similar git ignore files everywhere, i think it might be good to move some just up at the root of the repo

contract_manager/scripts/@types/contractManagerConfig.d.ts
benduran: this is a file that tells typescript that there are typings for this plain javascript file, even if it doesn't have one. this was required to allow typescript to build this project without erroring, after moving to a more strict typescript config (this is specifically for this file: `target_chains/ethereum/contracts/scripts/contractManagerConfig.js`)

apps/pyth-lazer-agent/.gitignore
benduran: good callout. could probably just apply a blanket ignore and allow local packages to override as needed.

I can address in a separate PR, so as to not bloat this already excessively-large one 🫠

ali-behjati:
LGTM

PR #3169, opened 2025-10-31
feat(hip-3-pusher): Configurable price sources

merolish:
## Summary

Configurable price source waterfall for oracle/mark/external prices for a particular HIP-3 dex.

## Rationale

Supporting multiple HIP-3 dex customers.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/hip-3-pusher/tests/test_price_state.py
+++ b/apps/hip-3-pusher/tests/test_price_state.py
@@ -36,22 +36,28 @@ def get_config():
 
 
 def test_good_hl_price():
+    """
+    Pass through fresh HL oracle price.
+    """
     config = get_config()
     price_state = PriceState(config)
     now = time.time()
-    price_state.state["hl_oracle"][SYMBOL] = PriceUpdate("110000.0", now - price_state.stale_price_threshold_seconds / 2.0)
+    price_state.hl_oracle_state.put(SYMBOL, PriceUpdate("110000.0", now - price_state.stale_price_threshold_seconds / 2.0))
 
     oracle_px, _, _ = price_state.get_all_prices(DEX)
     assert oracle_px == {f"{DEX}:{SYMBOL}": "110000.0"}
 
 
 def test_fallback_lazer():
+    """
+    HL oracle price is stale, so fall back to fresh Lazer price.
+    """
     config = get_config()
     price_state = PriceState(config)
     now = time.time()
-    price_state.state["hl_oracle"][SYMBOL] = PriceUpdate("110000.0", now - price_state.stale_price_threshold_seconds - 1.0)
-    price_state.state["lazer"][1] = PriceUpdate("11050000000000", now - price_state.stale_price_threshold_seconds / 2.0)
-    price_state.state["lazer"][8] = PriceUpdate("99000000", now - price_state.stale_price_threshold_seconds / 2.0)
+    price_state.hl_oracle_state.put(SYMBOL, PriceUpdate("110000.0", now - price_state.stale_price_threshold_seconds - 1.0))
+    price_state.lazer_state.put(1, PriceUpdate("11050000000000", now - price_state.stale_price_threshold_seconds / 2.0))
+    price_state.lazer_state.put(8, PriceUpdate("99000000", now - price_state.stale_price_threshold_seconds / 2.0))
 
     oracle_px, _, _ = price_state.get_all_prices(DEX)
     assert oracle_px == {f"{DEX}:{SYMBOL}": "111616.16"}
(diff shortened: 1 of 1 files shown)


apps/hip-3-pusher/config/config.sample.toml
tejasbadadare, line 1: Maybe its just me, but i get nervous when i see a `config.toml` committed to git, even though I know its an example config :P The convention in the rest of the codebase is to name it `config.sample.toml` to disambiguate, i think we should follow that here too

apps/hip-3-pusher/src/pusher/config.py
```diff
@@ -50,6 +45,37 @@ def set_default_urls(self):
...
+    base_source: PriceSource
+    quote_source: PriceSource
+
+
+class ConstantSourceConfig(BaseModel):
+    source_type: Literal["constant"]
+    value: str
+
+
+PriceSourceConfig = Union[SingleSourceConfig, PairSourceConfig, ConstantSourceConfig]
```
tejasbadadare: FYI in python 3 you don't need to use the `Union` type, you can just express this as `SingleSourceConfig | PairSourceConfig | ConstantSourceConfig` for a more concise syntax.

apps/hip-3-pusher/src/pusher/hermes_listener.py
```diff
@@ -81,9 +82,6 @@ def parse_hermes_message(self, data):
             publish_time = price_object["publish_time"]
             logger.debug("Hermes update: {} {} {} {}", id, price, expo, publish_time)
             now = time.time()
-            if id == self.base_feed_id:
-                self.price_state.hermes_base_price = PriceUpdate(price, now)
-            if id == self.quote_feed_id:
-                self.price_state.hermes_quote_price = PriceUpdate(price, now)
+            self.price_state.state[self.SOURCE_NAME][id] = PriceUpdate(price, now)
```
tejasbadadare: It feels a bit strange to me that we are using a mono-state, but indexing into it with "source name" to distinguish between states of different modules. It would be safer and more explicit to have distinct state objects (which you can still nest within the mono-state, like how Hermes does it,) something like HermesState / LazerState / HLState. The state objects can be passed to the Listener objects during instantiation, which guarantees a listener is scoped to its specific state, and can only mutate the data within. 

In the price consumers, you can also more safely reference `price_state.hermes_state` instead of `price_state[self.SOURCE_NAME]`. The state objects could also live inside of the Listener classes themselves to reduce indirection if desired.

apps/hip-3-pusher/src/pusher/price_state.py
```diff
@@ -1,18 +1,17 @@
...
 
+@dataclass
 class PriceUpdate:
-    def __init__(self, price, timestamp):
-        self.price = price
-        self.timestamp = timestamp
-
-    def __str__(self):
-        return f"PriceUpdate(price={self.price}, timestamp={self.timestamp})"
+    price: float | str
```
tejasbadadare, line 13: Float or str? Should we conform to one or the other?

apps/hip-3-pusher/src/pusher/price_state.py
```diff
@@ -1,18 +1,17 @@
...
+    timestamp: float
```
tejasbadadare, line 14: Timestamps should probably be integer (or datetime)

apps/hip-3-pusher/src/pusher/price_state.py
```diff
@@ -24,60 +23,65 @@ class PriceState:
...
+            self.get_prices(self.price_config.mark, market_name),
+            self.get_prices(self.price_config.external, market_name)
+        )
+
+    def get_prices(self, symbol_configs: dict[str, list[PriceSourceConfig]], market_name: str):
+        pxs = {}
+        for symbol in symbol_configs:
+            for source_config in symbol_configs[symbol]:
+                # find first valid price in the waterfall
+                px = self.get_price(source_config)
```
tejasbadadare, line 77: Could you please check that our unit tests sufficiently test the new waterfall logic?

apps/hip-3-pusher/src/pusher/config.py
```diff
@@ -50,6 +45,37 @@ def set_default_urls(self):
...
+    source_id: str | int
+    exponent: Optional[int] = None
+
+
+class SingleSourceConfig(BaseModel):
+    source_type: Literal["single"]
+    source: PriceSource
+
+
+class PairSourceConfig(BaseModel):
```
tejasbadadare, line 73: I guess generally we would leave "combining feeds" up to SEDA? Or would we use this PairSourceConfig for it?

apps/hip-3-pusher/tests/test_price_state.py
tejasbadadare, line 1: It's a bit difficult to follow what's happening in these tests, do you mind adding comments that explain the setup , action, and expected result? Even better would be to also update the test names to encode this information, e.g. `test_lazer_price_falls_back_to_hyperliquid_upon_stale_price`.

apps/hip-3-pusher/config/config.sample.toml
merolish: Ah, good call.

apps/hip-3-pusher/src/pusher/config.py
merolish: This is a "you're absolutely right" moment, cough.

apps/hip-3-pusher/src/pusher/hermes_listener.py
merolish: I like it, will change.

apps/hip-3-pusher/src/pusher/price_state.py
merolish: In this case I've been using time.time() which is indeed a float, I guess to allow differing precisions?

apps/hip-3-pusher/src/pusher/price_state.py
merolish: The tests can definitely stand to be expanded.

apps/hip-3-pusher/src/pusher/config.py
merolish: Set this up specifcally for the case (as in our internal publishers) where we need to quote in e.g. USDT. For more complicated transformations yes, hopefully SEDA can abstract it out.

apps/hip-3-pusher/src/pusher/price_state.py
merolish: HL messages have prices as a string-of-a-float; Lazer and Hermes messages have price as a string-of-an-int, and I was dividing by the exponent and keeping it as a float to not lose precision. We could keep the exponent in the PriceUpdate as well, perhaps.

tejasbadadare (approved):
LGTM, please add test improvements as you see fit

PR #3171, opened 2025-10-31
Chore(dev hub) add llm integration

aditya520:
## Summary

This PR adds LLM-friendly features to the Pyth Developer Hub, allowing users to easily share documentation pages with AI assistants like ChatGPT and Claude, and access documentation in formats optimized for LLM consumption.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/components/PageActions/index.tsx
+++ b/apps/developer-hub/src/components/PageActions/index.tsx
@@ -12,9 +12,6 @@ type PageActionOption = {
   id: string;
   label: string;
   icon: ReactNode;
-  url?: string;
-  type: "copy" | "markdown" | "llm";
-  llmProvider?: "openai" | "claude";
   ariaLabel?: string;
 };
 
@@ -29,33 +26,21 @@ const getPageActionOptions = (): PageActionOption[] => [
     id: "copy-page",
     label: "Copy Page",
     icon: <Copy />,
-    type: "copy",
-    ariaLabel: "Copy page content",
   },
   {
     id: "view-markdown",
     label: "View Markdown",
     icon: <Eye />,
-    type: "markdown",
-    ariaLabel: "View page as Markdown",
   },
   {
     id: "ask-chatgpt",
     label: "Ask in ChatGPT",
     icon: <OpenAiLogo />,
-    url: "https://chat.openai.com",
-    type: "llm",
-    llmProvider: "openai",
-    ariaLabel: "Ask in ChatGPT",
   },
   {
     id: "ask-claude",
     label: "Ask in Claude",
     icon: <ClaudeIcon />,
-    url: "https://claude.ai",
-    type: "llm",
-    llmProvider: "claude",
-    ariaLabel: "Ask in Claude",
   },
 ];
 
@@ -74,8 +59,7 @@ export function PageActions({ content, title, url }: PageActionsProps) {
   }
 
   function handleShare(option: PageActionOption) {
-    if (option.type === "llm" && option.url) {
-      const prompt = `Please read and analyze this documentation page:
+    const prompt = `Please read and analyze this documentation page:
 
         Title: ${title}
         URL: ${url}
@@ -85,27 +69,29 @@ export function PageActions({ content, title, url }: PageActionsProps) {
 
         Please provide a summary and answer any questions I might have about this content.`;
 
-      const encodedInstruction = encodeURIComponent(prompt);
-      const shareUrl =
-        option.llmProvider === "claude"
-          ? `https://claude.ai/new?q=${encodedInstruction}`
-          : `${option.url}?q=${encodedInstruction}`;
+    const encodedInstruction = encodeURIComponent(prompt);
 
+    if (option.id === "ask-claude") {
+      const shareUrl = `https://claude.ai/new?q=${encodedInstruction}`;
+      window.open(shareUrl, "_blank");
+    } else if (option.id === "ask-chatgpt") {
+      const shareUrl = `https://chat.openai.com?q=${encodedInstruction}`;
       window.open(shareUrl, "_blank");
     }
   }
 
   function handleActionClick(option: PageActionOption) {
-    switch (option.type) {
-      case "copy": {
+    switch (option.id) {
+      case "copy-page": {
         copy();
         break;
       }
-      case "markdown": {
+      case "view-markdown": {
         handleViewMarkdown();
         break;
       }
-      case "llm": {
+      case "ask-chatgpt":
+      case "ask-claude": {
         handleShare(option);
         break;
       }
@@ -117,7 +103,7 @@ export function PageActions({ content, title, url }: PageActionsProps) {
       <div className={styles.container}>
         {pageActionOptions.map((option, index) => {
           const isLast = index === pageActionOptions.length - 1;
-          const isCopyAction = option.type === "copy";
+          const isCopyAction = option.id === "copy-page";
           const showCheckIcon = isCopyAction && isCopied;
 
           return (


apps/developer-hub/src/app/llms.txt/route.ts
```diff
@@ -0,0 +1,38 @@
...
+      headers: {
+        "Content-Type": "text/plain; charset=utf-8",
+        "Cache-Control": "public, max-age=3600", // Cache for 1 hour
+      },
+    });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error("LLM Text route error:", error);
+    return new NextResponse("Internal server error", { status: 500 });
+  }
```
alexcambose: you actually don't need the try catch here as NextJS will automatically throw a 500 if something fails here.

apps/developer-hub/src/app/mdx/[...slug]/route.ts
```diff
@@ -0,0 +1,32 @@
...
+      headers: {
+        "Content-Type": "text/plain; charset=utf-8",
+        "Cache-Control": "public, max-age=3600", // Cache for 1 hour
+      },
+    });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error("MDX route error:", error);
+    return new NextResponse("Internal server error", { status: 500 });
+  }
```
Comment (alexcambose) [inline apps/developer-hub/src/app/mdx/[...slug]/route.ts]: same as https://github.com/pyth-network/pyth-crosschain/pull/3171/files#r2486552411

apps/developer-hub/src/components/LLMShare/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    name: "Ask Claude",
+    url: "https://claude.ai",
+    icon: ClaudeIcon,
+    type: "llm",
+  },
+];
+
+export function LLMShare({ content, title, url }: LLMShareProps) {
+  const [copiedStates, setCopiedStates] = useState<Record<string, boolean>>({});
+  const [selectedKey, setSelectedKey] = useState<string>("");
```
alexcambose: are we actually using selectedKey for something? seems we just use the dropdown stateless as a one time click.

apps/developer-hub/src/components/Pages/BasePage/index.tsx
```diff
@@ -6,18 +6,26 @@ import {
...
 
   const MDX = page.data.body;
+  const content = await getLLMText(page);
+  const title = page.data.title;
+  const url = page.url;
 
   return (
     <DocsPage toc={page.data.toc} full={page.data.full}>
-      <DocsTitle>{page.data.title}</DocsTitle>
+      <div className="flex items-center justify-between gap-4">
```
alexcambose: can we please use scss instead of tailwind

apps/developer-hub/src/components/LLMShare/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const logger = useLogger();
+  const dropdownOptions = getDropdownOptions();
+
+  async function handleCopy(key: string) {
```
alexcambose: I just did this now (you have to merge main here) but you can use `import { useCopy } from "@pythnetwork/component-library/useCopy";`
and then
`const { isCopied, copy } = useCopy(text)`

apps/developer-hub/src/components/LLMShare/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    id: "claude",
+    name: "Ask Claude",
+    url: "https://claude.ai",
+    icon: ClaudeIcon,
+    type: "llm",
+  },
+];
+
+export function LLMShare({ content, title, url }: LLMShareProps) {
+  const [copiedStates, setCopiedStates] = useState<Record<string, boolean>>({});
```
alexcambose: we have just 1 element with the copied state, not sure why we need the ability to manage multiple states in  an object here. But see this first: https://github.com/pyth-network/pyth-crosschain/pull/3171/files#r2486958147

apps/developer-hub/src/components/LLMShare/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+      window.open(shareUrl, "_blank");
+    }
+  }
+
+  function handleSelectionChange(newKey: string | number) {
+    setSelectedKey(String(newKey));
+    const option = dropdownOptions.find((o) => o.id === newKey);
+    if (option) {
+      if (option.type === "markdown") {
+        if (option.id === "view-markdown") {
```
alexcambose: we don't need all these nested if statements, just simple checks on id should be enough. eg: `    if(option?.id === "view-markdown")`

apps/developer-hub/src/components/LLMShare/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  content: string;
+  title: string;
+  url: string;
+};
+
+type DropdownOption = {
+  id: string;
+  name: string;
+  url?: string;
+  icon: React.ComponentType;
```
alexcambose: it'd be easier to make this ReactNode and `icon: <Eye/>`, and just render it as a simple var: `{icon}`

apps/developer-hub/src/components/LLMShare/index.tsx
aditya520: updated the code.

apps/developer-hub/src/components/PageActions/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+        URL: ${url}
+
+        Content:
+        ${content}
+
+        Please provide a summary and answer any questions I might have about this content.`;
+
+      const encodedInstruction = encodeURIComponent(prompt);
+      const shareUrl =
+        option.label === "Ask in Claude"
```
alexcambose: would make more sense to change the option type to something like `llm-claude`/`llm-chatgtpt` and check against that rather than comparing the label

apps/developer-hub/src/components/PageActions/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    return {
+      id: option.id,
+      label: option.label,
+      icon: option.icon,
+      onClick,
+      ariaLabel: option.ariaLabel ?? option.label,
+    };
+  });
+
```
alexcambose, line 100: instead of generating this map where the click handler is different, you can just have an onClick handler that just checks the clicked option type and executes different logic/functions

apps/developer-hub/src/lib/get-llm-text.ts
```diff
@@ -1,23 +1,37 @@
...
 
-import { source } from "./source";
-
 const processor = remark().use(remarkMdx).use(remarkInclude).use(remarkGfm);
 
-export async function getLLMText(page: InferPageType<typeof source>) {
+function resolveMdxPath(page: Page): string {
+  const parts: string[] = Array.isArray(page.path) ? page.path : [page.path];
+
+  const appRelPath = [process.cwd(), "content", "docs", ...parts] as [
```
alexcambose: looking at the path type this should always be a `string` right? so reducing it to just doing this should work?
```suggestion
function resolveMdxPath(page: Page): string {
  const appRelPath = [process.cwd(), "content", "docs", page.path] as [
```
I could also be wrong if the expected value doesn't match the type though, but just seems simpler

alexcambose (commented):
Looks good, some minor nits

apps/developer-hub/src/components/PageActions/index.tsx
```diff
@@ -14,14 +14,7 @@ type PageActionOption = {
...
   type: "copy" | "markdown" | "llm";
-  ariaLabel?: string;
-};
-
-type PageAction = {
-  id: string;
-  label: string;
-  icon: ReactNode;
-  onClick: () => void;
+  llmProvider?: "openai" | "claude";
```
alexcambose: nit looked more at the logic here, the way I would do this is have the object like:
(note I removed  type/llmProvider/url/make ariaLabel optional and default to the `label` if not defined in code)
```
[
...,
{
    id: "ask-chatgpt",
    label: "Ask in ChatGPT",
    icon: <OpenAiLogo />,
  },
  {
    id: "ask-claude",
    label: "Ask in Claude",
    icon: <ClaudeIcon />,
  },
]
```
and just switch the onClick logic based on id. The rest of the keys `type/llmProvider/url` don't add much value tbh and are just business logic rather than option config

alexcambose (approved):
last comment but looks good overall

PR #3179, opened 2025-11-06
chore: performed a minor bump all packages, since all were changed in the big dual package update

benduran:
## Summary

performed a minor semver bump of all packages

## Rationale

no logic changes were made to the packages, but adding dual publishing does change the output format, and technically the SemVer spec indicates that minor bumps or higher can carry with them breaking changes

## How has this been tested?

N / A. No code changes here, just package.json manifest changes

### DIFF
--- a/apps/api-reference/package.json
+++ b/apps/api-reference/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/api-reference",
-  "version": "0.1.0",
+  "version": "0.0.0",
   "private": true,
   "type": "module",
   "engines": {
--- a/apps/developer-hub/package.json
+++ b/apps/developer-hub/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/developer-hub",
-  "version": "0.1.0",
+  "version": "0.0.0",
   "private": true,
   "type": "module",
   "engines": {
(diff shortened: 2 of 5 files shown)


apps/api-reference/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/api-reference",
-  "version": "0.0.0",
+  "version": "0.1.0",
```
cprussin: No need to version the apps, the version field is actually only here to appease pnpm but is unused in vercel obviously

apps/entropy-tester/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/entropy-tester",
-  "version": "1.2.0",
+  "version": "1.3.0",
```
cprussin, line 3: Although this one does need to be versioned since this isn't a vercel app so please retain this change

apps/hermes/client/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pythnetwork/hermes-client",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
cprussin, line 3: Same for this one (this shouldn't even be in the `apps` folder lol)

PR #3186, opened 2025-11-11
fix(pyth-lazer-agent): Only send one success response per batch update

merolish:
## Summary

The jrpc handler is a bit tricky in terms of result chaining and sending responses. This is I think a sane refactor to handle batch updates in those cases where the publisher wants a response.

## Rationale

MIAX testing feedback.

## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/pyth-lazer-agent/src/jrpc_handle.rs
+++ b/apps/pyth-lazer-agent/src/jrpc_handle.rs
@@ -103,22 +103,36 @@ async fn handle_jrpc_inner<T: AsyncRead + AsyncWrite + Unpin>(
     match serde_json::from_slice::<PythLazerAgentJrpcV1>(receive_buf.as_slice()) {
         Ok(jrpc_request) => match jrpc_request.params {
             JrpcCall::PushUpdate(request_params) => {
-                handle_push_update(
-                    sender,
-                    lazer_publisher,
-                    request_params,
-                    jrpc_request.id.clone(),
-                )
-                .await
+                match lazer_publisher
+                    .push_feed_update(request_params.clone().into())
+                    .await
+                {
+                    Ok(()) => send_update_success_response(sender, jrpc_request.id).await,
+                    Err(err) => {
+                        send_update_failure_response(sender, request_params, jrpc_request.id, err)
+                            .await
+                    }
+                }
             }
-            JrpcCall::PushUpdates(request_params) => {
-                handle_batch_push_update(
-                    sender,
-                    lazer_publisher,
-                    &request_params,
-                    jrpc_request.id.clone(),
-                )
-                .await
+            JrpcCall::PushUpdates(request_params_batch) => {
+                for request_params in request_params_batch {
+                    match lazer_publisher
+                        .push_feed_update(request_params.clone().into())
+                        .await
+                    {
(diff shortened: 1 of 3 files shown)


apps/pyth-lazer-agent/src/jrpc_handle.rs
```diff
@@ -240,6 +249,33 @@ async fn handle_push_update<T: AsyncRead + AsyncWrite + Unpin>(
...
+            send_json(
+                sender,
+                &JrpcErrorResponse {
+                    jsonrpc: JsonRpcVersion::V2,
+                    error: JrpcError::SendUpdateError(request_params.clone()).into(),
+                    id: request_id.clone(),
+                },
+            )
+            .await?;
+            anyhow::bail!("Error processing batch update: {:?}", err);
```
Riateche: I think this `bail!` may not be needed here. It's not present in `handle_push_update`. Looks like it will result in client receiving two error responses.

apps/pyth-lazer-agent/src/jrpc_handle.rs
```diff
@@ -240,6 +249,33 @@ async fn handle_push_update<T: AsyncRead + AsyncWrite + Unpin>(
...
+            debug!("error while sending updates: {:?}", err);
+            send_json(
+                sender,
+                &JrpcErrorResponse {
+                    jsonrpc: JsonRpcVersion::V2,
+                    error: JrpcError::SendUpdateError(request_params.clone()).into(),
+                    id: request_id.clone(),
+                },
+            )
+            .await?;
```
Riateche: This whole body of the `for` loop can be moved out into a function to avoid code duplication with `handle_push_update`.

apps/pyth-lazer-agent/src/jrpc_handle.rs
```diff
@@ -240,6 +249,33 @@ async fn handle_push_update<T: AsyncRead + AsyncWrite + Unpin>(
     }
 }
 
+async fn handle_batch_push_update<T: AsyncRead + AsyncWrite + Unpin>(
+    sender: &mut Sender<T>,
+    lazer_publisher: &LazerPublisher,
+    batch_request: &[FeedUpdateParams],
+    request_id: JrpcId,
+) -> anyhow::Result<()> {
+    for request_params in batch_request.iter() {
```
Riateche:
```suggestion
    for request_params in batch_request {
```

Riateche (dismissed):

apps/pyth-lazer-agent/src/jrpc_handle.rs
merolish: Sure, just need to come up with a distinction between "this update failed to send" and "sending the error response itself failed" which has a distinction here. Errors upon errors.

apps/pyth-lazer-agent/src/jrpc_handle.rs
merolish: Ah, good point.

apps/pyth-lazer-agent/src/jrpc_handle.rs
merolish: Refactored.

PR #3191, opened 2025-11-12
feat(dev-hub) Pyth Pro pages Part-1

aditya520:
## How has this been tested?

- [ ] Added new tests

### DIFF
--- a/apps/developer-hub/src/components/PriceFeedIdsProTable/index.tsx
+++ b/apps/developer-hub/src/components/PriceFeedIdsProTable/index.tsx
@@ -26,12 +26,12 @@ export const PriceFeedIdsProTable = () => {
   }, []);
 
   const columns: ColumnConfig<Col>[] = [
-    { id: "Asset Type", name: "Asset Type", isRowHeader: true },
-    { id: "Description", name: "Description" },
-    { id: "Name", name: "Name" },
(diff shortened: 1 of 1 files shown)


cprussin (approved):

apps/developer-hub/src/components/PriceFeedIdsProTable/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+        setState(State.Failed(error));
+      });
+  }, []);
+
+  const columns: ColumnConfig<Col>[] = [
+    { id: "asset_type", name: "Asset Type" },
+    { id: "description", name: "Description" },
+    { id: "name", name: "Name" },
+    { id: "symbol", name: "Symbol" },
+    { id: "pyth_lazer_id", name: "Pyth Pro ID", isRowHeader: true },
```
alexcambose, line 33: nit maybe ux wise if this is going to be sorted by it should be the first column in the table?

aditya520: I thought about it before. Since the pyth_pro_id starts from 1 and sequentially incremented, it can confused with the index. Sicne folks try to read the symbol and description and then refer to the pro_id, it would be better we keep it on the right side.

PR #3399, opened 2026-01-20
feat(hermes-server): add option to allow reading the guardian set from ethereum

keyvankhademi:
## Summary

This PR adds new config parameters to allow choosing `pythnet` or `ethereum` as a source for fetching the guardian set.

## Rationale

After migrating from pythnet, it will shutdown and will no longer be usable for fetching the guardian set.

## How has this been tested?

- [ ] Added new tests

Manually tested the fetched guardian set to be identical to pythnet's.

### DIFF
--- a/apps/hermes/server/src/network/ethereum.rs
+++ b/apps/hermes/server/src/network/ethereum.rs
@@ -1,0 +1,299 @@
+//! This module provides functionality to fetch the Wormhole guardian set from
+//! Ethereum mainnet.
+//!
+//! Uses `alloy-sol-types` for type-safe ABI encoding/decoding of Wormhole
+//! contract calls.
+
+use {
+    crate::network::wormhole::GuardianSet,
+    alloy_primitives::Address,
+    alloy_sol_types::{sol, SolCall},
+    anyhow::{anyhow, Context, Result},
+    serde::{Deserialize, Serialize},
+    serde_json::json,
+};
+
+// Define the Wormhole contract interface using alloy's sol! macro
+sol! {
+    struct WormholeGuardianSet {
+        address[] keys;
+        uint32 expirationTime;
+    }
+
+    function getCurrentGuardianSetIndex() external view returns (uint32);
+
+    function getGuardianSet(uint32 index) external view returns (WormholeGuardianSet);
+}
+
+/// JSON-RPC request structure
+#[derive(Serialize)]
+struct JsonRpcRequest {
+    jsonrpc: &'static str,
+    method: &'static str,
+    params: serde_json::Value,
+    id: u64,
+}
+
+/// JSON-RPC response structure
+#[derive(Deserialize)]
+struct JsonRpcResponse {
+    result: Option<String>,
+    error: Option<JsonRpcError>,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcError {
+    code: i64,
+    message: String,
+}
+
+/// Makes an eth_call to a contract and returns the raw bytes
+async fn eth_call(
+    client: &reqwest::Client,
+    rpc_url: &str,
+    to: &str,
+    data: &[u8],
+) -> Result<Vec<u8>> {
+    let request = JsonRpcRequest {
+        jsonrpc: "2.0",
+        method: "eth_call",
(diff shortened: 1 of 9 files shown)


apps/hermes/server/src/network/ethereum.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonRpcError {
+    code: i64,
+    message: String,
+}
+
+/// Makes an eth_call to a contract and returns the raw bytes
+async fn eth_call(
```
ali-behjati: i'm surprised that alloy doesn't have a function for this :?

apps/hermes/server/src/network/pythnet.rs
```diff
@@ -234,6 +238,53 @@ where
...
+            .await
+        }
+        GuardianSetSource::Ethereum => {
+            tracing::debug!("Fetching guardian sets from Ethereum...");
+            let ethereum_rpc_addr = opts.wormhole.ethereum_rpc_addr.as_ref().ok_or_else(|| {
+                anyhow!(
+                    "Ethereum RPC address is required when guardian set source is 'ethereum'. \
+                         Set --wormhole-ethereum-rpc-addr or WORMHOLE_ETHEREUM_RPC_ADDR."
+                )
+            })?;
```
ali-behjati: errr i don't like doing validation like this here. isn't there any other way [like how does an enum work on clippy?]

ali-behjati (commented):
I have two concerns here, first is to whether we need to keep Pythnet or not? (i think we can get rid of it). 

Second is around the ETH contract we stick to here. the reason is that the WH main ETH is the trigger for the guardian transition and we are polling it every 60s and this means we can incur one minute of downtime when it happens. We might need to change it to another ETH contract [likely our custom one deployed in ETH or some other network]. 

p.s: In the future when we implement proper governance in Lazer, Lazer should become the source though.

apps/hermes/server/src/network/ethereum.rs
keyvankhademi: alloy has eth_call but I couldn't install full alloy because of a dependency issue with pinned solana sdk versions. Pavel is looking to see if there is a workaround.

keyvankhademi: Done.

apps/hermes/server/src/network/pythnet.rs
keyvankhademi: N/A anymore

ali-behjati (approved):
Nice!

apps/hermes/server/src/network/ethereum.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// Fetches the current and previous guardian sets from Ethereum.
+///
+/// This function mirrors the behavior of `fetch_guardian_sets_from_pythnet` in pythnet.rs
+/// but sources the data from Ethereum instead.
+pub async fn fetch_guardian_sets_from_ethereum(
+    ethereum_rpc_url: &str,
+    wormhole_contract_addr: &str,
+) -> Result<(u32, GuardianSet, Option<(u32, GuardianSet)>)> {
```
Riateche, line 34: It's unclear what's u32. Maybe replace the tuple with a struct with two fields?

apps/hermes/server/src/network/pythnet.rs
```diff
@@ -183,52 +111,25 @@ where
...
-            %previous,
-            "Retrieved Previous GuardianSet.",
-        );
+    // Update current guardian set
+    Wormhole::update_guardian_set(&*state, current_index, current_set).await;
 
-        Wormhole::update_guardian_set(&*state, bridge.guardian_set_index - 1, previous).await;
+    // Update previous guardian set if available
+    if let Some((prev_index, prev_set)) = previous {
+        Wormhole::update_guardian_set(&*state, prev_index, prev_set).await;
```
Riateche, line 132: I think it should be a single atomic update of both current and previous sets.

Riateche (approved):
Added some minor suggestions.

PR #3499, opened 2026-02-24
Pyth Lazer for Cardano

matej-douro:
## Summary

This is a big PR, merging Cardano work that was previously in a separate branch to `main`. There's more to do when it comes to SDK and building transactions off-chain, but contract itself should be in a place where the design is set and implemented.

## How has this been tested?

- [X] Added new tests

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3499" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


governance/xc_admin/packages/xc_admin_common/src/governance_payload/UpdateTrustedSigner.ts
```diff
@@ -42,3 +42,43 @@ export class UpdateTrustedSigner264Bit extends PythGovernanceActionImpl {
...
+
+  static decode(data: Buffer): UpdateTrustedSigner264Bit | undefined {
+    const decoded = PythGovernanceActionImpl.decodeWithPayload(
+      data,
+      "UpdateTrustedSigner",
+      this.layout,
+    );
+    if (!decoded) return undefined;
+
+    return new UpdateTrustedSigner264Bit(
```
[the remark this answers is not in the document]
matej-douro: Forgot `.toBeTruthy()` in tests 🤦.

lazer/contracts/cardano/validators/pyth_state.ak
```diff
@@ -0,0 +1,163 @@
...
+    },
+  )
+}
+
+pub type Action {
+  GovernanceAction(PreparedVAA)
+  PurgeExpiredWithdrawScripts
+}
+
+validator init(origin: OutputReference, home: ScriptHash) {
```
azellers, line 24: I believe this approach to minting an NFT ensures that it can be minted at most once, because `origin` is no longer available after the first mint(). This is good, because it means we cannot accidentally mint 2 copies of our supposedly unique NFT. But it also means if we ever accidentally burn the NFT, we have to redeploy everything and get all users to switch to the new NFT set. Let's make sure we're comfortable with that.

lazer/contracts/cardano/lib/state.ak
```diff
@@ -0,0 +1,163 @@
...
+  // output is locked to a script to enforce valid datum updates
+  expect Script(home) = output.address.payment_credential
+  // output should initialize the state
+  expect InlineDatum(data) = output.datum
+  with_initial(State { home, data })
+}
+
+fn burn_state(config: StateConfig, id: PolicyId, self: Transaction) {
+  // state and owner NFTs should be burned in lockstep
+  (assets.quantity_of(self.mint, id, config.owner_nft) == -1)?
```
azellers: If I understand correctly, this will only allow the owner of the `owner_nft` to burn our tokens. But how is that controlled? Is that just owned by whoever deployed them in the first place? 

Would it be possible to require governance for burns (and if so, should we)?

lazer/contracts/cardano/sdk/js/src/transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+    withdraw_script: new Uint8Array(),
+  });
+
+  return {
+    policy_id: PolicyId.fromBytes(minter.hash.hash),
+    tx: ctx.client
+      .newTx()
+      .attachScript(minter)
+      .mintAssets(
+        wormholeStateMint.mint(Assets.merge(stateNFT, ownerNFT), "Never"),
```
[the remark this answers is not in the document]
matej-douro: They use the same redeemer ATM, but obviously a typo, fixed.

lazer/contracts/cardano/lib/state.ak
matej-douro: Yes - but anyone can take our contract source and deploy it themselves anyway, so it's more about establishing trust in a specific deployment done by us.

I am leaning towards no, because it's a liveness issue, not a safety one, and our Cardano integration can only be as good as API used to get its inputs. Being able to burn it immediately makes it possible to quickly respond to incidents that couldn't be solved by governance in time.

Burning is really more of a practical measure - we could disallow burning completely, and simply notify users about new deployment/SDK to be used, but this lets us force them to stop using old deployment, and to reclaim ADA that would otherwise get permanently locked.

azellers (approved):

PR #3572, opened 2026-03-26
init(hyperliquid-recorder): Rust service that streams HL L2+trades data into CH

tejasbadadare:
A service that listens to the [StreamL2Book](https://www.quicknode.com/docs/hyperliquid/grpc-api/StreamL2Book) and [StreamData](https://www.quicknode.com/docs/hyperliquid/datasets/trades) streams from QuikNode and persists the updates in Clickhouse.
    
Some bells and whistles:
- gRPC / protobuf based integration
- Resilient reconnections
- Deduping of updates at both the app and CH level (ReplacingMergeTree), allows for redundant instances
- Prometheus + Grafana monitoring & alerting
- E2e local stack via docker-compose & Tilt

image
image


apps/hyperliquid-recorder/src/clickhouse.rs
```diff
@@ -0,0 +1,293 @@
...
+
+use crate::{
+    config::ClickHouseTarget,
+    models::{L2Snapshot, TradeRecord},
+};
+
+#[derive(Clone)]
+pub struct ClickHouseClient {
+    http_client: reqwest::Client,
+    target: ClickHouseTarget,
```
matej-douro, line 14: Would it make sense to use `clickhouse` crate to connect to the DB, or is it problematic in some way?

tejasbadadare: Yeah good point, not sure why Cursor decided to built it with an http client instead of the `clickhouse` crate, but it works just fine. 

Will ask it to use `clickhouse` in the next iteration

matej-douro (approved):

ali-behjati (commented):
It probably works but it's very janky in structure and quality. We need to properly refactor it when it gets to prod.

tejasbadadare:
> It probably works but it's very janky in structure and quality. We need to properly refactor it when it gets to prod.

Yeah, Ppanning to just throw it away when we make the actual hyperliquid publisher. We can also leave it on for research purposes.

PR #3603, opened 2026-04-08
feat: add upgrade method to WormholeExecutor contract

jayantk:
Add self-upgrade capability to the wormhole-executor-stellar contract, matching the pattern used by pyth-lazer-stellar. The upgrade method requires auth from the contract's own address, enabling self-governed upgrades via governance VAAs dispatched through execute_governance_action with target_contract set to the executor itself.

Changes:
- Added upgrade(env, new_wasm_hash) method to WormholeExecutor in lib.rs
- Added unit test: test_upgrade_authorized (auth passes, fails at deployer level with fake hash)
- Added unit test: test_upgrade_unauthorized (panics with auth error)
- Added integration test: test_governance_upgrade_dispatched_to_executor (full self-upgrade governance flow)

All existing tests continue to pass (57 unit tests + 19 integration tests).

### DIFF
--- a/lazer/contracts/stellar/contracts/integration-tests/src/test.rs
+++ b/lazer/contracts/stellar/contracts/integration-tests/src/test.rs
@@ -415,20 +415,18 @@ fn test_full_verification_and_payload_parsing() {
 fn test_governance_upgrade_dispatched_to_lazer() {
     let te = setup(1);
 
-    let wasm_digest = [0xAB; 32];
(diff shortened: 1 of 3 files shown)


jayantk:
the tests here are bad. We should actually set up enough mocking so that the upgrade path succeeds in the cases when it should, and we should test that it errors when it should fail (which we aren't doing!)

PR #3625, opened 2026-04-22
fix(dev-hub): degrade price feed IDs table when hermes-beta is down

aditya520:
Switch `Promise.all` → `Promise.allSettled` in `PriceFeedIdsCoreTable` so a `hermes-beta` outage no longer kills the whole table. Stable + Solana columns still render; Beta column stays empty on failure.

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3625" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


tejasbadadare (approved):
It works but its a bit hacky, it doesn't handle the case where hermes is down but hermes-beta is up. Better would be to allow both data sources to produce something like `Optional<Feed>` (maybe this happens in `getFeedsFromHermes` for better separation of concerns), and the rendering code can display blanks if the data is None.

PR #3626, opened 2026-04-22
fix(dev-hub): unbreak playground price feeds, format toggle, API key copy, and redact key from snippets

aditya520:
## Summary

- **Price Feeds panel:** swapped symbols URL to the CORS-enabled `history.pyth-lazer.dourolabs.app/history/v1/symbols`. The panel now renders the full feed catalog instead of "Failed to fetch".
- **Format toggle (JSON / Binary):** added `isolation: isolate` to `SingleToggleGroup`'s toggle button. The selected-state bubble had `z-index: -1` but no ancestor stacking context, so it was painting behind the page. Fixed in both dark and light mode.
- **API key copy:** rewrote the hint to invite paste-your-own-key for unlimited streams, kept rate-limit disclosure for the shared demo token, and added a "never stored" reassurance only when a user key is in use. Bumped the hint from 11px to 14px.
- **API key redaction in code snippets:** all four code generators (TypeScript, Python, Go, CLI) used to interpolate `config.accessToken` verbatim into the Monaco editor, leaking the user's key on video calls and screenshares even with `type="password"` on the input. Now each snippet reads the key from a `LAZER_TOKEN` env var instead — same convention the rest of the monorepo uses. Run is unaffected (it authenticates through `/api/playground/stream` using React state, not the rendered snippet).

## Test plan

- [ ] Load `/playground` — Price Feeds panel shows chips (BTC/USD, ETH/USD, …) with accurate category counts.
- [ ] Toggle Format between JSON and Binary in both dark and light themes — selected pill is visible; selected label is readable.
- [ ] Leave API key blank — hint reads "Using a shared demo token — rate-limited for fair use…".
- [ ] Paste any string into API key — hint flips to "Using your API key — no rate limits. Sent only to authorize your stream; never stored."
- [ ] Paste a recognizable fake key (e.g. `sk_live_SECRETKEY_xyz`) and switch through all four language tabs — none of them should contain the key. Each should reference `LAZER_TOKEN`.
- [ ] Copy / Download a snippet — downloaded file references `LAZER_TOKEN`, no raw key inside.
- [ ] Hit Run with the demo token — stream connects and messages arrive.

### DIFF
--- a/apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
+++ b/apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
@@ -48,7 +48,7 @@ export function AccessTokenInput({ className }: AccessTokenInputProps) {
             onChange={(event) => {
               updateConfig({ accessToken: event.target.value });
             }}
-            placeholder="Paste your API key for unlimited streams (optional)"
+            placeholder="Leave empty to use demo token (rate limited)"
             type="password"
             value={config.accessToken}
           />
@@ -56,8 +56,8 @@ export function AccessTokenInput({ className }: AccessTokenInputProps) {
 
         <span className={styles.hint}>
           {isUsingDemoToken
-            ? "Using a shared demo token — rate-limited for fair use. Paste your own API key for unlimited streams."
-            : "Using your API key — no rate limits. Sent only to authorize your stream; never stored."}
+            ? "Using a shared demo token is rate-limited for fair use. Paste your own API key for unlimited streams."
+            : "Using your API key. Sent only to authorize your stream and never stored."}
         </span>
       </div>
     </div>
--- a/apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
+++ b/apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
@@ -48,17 +48,17 @@ export function AccessTokenInput({ className }: AccessTokenInputProps) {
             onChange={(event) => {
               updateConfig({ accessToken: event.target.value });
             }}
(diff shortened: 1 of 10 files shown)


apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
```diff
@@ -48,17 +48,17 @@ export function AccessTokenInput({ className }: AccessTokenInputProps) {
             onChange={(event) => {
               updateConfig({ accessToken: event.target.value });
             }}
-            placeholder="Leave empty to use demo token (rate limited)"
+            placeholder="Paste your API key for unlimited streams (optional)"
```
benduran: nit: I like the old grammar, instead, especially since somebody can have an API token on the Free tier, which is rate limited

apps/developer-hub/src/components/Playground/AccessTokenInput/index.tsx
```diff
@@ -48,17 +48,17 @@ export function AccessTokenInput({ className }: AccessTokenInputProps) {
...
         </div>
 
-        {isUsingDemoToken && (
-          <span className={styles.hint}>
-            Using demo token (rate limited for testing)
-          </span>
-        )}
+        <span className={styles.hint}>
+          {isUsingDemoToken
+            ? "Using a shared demo token — rate-limited for fair use. Paste your own API key for unlimited streams."
```
benduran: nit: em-dash signals this was probably written by an AI. Also, it's an unnecessary mid-sentence break. Can just do the following:
`Using a shared demo token is rate-limited for fair use. Paste your own API key for unlimited streams.`

benduran (approved):

PR #3635, opened 2026-04-29
feat: add lazer hermes evm deployment

keyvankhademi:
## Summary

add lazer hermes evm deployment

Add a new wormhole contract version to support 1/2 + 1 guardians
Update contract manager to be able to deploy pyth contracts for lazer hermes

## How has this been tested?

- [ ] Added new tests

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3635" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/contract_manager/scripts/common.ts
+++ b/contract_manager/scripts/common.ts
@@ -17,17 +17,18 @@ import Web3 from "web3";
 import type { Contract } from "web3-eth-contract";
 import type { InferredOptionType } from "yargs";
 
-import type { PrivateKey } from "../src/core/base";
+import type { DeploymentType, PrivateKey } from "../src/core/base";
 import { getDefaultDeploymentConfig } from "../src/core/base";
 import { EvmChain } from "../src/core/chains";
+import type { EvmEntropyContract } from "../src/core/contracts";
 import {
-  EvmEntropyContract,
   EvmExecutorContract,
   EvmWormholeContract,
 } from "../src/core/contracts";
 import { DefaultStore } from "../src/node/utils/store";
 
 export type BaseDeployConfig = {
+  type: DeploymentType;
   gasMultiplier: number;
   gasPriceMultiplier: number;
   jsonOutputDir: string;
@@ -47,7 +48,7 @@ export async function deployIfNotCached(
   cacheKey?: string,
 ): Promise<string> {
   const runIfNotCached = makeCacheFunction(cacheFile);
-  const key = cacheKey ?? `${chain.getId()}-${artifactName}`;
+  const key = cacheKey ?? `${chain.getId()}-${artifactName}-${config.type}`;
   return runIfNotCached(key, async () => {
     const artifact = JSON.parse(
       readFileSync(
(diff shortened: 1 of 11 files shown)


contract_manager/src/core/base.ts
```diff
@@ -133,69 +139,128 @@ export function getDefaultDeploymentConfig(deploymentType: DeploymentType): {
     governanceChainId: number;
     governanceContract: string; // 32 byte address in 64 char hex format
     initialGuardianSet: string[]; // 20 byte addresses in 40 char hex format
+    // Source of the guardian set this receiver follows. Drives the quorum
+    // threshold (wormhole = 2/3+1, lazer = 1/2+1)
+    guardianSetSource: "wormhole" | "lazer";
```
guibescos: I think this field could be renamed to quorum: "two-thirds", "half" to decouple the quorum implementation from the governance and data sources. `initialGuardianSet` already differs between wormhole and lazer

ali-behjati (approved):
Very nice! my only comment is using deployment type on the WH contract too and also trying to softly fill it in for everything by checking the existing guardian set index (some chains are dead so you can't find it, but all EVM ones should be "stable")

contract_manager/src/core/contracts/evm.ts
```diff
@@ -543,19 +553,28 @@ export class EvmPriceFeedContract extends PriceFeedContract {
   constructor(
     public chain: EvmChain,
     public address: string,
+    public deploymentType?: DeploymentType,
```
guibescos, line 560: I don't love allowing this to be undefined.. is it too hard to retroactively tag as beta or stable the existing deployments

contract_manager/src/core/contracts/evm.ts
```diff
@@ -71,18 +71,27 @@ export class EvmWormholeContract extends WormholeContract {
 
   static fromJson(
     chain: Chain,
-    parsed: { type: string; address: string },
+    parsed: {
+      type: string;
+      address: string;
+      guardianSetSource?: "wormhole" | "lazer";
```
guibescos:
```suggestion
      guardianSetSource: "wormhole" | "lazer";
```
why not make this compulsory since after this PR all `EvmWormholeContract` s have this field

contract_manager/src/core/base.ts
```diff
@@ -133,69 +139,128 @@ export function getDefaultDeploymentConfig(deploymentType: DeploymentType): {
...
+      dataSources: [
+        {
+          emitterAddress:
+            "507974686e6574507974686e6574507974686e6574507974686e657450797468",
+          emitterChain: 26,
+        },
+      ],
+      governanceDataSource: {
+        emitterAddress:
+          "0000000000000000000000000000000000000000000000000000000000000004",
```
guibescos: should we set this to the Pythian Council (same as stable)

contract_manager/src/store/contracts/EvmWormholeContracts.json
```diff
@@ -2,961 +2,1165 @@
...
+    "chain": "sepolia",
+    "guardianSetSource": "lazer",
+    "type": "EvmWormholeContract"
+  },
+  {
+    "address": "0x237b7aFF1Af5d9F311f830234792d429355A58F3",
+    "chain": "ethereum",
+    "guardianSetSource": "lazer",
     "type": "EvmWormholeContract"
   }
```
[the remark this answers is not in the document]
guibescos: seems reasonable

guibescos (changes requested):
The governanceDataSource seems wrong. It should be the Pythian Council address

contract_manager/src/core/base.ts
keyvankhademi: it doesn't matter what we put here. We will only have to use this address when generating governance VAAs from router.

contract_manager/src/core/contracts/evm.ts
keyvankhademi: I changed this deploymentType after Ali's comment.

contract_manager/src/core/contracts/evm.ts
keyvankhademi: It would've been hard, needed a lot of manual work. Also some chains are dead. Anyway, we have changed it to deploymentType now.

contract_manager/src/core/base.ts
keyvankhademi: done

contract_manager/scripts/sync_wormhole_guardian_set.ts
```diff
@@ -147,9 +148,16 @@ async function main() {
...
+  const contracts = Object.values(DefaultStore.wormhole_contracts)
+  // Filter chains
+  .filter(
     (contract) => !chains || chains.includes(contract.getChain().getId()),
-  );
+  )
+  // Filter out lazer wormhole contracts
+  .filter((contract) => (
+    !(contract instanceof EvmWormholeContract) || 
+    (contract.deploymentType !== "lazer-staging" && contract.deploymentType !== "lazer-prod")
```
guibescos: feel like it'd read better like 

```typescript
  .filter(
    (contract) =>
      !(contract instanceof EvmWormholeContract) ||
      contract.deploymentType === undefined ||
      contract.deploymentType === "stable" ||
      contract.deploymentType === "beta",
  );
```

guibescos (approved):
I still think we should keep the Pythian Council as the governance emitter address.
When routers start fetching Pythian Council VAAs from wormhole, they will be able to just resign the VAA without changing any contents.

With your emitter address they need to update the emitter address before resigning.

contract_manager/scripts/common.ts
```diff
@@ -106,77 +107,77 @@ export function getWeb3Contract(
...
   },
   chain: {
-    type: "array",
-    string: true,
     desc: "Chains to upgrade the contract on",
+    string: true,
+    type: "array",
+  },
+  testnet: {
+    default: false,
```
Riateche, line 164: Maybe using testnet as default and switching to mainnet with an option would be safer.

contract_manager/scripts/common.ts
```diff
@@ -251,16 +252,23 @@ export function findEntropyContract(chain: EvmChain): EvmEntropyContract {
...
+    deploymentType === "stable" || deploymentType === "beta";
   for (const contract of Object.values(DefaultStore.wormhole_contracts)) {
+    if (!(contract instanceof EvmWormholeContract)) continue;
+    if (contract.getChain().getId() !== chain.getId()) continue;
     if (
-      contract instanceof EvmWormholeContract &&
-      contract.getChain().getId() === chain.getId()
-    ) {
+      contract.deploymentType === deploymentType || 
+      (isCanonicalWormhole && !contract.deploymentType))
```
Riateche, line 270: Is it possible to add `deploymentType` property to all contracts and not handle this special case of missing property in the code?

contract_manager/src/core/contracts/evm.ts
```diff
@@ -134,6 +143,11 @@ export class EvmWormholeContract extends WormholeContract {
       address: this.address,
       chain: this.chain.getId(),
       type: EvmWormholeContract.type,
+      // Only emit deploymentType for entries that have it set; pre-existing
+      // unlabeled entries round-trip unchanged.
+      ...(this.deploymentType !== undefined && {
+        deploymentType: this.deploymentType,
+      }),
```
Riateche, line 150: Can we set this for all of them?

contract_manager/src/core/base.ts
```diff
@@ -9,7 +9,7 @@ export type TxResult = {
   info: unknown; // chain specific info
 };
 
-export type DeploymentType = "stable" | "beta";
+export type DeploymentType = "stable" | "beta" | "lazer-staging" | "lazer-prod";
```
Riateche, line 13: Can you add a comment on what this property means?

Riateche (approved):

contract_manager/scripts/sync_wormhole_guardian_set.ts
keyvankhademi: fixed

contract_manager/scripts/common.ts
keyvankhademi: i didn't add this. I was just moved by formatter. I don't want to change the existing behavior.

contract_manager/scripts/common.ts
keyvankhademi: The problem is that we don't know which deployment was used. We can to go chain by chain and check them manually, or write a script that does this. Some chains also dead. It's a problem for another day.

contract_manager/src/core/contracts/evm.ts
keyvankhademi: same answer as the previous comment. We don't know the deployment types for existing contracts.

contract_manager/src/core/base.ts
keyvankhademi: done

PR #3649, opened 2026-04-30
feat: add integration test (pyth + wormhole core-bridge)

guibescos:
## Summary

When we vendored the SVM core-bridge, it came without any tests.
This PR adds some testing to core-bridge and its integration with pyth-solana-receiver.


guibescos (commented):
the test is failing

keyvankhademi (approved):

PR #3667, opened 2026-05-06
Replace ts-node with tsx in CLI packages for ESM/Node 24 compat

guibescos:
## Summary
- Replace ts-node with tsx in sui/cli, sui/cli-iota, and aptos/cli package.json scripts and dependencies
- Update contract_manager shell script to use tsx (keeping ts-node as a dependency since it is used programmatically in shell.ts)
- Update pnpm-lock.yaml

## Motivation
ts-node v10.9.2 does not support ESM natively and fails with ERR_UNKNOWN_FILE_EXTENSION in ESM packages. Node 24 further breaks ts-node by removing the --experimental-loader API. tsx is already in the workspace catalog and handles ESM natively.

## Test plan
- Verify pnpm install --frozen-lockfile passes
- Verify pnpm run cli works in pyth-sui-cli after building
- Verify no CLI logic changes — only runner and dependency updates

### DIFF
--- a/contract_manager/package.json
+++ b/contract_manager/package.json
@@ -306,7 +306,7 @@
   "scripts": {
     "build": "ts-duality --copyOtherFiles",
     "clean": "rm -rf ./dist",
-    "shell": "ts-node ./src/node/utils/shell.ts"
+    "shell": "tsx ./src/node/utils/shell.ts"
   },
   "ts-duality": {
     "extraExports": {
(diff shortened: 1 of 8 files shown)


guibescos (commented):
It throws 
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/gbescos/Documents/repos/pyth-crosschain/contract_manager/dist/esm/node/utils/preserve-native-fetch' imported from /Users/gbescos/Documents/repos/pyth-crosschain/contract_manager/dist/esm/node/utils/store.mjs

when running pnpm run cli for sui

PR #3692, opened 2026-05-13
fix(price-pusher): provision Loki + Promtail in sample monitoring stack

### LINKED ISSUE #2520: price_pusher - data sources incomplete for Grafana dash
It looks like there is a loki data source that has been omitted in this setup?

these are the default data sources that this comes with, how can we configure for our own pusher?

Image

0xghost42:
### Summary

Closes #2520.

The bundled `grafana-dashboard.sample.json` references a Loki data source (`uid: ads9ouz3jh4hsa`) for three panels — **Tx Hash**, **All Logs**, **Error Logs** — and uses it to populate the `chain` template variable. The sample stack only provisioned Prometheus, so on a fresh `docker-compose up` the dashboard renders with an unconfigured-data-source banner across all log panels and an empty chain selector. This is exactly what the issue reporter screenshot showed.

### Change

Make the sample stack actually run the bundled dashboard as-is.

- `datasource.sample.yml`: add a Loki entry pinned to the UID the dashboard already expects (`ads9ouz3jh4hsa`).
- `docker-compose.metrics.sample.yaml`: add a Loki service and a Promtail sidecar that ships Docker container logs into Loki.
- `promtail.sample.yml` (new): scrapes container logs via the Docker socket and labels each stream with `namespace=<container_name>` — exactly what the dashboard's `{namespace=~"$chain"}` filter selects on. JSON pino logs are parsed so the **Error Logs** panel's `detected_level = error` matcher works against pino's `level` field.
- `README.md`: update the *Running Locally with Docker* section to describe the new services and the one-container-per-chain pattern the dashboard expects.

### Verification

- `docker-compose -f docker-compose.metrics.sample.yaml config` is valid YAML and resolves all four services.
- `promtail -check-config -config.file=promtail.sample.yml` parses cleanly (`docker_sd_configs`, `relabel_configs`, `pipeline_stages` all accepted).
- `pnpm turbo fix --filter @pythnetwork/price-pusher` is clean (biome formatting only touches built artefacts).
- On a fresh `docker-compose -f docker-compose.metrics.sample.yaml up`, the imported dashboard now lists every pusher container's name in the `chain` selector and the log panels populate as soon as the pusher emits a line.

### Out of scope

I kept the Loki/Promtail/Grafana versions on `:latest` to match the existing Prometheus/Grafana pin convention in the same file. Happy to pin to specific tags in a follow-up if the team prefers reproducible versions.

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3692" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/price_pusher/README.md
+++ b/apps/price_pusher/README.md
@@ -293,7 +293,7 @@ pnpm run dev evm --config config.evm.mainnet.json --metrics-port 9091
 
 ### Running Locally with Docker
 
-You can run the monitoring stack (Prometheus and Grafana) using the provided docker-compose configuration:
+You can run the monitoring stack (Prometheus, Loki, Promtail, and Grafana) using the provided docker-compose configuration:
 
 1. Use the sample docker-compose file for metrics:
 
@@ -303,9 +303,11 @@ docker-compose -f docker-compose.metrics.sample.yaml up
 
 This will start:
 - Prometheus server on port 9090 with the alerts configured in alerts.sample.yml
+- Loki server on port 3100 (log aggregation)
+- Promtail, configured by promtail.sample.yml to ship every Docker container's logs into Loki and tag them with `namespace=<container_name>`
 - Grafana server on port 3000 with default credentials (admin/admin)
 
-The docker-compose.metrics.sample.yaml file includes a pre-configured Grafana dashboard (see the [Dashboard](#dashboard) section below) that displays all the metrics mentioned above. This dashboard provides monitoring of your price pusher operations with panels for configured feeds, active feeds, wallet balance, update statistics, and error tracking. The dashboard is automatically provisioned when you start the stack with docker-compose.
+The docker-compose.metrics.sample.yaml file includes a pre-configured Grafana dashboard (see the [Dashboard](#dashboard) section below) that displays all the metrics mentioned above plus log panels (Tx Hash, All Logs, Error Logs) backed by Loki. The dashboard is automatically provisioned when you start the stack with docker-compose; the `chain` template variable lists every container name Promtail has seen, so running one pusher container per chain gives you per-chain log views out of the box.
 
 ### Example Grafana Queries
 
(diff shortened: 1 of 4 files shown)


0xghost42:
Pushed `d154baf` addressing the two actionable Devin findings:

- **Pino numeric levels** — added a `template` stage in `promtail.sample.yml` that maps `10/20/30/40/50/60` to `trace/debug/info/warn/error/fatal` before the `labels` stage. Without this the Error Logs panel's `detected_level = error` matcher never fires because pino emits `level=50`, not `level=error`. Good catch — this directly affects whether the PR's own claim about the Error Logs panel holds in practice.
- **Prometheus UID** — set explicit `uid: edryyydtht14wa` on the Prometheus datasource in `datasource.sample.yml` so a fresh provision matches what the dashboard's metric panels hardcode. Same class of fix as the Loki UID this PR already pinned.

I left the `| logfmt` dashboard query issue alone — that lives in `grafana-dashboard.sample.json` (pre-existing) and feels like a separate dashboard cleanup PR; happy to follow up if you'd like it bundled in.

0xghost42:
Acknowledged on the namespace label finding — confirmed it's a pre-existing dashboard assumption: `apps/price_pusher/src/metrics.ts` doesn't emit a `namespace` label on any counter/gauge, and `prometheus.sample.yml`'s relabel block only sets `instance=price_pusher`. The dashboard's `namespace="$chain"` filter was designed for a Kubernetes deployment where the scrape's pod metadata supplies the label, so it works there but not in the docker-compose sample shipped with this repo.

Treating it the same as the `| logfmt` finding — out of scope for this PR (which is provisioning the missing Loki/Promtail layer), but worth a follow-up that either:

1. adds a static `namespace` label per scrape job in `prometheus.sample.yml` (and a matching label per docker-compose service), or
2. rewrites the dashboard's Prometheus panels to filter on `instance` / a different existing label.

Happy to bundle that into a separate dashboard-cleanup PR if you'd like — kept this one narrow so the Loki/Promtail provisioning lands cleanly first.

PR #3720, opened 2026-05-19
feat(price-pusher): emit `namespace` label on all Prometheus metrics

0xghost42:
## Summary

Make the sample Grafana dashboard's `namespace=\$chain` filters actually match emitted series, which they don't today.

## Background

`apps/price_pusher/grafana-dashboard.sample.json` queries every panel with `namespace="\$chain"`, e.g.

```promql
pyth_price_feeds_total{namespace="\$chain"}
pyth_price_last_published_time{namespace="\$chain"}
pyth_price_update_attempts_total{namespace="\$chain", status="success"}
```

But `apps/price_pusher/src/metrics.ts` only set `app="price_pusher"` as the default registry label, and `prometheus.sample.yml` doesn't add `namespace` via relabeling either. The result: every panel in the sample dashboard came up empty on a fresh deployment.

Devin flagged this during the #3692 review and I acknowledged it there as pre-existing dashboard behavior worth fixing in a separate PR. This is that PR.

## Wiring

- new CLI option `--metrics-namespace <name>` in `src/options.ts`, defaulted per chain command (`evm`, `sui`, `aptos`, `solana`) so single-chain deployments work without any configuration
- `PricePusherMetrics` constructor now takes a `namespace: string` parameter and sets it via `registry.setDefaultLabels`, so every existing metric series gains a `namespace` label without per-metric `labelNames` plumbing
- each chain command (`evm/sui/aptos/solana/command.ts`) reads the new arg and passes through to the constructor

Operators running multiple deployments of the same chain (e.g. several EVM networks against one Grafana instance) can now set `--metrics-namespace bsc-mainnet` / `--metrics-namespace polygon-mainnet` to disambiguate. Single-deployment setups keep working unchanged — the dashboard's `\$chain` template var will just match the chain command name out of the box.

Bumps `@pythnetwork/price-pusher` to `10.5.0`.

## Coordination heads-up

[#3689](https://github.com/pyth-network/pyth-crosschain/pull/3689) and [#3703](https://github.com/pyth-network/pyth-crosschain/pull/3703) (both open) also bump to `10.5.0`. Whichever of the three lands second/third will need a trivial rebase to `10.6.0` / `10.7.0`.

## Test plan

- [x] CLI `--help` exposes new `--metrics-namespace` option on each chain
- [x] Default (no `--metrics-namespace`): `curl :9090/metrics` shows `pyth_*{...namespace="evm",...}` (or `sui`/`aptos`/`solana`)
- [x] Override: `--metrics-namespace bsc-mainnet` yields `namespace="bsc-mainnet"` on every emitted series
- [x] Sample Grafana dashboard's `\$chain=\$namespace` panels populate end-to-end against a real pusher
- [ ] Maintainer signoff on the per-chain default values (`evm`/`sui`/`aptos`/`solana`)

Refs #3692.

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3720" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


0xghost42:
Fair point — `namespace` is exactly the label Kubernetes Prometheus service discovery auto-applies to scraped pods, and with the default `honor_labels: false` the operator-side label gets renamed to `exported_namespace` and the dashboard filter silently breaks. With `honor_labels: true` the operator wins but loses the k8s namespace context entirely.

Pushed `b37bd13` renaming everywhere to `chain`, which is what the dashboard's template variable was already named and avoids the collision:

- CLI option `--metrics-namespace` -> `--metrics-chain`
- `PricePusherMetrics(logger, namespace)` -> `PricePusherMetrics(logger, chain)` + inline comment noting why `namespace` was avoided
- `registry.setDefaultLabels({ app, namespace })` -> `setDefaultLabels({ app, chain })`
- All four chain command files (`evm/sui/aptos/solana/command.ts`) updated
- `grafana-dashboard.sample.json`: 14 panel queries (`namespace="$chain"` and `{namespace=~"$chain"}` Loki forms), the two `legendFormat: {{namespace}}` references, and the template variable's source-label query (`label: 'namespace'` -> `label: 'chain'`)

Default behaviour unchanged for single-deployment operators — the CLI defaults the value to the chain command name (`evm`/`sui`/`aptos`/`solana`), so a fresh pusher still works against the sample dashboard out of the box without setting the flag.

0xghost42:
Thanks @devin-ai-integration, both findings make sense — let me handle them separately since one belongs to the paired PR.

**Finding 1 (Grafana table panel)** — fixed in `8188538`. The "Price Feeds List" table's `organize.excludeByName` map already hid the `app` / `instance` / `job` triples produced by the `joinByField` transform; added `chain`, `chain#B`, `chain#C` to that map so the new default label doesn't leak three extra duplicate columns into the table.

**Finding 2 (Loki stream selector / `chain` template variable)** — handled by the paired #3692 Promtail/Loki provisioning PR, not this one. That PR sets up the Promtail scrape config to attach `chain=<chain_name>` as a stream label per container, which is what `{chain=~"$chain"}` and the `chain` template variable bind to. Standalone, this PR only changes Prometheus metric labels; the Loki side has nothing to bind to without #3692 landing first.

Happy to call out the dependency in the PR body if you'd like — figured it was already implied since the two PRs share the issue tag, but I can make it explicit.

PR #3790, opened 2026-06-04
feat(price-pusher): loop-iteration heartbeat metric + drop redundant price_id label

aditya520:
## What

Two related observability changes in `apps/price_pusher`:

1. **Add a liveness heartbeat metric** `pyth_pusher_loop_iterations_total` (counter), incremented once per completed controller loop iteration. A hung-but-not-crashed process becomes detectable as this counter going flat (`increase(pyth_pusher_loop_iterations_total[5m]) == 0`).
2. **Drop the redundant `price_id` label** from all 6 per-feed metrics, keeping `alias`. `price_id` and `alias` are enforced 1:1 unique in `price-config.ts`, so `price_id` carried no extra information while doubling label storage and cluttering Grafana label dropdowns. `pyth_price_update_attempts_total` is now `{alias, trigger, status}`.

## Why

Part of the price-pusher reliability/observability cleanup (PLA-57).

- The heartbeat is the prerequisite for re-pointing the Loki-based "Low number of logs" liveness alert to a metric, and for safely reducing steady-state log noise. There is no k8s liveness probe today, so that alert is the de-facto liveness signal.
- Dropping `price_id` is a fleet-wide ingest/storage byte reduction (~15k per-feed series each shed a ~66-char hex label) and de-noises dashboards.

## Scope / safety

- **No series-count change** — `price_id` was 1:1 with `alias`, so this trims label bytes, not series count. The larger series-count reductions (collapsing `update_attempts` to `{trigger,status}`, removing the static `pyth_configured_time_difference`) are deferred; the latter is coupled to the "Price Feed Not Updated" alert and must land together with a rule edit.
- **Known alerts unaffected** — they aggregate by `namespace,container` and none key on `price_id`.
- **Dashboards:** any panel using `price_id` as a field/legend should switch to `alias` (which remains). The shipped `grafana-dashboard.sample.json` has stale `price_id` transform entries, to be cleaned in a follow-up.

## Verification

`pnpm test:types`, `pnpm build`, and `biome check` all pass.

### DIFF
--- a/apps/price_pusher/src/controller.ts
+++ b/apps/price_pusher/src/controller.ts
@@ -28,6 +28,12 @@ export class Controller {
 
     // Set the number of price feeds if metrics are enabled
     this.metrics?.setPriceFeedsTotal(this.priceConfigs.length);
+
+    // Report unhealthy on /live if no loop iteration completes within a few
+    // pushing cycles (min 60s), so the k8s livenessProbe restarts a hung pod.
+    this.metrics?.setLivenessThresholdSeconds(
+      Math.max(this.pushingFrequency * 5, 60),
+    );
   }
 
   async start() {
--- a/apps/price_pusher/src/controller.ts
+++ b/apps/price_pusher/src/controller.ts
@@ -58,14 +58,12 @@ export class Controller {
 
         if (this.metrics && targetLatestPrice && sourceLatestPrice) {
           this.metrics.updateTimestamps(
-            priceId,
             alias,
             targetLatestPrice.publishTime,
             sourceLatestPrice.publishTime,
             priceConfig.timeDifference,
           );
           this.metrics.updatePriceValues(
-            priceId,
             alias,
             sourceLatestPrice.price,
             targetLatestPrice.price,
@@ -81,7 +79,7 @@ export class Controller {
 
         // Record update condition in metrics
         if (this.metrics) {
-          this.metrics.recordUpdateCondition(priceId, alias, priceShouldUpdate);
+          this.metrics.recordUpdateCondition(alias, priceShouldUpdate);
         }
 
         if (priceShouldUpdate == UpdateCondition.YES) {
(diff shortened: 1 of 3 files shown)


apps/price_pusher/src/controller.ts
```diff
@@ -81,7 +79,7 @@ export class Controller {
 
         // Record update condition in metrics
         if (this.metrics) {
-          this.metrics.recordUpdateCondition(priceId, alias, priceShouldUpdate);
```
tejasbadadare, line 84: Wondering why we're dropping alias and not priceId? And do we use priceId anywhere in the dashes or alerts?

apps/price_pusher/src/metrics.ts
```diff
@@ -88,6 +90,14 @@ export class PricePusherMetrics {
...
     });
 
+    // Liveness heartbeat: increments once per controller loop iteration so a
+    // hung (but not crashed) process is detectable via the metric going flat.
+    this.loopIterations = new Counter({
+      help: "Total number of controller loop iterations completed (liveness heartbeat)",
+      name: "pyth_pusher_loop_iterations_total",
+      registers: [this.registry],
+    });
+
```
tejasbadadare, line 103: I think it's better to expose a liveness endpoint. That way k8s can automatically recycle hung pods and we get a metric on that for free as well

apps/price_pusher/src/controller.ts
aditya520: We are dropping only price Ids, not aliases here.

aditya520: There will be one small dash fix, nothing related to alerts. I will fix the dash once we merge this.

apps/price_pusher/src/metrics.ts
aditya520: Good point.

tejasbadadare (approved):

PR #3825, opened 2026-06-23
@pythnetwork/pyth-sui-js: make SuiPythClient transport-agnostic (JSON-RPC + gRPC)

jayantk:
Resolves [[i-evtxxopo]]. Makes `@pythnetwork/pyth-sui-js` transport-agnostic so consumers can supply either a JSON-RPC client (`SuiJsonRpcClient`) or a gRPC client (`SuiGrpcClient`), both of which expose the `@mysten/sui` v2 unified `.core` API. Prerequisite that unblocks the Sui price pusher ([[i-vumtfomk]]).

## SDK (`target_chains/sui/sdk/js`)
- `@mysten/sui` bumped to `^2.7.0`, moved to `peerDependencies`. Major bump `3.0.0 → 4.0.0`.
- `SuiPythClient` accepts any v2 `.core`-exposing client (`SuiPythClientProvider = ClientWithCoreApi`) and reads exclusively through `.core`.
- **Fix (found via live validation):** `getPriceTableInfo` previously resolved `price_info` as a plain dynamic field, but on-chain it is a dynamic **object** field (it holds a `Table`). It is now resolved through the `0x2::dynamic_object_field::Wrapper<vector<u8>>` name type, decoding the backing table `object::ID` from the field value and reading the table's type for the `PriceIdentifier` package. Without this, `getPriceFeedObjectId` (the pusher hot path) derived a non-existent object id and failed on mainnet.
- Example gains `--transport jsonrpc|grpc` + `--network`; README has a v4 migration section.

## In-repo consumer migration (this iteration)
The new SDK needs a v2 client at runtime, so the three in-repo consumers are moved off the v1 client:
- **price_pusher**: constructs `SuiJsonRpcClient` (retains the legacy JSON-RPC methods these files call AND exposes `.core`); network inferred from the endpoint URL.
- **contract_manager**: `SuiChain.getProvider()` returns `SuiJsonRpcClient` (network from `isMainnet`).
- **target_chains/sui/cli**: type swaps + `fromB64 → fromBase64`. Required — keeping the CLI on v1 fails typecheck (cross-version `SuiJsonRpcClient`/`Transaction` mismatch against the now-v2 contract_manager).
- Consumers use the v2 `exports` map → tsconfig `moduleResolution` bumped `node → bundler` for price_pusher/contract_manager.
- Stale `eslint-disable` headers in the touched files converted to the `biome-ignore-all` headers the repo's linter actually reads (pre-existing latent violations that only surface because these files are now changed); filled in 4 placeholder `biome-ignore` explanations in `chains.ts`.

## Validation
Exercised the real SDK against live endpoints on **mainnet and testnet**, over **both JSON-RPC and gRPC** (public Sui fullnode gRPC-web). `getBaseUpdateFee`, `getPackageId`, `getWormholePackageId`, `getPriceFeedObjectId` and the `updatePriceFeeds` tx-build all succeed with **identical results across transports**. Note: `getPriceFeedObjectId`'s broad not-found `catch` is intentionally left as-is — narrowing by error code is risky given JSON-RPC vs gRPC throw different error shapes for a missing field (reviewer note #1, still deferred).

## Gate
`pnpm turbo build test:types test:unit` for SDK + price-pusher + contract-manager + sui-cli → 38/38 green. `biome ci --changed` → 0 errors. Merge-clean vs `origin/main`. npm publish still deferred to [[i-vumtfomk]].

### DIFF
--- a/apps/price_pusher/package.json
+++ b/apps/price_pusher/package.json
@@ -8,7 +8,7 @@
     "@injectivelabs/networks": "1.14.47",
     "@injectivelabs/sdk-ts": "1.14.50",
     "@injectivelabs/utils": "^1.14.48",
-    "@mysten/sui": "^1.3.0",
+    "@mysten/sui": "^2.7.0",
     "@pythnetwork/hermes-client": "workspace:*",
     "@pythnetwork/price-service-sdk": "workspace:^",
     "@pythnetwork/pyth-fuel-js": "workspace:*",
--- a/apps/price_pusher/src/sui/balance-tracker.ts
+++ b/apps/price_pusher/src/sui/balance-tracker.ts
@@ -1,4 +1,4 @@
-import type { SuiClient } from "@mysten/sui/client";
+import type { SuiJsonRpcClient } from "@mysten/sui/jsonRpc";
 import type { Logger } from "pino";
 
 import type {
@@ -14,14 +14,14 @@ import type { DurationInSeconds } from "../utils.js";
  */
 export type SuiBalanceTrackerConfig = {
   /** Sui client instance */
-  client: SuiClient;
+  client: SuiJsonRpcClient;
 } & BaseBalanceTrackerConfig;
 
 /**
  * Sui-specific implementation of the balance tracker
  */
 export class SuiBalanceTracker extends BaseBalanceTracker {
-  private client: SuiClient;
+  private client: SuiJsonRpcClient;
 
   constructor(config: SuiBalanceTrackerConfig) {
     super({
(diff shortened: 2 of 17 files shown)


tejasbadadare (approved):

John-Atha:
Hey @jayantk and @tejasbadadare, thanks for landing this! The PR notes npm publish of `4.0.0` is deferred to the linked price-pusher issue.

Any rough ETA on that publish?
We're migrating a service to the `@mysten/sui` v2 gRPC client, and the transport-agnostic `SuiPythClient` from this PR is exactly what we need. We'd much rather depend on published `4.0.0` than vendor `client.ts`. Thanks!

PR #3833, opened 2026-06-24
feat(price_pusher): add gRPC support to the Sui driver alongside JSON-RPC

jayantk:
Resolves [[i-vumtfomk]]. Now unblocked: [[i-evtxxopo]] (transport-agnostic `@pythnetwork/pyth-sui-js`, #3825) is merged on `main`, so `SuiPythClient` accepts either a JSON-RPC or gRPC v2 client.

## What

Adds a per-instance `--endpoint-type {json-rpc|grpc}` CLI flag (default `json-rpc`, backwards compatible) and a required `--network {mainnet|testnet|devnet|localnet}` flag to the Sui price pusher, and refactors the three Sui driver files to read and write exclusively through the `@mysten/sui` v2 unified `.core` API. Both `SuiJsonRpcClient` (`@mysten/sui/jsonRpc`) and the experimental `SuiGrpcClient` (`@mysten/sui/grpc`) expose `.core`, so a single code path serves both transports.

## Why

Sui Foundation is decommissioning JSON-RPC: public endpoints off July 2026 (testnet wk of Jul 6, mainnet wk of Jul 20), removed from full nodes by mid-Oct 2026. gRPC is the replacement. This keeps the same binary usable for both transports during the migration window.

## Changes

- `src/sui/command.ts` — new `--endpoint-type` and required `--network` flags; threaded into the provider factory, listener, and pusher.
- `src/sui/sui.ts` —
  - `createSuiProvider(endpointType, network, url)` returns a `ClientWithCoreApi`, constructing `SuiGrpcClient`/`SuiJsonRpcClient` per the flag. The network label is now an explicit parameter (no longer inferred from the URL), and an unknown `endpointType` throws rather than silently falling back to JSON-RPC.
  - All RPC calls moved off the legacy JSON-RPC method shapes onto `.core`. Execution effects are read from the new `TransactionEffects` shape via small `getExecutedTransaction` / `changedObjectToRef` helpers.
  - `sendTransactionBlock` now checks `effects.status.error` after unwrapping the result: the `.core` API returns a `FailedTransaction` for on-chain failures rather than throwing, so failed pushes are surfaced as errors (and the gas object refreshed) instead of being logged as successes.
  - The nested-struct normalizer `getStructFields` is now exported from `@pythnetwork/pyth-sui-js` and imported here, removing the duplicate copy that the SDK already owns.
- `target_chains/sui/sdk/js/{client,index}.ts` — export `getStructFields` from the SDK.
- Sample configs + README — add the `network` field / `--network` flag.

## Review feedback addressed (human-iteration)

1. `createSuiProvider` errors on an unknown `endpointType` instead of defaulting to JSON-RPC.
2. Network is an explicit `--network` parameter, no longer inferred from the endpoint URL.
3. `getStructFields` consolidated into the SDK and re-exported (was a verbatim copy).
4. Failed Sui executions (Devin): `sendTransactionBlock` now checks execution status so failures aren't reported as successful pushes.

## Validation

`pnpm turbo test:types test:unit build --filter=@pythnetwork/price-pusher --filter=@pythnetwork/pyth-sui-js` → 24/24 green; biome clean on changed files. Merge clean against `origin/main`. gRPC path remains type-checked but not exercised against a live Sui gRPC endpoint (none available in this env).

### DIFF
--- a/apps/price_pusher/src/sui/command.ts
+++ b/apps/price_pusher/src/sui/command.ts
@@ -28,8 +28,19 @@ export default {
       description:
         "RPC endpoint URL for sui. The pusher will periodically" +
         "poll for updates. The polling interval is configurable via the " +
-        "`polling-frequency` command-line argument.",
+        "`polling-frequency` command-line argument. For `--endpoint-type grpc` " +
+        "this is the gRPC-web base URL.",
       required: true,
+      type: "string",
+    } as Options,
+    "endpoint-type": {
+      choices: ["json-rpc", "grpc"],
+      default: "json-rpc",
+      description:
+        "Transport to use for the Sui RPC endpoint. Sui Foundation is deprecating " +
+        "JSON-RPC (public endpoints off July 2026, removed by mid-Oct 2026); use " +
+        "`grpc` to migrate. `grpc` relies on @mysten/sui's experimental SuiGrpcClient.",
+      required: false,
       type: "string",
     } as Options,
     "gas-budget": {
@@ -44,6 +55,14 @@ export default {
         "Gas objects to ignore when merging gas objects on startup -- use this for locked objects.",
       required: false,
       type: "array",
+    } as Options,
+    network: {
+      choices: ["mainnet", "testnet", "devnet", "localnet"],
+      description:
+        "Sui network label for the RPC client. This selects the network the " +
+        "endpoint serves; it does not change which `--endpoint` URL is used.",
+      required: true,
+      type: "string",
     } as Options,
     "num-gas-objects": {
       default: 30,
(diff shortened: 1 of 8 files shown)


apps/price_pusher/src/sui/command.ts
```diff
@@ -45,6 +56,14 @@ export default {
       required: false,
       type: "array",
     } as Options,
+    network: {
+      choices: ["mainnet", "testnet", "devnet", "localnet"],
+      description:
+        "Sui network label for the RPC client. This selects the network the " +
+        "endpoint serves; it does not change which `--endpoint` URL is used.",
+      required: true,
```
aditya520, line 64: Won't that break existing pusher deployment?

aditya520 (approved):

apps/price_pusher/src/sui/command.ts
jayantk: we'll have to add --network as a CLI flag to the deployment whenever we release this. we don't have CD for the code though. we have to actually do a price pusher release before the code will get picked up, so nothing will break until then.

PR #3848, opened 2026-06-25
feat(contract_manager): add gRPC support for Sui chains

jayantk:
## What

Adds gRPC support to `contract_manager`'s Sui chain/contract code, alongside the existing JSON-RPC client — the same dual-protocol pattern as the price pusher ([[i-vumtfomk]]). Sui Foundation turns off public JSON-RPC endpoints in July 2026 (removed entirely by mid-Oct 2026), so ops tooling needs a gRPC path.

Resolves [[i-mjqwrtbw]] (child of audit [[i-vurepyya]]).

## Changes

- **`contract_manager/src/core/chains.ts`**
  - `SuiChain` and `IotaChain` gain an `endpointType` field (`json-rpc` | `grpc`), parsed from / serialised to the chain config (defaults to `json-rpc`).
  - `SuiChain.getProvider()` now constructs a `SuiGrpcClient` or `SuiJsonRpcClient` via a `createSuiProvider` factory and returns the unified `ClientWithCoreApi`. All reads/writes go through the transport-agnostic `.core` API (`getBalance`, `getObject`, `simulateTransaction`, `signAndExecuteTransaction`, `waitForTransaction`).
  - Governance/state struct parsing now uses `getStructFields` from `@pythnetwork/pyth-sui-js`, which normalises the JSON shape (it differs between JSON-RPC `{ type, fields }` and gRPC flattened). Removed the now-dead `hasStructField`/`getStateObject` helpers.
  - `publishPackage`/`initLazerContract` derive package/UpgradeCap/State ids from `.core` effects (`changedObjects` + `objectTypes`) instead of JSON-RPC `objectChanges`. Object-type matching uses a type **suffix** to stay robust to address normalisation across transports.
  - A single exported `executeSuiTransaction(provider, tx, keypair)` helper holds the full sign-and-execute path (simulate, 2× gas budget, sign+execute, on-chain error check, `waitForTransaction`). `SuiChain.executeTransaction` and the Sui contract classes all route through it.
- **`contract_manager/src/core/contracts/sui.ts`**
  - `SuiPriceFeedContract` / `SuiWormholeContract` migrated to `.core` (price-feed reads, data-source dynamic field via BCS-encoded name, governance/state reads). They delegate transaction execution to `this.chain.executeTransaction`, so the simulate / 2× gas / error-check / wait logic is shared with `SuiChain` and cannot drift.
- **`target_chains/sui/cli/` (downstream consumer)**
  - `pyth-sui-cli`'s deploy/upgrade helpers (`publishPackage`, `initPyth` in `pyth_deploy.ts`; `upgradePyth`, `migratePyth` in `upgrade_pyth.ts`) consumed `SuiChain.getProvider()`, whose return type changed from `SuiJsonRpcClient` to `ClientWithCoreApi`. That broke the `build` CI check (the `--filter=@pythnetwork/contract-manager` build never compiled this consumer, so it slipped through local verification). Migrated all four helpers onto the same `.core` API via the exported `executeSuiTransaction` helper: ids are now read from `.core` effects (`changedObjects` + `objectTypes`, suffix-matched) and on-chain failure throws (matching contract-manager). The redundant post-execution `effects.status.status` checks in `cli.ts` were removed (the helper now throws); the migrate command preserves its operator-facing "do the migration manually" guidance via a `try/catch`.
- **Store JSON** (`SuiChains.json`, `IotaChains.json`): added `endpointType: "json-rpc"` to every entry.

## Decisions / scope notes

- **IOTA has no gRPC client.** `@iota/iota-sdk` (v1.14.0, range `^1.11.0`) ships only the legacy JSON-RPC `IotaClient` — no `./grpc` export, no `.core` API. So `IotaChain` keeps using `IotaClient` for `json-rpc` and `getProvider()` throws a clear error on `grpc`. The field is added for config-schema symmetry and future readiness; re-check the IOTA SDK before promising IOTA gRPC.
- **No dependency bump needed.** The existing `@mysten/sui: ^2.7.0` range already resolves to a gRPC-shipping 2.x (the price pusher imports `@mysten/sui/grpc` on the same pin).
- **gRPC sibling chain entries deferred.** Per the issue, dropping the hardcoded `fullnode.*.sui.io` URLs / adding gRPC siblings is gated on validated gRPC endpoint URLs from the provider-validation issue, which aren't available yet. Flipping a chain to gRPC is now a one-field config change (`endpointType: "grpc"` + gRPC `rpcUrl`).
- JSON-RPC code paths are unchanged in behavior. One intentional improvement: `executeTransaction` now throws on on-chain execution failure (the `.core` API returns a `FailedTransaction` rather than throwing), matching the price pusher.

## Validation

- `pnpm tsc --project tsconfig.build.json --noEmit`: clean.
- `turbo run build --filter=pyth-sui-cli`: 34/34 successful (this is the build that the CI `build` check failed on; it now passes, exercising both `contract-manager` and the migrated CLI consumer).
- biome: changed files formatted; no new lint violations introduced (the remaining `cli.ts` `noConsole`/`useNodejsImportProtocol`/`useAwait` diagnostics are pre-existing — identical count before and after this change).
- `git merge-tree origin/main HEAD`: clean.
- End-to-end gRPC validation on a live Sui mainnet endpoint is a manual ops step pending a validated gRPC URL.

### DIFF
--- a/contract_manager/src/core/contracts/sui.ts
+++ b/contract_manager/src/core/contracts/sui.ts
@@ -9,10 +9,12 @@
 /* eslint-disable @typescript-eslint/no-base-to-string */
 import { parseVaa } from "@certusone/wormhole-sdk";
 import { uint8ArrayToBCS } from "@certusone/wormhole-sdk/lib/cjs/sui";
+import { bcs } from "@mysten/sui/bcs";
+import type { ClientWithCoreApi } from "@mysten/sui/client";
 import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
 import { Transaction } from "@mysten/sui/transactions";
 import { SUI_CLOCK_OBJECT_ID } from "@mysten/sui/utils";
-import { SuiPythClient } from "@pythnetwork/pyth-sui-js";
+import { getStructFields, SuiPythClient } from "@pythnetwork/pyth-sui-js";
 import type { DataSource } from "@pythnetwork/xc-admin-common";
 import {
   decodeGovernancePayload,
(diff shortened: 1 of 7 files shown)


aditya520 (approved):

contract_manager/src/core/contracts/sui.ts
```diff
@@ -114,57 +127,41 @@ export class SuiPriceFeedContract extends PriceFeedContract {
...
-      price: { fields: { magnitude: string; negative: boolean } };
-      conf: string;
-      timestamp: string;
-    };
-  }) {
-    let expo = priceInfo.fields.expo.fields.magnitude;
-    if (priceInfo.fields.expo.fields.negative) expo = "-" + expo;
-    let price = priceInfo.fields.price.fields.magnitude;
-    if (priceInfo.fields.price.fields.negative) price = "-" + price;
+  private parsePrice(priceInfo: unknown) {
```
aditya520, line 130: Shouldn't we parse this into a generate `Move` type instead of json?

aditya520 (approved):

PR #3863, opened 2026-06-30
PythLazer.sol: add public getters for unreadable trusted-signer state

jayantk:
Resolves [[i-sghnfsil]].

### `lazer/contracts/evm/src/PythLazer.sol`
- Add `getTrustedSigners() external view returns (TrustedSignerInfo[] memory)` — returns only populated slots (skips `pubkey == address(0)`), preserving slot order; length ≤ 100.
- Add `getTrustedSignerExpiry(address) external view returns (uint256)` — reads `trustedSignerToExpiresAtMapping[signer]`, exposing a named function rather than flipping the mapping to `public`.
- Bump `version()` from `"0.1.1"` to `"0.2.0"`.

Both additions are pure view functions with no new state — **storage layout is unchanged** (verified, see below). This is upgrade-safe.

Audit per parent guidance: the only previously-unreadable state was `trustedSigners` (internal, no getter) and `trustedSignerToExpiresAtMapping` (default-internal); both are now covered. `verification_fee` is already `public`, and inherited `owner()` / `proxiableUUID()` are readable. No other gaps.

### `contract_manager/src/core/contracts/evm_abis.ts`
- Added `getTrustedSigners` and `getTrustedSignerExpiry` entries to the hand-written `LAZER_ABI`. All existing entries left intact.

### `lazer/contracts/evm/test/PythLazer.t.sol`
Test names use the file's snake_case convention (matching `test_update_add_signer`):
- `test_get_trusted_signers_empty` — zero-length array on empty state.
- `test_get_trusted_signers_insertion_order` — N signers returned in insertion/slot order with no zero entries.
- `test_get_trusted_signers_removal_and_reuse` — removing a signer (`expiresAt = 0`) frees its slot, which is skipped and then reused by a subsequent add.
- `test_get_trusted_signer_expiry` — `0` for unknown, stored expiry for known, `0` again after removal.
- `test_version` — asserts `"0.2.0"`.

## Verification

- `forge test --ffi` (in `lazer/contracts/evm/`): 20/20 pass — all 18 tests in `PythLazer.t.sol` (incl. the 5 new) and 2 in `PythLazerApi.t.sol`.
- Storage layout unchanged: compared `forge inspect PythLazer storage-layout` before vs. after — byte-identical (slots: `trustedSigners` @0, `verification_fee` @200, `trustedSignerToExpiresAtMapping` @201).
- `prettier --check` clean on the modified `evm_abis.ts`.
- `git merge-tree origin/main HEAD`: no conflict.

## Out of scope (per issue)

Contract deployment/upgrade and switching `EvmLazerContract.getTrustedSigners()` ([[i-hnsfnbfk]]) off storage-slot reads are deferred until all deployed proxies are upgraded.

### DIFF
--- a/lazer/contracts/evm/test/PythLazer.t.sol
+++ b/lazer/contracts/evm/test/PythLazer.t.sol
@@ -42,6 +42,86 @@ contract PythLazerTest is Test {
         assert(!pythLazer.isValidSigner(address(2)));
     }
 
+    function test_get_trusted_signers_empty() public view {
+        PythLazer.TrustedSignerInfo[] memory signers = pythLazer
+            .getTrustedSigners();
+        assertEq(signers.length, 0);
+    }
+
+    function test_get_trusted_signers_insertion_order() public {
+        address signerA = address(10);
+        address signerB = address(11);
+        address signerC = address(12);
+        uint256 expiryA = block.timestamp + 1000;
+        uint256 expiryB = block.timestamp + 2000;
+        uint256 expiryC = block.timestamp + 3000;
+
+        vm.startPrank(owner);
+        pythLazer.updateTrustedSigner(signerA, expiryA);
+        pythLazer.updateTrustedSigner(signerB, expiryB);
+        pythLazer.updateTrustedSigner(signerC, expiryC);
+        vm.stopPrank();
+
+        PythLazer.TrustedSignerInfo[] memory signers = pythLazer
+            .getTrustedSigners();
+        assertEq(signers.length, 3);
+        assertEq(signers[0].pubkey, signerA);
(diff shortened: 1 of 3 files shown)


lazer/contracts/evm/test/PythLazer.t.sol
```diff
@@ -42,6 +42,86 @@ contract PythLazerTest is Test {
         assert(!pythLazer.isValidSigner(address(2)));
     }
 
+    function test_get_trusted_signers_empty() public view {
```
jayantk, line 45: (i don't know why these are snake case given that everything else is camelCase, but that seems to be the convention in this file)

Riateche (approved):

PR #3885, opened 2026-07-09
fix(price_pusher): bound the EVM receipt tracker to stop the getTransactionByHash leak

aditya520:
## What
Replace the EVM pusher's fire-and-forget `waitForTransactionReceipt` with a **bounded, per-nonce, cancellable** `getTransactionReceipt` poller.

## Why
On viem 2.24.3, `waitForTransactionReceipt`'s timeout rejects without tearing down its internal `eth_getTransactionByHash` / block poll, so a tx that never becomes findable leaks a detached poller — one per push, forever. Measured: **~1,157 `getTransactionByHash`/s on Soneium (green), ~197M/day fleet-wide**.

## Fix
- Polls `getTransactionReceipt` only (**zero `getTransactionByHash`**), bounded by a deadline and torn down via `AbortController` — no accumulation.
- **Per-nonce**: same-nonce escalations keep every competing hash watched (whichever lands is logged, incl. the original winning the race); a nonce advance aborts the resolved nonce's stragglers.
- Preserves the `Price update successful` `{hash}` log; no hot-path change (nonce/gas still driven by `getTransactionCount`).
- Timeout/interval are CLI-configurable (`--receipt-wait-timeout` / `--receipt-poll-interval`); defaults reproduce prior behavior.

## Tests
6-scenario suite driving the real `EvmPricePusher`; also verified end-to-end against anvil.

Pairs with deployments#5120 (eRPC hedge-off) — independent, either order.

### DIFF
--- a/apps/price_pusher/package.json
+++ b/apps/price_pusher/package.json
@@ -226,5 +226,5 @@
   },
   "type": "module",
   "types": "./dist/index.d.ts",
-  "version": "11.2.2"
+  "version": "11.3.0"
 }
--- a/apps/price_pusher/src/evm/__tests__/receipt-tracking.test.ts
+++ b/apps/price_pusher/src/evm/__tests__/receipt-tracking.test.ts
@@ -203,73 +203,72 @@ describe("EvmPricePusher receipt tracking (leak fix)", () => {
     expect(chain.calls.getTransactionByHash).toBe(0);
   });
 
-  it("superseded nonce (gas escalation): single-flight, old tracker is cancelled", async () => {
+  it("same-nonce escalation: both hashes stay watched; the OLD tx winning the race is still logged (Codex)", async () => {
+    // On a same-nonce gas escalation the original and repriced tx compete for
+    // the nonce, and a miner may include the OLD one over the replacement. The
+    // original's tracker must NOT be cancelled, or its landing is never logged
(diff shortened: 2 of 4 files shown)


aditya520:
Addressed the automated review in `642c48e43`:

**Codex (P2 — same-nonce race):** Fixed. Receipt tracking is now **per-nonce** instead of single-flight-per-hash. A same-nonce gas escalation now *adds* a tracker rather than aborting the previous one, so both the original and the repriced tx stay watched and whichever lands (including the old tx winning the mempool race) logs `Price update successful`. A nonce *advance* aborts the resolved nonce's stragglers (each does one final lookup on abort, so the landed hash is still logged). Still bounded — at most one nonce tracked at a time and every tracker self-terminates at the deadline. Added two regression tests: old-tx-wins-the-race is logged, and nonce-advance aborts stragglers while still logging the landed tx.

**Devin (`dispose()` not on `IPricePusher`):** Intentionally left off the interface. There's no graceful-shutdown path in the controller today (it loops forever; k8s SIGTERM kills the process, so a pending timer is moot), and adding `dispose()` to `IPricePusher` would force every driver (solana/sui/aptos/near/…) to implement it for no current benefit. `dispose()` remains a lifecycle/test helper. Happy to wire it into a shutdown path as a follow-up if we add one.

tejasbadadare (approved):
Nice one, thanks!

PR #3905, opened 2026-07-17
lazer/iota: Implement contract_manager integration

matej-douro:
## Summary

Adds support for Iota Lazer contract to contract manager, including equivalent of management script modified from Sui version.

## How has this been tested?

- [ ] Added new tests

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3905" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


contract_manager/scripts/manage_iota_lazer_contract.ts
```diff
@@ -0,0 +1,1073 @@
...
+function getIotaNetworkName(chainId: string): string {
+  const network = chainId.replace(/^iota_/, "");
+  if (network === chainId || network.length === 0) {
+    throw new Error(
+      `Cannot derive IOTA network name from chain ID '${chainId}'`,
+    );
+  }
+  return network;
+}
+
```
jayantk, line 184: we usually don't have the contract deployment script in contract manager because it's been difficult to manage the build tooling dependencies. (i note below that something is calling the iota build script somewhere). The usual split is something like a deploy script that sits alongside the contract, and then we have the ongoing management in contract manager.

I don't feel particularly strongly about this -- honestly i would love to have everything controlled by contract manager, but it feels like doing that properly requires some more fancy build tooling. Something like contract artifacts getting built automatically and published somewhere that the contract manager can hook into...

jayantk (approved):
LGTM but please do consider if it makes sense to split out the deployment into a separate script (see inline)

contract_manager/scripts/manage_iota_lazer_contract.ts
```diff
@@ -0,0 +1,1073 @@
...
+async function readIfExists(file: string): Promise<Buffer | undefined> {
+  try {
+    return await readFile(file);
+  } catch (error) {
+    if (isMissingFile(error)) return undefined;
+    throw error;
+  }
+}
```
tejasbadadare, line 54: Mental note to factor out this kind of boilerplate down the line

tejasbadadare (approved):

contract_manager/scripts/manage_iota_lazer_contract.ts
matej-douro: I see - current version of Sui and IOTA scripts sort of thread respective CLIs through, to make it possible to get info about package to other subcommands too, e.g. to propose contract upgrades to a specific hash.

I would love to find some good boundary for this though, fully agree - I've mentioned to Tejas that CM feels a little bit like service manager to me in terms of responsibilities 😅 We have some pretty high-level tasks shared across most contracts, but individual details vary wildly, use different toolchains and dependencies, and trying to thread them through monolithic CM makes it hard to integrate with their ecosystems nicely (e.g. Sui and IOTA package management vs. CM contract store).

PR #3908, opened 2026-07-17
feat(developer-hub): search Pyth Pro feeds by hermes_id / nasdaq_symbol / cmc_id

aditya520:
## Summary

Extends the Pyth Pro **Price Feed IDs** page search box so users can find a feed by pasting any of the three additional identifiers that the Symbology API returns per feed:

- `hermes_id` (nullable hex string) — pasting `e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43` now surfaces the BTC/USD row, and the `0x`-prefixed form works too.
- `nasdaq_symbol` (nullable) — pasting `BKNG` surfaces the BKNG equity feed.
- `cmc_id` (nullable numeric) — pasting a CoinMarketCap ID matches. The existing exact-`pyth_lazer_id` priority still wins ties.

## Changes

- `PriceFeedIdsProTable/index.tsx`: extend `pythProSchema` with the three new nullable/optional identifier fields; replace the inline search predicate with an imported `filterFeedsBySearch`; update the `SearchInput` placeholder to `"Search by symbol, name, ID, or hex feed ID"`.
- `PriceFeedIdsProTable/search.ts` (new): module-scoped, module-tested `filterFeedsBySearch(items, searchString)`. Adds `hermes_id` / `nasdaq_symbol` / `cmc_id` to the `matchSorter` keys in the single-term path, and to the multi-term `termMatchesItem` — substring match for `hermes_id` (with leading `0x`/`0X` stripped from the term) and `nasdaq_symbol`; exact match for `cmc_id` on numeric terms. Skips fields that are `null` / `undefined`. Preserves exact-`pyth_lazer_id` priority in the multi-term path.
- `PriceFeedIdsProTable/search.test.ts` (new): 15 cases covering hex `hermes_id` (raw / `0x` / `0X`), `nasdaq_symbol` on an equity, `cmc_id` numeric, multi-term OR across mixed identifiers, priority of exact `pyth_lazer_id` over `hermes_id` substring collision, null-identifier feed doesn't blow up, and existing symbol / name / description matching regressions.

## Test plan

- [x] `pnpm turbo run test:types --filter @pythnetwork/developer-hub` — pass (includes `next build`)
- [x] `pnpm turbo run test:lint test:format --filter @pythnetwork/developer-hub` — pass
- [x] `NODE_OPTIONS=--experimental-vm-modules pnpm jest PriceFeedIdsProTable` — 15/15 pass
- [x] Verified against live API (`https://pyth.dourolabs.app/v1/symbols`): the three new fields are present on every row, BKNG's `nasdaq_symbol` is `"BKNG"`, BTC's `hermes_id` is `e62df6...`.

### DIFF
--- a/apps/developer-hub/src/components/PriceFeedIdsProTable/search.ts
+++ b/apps/developer-hub/src/components/PriceFeedIdsProTable/search.ts
@@ -56,7 +56,12 @@ export const filterFeedsBySearch = <T extends SearchableFeed>(
     // Numeric ID match - exact match for numeric terms against pyth_lazer_id or cmc_id
     if (/^\d+$/.test(term)) {
       if (String(item.pyth_lazer_id) === term) return true;
-      if (item.cmc_id != null && String(item.cmc_id) === term) return true;
+      if (
+        item.cmc_id !== null &&
+        item.cmc_id !== undefined &&
+        String(item.cmc_id) === term
+      )
+        return true;
       return false;
     }
 
@@ -75,7 +80,7 @@ export const filterFeedsBySearch = <T extends SearchableFeed>(
 
     if (item.nasdaq_symbol?.toLowerCase().includes(term)) return true;
 
-    if (item.hermes_id != null) {
+    if (item.hermes_id !== null && item.hermes_id !== undefined) {
       const hermesTerm = stripHexPrefix(term);
       if (hermesTerm && item.hermes_id.toLowerCase().includes(hermesTerm)) {
         return true;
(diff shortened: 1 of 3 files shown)


apps/developer-hub/src/components/PriceFeedIdsProTable/search.ts
```diff
@@ -0,0 +1,118 @@
...
+      symbol.includes(term) ||
+      name.includes(term) ||
+      description.includes(term)
+    ) {
+      return true;
+    }
+
+    if (item.nasdaq_symbol?.toLowerCase().includes(term)) return true;
+
+    if (item.hermes_id != null) {
```
benduran: nit: we should use `!==` instead of `!=` for strictness

benduran (approved):

PR #3912, opened 2026-07-20
docs(developer-hub): replace Proxy page with Frontend Auth guide

aditya520:
## Summary

Replaces the beta **Proxy** API page with a **Frontend Authentication** guide. It documents the short-lived-JWT flow that browser and frontend apps should use instead of embedding the long-lived Pro API key. The Pyth Pro History API starts requiring authentication on **July 24, 2026**, so this pattern needs to be in the docs before then.

Live testing confirmed the `/auth/token` JWT is a **Pyth Pro-wide** credential: it authenticates the **History, REST, and WebSocket** APIs. So it also replaces the removed, unauthenticated Proxy service — a frontend streams over WebSocket, or queries REST and History, with a short-lived JWT instead of the proxy.

## How the JWT is presented

- **History and REST APIs** (HTTP): `Authorization: Bearer <JWT>` header.
- **WebSocket API**: `?ACCESS_TOKEN=<JWT>` query parameter on the connection URL — a browser cannot set an `Authorization` header on a WebSocket.
- The raw `PRO_API_KEY` never leaves the backend; it is used only to mint JWTs via `POST /auth/token`.

## Files

**New page**

- `content/docs/price-feeds/pro/frontend-auth.mdx` — overview; "when to use which pattern" table; "replacing the Proxy" callout; the four-step mint flow; `POST /auth/token` reference (request `ttl_seconds`; 200 response `access_token` / `expires_at` / `token_type`; 401/404 errors); Node/Express backend example (`POST /pyth-token`); browser TypeScript example (in-memory JWT cache with a refresh margin); the REST and WebSocket transports; "protecting the token endpoint" guidance; related links.

**Proxy removal + redirect**

- `content/docs/price-feeds/pro/api/proxy.mdx` — deleted.
- `content/docs/price-feeds/pro/api/index.mdx` — drop the `Proxy (Beta)` row from the Services table; add a note pointing frontend apps to Frontend Authentication.
- `content/docs/price-feeds/pro/api/meta.json` — drop the `Proxy (Beta)` nav entry.
- `apps/developer-hub/next.config.js` — permanent redirect `/price-feeds/pro/api/proxy` → `/price-feeds/pro/frontend-auth`.

**Retargeted links**

- `content/docs/price-feeds/pro/api/history.mdx` — replace the inline `JWT Authentication` section with a pointer to the new page; top callout links to it.
- `content/docs/price-feeds/pro/api/rest.mdx` — top callout and the Authentication section point to the new page; drop the "run their own proxy" phrasing.
- `content/docs/price-feeds/pro/acquire-api-key.mdx` — the "browser apps must not embed the key" paragraph and Next Steps point to Frontend Authentication.
- `content/docs/price-feeds/pro/faq.mdx` — the "authenticate from a frontend app" answer points to the JWT flow; new "What happened to the Proxy API?" entry.

**Streaming transport**

- `content/docs/price-feeds/pro/subscribe-to-prices.mdx` — document the browser WebSocket transport: pass the JWT as the `ACCESS_TOKEN` query parameter (a browser can't set the `Authorization` header on a WebSocket).

**Nav**

- `content/docs/price-feeds/pro/meta.json` — add `frontend-auth` under How-To Guides, after `acquire-api-key`.

## Invariants

- `grep -rn 'price-feeds/pro/api/proxy' apps/developer-hub/content` → zero (the deleted slug survives only as the redirect source in `next.config.js`).
- `grep -rn 'history#jwt-authentication' apps/developer-hub/content` → zero.
- No page embeds `PRO_API_KEY` in a browser snippet — the remaining references are server-side (curl, backend `process.env`, or explicit "don't embed" warnings).
- `POST /auth/token` request/response fields match the Auth OpenAPI spec.

## Testing

- **JWT scope, live:** minted a token via `POST /auth/token`, then confirmed History, REST (`POST /v1/latest_price`), and WebSocket (`/v1/stream`) all accept it. A bogus token is rejected with 401; the WebSocket accepts the JWT via both the `Authorization` header and the `ACCESS_TOKEN` query parameter. This is the basis for documenting the JWT as Pyth Pro-wide.
- **Docs tooling:** lint, types, and build for `@pythnetwork/developer-hub` run in CI on each push.

## Reviewers

Please flag **@aditya520** and **@jayantk** for content review before merge.


apps/developer-hub/content/docs/price-feeds/pro/frontend-auth.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: Frontend Authentication
+description: Use short-lived tokens to call Pyth Pro from browser apps
+slug: /price-feeds/pro/frontend-auth
+---
+
+import { Callout } from "fumadocs-ui/components/callout";
+
+This guide explains how to call Pyth Pro from browser and frontend apps without shipping your `PRO_API_KEY` to users. Anything in a browser bundle is public, so your backend mints a short-lived JWT with `POST /auth/token` and hands it to the browser. The browser then presents that JWT in place of the API key — as an `Authorization: Bearer` header on the History and REST APIs, and as the `ACCESS_TOKEN` query parameter on a WebSocket connection.
```
azellers: We now accept the 'sec-websocket-protocol' header, which should be preferred over the ACCESS_TOKEN query param. I think we should treat ACCESS_TOKEN as deprecated and undocumented.  For the sec-websocket-protocol approach, we can say something like:

The browser WebSocket API doesn't allow custom headers, so pass the token through the subprotocol list instead. Offer the marker `pyth-lazer-auth` immediately followed by your token:

```
  const ws = new WebSocket(url, ["pyth-lazer-auth", "<YOUR_TOKEN>"]);
```

  The server authenticates from that value and completes the handshake by echoing back the `pyth-lazer-auth` subprotocol. This is expected — your client should accept it and does not need to do anything with it.

apps/developer-hub/content/docs/price-feeds/pro/frontend-auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+  the browser uses it over the WebSocket, REST, and History APIs — the raw API
+  key stays server-side.
+</Callout>
+
+## When to use which pattern
+
+| Client                | Pattern                                                                                                                       |
+| --------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| Server-to-server      | Present the long-lived Pro API key directly as `Authorization: Bearer <PRO_API_KEY>`. Simplest — prefer this whenever the client can hold a secret. |
+| Browser / frontend    | Your backend mints a JWT via `POST /auth/token` and returns it to the browser, which presents it in place of the API key: `Authorization: Bearer <JWT>` on the History and REST APIs, or `?ACCESS_TOKEN=<JWT>` on a WebSocket. |
```
azellers: again, prefer `pyth-lazer-auth` via sec-websocket-protocol

apps/developer-hub/content/docs/price-feeds/pro/frontend-auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+const res = await fetch(
+  "https://pyth.dourolabs.app/v1/fixed_rate@200ms/price?ids=1&timestamp=1704067200000000",
+  { headers: { Authorization: `Bearer ${jwt}` } },
+);
+console.log(await res.json());
+```
+
+The same JWT works against the REST API on `https://pyth-lazer.dourolabs.app` — send it as `Authorization: Bearer <JWT>`, exactly as on the History API.
+
+For the WebSocket API, a browser cannot set an `Authorization` header on the connection, so pass the JWT as the `ACCESS_TOKEN` query parameter instead:
```
azellers: same as above

azellers (approved):

PR #3919, opened 2026-07-21
fix(price-pusher): identify success logs by message, not hash presence

aditya520:
## What

Fixes the 3 failing tests in `src/evm/__tests__/receipt-tracking.test.ts` that are currently red on `main` ([run 29835506529](https://github.com/pyth-network/pyth-crosschain/actions/runs/29835506529)). Test file only, no source changes, so the noise reduction from #3915 is fully preserved.

## Why the tests broke

`receipt-tracking.test.ts` is the regression suite for the bounded-receipt-tracker fix (#3885). It uses the `Price update successful` line as the observable signal that a tx was detected as landed. For example, the same-nonce escalation case uses it to prove the original tx's tracker was not cancelled when the old tx wins the mempool race.

#3915 moved that line from `info` to `debug`, and the assertions read the info channel.

The fix is not a simple channel swap. The harness's `record()` captured only the first argument and discarded the message, so `hashLogs()` identified a success purely by the presence of a `hash` field. That is unambiguous on `info`, where the success line was the only hash-bearing writer, but not on `debug`, which also carries:

- `evm.ts:301` `{ hash }` "Price update sent", on every send
- `evm.ts:515` `{ hash }` gave up at deadline
- `logger.error`, which the harness aliases into the same bucket

So the harness now captures the message and filters on it:

```ts
const SUCCESS_MSG = "Price update successful";
const successLogs = (entries: LogEntry[]): LogEntry[] =>
  entries.filter((entry) => entry.msg === SUCCESS_MSG && entry.hash !== undefined);
```

Five assertion sites move from `hashLogs(logs.info)` to `successLogs(logs.debug)`.

## Slightly stronger than before

Two assertions were passing vacuously after #3915. The never-lands and reverted-tx cases assert *zero* success logs, and `logs.info` was unconditionally empty once the line moved to debug, so they asserted nothing. They now check a channel that actually receives traffic.

## Verification

| Check | Result |
|---|---|
| `turbo run test --filter=@pythnetwork/price-pusher` | 23/23 tasks, 21/21 tests |
| Suite run 10x sequentially | 10/10 green |
| Suite run 4x concurrently (loaded-runner sim) | 4/4 green |
| Mutation: rename the log message | 3 tests fail, as expected |
| Mutation: log success on reverted txs | reverted-tx test fails, as expected |

The timing checks are because this suite uses real timers and `sleep()`, and the CI log showed a worker not exiting gracefully.

## Not addressed here

The bundled `grafana-dashboard.sample.json` "Tx Hash" panel scrapes all three of `Price update successful` (EVM), `Transaction confirmed` (Aptos) and `Successfully updated price...` (Sui) from Loki. #3915 moved all three to `debug`, and pushers default to `log-level: info`, so that panel now returns nothing. Left alone deliberately per discussion, flagging it so it is a known state rather than a surprise. The comments above those three call sites still describe the old info-level behaviour.

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/3919" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/developer-hub/turbo.json
+++ b/apps/developer-hub/turbo.json
@@ -4,6 +4,13 @@
   "tasks": {
     "build": {
       "dependsOn": ["generate:docs", "generate:changelog"],
+      "outputs": [
+        "lib/**",
+        "dist/**",
+        ".next/**",
+        "!.next/cache/**",
+        ".source/**"
+      ],
       "env": [
         "VERCEL_ENV",
         "COOKIE_SIGNING_SECRET",
--- a/apps/developer-hub/turbo.json
+++ b/apps/developer-hub/turbo.json
@@ -4,13 +4,6 @@
   "tasks": {
     "build": {
       "dependsOn": ["generate:docs", "generate:changelog"],
-      "outputs": [
-        "lib/**",
-        "dist/**",
-        ".next/**",
-        "!.next/cache/**",
-        ".source/**"
-      ],
       "env": [
         "VERCEL_ENV",
         "COOKIE_SIGNING_SECRET",
@@ -21,6 +14,13 @@
         "PLAYGROUND_MAX_STREAM_DURATION_MS",
         "PLAYGROUND_RATE_LIMIT_WINDOW_MS",
         "PLAYGROUND_RATE_LIMIT_MAX_REQUESTS"
+      ],
+      "outputs": [
+        "lib/**",
+        "dist/**",
+        ".next/**",
+        "!.next/cache/**",
+        ".source/**"
       ]
     },
     "fix:lint": {
--- a/apps/price_pusher/src/evm/__tests__/receipt-tracking.test.ts
+++ b/apps/price_pusher/src/evm/__tests__/receipt-tracking.test.ts
@@ -1,6 +1,7 @@
 import type { HermesClient, UnixTimestamp } from "@pythnetwork/hermes-client";
 import type { Logger } from "pino";
 
+import type { ReceiptOutcome } from "../evm.js";
 import { EvmPricePusher } from "../evm.js";
 import type { GasPriceConfig } from "../gas-price.js";
 import type { PythContract } from "../pyth-contract.js";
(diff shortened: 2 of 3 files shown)


aditya520:
Pushed a second commit (c493066cb) for an unrelated failure this PR exposed.

Once the price-pusher tests passed, the run got further than any recent CI run has and hit `@pythnetwork/developer-hub#test:types` failing with six TS errors, all downstream of `Cannot find module '../../.source/server'`.

**Not caused by this PR.** It is a latent turbo cache-correctness bug on `main`. developer-hub's `build` task inherits the root `outputs` list (`lib/**`, `dist/**`, `.next/**`), which omits `.source`, the gitignored directory the fumadocs MDX plugin generates during `next build`. On a build **cache hit** turbo skips the build and restores only the declared outputs, so `.source` never lands on disk and `tsc` fails.

That makes it non-deterministic rather than reliably broken:

| Run | `developer-hub:build` | Result |
|---|---|---|
| [29780119119](https://github.com/pyth-network/pyth-crosschain/actions/runs/29780119119) (2026-07-20) | cache miss, built for real | green, 91/91 |
| [29846229275](https://github.com/pyth-network/pyth-crosschain/actions/runs/29846229275) (this PR) | cache hit | 6 TS errors |

It stayed hidden because every recent run on every branch died earlier at `price-pusher#test:unit` and never reached this task.

Fix is adding `.source/**` to that task's `outputs`. Verified locally by forcing a build cache hit: `tsc` reproduces all six CI errors byte-for-byte before the change and exits 0 after, with `.source` restored from cache.

Happy to split this into its own PR if reviewers would rather keep the two concerns separate.

tejasbadadare (commented):
I feel like there's a code smell here — the test broke because a log line drifted, which suggests a brittle test. I suggest making the tracker's outcome observable through a non-log mechanism/seam and asserting on that.

matej-douro (approved):

PR #3922, opened 2026-07-22
feat(xc-admin-cli): propose integrity-pool reward withdraw

guibescos:
Adds a `withdraw-integrity-pool-rewards` multisig command to xc-admin-cli that proposes the integrity-pool `withdraw` instruction through the Squads vault.

## Details
- Extends the bundled `integrity-pool.json` IDL with the `withdraw` instruction and the `PoolConfig` account so Anchor can build the instruction and read the reward-custody mint on-chain.
- The command takes a required `--destination` (a PYTH token account) and an optional `--amount` in raw base units. When `--amount` is omitted it reads the reward-custody token balance and withdraws the full amount.
- `reward_program_authority` is resolved to the multisig vault authority PDA (same pattern as `set-publisher-stake-account`); `pool_config` is auto-resolved from its PDA seeds and `pool_reward_custody` is derived as the pool_config ATA of the pool's `pyth_token_mint`.
- Before proposing, the built instruction is simulated (`sigVerify: false`, since the authority signature is produced by the vault PDA at execution time) and the command aborts if the simulation fails.

## Verification
Simulated the built instruction against Solana mainnet-beta (pool_config `BE8Xq1iHSQYKG8CZorZsrTnxyeHKQLzCWm4dYKdFkVEL`, reward custody `FbSivnCf6iXfXns6osFih2MgX8icmSKx9iba3gyFpiCT`): the transaction returned `err: null` with program logs showing `Instruction: Withdraw` and a successful SPL token transfer CPI.

### DIFF
--- a/governance/xc_admin/packages/xc_admin_cli/src/index.ts
+++ b/governance/xc_admin/packages/xc_admin_cli/src/index.ts
@@ -61,6 +61,8 @@ import {
   SYSVAR_CLOCK_PUBKEY,
   SYSVAR_RENT_PUBKEY,
   SystemProgram,
+  TransactionMessage,
+  VersionedTransaction,
 } from "@solana/web3.js";
 import SquadsMesh from "@sqds/mesh";
 import { program } from "commander";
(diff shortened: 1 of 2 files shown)


KemarTiti (approved):
lgtm

PR #3937, opened 2026-07-28
feat(evm): add MigrateGovernanceAndWormhole governance action

keyvankhademi:
## Summary
- Adds Target governance action **10** `MigrateGovernanceAndWormhole` for legacy → pro-compatible in-place cutover.
- Atomically sets wormhole (no dual-VAA re-verify), replaces valid data sources, sets the **new** governance emitter, bumps `governanceDataSourceIndex`, and resets `lastExecutedGovernanceSequence` to `0`.
- Fee is **not** part of this action — set to `0` beforehand via existing `SetFee`.
- Leaves `SetWormholeAddress` dual-verify and `AuthorizeGovernanceDataSourceTransfer` unchanged (handoff cannot cross mismatched attestor sets).

Wire format (after PTGM header):
`wormhole(20) | dataSources* | governanceEmitter(chain u16 + addr 32) | governanceDataSourceIndex(u32)`

Sibling PRs:
- xc_admin_common codec
- contract_manager migrate tooling

## Test plan
- [ ] `forge test --match-contract PythGovernance`
- [ ] Happy path: wormhole, data sources, governance DS/index, sequence `0`; fee unchanged
- [ ] Reverts: zero/no-code wormhole, zero governance emitter, stale index, chainId `0`, trailing/truncated payload
- [ ] Confirm actions 6 and authorize-transfer behavior unchanged

### DIFF
--- a/target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
+++ b/target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
@@ -260,8 +260,9 @@ abstract contract PythGovernance is
         emit WormholeAddressSet(oldWormholeAddress, address(wormhole()));
     }
 
-    /// @dev Sets wormhole address, data sources, and fee without dual-VAA re-verify.
+    /// @dev Sets wormhole address and data sources without dual-VAA re-verify.
     /// Used for legacy → pro-compatible migration where guardian sets do not overlap.
+    /// Fee is set separately via SetFee before migration.
     function setWormholeAddressAndDataSources(
         SetWormholeAddressAndDataSourcesPayload memory payload
     ) internal {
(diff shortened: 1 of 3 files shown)


target_chains/ethereum/contracts/contracts/pyth/PythGovernance.sol
```diff
@@ -105,6 +105,14 @@ abstract contract PythGovernance is
             setTransactionFee(parseSetTransactionFeePayload(gi.payload));
         } else if (gi.action == GovernanceAction.WithdrawFee) {
             withdrawFee(parseWithdrawFeePayload(gi.payload));
+        } else if (
+            gi.action == GovernanceAction.MigrateGovernanceAndWormhole
+        ) {
+            if (gi.targetChainId == 0)
```
guibescos, line 111: Is the reason to exclude targetChainId 0 (which means all chains) that the wormhole address is almost surely different on every chain?

keyvankhademi: yes.
1. as you said the wormhole address is different on every chain, a targetChainId = 0 payload can be brick other contracts
2. governanceDataSourceIndex is also per chain (i'm not sure if they are all on the same number or not) but we should set it to `current + 1` for every chain.

PR #3997, opened 2026-08-18
chore(entropy): flag deprecated Entropy deployments and hide them from the chainlist

aditya520:
Entropy support for seven chains ended on **August 15, 2026**
([deprecation notice](https://dev-forum.pyth.network/t/deprecation-notice-pyth-entropy-deprecation-for-selected-chains-august-15-2026/816)).
The keepers stop serving them in `pyth-network/deployments`; this is the pyth-crosschain side.

Review feedback on the first revision of this PR was to **keep** the deployments in
`contract_manager` and mark them deprecated rather than delete them, and to drive the docs
chainlist off that flag. That is what this does — no deployment record is removed.

## What changed

- **`contract_manager/src/store/contracts/EvmEntropyContracts.json`** — every deployment
  covered by the notice now carries `"deprecated": true`: the 11 records for the seven chains
  (ZetaChain, Unichain, Taiko, Tabi, Story, Blast, Sei EVM — mainnet and/or testnet each), plus
  Etherlink mainnet and testnet, which the notice also covers and which the chainlist has
  hidden since August 1. Keeping the records means `transfer_balance_entropy_chains.ts`,
  `fetch_accrued_entropy_dao_fees.ts`, and `entropy-accept-admin-and-ownership.ts` still reach
  these chains, so any remaining provider balance or accrued DAO fee can still be swept.
- **`taiko_mainnet` is added** as a new record, also flagged. Taiko is one of the seven chains
  but its Entropy deployment was never registered here, so there was nothing to flag and the
  chainlist would have had to keep hiding it by name. The address
  `0x26DD80569a8B23768A1d80869Ed7339e07595E85` comes from the live keeper config and was
  verified on-chain (see below).
- **`EvmEntropyContract`** round-trips `deprecated` through `fromJson` / `toJson`, so
  `Store.saveAllContracts()` does not silently erase the flags the next time a deploy script
  saves the store. Verified by loading the store and diffing `toJson()` against the file.
- **`contract_manager/src/utils/utils.ts`** exports `evmEntropyContracts`, next to the
  existing `evmChains` / `evmLazerContracts` / `evmPriceFeedContracts` exports the docs
  already consume.
- **`apps/developer-hub/src/components/EntropyTable/entropy-api-data-fetcher.tsx`** — the
  hardcoded `HIDDEN_CHAINS` list is gone; the chainlist now hides whatever `contract_manager`
  flags as deprecated. Fortuna names chains itself and reports `network_id: 0` for chains whose
  RPC it cannot reach (`taiko`, `tabi-testnet` and `etherlink-testnet` today), so a deployment
  is matched on either the network id or the chain id spelled Fortuna-style (`sei_evm_mainnet`
  → `sei-evm`). Three unit tests cover the live, matched-by-id, and matched-by-name cases.
- **Docs and samples** — `debug-callback-failures.mdx` and `apps/entropy-tester/config.sample.json`
  used Blast as their worked example; both now use Base. A changelog entry announces the
  removal.
- **`apps/entropy-explorer/src/entropy-deployments.tsx` is deliberately untouched.**
  `GET /v1/logs` keeps returning history for these chains (~1M rows for Blast alone) and the
  Explorer throws `Invalid chain id` for any row it cannot map to a deployment, which fails the
  whole page. This matches the B3 and Sanko sunsets, which also kept their Explorer entries.

## Verification

- **Chainlist output is unchanged.** Replaying both live Fortuna configs
  (`fortuna.dourolabs.app` and `fortuna-staging.dourolabs.app`) through the new filter hides
  exactly the 14 names the deleted `HIDDEN_CHAINS` set hid — `blast`, `blast-testnet`,
  `etherlink`, `etherlink-testnet`, `sei-evm`, `sei-evm-testnet`, `story`, `story-testnet`,
  `tabi-testnet`, `taiko`, `unichain`, `unichain-sepolia`, `zetachain`, `zetachain-testnet` —
  and keeps all 23 other chains.
- **Taiko address checked on-chain** against `https://rpc.mainnet.taiko.xyz` (chain id
  `0x28c58` = 167000): the address has code, `version()` returns `0.4.0`, and
  `getDefaultProvider()` returns `0x52DeaA1c84233F7bb8C8A45baeDE41091c616506`, the same default
  provider as the Base mainnet deployment.
- **Store round-trip**: 42 Entropy contracts load, 14 report `deprecated`, and re-serializing
  them reproduces `EvmEntropyContracts.json` exactly.
- `pnpm turbo test:types` (includes the developer-hub production build), `pnpm run test:unit`
  in `apps/developer-hub` (6 suites / 69 tests), `biome ci --changed`, and
  `pre-commit run --from-ref origin/main --to-ref HEAD` all pass.

Sei price feeds are untouched: this is Entropy-only. Sei keeps its entries in `EvmChains.json`,
`EvmWormholeContracts.json`, `EvmExecutorContracts.json`, and `EvmLazerContracts.json`.

### DIFF
--- a/apps/developer-hub/content/changelog/2026-08-18-entropy-removed-from-seven-chains.mdx
+++ b/apps/developer-hub/content/changelog/2026-08-18-entropy-removed-from-seven-chains.mdx
@@ -1,0 +1,38 @@
+---
+title: Entropy removed from seven deprecated chains
+date: 2026-08-18
+product: entropy
+type: breaking-change
+area: network
+---
+
+Following the **August 15, 2026** deprecation date, Pyth Entropy no longer runs
+on seven of the chains named in the deprecation notice. The default provider has
+stopped serving requests on these networks, and they no longer appear in the
+Entropy chainlist.
+
+| Chain     | Networks removed    |
+| --------- | ------------------- |
+| ZetaChain | Mainnet and testnet |
+| Unichain  | Mainnet and testnet |
+| Taiko     | Mainnet             |
+| Tabi      | Testnet             |
+| Story     | Mainnet and testnet |
+| Blast     | Mainnet and testnet |
+| Sei EVM   | Mainnet and testnet |
+
+Entropy is deprecated on all seven chains. Taiko only ever ran an Entropy keeper
+on mainnet and Tabi only on testnet, so those are the only networks with anything
+to remove.
+
+**Why it matters:** requests to the default provider on these networks are no
+longer fulfilled. Entropy on every other supported network is unaffected, as are
+Pyth price feeds on all of the chains above — including Sei, where price feeds
+remain fully supported.
+
+**What to do:** if your application still calls Entropy on one of these chains,
+migrate to a [supported network](/entropy/chainlist). Past requests on these
+networks stay browsable in the Entropy Explorer.
+
+- [Deprecation notice: Pyth Entropy deprecation for selected chains (August 15, 2026)](https://dev-forum.pyth.network/t/deprecation-notice-pyth-entropy-deprecation-for-selected-chains-august-15-2026/816)
+- [Entropy chainlist](/entropy/chainlist)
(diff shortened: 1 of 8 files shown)


KemarTiti (approved):
lgtm

PR #4007, opened 2026-08-20
feat: migrate wormhole to new guardian set

guibescos:
## Summary

This PR enables migrating the core-bridge contract from Wormhole guardian sets to Pyth Pro multisig guardians.

## Rationale

We want to move off the Wormhole guardian set.


target_chains/solana/programs/core-bridge/src/legacy/processor/mod.rs
```diff
@@ -63,7 +65,10 @@ pub fn process_legacy_instruction<'info>(
             PostMessageUnreliable::process_instruction(program_id, account_infos, ix_data)
         }
         LegacyInstruction::UpdateGuardianSetTtl => {
-            UpdateGuardianSetTtl::process_instruction(program_id, account_infos, ix_data)
+            CloseGuardianSet::process_instruction(program_id, account_infos, ix_data)
```
guibescos: check what happens with existing SignatureSet accounts

keyvankhademi (approved):

PR #4008, opened 2026-08-20
feat(contract-manager): add EVM Pyth Pro cutover tooling

keyvankhademi:
## Summary

Adds the tooling to migrate legacy EVM price feed proxies to Pyth Pro verification **in place**. The consumer-facing proxy address never changes; what changes is what it verifies against — implementation, data sources, and wormhole pointer. The governance emitter is deliberately left alone.

Four phases, each with `--dry-run`:

| Script | Does |
| --- | --- |
| `preflight_evm_pro_cutover.ts` | read-only readiness report |
| `deploy_evm_pro_cutover.ts` | Pro receiver + `PythUpgradable` implementation |
| `propose_evm_pro_cutover.ts` | the three-action governance proposal |
| `execute_evm_pro_cutover.ts` | execute, verify, flip the store entry |

Shared logic is in `pro_cutover.ts`. `proposal_review.ts` holds the review checks so they can be exercised without a live proposal.

`check_proposal.ts` also gains a `SetDataSources` branch — it previously had none, so it silently skipped the middle action of every cutover — plus per-chain grouping and a non-zero exit on critical findings.

## Rationale

Three things drove the design and are worth a reviewer's attention.

**Governance payloads are chain-scoped, not address-scoped.** `EvmUpgradeContract`, `SetDataSources` and `EvmSetWormholeAddress` all carry a `targetChainId` and no target address. So a proposal is per chain, while execution, verification and store updates are per address. `findPriceFeedContracts` returns every matching contract rather than one; sepolia hosts two legacy proxies in lockstep at the same governance sequence.

**`executeVaa` broadcasts, which is a hazard here.** It submits a VAA to every contract on the chain whose governance emitter matches. `stable` and `pro-compatible-production` share emitter `5635979a…5b12e9e`, and most chains host a Pro proxy beside the legacy one. A cutover VAA sent to a Pro proxy fails verification and reverts, aborting the run. The execute script targets explicit proxy addresses instead.

**The gap between VAA2 and VAA3 is not merely "stale".** After `SetDataSources` but before `SetWormholeAddress`, the proxy holds Pro data sources while still verifying against its old wormhole, so it can verify **no** price update at all: Pro updates are signed by a guardian set the old wormhole does not know, and pythnet updates come from emitters that are no longer valid data sources. Stored prices still read, but any consumer with a staleness check reverts. The two actions therefore go in one Multicall3 `aggregate3` call with `allowFailure: false`, so that state cannot exist. 72 of 76 chains have Multicall3, including all 19 currently ready. The other four get two transactions and accept the window; `--force-sequential` does the same everywhere. A proxy stranded in that state is detected and repaired by sending VAA3 alone.

Resumability comes from each proxy's `lastExecutedGovernanceSequence` rather than a cache file, so a re-run skips exactly what already landed.

### Store changes

Nine dead or paywalled RPC endpoints replaced. Each replacement was validated on three things before being applied: `eth_chainId` matches the store's `networkId`, `eth_blockNumber` works, and an `eth_call` to that chain's Pyth proxy returns `version()`. The `networkId` check matters — working Manta endpoints report `3441006` while `manta_testnet` claimed `3441005`, so a naive swap would have silently repointed the entry at a different network.

`manta_testnet` is retired: `3441005` is the superseded Manta Pacific testnet, and `manta_sepolia` already covers Manta testnet at `3441006` with a live RPC and its own proxy. Removed from the chain store and its two contract entries; the store loads with no dangling references.

## How has this been tested?

- [ ] Added new tests

`contract_manager` has no test runner, so verification was by dry run against live RPCs and real on-chain state. `tsc` is unchanged at its 3 pre-existing errors and biome is clean on every file here.

- **preflight** across all 89 mainnet chains and every status
- **deploy** `--dry-run` across `READY` / `NEEDS RECEIVER` / `BLOCKED`, plus its guards (missing key, non-Pro deployment type). Artifact size check reports 24,067 bytes, the dual-verify-removed implementation
- **propose** `--dry-run` building all three payloads against a real implementation address read from the live ethereum proxy's ERC1967 slot, round-trip decoded and compared field by field. Rejections verified: no code at the address, an address that is not a `PythUpgradable`, missing ops key, and the staging vault against a production deployment
- **vault emitter check** verified live — the production vault emits from `5635979a…5b12e9e`, matching the config and the legacy proxies
- **check_proposal** against real mainnet proposals #312 and #311 (no findings, exit 0), plus a 12-case harness over `proposal_review.ts` covering every finding and its negative cases
- **execute** `--dry-run` against real VAAs (sequences 895–897) fetched from the live governance emitter: fetched, decoded, correctly identified as `WithdrawFee` rather than cutovers, and skipped

Two supporting scans informed the design: Multicall3 presence across all chains with a legacy proxy (72 present, 4 absent), and `lastExecutedGovernanceSequence` across all 75 readable proxies (maximum 897, which confirms pyth-direct's 10,000 starting point clears every target).

**Not tested:** the cutover execution path itself — Multicall3 encoding, pending-work calculation and post-execution verification — cannot run until a real cutover proposal exists. First run should be sepolia, which needs `--deployment-type pro-compatible-production` despite being a testnet, because its Pro deployment uses the production governance emitter. It also has two legacy proxies, exercising the multi-proxy path no mainnet chain will.

---

<a href="https://app.devin.ai/review/pyth-network/pyth-crosschain/pull/4008" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


apps/developer-hub/src/components/EvmContractsStatusTable/index.tsx
```diff
@@ -59,12 +59,14 @@ const buildStatuses = (
...
-    ) {
+    // Legacy deployments are everything that is not a Pro deployment. They carry no
+    // deploymentType historically, and "stable"/"beta" once labelled, so match on the
+    // absence of a Pro type rather than on the absence of the key.
+    const deploymentType =
+      "deploymentType" in contract ? contract.deploymentType : undefined;
+    if (deploymentType === "pro-compatible-production") {
       upgraded.set(contract.chain, contract.address);
-    } else if (!("deploymentType" in contract)) {
+    } else if (!deploymentType?.startsWith("pro-compatible")) {
```
guibescos, line 69: `deploymentType == stable` ?

contract_manager/scripts/check_proposal.ts
```diff
@@ -73,6 +81,11 @@ function pythContractsOn(chain: EvmChain): EvmPriceFeedContract[] {
 const parser = yargs(hideBin(process.argv))
   .usage("Usage: $0 --cluster <cluster_id> --proposal <proposal_address>")
   .options({
+    "allow-governance-transfer": {
```
guibescos, line 84: Do we need this? I don't think we're actually doing any governance transfers in the proposal

contract_manager/scripts/proposal_review.ts
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Works out which deployment the given data sources belong to.
+ * @param {DataSource[]} dataSources The data sources named by a `SetDataSources` action.
+ * @returns The matching deployment type, or undefined if they match none of them.
+ */
+export function identifyDataSources(
+  dataSources: DataSource[],
+): DeploymentType | undefined {
+  return DEPLOYMENT_TYPES.find((type) =>
```
guibescos, line 64: `toDeploymentType`?

contract_manager/scripts/check_proposal.ts
```diff
@@ -166,6 +179,19 @@ async function main() {
           }
         }
       }
+      if (instruction.governanceAction instanceof SetDataSources) {
+        const { targetChainId, dataSources } = instruction.governanceAction;
+        console.log(`\nVerifying SetDataSources on ${targetChainId}`);
+        for (const source of dataSources) {
+          console.log(`  ${source.emitterChain}:${source.emitterAddress}`);
```
guibescos, line 186: need a `/n` after this

contract_manager/scripts/check_proposal.ts
```diff
@@ -474,6 +500,45 @@ async function main() {
       }
     }
   }
+
```
guibescos, line 503: it didn't print the hash for testnet upgrades

guibescos (approved):

PR #4031, opened 2026-08-28
chore: remove wormhole hack

guibescos:
## Summary

Remove migration features from the wormhole receiver program so it's easier to maintain moving forward.


target_chains/solana/programs/core-bridge/src/legacy/instruction/mod.rs
```diff
@@ -42,8 +42,6 @@ pub enum LegacyInstruction {
     /// kept as a placeholder so the remaining variants keep their original serialized
     /// discriminants.
     _RemovedUpdateGuardianSetTtl,
-    /// Close a guardian set containing a legacy guardian.
-    CloseGuardianSet,
```
guibescos, line 46: removed

keyvankhademi (approved):

